PR #24, opened 2015-03-30
Add localization for zh-rTW

bucker:
Add localization for zh-rTW.


lkorth:
@bucker thanks for the contribution!

We'll look at merging it in, we need to make sure we can maintain these translations through future changes to the SDK. We also have a [few more strings](https://github.com/braintree/android-card-form/blob/master/CardForm/src/main/res/values/strings.xml) that are part of the card form that we will need to translate before we can support zh-rTW.

bucker:
Hi @lkorth great, thanks!
Also if there's anything I can help, please let me know.

PR #26, opened 2015-04-13
Fix crash when completing the purchase and main activity is destroyed

alexeystep:
Initial conditions:
- Turn on "Do not keep activities" flag in Developer Options

Scenario:
 - Press Drop-In button
 - Enter credit card information e.g. 4242 4242 4242 4242
 - Press Buy button

Expected:
 - Main Activity is in the foreground displaying the nonce
Actual:
 - Demo app crashes


lkorth:
Thanks for the contribution @alexeystep!

PR #50, opened 2015-10-04
Bump Google Play Services version to 8.1.0 or above

nvanbenschoten:
In reference to https://code.google.com/p/android/issues/detail?id=187483, clients depending on both Google Play Services 8.1.0 and on libraries who depend on Google Play Services < 8.1.0 will fail building, giving the dex error:

```
Unknown source file : Uncaught translation error: com.android.dx.cf.code.SimException: com.android.dx.rop.cst.CstMethodRef cannot be cast to com.android.dx.rop.cst.CstInterfaceMethodRef
Unknown source file : Uncaught translation error: com.android.dx.cf.code.SimException: com.android.dx.rop.cst.CstMethodRef cannot be cast to com.android.dx.rop.cst.CstInterfaceMethodRef
Unknown source file : 2 errors; aborting
```

This pull request is bumping up the gms wallet version in the Android Braintree SDK from 7.8.0 to 8.1.0. At first glance, it looks like this update is completely stable, and there don't seem to be any breaking changes.

### DIFF
--- a/BraintreeApi/build.gradle
+++ b/BraintreeApi/build.gradle
@@ -26,7 +26,7 @@ configurations {
 
 dependencies {
     compile 'com.google.code.gson:gson:[2.2.4,3.0)'
-    compile 'com.google.android.gms:play-services-wallet:[7.5.0,8.0.0)'
+    compile 'com.google.android.gms:play-services-wallet:[8.1.0,9.0.0)'
 
     compile project(':BraintreeData')
     compile files('libs/' + paypalVersion)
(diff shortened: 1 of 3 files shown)


nvanbenschoten:
Based on the [Play Services Release History](https://developers.google.com/android/guides/releases), it doesn't look like the wallet API was touched between these two versions.

lkorth:
@nvanbenschoten: Thanks for the PR! We have an update going out today to that already includes this change.

One thing worth noting, since the dependency on Google Play Services Wallet is optional in `braintree_android`, using this library in an app that uses `8.1.0` does not generate a compile error, but method calls to the wallet API appear to fail.

PR #128, opened 2016-11-08
Fix NPE with connectivityManager.getActiveNetworkInfo() in AnalyticsEvent

evastos:
When called consecutively, `connectivityManager.getActiveNetworkInfo()` may return different values. Storing the `NetworkInfo` object in a variable fixes wild NPEs in `AnalyticsEvent.getNetworkType`.

[StackOverflow link](http://stackoverflow.com/questions/28872310/connectivitymanager-getactivenetworkinfo-returns-null)


lkorth:
Thanks for the PR @evastos!

PR #130, opened 2016-11-22
Migrate lint check to new PSI API

felipecsl:
Custom lint checks using the old and deprecated Lombok AST API cause an incompatibility between the Android lint and Retrolambda as described here https://github.com/evant/gradle-retrolambda/issues/96

Migrating to the new API fixes the issue.

### DIFF
--- a/Lint/src/main/java/com/braintreepayments/lint/BetaDetector.java
+++ b/Lint/src/main/java/com/braintreepayments/lint/BetaDetector.java
@@ -1,83 +1,50 @@
 package com.braintreepayments.lint;
 
-import com.android.annotations.NonNull;
-import com.android.annotations.Nullable;
 import com.android.tools.lint.detector.api.Category;
 import com.android.tools.lint.detector.api.Detector;
-import com.android.tools.lint.detector.api.Detector.JavaScanner;
 import com.android.tools.lint.detector.api.Implementation;
 import com.android.tools.lint.detector.api.Issue;
 import com.android.tools.lint.detector.api.JavaContext;
 import com.android.tools.lint.detector.api.Scope;
 import com.android.tools.lint.detector.api.Severity;
-import com.android.tools.lint.detector.api.Speed;
 import com.android.tools.lint.detector.api.TextFormat;
+import com.intellij.psi.JavaElementVisitor;
+import com.intellij.psi.PsiMethod;
+import com.intellij.psi.PsiMethodCallExpression;
 
 import java.util.Collections;
 import java.util.List;
 
-import lombok.ast.AstVisitor;
-import lombok.ast.MethodInvocation;
-import lombok.ast.Node;
+public class BetaDetector extends Detector implements Detector.JavaPsiScanner {
+  // This would normally be private final, but has exposed here for testing purposes, since there
+  // seems to be no way to inject it via the Detector constructor using LintDetectorTest
+  static List<String> METHODS = Collections.emptyList();
+  private static final Implementation IMPLEMENTATION = new Implementation(BetaDetector.class,
+      Scope.JAVA_FILE_SCOPE);
+  static final Issue ISSUE = Issue.create(
+      "com.braintreepayments.beta",
+      "API is in beta",
+      "This API is currently in beta and subject to change. Braintree makes no guarantees about " +
+          "future compatibility or existence of this API.",
+      Category.CORRECTNESS,
+      6,
+      Severity.ERROR,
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.3'
+        classpath 'com.android.tools.build:gradle:2.2.2'
```
lkorth: This upgrade is not possible at the moment, Robolectric 3.0 does not work with gradle build plugin 2.2.2 and upgrading Robolectric to 3.1.4 causes a conflict with PowerMock.

Lint/src/main/java/com/braintreepayments/lint/BetaDetector.java
```diff
@@ -1,46 +1,37 @@
...
 
-import lombok.ast.AstVisitor;
-import lombok.ast.MethodInvocation;
-import lombok.ast.Node;
-
-public class BetaDetector extends Detector implements JavaScanner {
-
-    private static final List<String> METHODS = Collections.emptyList();
-
+public class BetaDetector extends Detector implements Detector.JavaPsiScanner {
```
lkorth: Newline after this.

Lint/src/main/java/com/braintreepayments/lint/BetaDetector.java
```diff
@@ -1,46 +1,37 @@
...
+    // This would normally be private final, but has exposed here for testing purposes, since there
+    // seems to be no way to inject it via the Detector constructor using LintDetectorTest
```
lkorth: This seems fine, I don't think it really needs a comment.

Lint/src/test/java/com/braintreepayments/lint/BetaDetectorTest.java
```diff
@@ -0,0 +1,64 @@
...
+import com.android.tools.lint.checks.infrastructure.LintDetectorTest;
+import com.android.tools.lint.detector.api.Detector;
+import com.android.tools.lint.detector.api.Issue;
+
+import java.util.Collections;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class BetaDetectorTest extends LintDetectorTest {
```
lkorth: Newline here.

build.gradle
felipecsl: Hmmm weird, we have Robolectric 3.0 working fine with the plugin v2.2.2, anyways this is unrelated so I'll just revert it

felipecsl:
PR updated

felipecsl:
would it be possible to get this out in a SNAPSHOT release of the library?

build.gradle
lkorth: The difference is probably an Android application vs library. The location of the merged manifest changed in 2.2.2 and Robolectric has the path hard coded to the old path in 3.0 for libraries.

lkorth:
I've released this in `2.3.12-SNAPSHOT`. We have had issues with SNAPSHOTs being cleaned up after a short period of time, so I can't say how long it will stick around, but it should be available for a few days.

felipecsl:
Since the latest release is `2.3.12`, i'd expect the current snapshot to be `2.3.13-SNAPSHOT`, no?

felipecsl:
Also I can't find  `2.3.12-SNAPSHOT` anywhere

lkorth:
Sorry about the wait @felipecsl, I've just returned after some leave.

No, we don't currently version snapshots like that since we don't know the next version number at the time, but it is a bit backwards compared to most libraries' snapshot versioning.

I've released `2.3.12-SNAPSHOT` again and you can find it https://oss.sonatype.org/content/repositories/snapshots/com/braintreepayments/api/braintree/2.3.12-SNAPSHOT/

I'll make a full release soon.

lkorth:
Running lint with this change spits out quite a bit of noise because of missing classes and results in the custom rules not being loaded.

```
Could not load custom rule jar file /braintree-android/Demo/build/intermediates/exploded-aar/com.braintreepayments.api/Braintree/2.3.12-SNAPSHOT/jars/lint.jar
java.lang.NoClassDefFoundError: com/android/tools/lint/detector/api/Detector$JavaPsiScanner
```

Since the custom rules aren't currently being used for anything and I haven't seen any workable fix for this, I've just removed them for now.

The 2.4.0 release is going out right now with these changes.

PR #143, opened 2017-04-11
Include cause in ConfigurationException

brianguertin:
So, for debugging purposes, I would prefer to know the cause of this exception when there is one, and I don't see any harm in including it :)

### DIFF
--- a/Braintree/src/main/java/com/braintreepayments/api/exceptions/ConfigurationException.java
+++ b/Braintree/src/main/java/com/braintreepayments/api/exceptions/ConfigurationException.java
@@ -4,9 +4,11 @@ package com.braintreepayments.api.exceptions;
  * Error class thrown when a configuration value is invalid
  */
 public class ConfigurationException extends BraintreeException {
+    
     public ConfigurationException(String message, Throwable cause) {
         super(message, cause);
     }
+    
     public ConfigurationException(String message) {
         super(message);
     }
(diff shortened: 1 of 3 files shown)


quinnjn:
Hi @brianguertin! Thanks for the PR!

I'm intrigued that you had to debug something in `BraintreeFragment`. Do you mind explaining what you were debugging? Also, how would knowing the stack trace have made this debugging easier?

brianguertin:
Oh, just noticed your comment.

The reason I want to differentiate is because I prefer handle network connectivity issues in a different manner (E.g. informing the user they are offline or have no internet connectivity). I also ensure that general connectivity issues don't raise alerts in our app's bug tracking tools.

Currently, `ConfigurationException` swallows connectivity issues, so I can't detect them and handle them differently.

Braintree/src/main/java/com/braintreepayments/api/exceptions/ConfigurationException.java
```diff
@@ -4,6 +4,9 @@
  * Error class thrown when a configuration value is invalid
  */
 public class ConfigurationException extends BraintreeException {
+    public ConfigurationException(String message, Throwable cause) {
```
lkorth: Needs a new line before this line.

Braintree/src/main/java/com/braintreepayments/api/exceptions/ConfigurationException.java
```diff
@@ -4,6 +4,9 @@
...
+        super(message, cause);
+    }
```
lkorth: Needs a new line after this line.

lkorth:
Thanks again for the PR @brianguertin! Just two whitespace changes and we'll get this merged in.

PR #159, opened 2017-08-10
Add DOM storage & DB storage to ThreeDSecureWebView

apascual:
There are some banks which 3DSecure website require web storage in order to work properly, specially when another 3rd party app is required in the process, such as digipass or bankid.


lkorth:
Thanks for the PR @apascual!

PR #170, opened 2017-11-17
optimization with regrex

Pavneet-Sing:
eliminate spaces with regex henceforth no loop and trim

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/Settings.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/Settings.java
@@ -121,11 +121,7 @@ public class Settings {
     }
 
     public static String[] getAndroidPayAllowedCountriesForShipping(Context context) {
-        String[] countries = getPreferences(context).getString("android_pay_allowed_countries_for_shipping", "US").split(",");
-        for(int i = 0; i < countries.length; i++) {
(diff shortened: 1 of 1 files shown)


Demo/src/main/java/com/braintreepayments/demo/Settings.java
```diff
@@ -121,11 +121,7 @@ public static String getAndroidPayCurrency(Context context) {
     }
 
     public static String[] getAndroidPayAllowedCountriesForShipping(Context context) {
-        String[] countries = getPreferences(context).getString("android_pay_allowed_countries_for_shipping", "US").split(",");
-        for(int i = 0; i < countries.length; i++) {
-            countries[i] = countries[i].trim();
-        }
-
+        String[] countries = getPreferences(context).getString("android_pay_allowed_countries_for_shipping", "US").split("\\s*,\\s*");
```
Pavneet-Sing: although for edge case (space at the beginning and end ) it should be

String[] countries = getPreferences(context).getString("android_pay_allowed_countries_for_shipping", "US").trim().split("\\s*,\\s*");

crookedneighbor:
Thanks for the PR, but we're going to leave this as is for now.

PR #273, opened 2019-10-07
[Proof of Concept] Create new BraintreeFragment when authorization string doesn't match

krunk4ever:

### Summary of changes

Currently when we call `BraintreeFragment.newInstance()`, it can sometimes return a cached instance which is fine when all the other parameters are the same. However, when I call `BraintreeFragment.newInstance()` with a different `authorization` string, it should NOT returned the cached instance and instead create a new one with the new `authorization` string.

The PR below is untested and is more of a proof of concept.

I also noted that for FragmentTransactions, we would need to update `add` with `replace` when there's an existing fragment associated to `TAG`.

Example:

```kt
val activity = getAppCompatActivity()
val bf1 = BraintreeFragment.newInstance(activity, "123456")
val bf2 = BraintreeFragment.newInstance(activity, "abcedf")
```

Both calls would actually return an instance of `BraintreeFragment` where the `authorization` is `"123456"`

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/Braintree/src/main/java/com/braintreepayments/api/BraintreeFragment.java
+++ b/Braintree/src/main/java/com/braintreepayments/api/BraintreeFragment.java
@@ -170,7 +170,8 @@ public class BraintreeFragment extends BrowserSwitchFragment {
         }
 
         BraintreeFragment braintreeFragment = (BraintreeFragment) fragmentManager.findFragmentByTag(TAG);
-        if (braintreeFragment == null) {
+        
+        if (braintreeFragment == null || !braintreeFragment.hasMatchingAuthorization(authorization)) {
             braintreeFragment = new BraintreeFragment();
             Bundle bundle = new Bundle();
 
@@ -187,6 +188,7 @@ public class BraintreeFragment extends BrowserSwitchFragment {
 
             try {
                 if (VERSION.SDK_INT >= VERSION_CODES.N) {
+                    // TODO: where we use `add`, we might nee to use `replace` if an existing fragment for that TAG exists
                     try {
                         fragmentManager.beginTransaction().add(braintreeFragment, TAG).commitNow();
                     } catch (IllegalStateException | NullPointerException e) {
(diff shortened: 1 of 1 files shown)


Braintree/src/main/java/com/braintreepayments/api/BraintreeFragment.java
```diff
@@ -558,6 +560,10 @@ public void onActivityResult(final int requestCode, int resultCode, Intent data)
     public boolean hasFetchedPaymentMethodNonces() {
         return mHasFetchedPaymentMethodNonces;
     }
+    
+    public boolean hasMatchingAuthorization(String authorization) {
+        return mAuthorization.toString() == authorization;
```
krunk4ever: we probably want to do a null check here

krunk4ever:
cc @sestevens

Epreuve:
@krunk4ever Can you elaborate on what sort of use case you have where this would be useful?  I'm assuming front-loading fragments for some differing flows or something around offering different payment methods or currencies, but I'd like to understand the specifics you had in mind to be sure.

Provided the use case is something we agree the SDK could use in terms of this feature, we'd require that tests are added to confirm the behavior, and addressing the NPE you brought up ideally as well.

krunk4ever:
@Epreuve By default we use single static `tokenization key` for vaulting credit cards. However, for SCA/3DS2, the `tokenization key` doesn't work. Instead we need to fetch a `client token` from the server and build a new `BraintreeFragment`.

Epreuve:
Ah, makes sense.  I figured it was something like that.  I do think that something like this might be useful overall, exactly because of this sort of edge case (where you have a handful of possible merchant account IDs that can be used instead of just one).

I'm going to reach out to some colleagues as I believe might be an alternative, but also bring up this as another approach (if my memory serves me, both seem useful in different scenarios).

Braintree/src/main/java/com/braintreepayments/api/BraintreeFragment.java
v-sulimov: You perform reference check(not the actual value) between two String's. I think .equals() call should be here instead.

scannillo:
Hi @krunk4ever. Version 3.7.1 of `braintree_android` was released with this fix. Thank you for bringing this to our attention and your contribution!

https://github.com/braintree/braintree_android/releases/tag/3.7.1

krunk4ever:
@scannillo do you mind pointing to the commit that fix this. want to double check the logic before i upgrade. thanks!

scannillo:
https://github.com/braintree/braintree_android/commit/4b7f3af52f84bdc2ef7f061e8985f0d3c1753d1d

PR #334, opened 2021-03-09
Three d secure updates

sarahkoop:
### Summary of changes

 * The `shippingMethod` property on `ThreeDSecureRequest` is now an enum rather than a string. Possible values:
    * `SAME_DAY`
    * `EXPEDITED`
    * `PRIORITY`
    * `GROUND`
    * `ELECTRONIC_DELIVERY`
    * `SHIP_TO_STORE`
  * Change default `versionRequested` on `ThreeDSecureRequest` to `ThreeDSecureVersion.VERSION_2`
  * Rename `ThreeDSecureRequest#uiCustomization()` to `ThreeDSecureRequest#v2UiCustomization()`

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire 
- @sarahkoop


scannillo (approved):

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
```diff
@@ -33,15 +33,15 @@
...
     private ThreeDSecurePostalAddress mBillingAddress;
-    private @ThreeDSecureVersion String mVersionRequested = VERSION_1;
+    private @ThreeDSecureVersion String mVersionRequested = VERSION_2;
     private @ThreeDSecureAccountType String mAccountType;
     private ThreeDSecureAdditionalInformation mAdditionalInformation;
     private boolean mChallengeRequested = false;
     private boolean mDataOnlyRequested = false;
     private boolean mExemptionRequested = false;
-    private UiCustomization mUiCustomization;
+    private UiCustomization mv2UiCustomization;
```
sestevens, line 44:
```suggestion
    private UiCustomization mV2UiCustomization;
```
This hardly matters, but it looks like we're capitalizing the `v` in `mV1UiCustomization`.

sestevens:
Should we add a couple sentences about these changes to the migration guide as well?

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
sarahkoop: This needed to be changed in a few places 👍

sarahkoop:
> Should we add a couple sentences about these changes to the migration guide as well?

There wasn't an explicit section of the migration guide to discuss these changes, so I added them as comments above the creation of the `ThreeDSecureRequest` in the code snippet. Open to other ideas of where to note these changes in the guide.

v4_MIGRATION_GUIDE.md
```diff
@@ -733,12 +733,15 @@ public class ThreeDSecureActivity extends AppCompatActivity {
     ThreeDSecureAdditionalInformation additionalInformation = new ThreeDSecureAdditionalInformation()
         .accountId("account-id");
 
+    // ThreeDSecureRequest default version requested is now VERSION_2
+    // uiCustomization() for 3DS v2 has been renamed to v2UiCustomization()
+    // shippingMethod property is now an enum
```
sestevens, line 738: I think what you have here works, but we could also put these notes above the whole 3DS code snippet. Maybe right before "To use the feature, instantiate a `ThreeDSecureClient`".

Feel free to copy from the [iOS migration guide](https://github.com/braintree/braintree_ios/blob/master/V5_MIGRATION.md#3d-secure), too. 🙂

sestevens (approved):

v4_MIGRATION_GUIDE.md
sarahkoop: The iOS migration guide seemed clearer, so I updated to follow that 👍

PR #337, opened 2021-03-10
Cleanup Google Pay Client

sarahkoop:
### Summary of changes

 - Make `GooglePayClient#tokenize()` package-private
 - Remove unneeded activity parameter from methods

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop

### DIFF
--- a/GooglePay/src/main/java/com/braintreepayments/api/GooglePayClient.java
+++ b/GooglePay/src/main/java/com/braintreepayments/api/GooglePayClient.java
@@ -44,8 +44,8 @@ public class GooglePayClient {
     private static final String CARD_PAYMENT_TYPE = "CARD";
     private static final String PAYPAL_PAYMENT_TYPE = "PAYPAL";
 
-    private BraintreeClient braintreeClient;
-    private GooglePayInternalClient internalGooglePayClient;
+    private final BraintreeClient braintreeClient;
+    private final GooglePayInternalClient internalGooglePayClient;
 
     public GooglePayClient(BraintreeClient braintreeClient) {
         this(braintreeClient, new GooglePayInternalClient());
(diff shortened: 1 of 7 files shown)


GooglePay/src/main/java/com/braintreepayments/api/GooglePayClient.java
```diff
@@ -209,12 +207,10 @@ public void onResult(@Nullable Configuration configuration, @Nullable Exception
      * Call this method when you've received a successful {@link PaymentData} response in your
      * activity or fragment's {@code onActivityResult} method to get a {@link GooglePayCardNonce}
      * or {@link PayPalAccountNonce}.
-     *
-     * @param activity Android FragmentActivity
      * @param paymentData {@link PaymentData} from the Intent in {@code onActivityResult} method.
      * @param callback {@link GooglePayOnActivityResultCallback}
      */
```
sarahkoop, line 212: We've been leaving doc strings even when making methods package-private to continue to provide a description of the method. Do people agree that's helpful or is the current description confusing?  Should the strings be updated to a more internal description of the method or left as is?

sestevens: My personal preference would be to remove doc comments from internal methods. My suspicion is that they'll get out of date quickly and cause more confusion than it's worth.

sestevens (approved):

GooglePay/src/main/java/com/braintreepayments/api/GooglePayClient.java
scannillo: Since the docstring is still accurate to what the method is doing - I think we could keep it. I think it's better than not to have doc comments, even for internal things. It can save a lot of time, especially for new people to the code base.

But in places if the comment is outdated - no comment is better than an incorrect one.

scannillo (approved):

GooglePay/src/main/java/com/braintreepayments/api/GooglePayClient.java
sarahkoop: This comment wasn't completely accurate, so I updated it. I think going forward, though, I would also lean towards removing non-public comments because I don't think they are always updated. We can do another pass of all the doc strings to make sure both public and non-public methods have accurate docs (if they exist).

PR #338, opened 2021-03-11
3DS2 UI Customization Wrapper

sarahkoop:
### Summary of changes

 - Wrap Cardinal UI customization classes in Braintree classes
 - Mark Cardinal dependency as `implementation` instead of `api`
 - Follows iOS PR [569](https://github.com/braintree/braintree_ios/pull/569)

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@
 * Add `setCountryCode` to `GooglePayRequest`
 * Add Google Pay support for Elo cards. 
 * Add `VenmoRequest` 
+* Add new classes for 3DS2 UI Customization:
+  * `ThreeDSecureV2UiCustomization`
+  * `ThreeDSecureV2ButtonCustomization`
+  * `ThreeDSecureV2LabelCustomization`
+  * `ThreeDSecureV2TextBoxCustomization`
+  * `ThreeDSecureV2ToolbarCustomization`
 * Breaking Changes
   * Make `AmericanExpressRewardsBalance#fromJson()` package-private
   * Make `TYPE` and `API_RESOURCE_KEY` in `CardNonce` package-private
@@ -44,7 +50,7 @@
     * `ELECTRONIC_DELIVERY`
     * `SHIP_TO_STORE`
   * Change default `versionRequested` on `ThreeDSecureRequest` to `ThreeDSecureVersion.VERSION_2`
-  * Rename `uiCustomization` on `ThreeDSecureRequest` to `v2UiCustomization`
+  * Rename `uiCustomization` on `ThreeDSecureRequest` to `v2UiCustomization` and change parameter to `ThreeDSecureV2UiCustomization`
   * Replace `VenmoClient#authorizeAccount()` with `VenmoClient#tokenizeVenmoAccount()`
   * Rename `VenmoAuthorizeAccountCallback` to `VenmoTokenizeAccountCallback`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@
     * `SHIP_TO_STORE`
   * Change default `versionRequested` on `ThreeDSecureRequest` to `ThreeDSecureVersion.VERSION_2`
   * Rename `uiCustomization` on `ThreeDSecureRequest` to `v2UiCustomization` and change parameter to `ThreeDSecureV2UiCustomization`
+  * Change Cardinal dependency from `api` to `implementation`
   * Replace `VenmoClient#authorizeAccount()` with `VenmoClient#tokenizeVenmoAccount()`
   * Rename `VenmoAuthorizeAccountCallback` to `VenmoTokenizeAccountCallback`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@
     * `SHIP_TO_STORE`
   * Change default `versionRequested` on `ThreeDSecureRequest` to `ThreeDSecureVersion.VERSION_2`
   * Rename `uiCustomization` on `ThreeDSecureRequest` to `v2UiCustomization` and change parameter to `ThreeDSecureV2UiCustomization`
+  * Update setters on `V1UiCustomization` to remove method chaining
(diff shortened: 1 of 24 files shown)


ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
```diff
@@ -293,10 +291,7 @@ public ThreeDSecureV1UiCustomization getV1UiCustomization() {
         return mV1UiCustomization;
     }
 
-    public ThreeDSecureRequest() {
-        // NOTE: this is a temporary fix for a null-pointer bug introduced by Cardinal 2.2.3-2
```
sarahkoop, line 297: This bug still exists in Cardinal 2.2.5-1, but setting an empty `UiCustomization` here is no longer necessary, since one is instantiated when a `ThreeDSecureV2UiCustomization` is created.

scannillo:
So we added those wrapper UICustomization classes in iOS mainly just to get around the issue with SPM not supporting the `CardinalMobile.framework`. The wrapper classes allow us to avoid calling the actual CardinalMobile classes until runtime.

So, I wanted to raise the question of - do we need these wrappers in Android? Yes - it creates parity with iOS. But it also a lot of extra code and will require us to make updates whenever Cardinal makes changes to `UICustomization`. 

I can see both ways, but wanted to see if we had thought on this!

Demo/src/main/java/com/braintreepayments/demo/CardFragment.java
```diff
@@ -389,15 +389,15 @@ private ThreeDSecureRequest threeDSecureRequest(PaymentMethodNonce paymentMethod
         ThreeDSecureAdditionalInformation additionalInformation = new ThreeDSecureAdditionalInformation()
                 .accountId("account-id");
 
-        ToolbarCustomization toolbarCustomization = new ToolbarCustomization();
-        toolbarCustomization.setHeaderText("Braintree 3DS Checkout");
-        toolbarCustomization.setBackgroundColor("#FF5A5F");
-        toolbarCustomization.setButtonText("Close");
-        toolbarCustomization.setTextColor("#222222");
-        toolbarCustomization.setTextFontSize(18);
+        ThreeDSecureV2ToolbarCustomization toolbarCustomization = (ThreeDSecureV2ToolbarCustomization) new ThreeDSecureV2ToolbarCustomization()
```
sestevens, line 392:
```suggestion
        ThreeDSecureV2ToolbarCustomization toolbarCustomization = new ThreeDSecureV2ToolbarCustomization()
```
I think that can be removed?

sestevens (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 * Add `setCountryCode` to `GooglePayRequest`
 * Add Google Pay support for Elo cards. 
+* Add new classes for 3DS2 UI Customization:
```
sestevens, line 7: Can we make a note of this in the migration guide as well?

sestevens:
> So, I wanted to raise the question of - do we need these wrappers in Android? Yes - it creates parity with iOS. But it also a lot of extra code and will require us to make updates whenever Cardinal makes changes to UICustomization.

I like having the wrapper classes. I think the parity with iOS is good, and it gives us more control over the API. We can do things like use builders, which matches the rest of the SDK, and give the classes more specific names to indicate that they have to do with 3DS2. `ThreeDSecureV2ToolbarCustomization` is much more specific than `ToolbarCustomization`, for example.

sarahkoop:
> So we added those wrapper UICustomization classes in iOS mainly just to get around the issue with SPM not supporting the `CardinalMobile.framework`. The wrapper classes allow us to avoid calling the actual CardinalMobile classes until runtime.
> 
> So, I wanted to raise the question of - do we need these wrappers in Android? Yes - it creates parity with iOS. But it also a lot of extra code and will require us to make updates whenever Cardinal makes changes to `UICustomization`.
> 
> I can see both ways, but wanted to see if we had thought on this!

> > So, I wanted to raise the question of - do we need these wrappers in Android? Yes - it creates parity with iOS. But it also a lot of extra code and will require us to make updates whenever Cardinal makes changes to UICustomization.
> 
> I like having the wrapper classes. I think the parity with iOS is good, and it gives us more control over the API. We can do things like use builders, which matches the rest of the SDK, and give the classes more specific names to indicate that they have to do with 3DS2. `ThreeDSecureV2ToolbarCustomization` is much more specific than `ToolbarCustomization`, for example.

The other main benefit of wrapping the cardinal classes is that it allows us to mark the cardinal dependency as `implementation` rather than `api` in the `ThreeDSecure` module.

scannillo:
> The other main benefit of wrapping the cardinal classes is that it allows us to mark the cardinal dependency as implementation rather than api in the ThreeDSecure module.

Oh this is huge. Would this prevent merchants from needing to add the Cardinal credentials to their gradle?

scannillo (approved):

CHANGELOG.md
sarahkoop: Added this to the migration guide 👍

Demo/src/main/java/com/braintreepayments/demo/CardFragment.java
sarahkoop: This casting was necessary when we were following the builder/method chaining pattern with a base class, so we decided to rethink the design of these classes and move away from the builder pattern. Abandoning the builder pattern in the SDK will make it more Kotlin-friendly moving forward. We updated the v1 and v2 UI customization classes to follow the standard Java getter/setter pattern in this PR, and we plan to update the other builder pattern classes in future PRs.

sestevens: That makes sense 👍

sarahkoop:
> Oh this is huge. Would this prevent merchants from needing to add the Cardinal credentials to their gradle?

Unfortunately, the credentials are still needed. But, I think this is a step in the direction of being able to remove those credentials for the merchant.

PR #341, opened 2021-03-18
ThreeDSecureResult

sshropshire:
### Summary of changes

 - Add `ThreeDSecureResult` to project to mirror iOS SDK ([braintree_ios #564](https://github.braintreeps.com/team-sdk/braintree-ios/pull/564))

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop
- @sshropshire


Demo/src/main/java/com/braintreepayments/demo/CardFragment.java
```diff
@@ -330,9 +332,10 @@ private void handlePaymentMethodNonceCreated(PaymentMethodNonce paymentMethodNon
             mThreeDSecureRequested = true;
             mLoading = ProgressDialog.show(activity, getString(R.string.loading), getString(R.string.loading), true, false);
 
-            threeDSecureClient.performVerification(activity, threeDSecureRequest(paymentMethodNonce), (request, lookup, error) -> {
-                if (request != null && lookup != null) {
-                    threeDSecureClient.continuePerformVerification(activity, request, lookup, this::handleThreeDSecureResult);
+            ThreeDSecureRequest threeDSecureRequest = threeDSecureRequest(paymentMethodNonce);
+            threeDSecureClient.performLookup(activity, threeDSecureRequest, (threeDSecureResult, error) -> {
+                if (threeDSecureResult != null) {
+                    threeDSecureClient.initiateChallengeWithLookup(activity, threeDSecureRequest, threeDSecureResult, (paymentMethodNonce1, error1) -> handleThreeDSecureResult(paymentMethodNonce1, error1));
```
sestevens, line 338:
```suggestion
                    threeDSecureClient.initiateChallengeWithLookup(activity, threeDSecureRequest, threeDSecureResult, this::handleThreeDSecureResult);
```
Any reason not to use that syntax? It's a little more concise.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
```diff
@@ -84,34 +55,35 @@ public void onResult(ThreeDSecureRequest request, ThreeDSecureLookup lookup, Exc
      *
      * @param activity Android FragmentActivity
      * @param request  the {@link ThreeDSecureRequest} with information used for authentication.
+     * @param callback {@link ThreeDSecureResultCallback}
      */
-    public void performVerification(final FragmentActivity activity, final ThreeDSecureRequest request, final ThreeDSecureLookupCallback callback) {
+    public void performLookup(final FragmentActivity activity, final ThreeDSecureRequest request, final ThreeDSecureResultCallback callback) {
```
sestevens, line 60: How come we're changing this from `performVerification` to `performLookup`? My understanding is the lookup is just one part of the 3DS verification process.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
```diff
@@ -139,33 +111,35 @@ public void onResult(String consumerSessionId, Exception error) {
...
+     *
+     * @param activity Android FragmentActivity
+     * @param request  the {@link ThreeDSecureRequest} with information used for authentication.
+     * @param result   the {@link ThreeDSecureResult} returned for this request.
+     *                 Contains information about the 3DS verification request that will
+     *                 be invoked in this method.
+     * @param callback {@link ThreeDSecureResultCallback}
      */
-    public void continuePerformVerification(final FragmentActivity activity, final ThreeDSecureRequest request, final ThreeDSecureLookup threeDSecureLookup, final ThreeDSecureResultCallback callback) {
+    public void initiateChallengeWithLookup(final FragmentActivity activity, final ThreeDSecureRequest request, final ThreeDSecureResult result, final ThreeDSecureResultCallback callback) {
```
sestevens, line 123: Nitpick - on iOS we call this `initializeChallengeWithLookup`. Maybe we should update this to match?

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
```diff
@@ -201,126 +175,17 @@ public void onResult(@Nullable Configuration configuration, @Nullable Exception
...
     }
 
-    /**
-     * Creates a stringified JSON object containing the information necessary to perform a lookup
-     *
-     * @param context Android Context
-     * @param request  the {@link ThreeDSecureRequest} that has a nonce and an optional UI customization.
-     * @param callback {@link ThreeDSecurePrepareLookupCallback}
-     */
-    public void prepareLookup(final Context context, final ThreeDSecureRequest request, final ThreeDSecurePrepareLookupCallback callback) {
```
sestevens, line 211: Is this method being replaced with something else? I believe it's used by merchants who do server-side lookups for 3DS, so I don't think it can go away entirely.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
sestevens: Actually, I think I'm getting confused. On iOS, there's a method called `initializeChallengeWithLookup` that's used by merchants who do server-side 3DS lookups.

This looks like it's used to continue the regular 3DS flow after inspecting the lookup response. Is that right?

I guess I want to understand the name changes better. `performVerification` and `continuePeformVerification` seem to go together nicely. Is there a reason why we're changing them?

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureActivity.java
```diff
@@ -13,7 +13,7 @@
 
 public class ThreeDSecureActivity extends AppCompatActivity implements CardinalValidateReceiver {
```
sestevens, line 14: What does this class do? Could we add a doc comment? I'm confused about how this is used in relation to `ThreeDSecureClient`.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
sshropshire: This one was a bit weird for us too. We had `performVerification` and `continuePerformVerificaton` before, but we didn't see an equivalent in iOS. We went digging to see if we could find something comparable and came up with these terms. We realize that everything on iOS goes through a PaymentFlowDriver extension that has a `startPaymentFlow` method. We'd like to have some form of parity with the naming, but we'd be open to reverting to `performVerification` and `continuePerformVerification`.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
sshropshire: Right yeah there doesn't seem to be an iOS equivalent for this method. We figured if it's requested we can always add back and potentially add it to iOS?

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureActivity.java
sshropshire: I guess we could make a doc string? It's the intermediary between Cardinal and the SDK when using 3DS v2 flow.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
sestevens: Yeah, it's a bit different on iOS because we use `BTPaymentFlowDriver`. So `performVerification` is equivalent to `startPaymentFlow` (with a `BTThreeDSecureRequest` being passed in`).

Merchants also have to implement `BTThreeDSecureRequestDelegate`, which has the [`onLookupComplete`](https://braintree.github.io/braintree_ios/current/Protocols/BTThreeDSecureRequestDelegate.html#/c:objc(pl)BTThreeDSecureRequestDelegate(im)onLookupComplete:lookupResult:next:) method. One of the params that we pass to the merchant in the `onLookupComplete` method is a function called `next`. The merchant calls this when they're ready to continue the 3DS flow, which is the equivalent of calling `continuePerformVerification` on Android.

sestevens: One way have more parity between platforms would be to update the callback in `performVerification` to pass the merchant a `next` function that they call instead of calling `continuePerformVerification`.

sestevens: But it might also make sense to just keep `performVerification` and `continuePerformVerification` for the sake of minimizing breaking changes for merchants.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureActivity.java
sestevens: Why is it public?

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
sarahkoop: We reverted the renaming of these methods - so they are back to `performVerification` and `continuePerformVerification`

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
sarahkoop: Good catch - we added back the `prepareLookup` and `initializeChallengeWithLookupResponse` methods and did some refactoring for code reuse.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureActivity.java
sshropshire: That's a good call. We looked into this and it turns out to be a Java / Android limitation. `ThreeDSecureActivity` needs to be publicly accessible by the Android system in order to start the activity, which in turn starts Cardinal. We do believe this can be removed in the future, once Cardinal provides us with an update where the response is returned via an Android Activity result.

sshropshire: > Could we add a doc comment? 

We're going to go ahead and add the doc strings! 🙌

ThreeDSecure/build.gradle
```diff
@@ -30,6 +30,7 @@ android {
 
 dependencies {
     api project(':BraintreeCore')
+    api project(':Card')
```
scannillo, line 33: Does this mean a merchant using 3DS, also doesn't need to add `implementation 'com.braintreepayments.api:card:4.0.0-beta1'` to their gradle?

sshropshire: Yup! Merchant can still add card to their gradle, but they aren't required to do so. Gradle will always pick the newest version.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureActivity.java
sestevens: Awesome, thanks for the clarification!

sestevens (approved):
Looks good!

ThreeDSecure/build.gradle
scannillo: Awesome!

scannillo (approved):
🍕

PR #343, opened 2021-03-22
[DRAFT] v4 Feedback

sestevens:
### Summary of changes

 - PR for feedback as we work on a Kotlin sample app using the v4 beta

 ### Checklist

- @sestevens 
- @scannillo

### DIFF
--- a/v4_MIGRATION_GUIDE.md
+++ b/v4_MIGRATION_GUIDE.md
@@ -44,6 +44,8 @@ In the `AndroidManifest.xml`, migrate the `intent-filter` from your v3 integrati
 </activity>
 ``` 
 
+FEEDBACK: `onBrowserSwitchResult` crashes if `browserSwitchResult` is null, but in Kotlin we could just prevent devs from passing in null. Maybe we can add a null-check for Java too.
+
 ## American Express
 
 The American Express feature is now supported by implementing the following dependencies:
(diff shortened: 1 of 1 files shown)


v4_MIGRATION_GUIDE.md
```diff
@@ -44,6 +44,8 @@ In the `AndroidManifest.xml`, migrate the `intent-filter` from your v3 integrati
 </activity>
 ``` 
 
+FEEDBACK: `onBrowserSwitchResult` crashes if `browserSwitchResult` is null, but in Kotlin we could just prevent devs from passing in null. Maybe we can add a null-check for Java too.
```
sarahkoop, line 47: We added null checks for these methods in #345

v4_MIGRATION_GUIDE.md
```diff
@@ -286,7 +288,7 @@ The Google Pay feature is now supported in a single dependency:
 
 ```groovy
 dependencies {
-  implementation 'com.braintreepayments.api:google-pay:4.0.0-beta1'
+  implementation 'com.braintreepayments.api:google-payment:4.0.0-beta1' // we think this should probably be google-pay, but it's deployed as google-payment
```
sarahkoop, line 291: This was fixed in #335, so the next beta release should have the correct `google-pay` package name

v4_MIGRATION_GUIDE.md
```diff
@@ -331,6 +334,7 @@ public class GooglePayActivity extends AppCompatActivity {
       .billingAddressRequired(true)
       .googleMerchantId("merchant-id-from-google");
 
+// FEEDBACK: Why do merchants need success and error? Is there ever a case where you would have success = true and a non-null error OR success = false and a null error?
```
sarahkoop, line 337: Good question - there wasn't a case where we were passing back true and no error, so we updated this callback to only have an error parameter. We opened a draft PR (#346) for this change and the other migration guide fixes (and any other future feedback)

sestevens:
Closing this - I think all the feedback has been addressed! 🙂

PR #347, opened 2021-03-24
[V3] Update Cardinal and Add PayPal to Intent Queries

sshropshire:
### Summary of changes

 - Bump Cardinal version to `2.2.5-2` in `3.x` branch (`v4` to follow)
 - Add PayPal to `queries` element in `AndroidManifest.xml`

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop
- @sshropshire

### DIFF
--- a/Braintree/src/main/AndroidManifest.xml
+++ b/Braintree/src/main/AndroidManifest.xml
@@ -7,6 +7,7 @@
 
     <queries>
         <package android:name="com.venmo" />
+        <package android:name="com.paypal.android.p2pmobile" />
     </queries>
 
     <application>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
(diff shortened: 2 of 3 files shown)


Braintree/src/main/AndroidManifest.xml
```diff
@@ -7,6 +7,7 @@
 
     <queries>
         <package android:name="com.venmo" />
+        <package android:name="com.paypal.android.p2pmobile" />
```
sestevens, line 10: Why are we adding this? I didn't think we ever switched to the PayPal app anymore.

sestevens (approved):

scannillo (approved):

Braintree/src/main/AndroidManifest.xml
sshropshire: Yeah Preferred Payments still checks if PayPal is installed similar to Venmo, so we went ahead and added them in.

sestevens: Oh, nice! Makes sense.

PR #356, opened 2021-04-07
Refactor Payment Method Nonce Parsing

sshropshire:
### Summary of changes

 - Create `PaymentMethodNonce` interface
 - Create `BraintreeNonce` type to parse nonces for `PaymentMethodClient#getPaymentMethodNonces` callback.
 - Remove inheritance from all PaymentMethodNonce types in favor of `PaymentMethodNonce` interface

### Notes
  - The original implementation overloaded the `PaymentMethodNonce` base class heavily by accounting for both `PaymentMethodClient` response JSON as well as `Tokenization` response JSON. Removing this pattern allows us to move feature-related classes out of `:BraintreeCore` and into their respective modules.

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop
- @sshropshire


sestevens:
I'm a little confused about why we need the `BraintreeNonce` base class and the `PaymentMethodNonce` interface. I understand that we want the specific nonce classes to live in the respective modules (e.g., `PayPalAccountNonce` should live in the PayPal module). But since `BraintreeNonce` and `PaymentMethodNonce` both live in Core, won't all of the other modules have access to both? Why not just use a base class instead of base class and interface?

sshropshire:
> I'm a little confused about why we need the `BraintreeNonce` base class and the `PaymentMethodNonce` interface.

It has to do with [PaymentMethodClient](https://github.com/braintree/braintree_android/blob/master/BraintreeCore/src/main/java/com/braintreepayments/api/PaymentMethodClient.java). The response JSON schema for `GET /v1/payment_methods` is different from a result you would get from tokenizing a payment method.

The original implementation has a base class and it seems like the goal was to keep the JSON parsing DRY, but as a side-effect, it creates a rigid inheritance structure. Since it's only parsing three JSON properties in the base class, an interface would give us more flexibility while keeping the same "base" type pattern.

BraintreeCore/src/test/java/com/braintreepayments/api/PaymentMethodClientUnitTest.java
```diff
@@ -0,0 +1,143 @@
...
+        PaymentMethodClient sut = new PaymentMethodClient(braintreeClient);
+
+        GetPaymentMethodNoncesCallback callback = mock(GetPaymentMethodNoncesCallback.class);
+        sut.getPaymentMethodNonces(false, callback);
+
+        verify(callback).onResult(paymentMethodNoncesCaptor.capture(), (Exception) isNull());
+
+        List<BraintreeNonce> paymentMethodNonces = paymentMethodNoncesCaptor.getValue();
+        assertEquals("Visa", paymentMethodNonces.get(0).getTypeLabel());
+        assertEquals("ending in ••11", paymentMethodNonces.get(0).getDescription());
```
sestevens, line 80: So, we had [an issue in iOS Drop-in](https://github.com/braintree/braintree-ios-drop-in/issues/188) about the `description` on payment method nonces not being localized. For cards specifically, the description that comes back from the Gateway is in English (i.e., "ending in ••11").

On iOS, we decided to remove `description` from the payment method nonce class in the base SDK, and add a description property to an extension on payment method nonces in Drop-in. https://github.com/braintree/braintree-ios-drop-in/blob/76a657e1b95f78643dfd4395f8e2c79d1c08c3b1/Sources/BraintreeDropIn/BTPaymentMethodNonce%2BDropIn.m#L23

We have some client-side logic to check the type of the nonce and use the relevant property for that type. (So for Venmo, it's the username, for PayPal it's the email address, for cards it's just the last four digits.)

I'm thinking we should do something similar in Android?

BraintreeModels/src/main/java/com/braintreepayments/api/PaymentMethodNonce.java
```diff
@@ -1,186 +1,29 @@
...
-     *
-     * @param json {@link JSONObject} representation of a {@link PaymentMethodNonce}.
-     * @param type The {@link String} type of the {@link PaymentMethodNonce}.
-     * @return {@link PaymentMethodNonce}
-     * @throws JSONException if parsing fails
-     */
-    @Nullable
-    static PaymentMethodNonce parsePaymentMethodNonces(JSONObject json, String type) throws JSONException {
-        switch (type) {
-            case CardNonce.TYPE:
```
sestevens, line 130: I'm a little concerned that by removing this logic and giving merchants just generic `BraintreeNonce` objects, there will be some info merchants will no longer be able to access on vaulted payment methods. I'm not 100% sure how much of a concern this is since it depends on what merchants are doing with those vaulted payment methods. But it might be safer to make sure that they can access all of the same information as before.

I know we want to separate out the specific nonce-types into their respective modules, so that they don't all live in Core. Would it be possible to do a run-time check for those classes in Core so that we can still parse the vaulted payment method response into the specific nonce types? We'd still need to return the base class to the merchant, but they'd be able to cast to the more specific type of nonce if needed, and all of the relevant info would be there.

sestevens:
> It has to do with [PaymentMethodClient](https://github.com/braintree/braintree_android/blob/master/BraintreeCore/src/main/java/com/braintreepayments/api/PaymentMethodClient.java). The response JSON schema for `GET /v1/payment_methods` is different from a result you would get from tokenizing a payment method.

That's interesting! So the response from `GET /v1/payment_methods` doesn't have all the properties specific to each payment method?

sshropshire:
> That's interesting! So the response from `GET /v1/payment_methods` doesn't have all the properties specific to each payment method?

Actually all the nonce properties are there, but it's the top-level property that differs. For PaymentMethodClient, you get back a list of nonces under a property called `paymentMethods`. For tokenization responses, the top-level property is related to the payment method type e.g. `creditCards` for Card, `paypalAccounts` for PayPal, etc.

There's also GraphQL tokenize responses that have a top-level `data` property, and `GooglePayNonce` [has it's own structure](https://github.com/braintree/braintree_android/blob/7dbd7d9bc030b5d0a57b6b288537257f8e285a5f/TestUtils/src/main/java/com/braintreepayments/api/Fixtures.kt#L1850) that's unlike any of the others.

BraintreeCore/src/test/java/com/braintreepayments/api/PaymentMethodClientUnitTest.java
sshropshire: Yeah I agree we definitely should do something similar to match iOS localization. That'll be super useful for custom integrations too if we handle it in the base of the sdk.

BraintreeModels/src/main/java/com/braintreepayments/api/PaymentMethodNonce.java
sshropshire: Yeah I like that idea. Right now we have [copy constructors](https://github.com/braintree/braintree_android/blob/8825a4d50ded576dd7fcefc1962bc61e600d25e2/GooglePay/src/main/java/com/braintreepayments/api/GooglePayCardNonce.java#L44) for the merchant to do a cast-like operation to access nonce-type specific information. We could leverage `compileOnly` here to take the burden off of the merchant, then they can use `instanceOf` for introspection.

BraintreeCore/src/test/java/com/braintreepayments/api/PaymentMethodClientUnitTest.java
scannillo: This can probably be completed in a separate PR, yeh?

BraintreeModels/src/main/java/com/braintreepayments/api/PaymentMethodNonce.java
```diff
@@ -1,186 +1,29 @@
...
-        return json.getJSONArray(apiResourceKey).getJSONObject(0);
-    }
-
-    @CallSuper
-    void fromJson(JSONObject json) throws JSONException {
-        mNonce = json.getString(PAYMENT_METHOD_NONCE_KEY);
-        mDescription = json.getString(DESCRIPTION_KEY);
-        mDefault = json.optBoolean(PAYMENT_METHOD_DEFAULT_KEY, false);
-    }
+public interface PaymentMethodNonce extends Parcelable {
```
scannillo, line 5: Docstring for this interface?

BraintreeModels/src/main/java/com/braintreepayments/api/TokenizeCallback.java
```diff
@@ -0,0 +1,14 @@
+package com.braintreepayments.api;
+
+
+/**
+ * Communicates JSON response from a tokenization request on the main thread.
+ */
+interface TokenizeCallback {
+
+    /**
+     * @param tokenizationResponse parsed {@link BraintreeNonce} from the HTTP request.
```
scannillo, line 10: I think this could clarified - this string is the raw JSON of the tokenization response as a string?

CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* Add `BraintreeNonce` result type for `GetPaymentMethodNoncesCallback`
```
scannillo, line 5: Nitpick / take it or leave it

```suggestion
* Replace `PaymentMethodNonce` with `BraintreeNonce` result type for `GetPaymentMethodNoncesCallback`
```

scannillo:
I'm still confused about the same question Susan asked earlier - why we need both the `PaymentMethodNonce` interface & the `BraintreeNonce` base class. I know you mentioned that "_The response JSON schema for GET /v1/payment_methods is different from a result you would get from tokenizing a payment method_", but I'm still not sure why this results in us needing both.

Can you clarify a little bit more? Thanks!

demerino:
I like the idea of moving all of the `nonce objects` to the modules and having them extend the base class.

Looking at iOS, the base class is pretty basic and doesn't include any parsing logic. Maybe following that model would be a good compromise here:

BraintreeCore/src/test/java/com/braintreepayments/api/PaymentMethodClientUnitTest.java
sarahkoop: We've added a TODO for this 👍

CHANGELOG.md
sarahkoop: Removed this line since we removed `BraintreeNonce`

BraintreeModels/src/main/java/com/braintreepayments/api/PaymentMethodNonce.java
sarahkoop: We reverted this back to a base class so the original doc strings apply

BraintreeModels/src/main/java/com/braintreepayments/api/TokenizeCallback.java
sarahkoop: We re-worded this

sshropshire:
@sestevens @scannillo @demerino thanks for all the feedback! We circled back to see how we could simplify the design to eliminate confusion with `BraintreeNonce`.

What we found (with a bit of help from our iOS implementation) is that it doesn't make sense to inherit the JSON parsing behavior in the `PaymentMethodNonce` base class. By making the nonces pure data types and moving the JSON parsing into static factory methods, we were able to achieve the compromise we needed.

We also took note of other potential enhancements that we can add in another pull request. Feel free to revisit this PR! ✌️

BraintreeModels/src/main/java/com/braintreepayments/api/PaymentMethodNonce.java
```diff
@@ -176,11 +80,42 @@ public void writeToParcel(Parcel dest, int flags) {
...
+            return new PaymentMethodNonce(in);
+        }
+
+        @Override
+        public PaymentMethodNonce[] newArray(int size) {
+            return new PaymentMethodNonce[size];
+        }
+    };
+
+    static @PaymentMethodType int paymentMethodTypeFromString(String typeString) {
```
demerino, line 105: Why not have this in the PaymentMethodType.java file?

BraintreeModels/src/main/java/com/braintreepayments/api/PaymentMethodNonce.java
```diff
@@ -2,60 +2,60 @@
...
-    public String getNonce() {
+    public String getString() {
         return mNonce;
     }
 
     /**
      * @return The description of this PaymentMethod for displaying to a customer, e.g. 'Visa ending in...'
      */
     public String getDescription() {
+        // TODO: find a way to localize this, or move to Drop-In
```
demerino, line 58: I think we want to just remove this (if we want to, we can add it back in later based on merchant feedback)

demerino (approved):
Yeah, I like this new approach.

BraintreeCore/src/main/java/com/braintreepayments/api/PaymentMethodClient.java
```diff
@@ -26,18 +31,46 @@ public PaymentMethodClient(BraintreeClient braintreeClient) {
...
+        }
+
+        List<PaymentMethodNonce> paymentMethodNonces = new ArrayList<>();
+        JSONObject json;
+        PaymentMethodNonce paymentMethodNonce;
+        for(int i = 0; i < paymentMethods.length(); i++) {
+            json = paymentMethods.getJSONObject(i);
+            // TODO: leverage compileOnly gradle dependencies to return strongly-typed nonces (e.g. CardNonce, PayPalAccountNonce etc.)
+            paymentMethodNonce = PaymentMethodNonce.fromJSON(json);
+            if (paymentMethodNonce.getType() != PaymentMethodType.GOOGLE_PAY) {
```
sestevens, line 56: Why are we filtering out Google Pay nonces?

BraintreeCore/src/main/java/com/braintreepayments/api/PaymentMethodClient.java
```diff
@@ -26,18 +31,46 @@ public PaymentMethodClient(BraintreeClient braintreeClient) {
...
     /**
      * Retrieves the current list of {@link PaymentMethodNonce}s for the current customer.
      * <p>
      * When finished, the {@link java.util.List} of {@link PaymentMethodNonce}s will be sent to {@link
      * GetPaymentMethodNoncesCallback}
-     *
-     * @param context Android Context
-     * @param defaultFirst when {@code true} the customer's default payment method will be first in the list, otherwise
+     *  @param defaultFirst when {@code true} the customer's default payment method will be first in the list, otherwise
      *        payment methods will be ordered my most recently used.
```
sestevens, line 70:
```suggestion
     *        payment methods will be ordered by most recently added.
```

Not related to this PR, but this method actually returns the most recently _added_ payment methods, not the most recently used. We had [an issue about this](https://github.com/braintree/braintree-ios-drop-in/issues/180) on iOS Drop-in.

BraintreeCore/src/main/java/com/braintreepayments/api/PaymentMethodClient.java
sshropshire: 🤷‍♂️  we kept the logic the same. Here's the [OG test](https://github.com/braintree/braintree_android/blob/23f48a492fbb7f1b900ccc0b2a44c9d7512f5add/BraintreeCore/src/test/java/com/braintreepayments/api/PaymentMethodUnitTest.java#L119).

BraintreeModels/src/main/java/com/braintreepayments/api/PaymentMethodNonce.java
sestevens: +1 to this idea. Could we even use `StringDef` instead of `IntDef` so that the enum is backed by the associated string instead of an int?

sshropshire: Yeah, I agree we should move it. The only concern with `StringDef` is it doesn't account for `PaymentMethodType.UNKNOWN`, so we'd still need to have some switching logic to convert from `String` to `PaymentMethodType`.

BraintreeCore/src/main/java/com/braintreepayments/api/PaymentMethodClient.java
sestevens: OK, sounds good!

sestevens (approved):

BraintreeModels/src/main/java/com/braintreepayments/api/PaymentMethodNonce.java
sshropshire: We tried to move it and I forgot that the compiler barked when we tried to add a static method to an interface. We may just make the method `private` for now and keep it in PaymentMethodNonce.java until we can find a more suitable solution.

BraintreeModels/src/main/java/com/braintreepayments/api/PaymentMethodNonce.java
sarahkoop: Removed this 👍

PR #364, opened 2021-04-21
Venmo Payment Context

sshropshire:
### Summary of changes

 - Use payment context in Venmo flow

 ### Checklist

 - [X] Added a changelog entry

- @sarahkoop
- @sshropshire

### DIFF
--- a/Venmo/src/main/java/com/braintreepayments/api/VenmoAccountNonce.java
+++ b/Venmo/src/main/java/com/braintreepayments/api/VenmoAccountNonce.java
@@ -19,6 +19,9 @@ public class VenmoAccountNonce extends PaymentMethodNonce {
     private static final String VENMO_DETAILS_KEY = "details";
     private static final String VENMO_USERNAME_KEY = "username";
 
+    private static final String VENMO_PAYMENT_METHOD_ID_KEY = "paymentMethodId";
+    private static final String VENMO_PAYMENT_METHOD_USERNAME_KEY = "userName";
+    
     private final String username;
 
     static VenmoAccountNonce fromJSON(JSONObject inputJson) throws JSONException {
@@ -29,11 +32,22 @@ public class VenmoAccountNonce extends PaymentMethodNonce {
             json = inputJson;
         }
 
-        String nonce = json.getString(PAYMENT_METHOD_NONCE_KEY);
-        boolean isDefault = json.optBoolean(PAYMENT_METHOD_DEFAULT_KEY, false);
+        String nonce;
+        boolean isDefault;
+        String username;
 
-        JSONObject details = json.getJSONObject(VENMO_DETAILS_KEY);
-        String username = details.getString(VENMO_USERNAME_KEY);
+        if (json.has(VENMO_PAYMENT_METHOD_ID_KEY)) {
+            isDefault = false;
+            nonce = json.getString(VENMO_PAYMENT_METHOD_ID_KEY);
+            username = json.getString(VENMO_PAYMENT_METHOD_USERNAME_KEY);
+
+        } else {
+            nonce = json.getString(PAYMENT_METHOD_NONCE_KEY);
+            isDefault = json.optBoolean(PAYMENT_METHOD_DEFAULT_KEY, false);
+
+            JSONObject details = json.getJSONObject(VENMO_DETAILS_KEY);
+            username = details.getString(VENMO_USERNAME_KEY);
+        }
 
         return new VenmoAccountNonce(nonce, username, isDefault);
     }
--- a/Venmo/src/main/java/com/braintreepayments/api/VenmoAccountNonce.java
+++ b/Venmo/src/main/java/com/braintreepayments/api/VenmoAccountNonce.java
@@ -40,7 +40,6 @@ public class VenmoAccountNonce extends PaymentMethodNonce {
             isDefault = false;
             nonce = json.getString(VENMO_PAYMENT_METHOD_ID_KEY);
             username = json.getString(VENMO_PAYMENT_METHOD_USERNAME_KEY);
-
         } else {
             nonce = json.getString(PAYMENT_METHOD_NONCE_KEY);
             isDefault = json.optBoolean(PAYMENT_METHOD_DEFAULT_KEY, false);
--- a/Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
+++ b/Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
@@ -5,7 +5,6 @@ import android.content.Context;
 import android.content.Intent;
 import android.net.Uri;
 import android.text.TextUtils;
-import android.util.Log;
 
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
--- a/Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
+++ b/Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
@@ -94,6 +94,7 @@ public class VenmoClient {
                 }
 
                 if (request.getPaymentMethodUsage() == VenmoPaymentMethodUsage.UNSPECIFIED) {
+                    braintreeClient.sendAnalyticsEvent("pay-with-venmo.app-switch.failed");
                     callback.onResult(new BraintreeException("Payment method usage must be set."));
                 }
 
@@ -123,6 +124,7 @@ public class VenmoClient {
                     public void success(String responseBody) {
                         String paymentContextId = parsePaymentContextId(responseBody);
                         if (paymentContextId == null || TextUtils.isEmpty(paymentContextId)) {
+                            braintreeClient.sendAnalyticsEvent("pay-with-venmo.app-switch.failed");
                             callback.onResult(new BraintreeException("Failed to fetch a Venmo paymentContextId while constructing the requestURL."));
                             return;
                         }
@@ -131,6 +133,7 @@ public class VenmoClient {
 
                     @Override
                     public void failure(Exception exception) {
+                        braintreeClient.sendAnalyticsEvent("pay-with-venmo.app-switch.failed");
                         callback.onResult(exception);
                     }
                 });
@@ -167,6 +170,7 @@ public class VenmoClient {
      */
     public void onActivityResult(final Context context, int resultCode, Intent data, final VenmoOnActivityResultCallback callback) {
         if (resultCode == AppCompatActivity.RESULT_OK) {
+            braintreeClient.sendAnalyticsEvent("pay-with-venmo.app-switch.success");
             String paymentContextId = sharedPrefsWriter.getVenmoPaymentContextId(context);
             if (paymentContextId != null) {
                 JSONObject params = new JSONObject();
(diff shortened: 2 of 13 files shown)


Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -91,59 +93,141 @@ public void onResult(@Nullable Configuration configuration, @Nullable Exception
                     return;
                 }
 
-                sharedPrefsWriter.persistVenmoVaultOption(activity, request.getShouldVault() && braintreeClient.getAuthorization() instanceof ClientToken);
+                if (request.getPaymentMethodUsage() == VenmoPaymentMethodUsage.UNSPECIFIED) {
+                    braintreeClient.sendAnalyticsEvent("pay-with-venmo.app-switch.failed");
+                    callback.onResult(new BraintreeException("Payment method usage must be set."));
```
sestevens, line 98: Is there any reason that we're throwing an exception instead of just requiring this to be passed in when constructing a `VenmoRequest`? It seems like that would allow us to remove the `UNSPECIFIED` case from the enum and maybe alleviate some merchant confusion.

Venmo/src/main/java/com/braintreepayments/api/VenmoAccountNonce.java
```diff
@@ -31,11 +34,22 @@ static VenmoAccountNonce fromJSON(JSONObject inputJson) throws JSONException {
...
+        boolean isDefault;
+        String username;
 
-        JSONObject details = json.getJSONObject(VENMO_DETAILS_KEY);
-        String username = details.getString(VENMO_USERNAME_KEY);
+        if (json.has(VENMO_PAYMENT_METHOD_ID_KEY)) {
+            isDefault = false;
+            nonce = json.getString(VENMO_PAYMENT_METHOD_ID_KEY);
+            username = json.getString(VENMO_PAYMENT_METHOD_USERNAME_KEY);
+
```
sestevens, line 45:
```suggestion
```

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -91,59 +93,141 @@ public void onResult(@Nullable Configuration configuration, @Nullable Exception
...
+                } catch (JSONException e) {
+                    callback.onResult(new BraintreeException("unexpected error"));
+                }
+
+                final String finalVenmoProfileId = venmoProfileId;
+                braintreeClient.sendGraphQLPOST(params.toString(), new HttpResponseCallback() {
+                    @Override
+                    public void success(String responseBody) {
+                        String paymentContextId = parsePaymentContextId(responseBody);
+                        if (paymentContextId == null || TextUtils.isEmpty(paymentContextId)) {
```
sestevens, line 126:
```suggestion
                        if (TextUtils.isEmpty(paymentContextId)) {
```
I believe `TextUtils.isEmpty` will handle the null check.

Venmo/src/main/java/com/braintreepayments/api/VenmoPaymentMethodUsage.java
```diff
@@ -0,0 +1,21 @@
+package com.braintreepayments.api;
+
+import androidx.annotation.IntDef;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/**
+ * Enum representing different types of supported Braintree payment methods.
```
sestevens, line 9: This makes it sound like this enum will contain payment methods like PayPal, Venmo, Card, etc. On iOS, the equivalent doc string says "Usage type for the tokenized Venmo account". Maybe we could do the same here?

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
sarahkoop: I think this was the requested behavior on the Android side - possibly to align the requests with iOS? But I agree it would make it easier on merchants to require `paymentMethodUsage` in the request. @demerino Is `UNSPECIFIED` still required here or can we simplify this?

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
sarahkoop: Yes good call

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
sestevens: I'm fairly certain this is a case where it's OK for iOS and Android to differ! We can add a NEXT_MAJOR_VERSION note on iOS to align with Android later on.

sshropshire: Went ahead and removed `UNSPECIFIED`!

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -171,11 +254,35 @@ private Intent getLaunchIntent(Configuration configuration, String profileId) {
...
+        }
 
         return venmoIntent;
     }
 
+    private void vaultVenmoAccountNonce(String nonce, final VenmoOnActivityResultCallback callback) {
+        VenmoAccount venmoAccount = new VenmoAccount();
+        venmoAccount.setNonce(nonce);
+
+        tokenizationClient.tokenizeREST(venmoAccount, new TokenizeCallback() {
```
sestevens, line 267: It's kind of confusing to me that we're calling `tokenizationClient.tokenizeREST` here since we're vaulting, not tokenizing. Would it make sense to call this something else?

sestevens: I guess it is still giving back a nonce, so maybe it makes sense?

Venmo/src/main/java/com/braintreepayments/api/VenmoRequest.java
```diff
@@ -8,6 +8,12 @@
     private boolean shouldVault;
     private String profileId;
 
+    private @VenmoPaymentMethodUsage int paymentMethodUsage;
+
+    public VenmoRequest(@VenmoPaymentMethodUsage int paymentMethodUsage) {
```
sestevens, line 13: Should we add a doc comment here?

Venmo/src/main/java/com/braintreepayments/api/VenmoRequest.java
```diff
@@ -41,5 +47,31 @@ public boolean getShouldVault() {
...
+            default:
+                return null;
+        }
+    }
+
+    /**
+     * Set {@link VenmoPaymentMethodUsage} for the tokenized Venmo account: either multi-use or single use.
+     * Defaults to {@link VenmoPaymentMethodUsage#MULTI_USE}
+     */
+    public void setPaymentMethodUsage(@VenmoPaymentMethodUsage int paymentMethodUsage) {
```
sestevens, line 73: We can remove this now that we have it in the constructor, right?

Venmo/src/test/java/com/braintreepayments/api/VenmoClientUnitTest.java
```diff
@@ -93,15 +98,96 @@ public void showVenmoInGooglePlayStore_sendsAnalyticsEvent() {
...
+        JSONObject variables = graphQLJSON.getJSONObject("variables");
+        JSONObject input = variables.getJSONObject("input");
+        assertEquals("SINGLE_USE", input.getString("paymentMethodUsage"));
+        assertEquals("sample-venmo-merchant", input.getString("merchantProfileId"));
+        assertEquals("MOBILE_APP", input.getString("customerClient"));
+        assertEquals("CONTINUE", input.getString("intent"));
+    }
+
+    @Test
+    public void tokenizeVenmoAccount_whenCreatePaymentContextSucceeds_persitsPaymentContextIdThenStartsVenmoActivityAndSendsAnalytics() throws JSONException {
```
sestevens, line 137: Is this still right? I thought we were no longer persisting the `paymentContextId`, since we're checking for it in the return URL instead.

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
sshropshire: Yeah it was `TokenizationClient#tokenize()` in `v3` we kind of just went with it. We split it into two methods, `tokenizeRest` and `tokenizeGraphQL` to get rid of a dependency on `CardNonce` which is in the `Card` module now.

sshropshire: `CardClient` now makes the decision on whether or not to use GraphQL.

Venmo/src/test/java/com/braintreepayments/api/VenmoClientUnitTest.java
sshropshire: Right yeah good call. We updated the code and not the test name. We can take that out.

Venmo/src/main/java/com/braintreepayments/api/VenmoRequest.java
sshropshire: We technically don't need the getter either may take that out for now since we can always add back.

sshropshire: Actually will keep there are good docs on that property.

Venmo/src/main/java/com/braintreepayments/api/VenmoRequest.java
```diff
@@ -8,6 +8,17 @@
     private boolean shouldVault;
     private String profileId;
 
+    private final @VenmoPaymentMethodUsage int paymentMethodUsage;
+
+    /**
+     * Request to tokenize an existing Venmo account.
```
sestevens, line 14:
```suggestion
     * Request to tokenize a Venmo account.
```

Do we need the word "existing" here?

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
sestevens: OK, so you don't think anyone will be confused in the future that `tokenizeREST` is actually vaulting in this case?

Venmo/src/main/java/com/braintreepayments/api/VenmoRequest.java
sshropshire: Nope! Agreed here it's a bit verbose.

demerino (approved):

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
sshropshire: I think so yeah we could re-name. I think `TokenizationClient` could be re-named here to `ApiClient` as well to mirror iOS more. Would `ApiClient#vault()` make sense?

sestevens: Sure, that would make sense to me!

sestevens (approved):
Looks good!

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
sestevens: I'm OK with making that change in a separate PR if that's easier -- I know this PR needs to go out soon.

sshropshire: Made a ticket!  🙌

PR #365, opened 2021-04-22
Spike -- Adding PayPal Checkout SDK To Braintree SDK

CodyEngel:
### Summary of changes
Updated PayPalClient to launch the Checkout SDK for single transactions while retaining the current functionality for billing agreements. Looking forward to feedback to understand what is left to formalize this integration 🙏

 - Updated PayPalClient to launch Checkout experience.
 - Updated PayPalClient to tokenize the request similar to how the web experience works today.

 ### Checklist

 - [ ] Added a changelog entry

- @CodyEngel

### DIFF
--- a/PayPal/build.gradle
+++ b/PayPal/build.gradle
@@ -32,6 +32,10 @@ dependencies {
     implementation 'androidx.appcompat:appcompat:1.2.0'
     implementation project(':BraintreeDataCollector')
 
+    implementation('com.paypal.checkout:android-sdk:0.1.0') {
+        exclude group: 'com.paypal.android.sdk'
+    }
+
     testImplementation 'org.robolectric:robolectric:4.1'
     testImplementation 'org.powermock:powermock-module-junit4:1.6.6'
     testImplementation 'org.powermock:powermock-module-junit4-rule:1.6.6'
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
@@ -1,14 +1,27 @@
 package com.braintreepayments.api;
 
+import android.content.Context;
+import android.content.Intent;
 import android.net.Uri;
 import android.text.TextUtils;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
 import androidx.fragment.app.FragmentActivity;
 
+import com.paypal.checkout.PayPalCheckout;
+import com.paypal.checkout.approve.Approval;
+import com.paypal.checkout.approve.OnApprove;
+import com.paypal.checkout.config.CheckoutConfig;
+import com.paypal.checkout.config.Environment;
+import com.paypal.checkout.createorder.CreateOrder;
+import com.paypal.checkout.createorder.CreateOrderActions;
+
 import org.json.JSONException;
 import org.json.JSONObject;
+
+import java.util.Locale;
 
 /**
  * Used to tokenize PayPal accounts. For more information see the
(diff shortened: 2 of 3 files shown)


PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
```diff
@@ -113,12 +126,33 @@ public void onResult(@Nullable final Configuration configuration, @Nullable Exce
...
+                PayPalCheckout.setConfig(checkoutConfig);
                 sendPayPalRequest(activity, payPalCheckoutRequest, callback);
             }
         });
 
     }
 
+    private String getPayPalReturnUrl(Context context) {
+        if (context != null) {
+            return String.format("%s://paypalpay", context.getPackageName().toLowerCase(Locale.ROOT));
```
sshropshire, line 151: We do [strip out underscores](https://github.com/braintree/braintree_android/blob/eb04b3794b7fba185f4dd9a34a8c7242934f09a1/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java#L152) when performing a similar operation in `BraintreeClient`, though I can't remember exactly why. Couldn't hurt to do the same here for consitency.

PayPal/build.gradle
```diff
@@ -32,6 +32,10 @@ dependencies {
     implementation 'androidx.appcompat:appcompat:1.2.0'
     implementation project(':BraintreeDataCollector')
 
+    implementation('com.paypal.checkout:android-sdk:0.1.0') {
+        exclude group: 'com.paypal.android.sdk'
```
sshropshire, line 36: General knowledge: Is the base PayPal Android SDK not needed for native checkout?

sarahkoop:
This looks great! A few things we noticed when testing this out:
- When testing with the `SANDBOX_TOKENIZATION_KEY` found in `Settings.java` we get a crash due to    `kotlin.UninitializedPropertyAccessException: lateinit property accessToken has not been initialized`. This tokenization key is for an account not setup for this, but there should probably be some graceful error handling here.
- It doesn't seem like there is currently any way to be notified of user cancelation, either from the payment flow or from exiting out of the error screen. How should that be handled? 
- Maybe outside the scope of this, but wondering about accessibility of the payment modal - Should there be a visible cancel button in addition to clickable gray area?

PayPal/build.gradle
CodyEngel: Braintree and the Checkout Android SDK have a dependency on magnes so this resolves the duplicate dependency, we could probably exclude on the artifact id as well though.

PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
CodyEngel: Gotcha good to know, I wonder if that was added to guard against package names not matching up with valid application IDs?

I wonder if there are other edge cases that could come up by creating the returnUrl this way (thinking about if an application id doesn't match up with a package). Can definitely dig into this a bit more, stripping the underscores sounds reasonable though 👍

CodyEngel:
Hey @sarahkoop thanks for the feedback.
> When testing with the SANDBOX_TOKENIZATION_KEY found in Settings.java we get a crash due to kotlin.UninitializedPropertyAccessException: lateinit property accessToken has not been initialized. This tokenization key is for an account not setup for this, but there should probably be some graceful error handling here.

Great catch and agree that should be handled more gracefully. I'd have to check but I'm guessing we wouldn't get a client ID back if the tokenization key isn't setup for PayPal so that could be a spot where we could fail more gracefully. **For the Braintree SDK, what would that look like?** The Checkout SDK will throw exceptions for integration errors that would crash an application so it's apparent to the developer they are missing a step. Not sure if that'd be appropriate here as well? We could provide a more meaningful exception.

> It doesn't seem like there is currently any way to be notified of user cancelation, either from the payment flow or from exiting out of the error screen. How should that be handled?

Good question the SDK has an `OnCancel` and `OnError` callback so we can hook those into the Braintree SDK as well. Would be interested in what behavior would make sense for Braintree when it comes to those specific events.

> Maybe outside the scope of this, but wondering about accessibility of the payment modal - Should there be a visible cancel button in addition to clickable gray area?

I know some folks on the native experience side of the SDK have been work on an accessibility audit. I am not sure where they are at with it, but I'll get more info. Making it easy to cancel out the pay sheet with a screen reader makes a lot of sense.

PR #368, opened 2021-04-23
Add Card Autofill to Demo App

sarahkoop:
### Summary of changes

 - Add autofill button to `CardFragment`
 - Upgrade demo app to use the latest `card-form`

 ### Checklist

- @sarahkoop

### DIFF
--- a/Demo/src/androidTest/java/com/braintreepayments/demo/test/CardTest.java
+++ b/Demo/src/androidTest/java/com/braintreepayments/demo/test/CardTest.java
@@ -47,7 +47,7 @@ public class CardTest extends TestHelper {
                 .commit();
 
         onDevice(withText("Card Number")).perform(setText("4111111111111111"));
-        onDevice(withText("Expiration Date")).perform(setText("01/27"));
+        onDevice(withText("Expiration Date")).perform(setText(validExpirationText()));
         onDevice(withText("CVV")).perform(setText("123"));
         onDevice(withText("Postal Code")).perform(setText("12345"));
         onDevice(withText("Purchase")).perform(click());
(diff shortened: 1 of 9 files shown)


Demo/src/androidTest/java/com/braintreepayments/demo/test/CardTest.java
```diff
@@ -47,7 +47,7 @@ public void tokenizesACard() {
                 .commit();
 
         onDevice(withText("Card Number")).perform(setText("4111111111111111"));
-        fillInExpiration();
+        onDevice(withText("Expiration Date")).perform(setText("01/27"));
```
sarahkoop, line 50: I opted to hard code a future expiration date rather than change the helper method used to generate a valid expiration year. That helper method is used elsewhere, and I think there is some value in seeing the test inputs when scanning the file. Open to other thoughts on this.

sshropshire: I'm indifferent. We could take the middle ground route and create a method like `ExpirationDateHelper.validExpirationText()`. Then you can call
```
onDevice(withText("Expiration Date")).perform(setText(validExpirationText()));
```
after using a static import.

Demo/src/main/java/com/braintreepayments/demo/AutofillHelper.java
```diff
@@ -0,0 +1,63 @@
+package com.braintreepayments.demo;
+
+import android.text.Editable;
+
+import com.braintreepayments.cardform.view.CardForm;
+
+public class AutofillHelper {
```
sshropshire, line 7: 🔥

sshropshire (approved):

Demo/src/androidTest/java/com/braintreepayments/demo/test/CardTest.java
scannillo: Also indifferent on this! Steven's idea is like what we're doing on iOS, FWIW.

scannillo (approved):

PR #370, opened 2021-04-27
Refactor Payment Method Build JSON

sshropshire:
### Summary of changes

 - Refactor `PaymentMethod` base class to remove JSON building inheritance and `buildGraphQL` method that was only used by `Card`.

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop
- @sshropshire

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/TokenizationClient.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/TokenizationClient.java
@@ -59,22 +59,27 @@ class TokenizationClient {
                 TokenizationClient.PAYMENT_METHOD_ENDPOINT + "/" + paymentMethod.getApiPath());
 
         paymentMethod.setSessionId(braintreeClient.getSessionId());
-        braintreeClient.sendPOST(url, paymentMethod.buildJSON().toString(), new HttpResponseCallback() {
 
-            @Override
-            public void success(String responseBody) {
-                try {
-                    callback.onResult(new JSONObject(responseBody), null);
-                } catch (JSONException exception) {
+        try {
+            braintreeClient.sendPOST(url, paymentMethod.buildJSON().toString(), new HttpResponseCallback() {
+
+                @Override
+                public void success(String responseBody) {
+                    try {
+                        callback.onResult(new JSONObject(responseBody), null);
+                    } catch (JSONException exception) {
+                        callback.onResult(null, exception);
+                    }
+                }
+
+                @Override
+                public void failure(Exception exception) {
                     callback.onResult(null, exception);
                 }
-            }
-
-            @Override
-            public void failure(Exception exception) {
-                callback.onResult(null, exception);
-            }
-        });
+            });
+        } catch (JSONException exception) {
+            callback.onResult(null, exception);
+        }
     }
 
     static String versionedPath(String path) {
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/TokenizationClient.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/TokenizationClient.java
@@ -98,7 +98,7 @@ class TokenizationClient {
         String url = TokenizationClient.versionedPath(
                 TokenizationClient.PAYMENT_METHOD_ENDPOINT + "/" + paymentMethod.getApiPath());
 
-        braintreeClient.sendPOST(url, paymentMethod.buildJSON(), new HttpResponseCallback() {
+        braintreeClient.sendPOST(url, paymentMethod.buildTokenizationJSON().toString(), new HttpResponseCallback() {
 
             @Override
             public void success(String responseBody) {
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/TokenizationClient.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/TokenizationClient.java
@@ -66,15 +66,15 @@ class TokenizationClient {
 
     private static void tokenizeGraphQL(final BraintreeClient braintreeClient, final GraphQLTokenizable graphQLTokenizable, final TokenizeCallback callback) {
         braintreeClient.sendAnalyticsEvent("card.graphql.tokenization.started");
-        final String payload;
+        final JSONObject payload;
         try {
-            payload = graphQLTokenizable.buildGraphQL(braintreeClient.getAuthorization());
+            payload = graphQLTokenizable.buildGraphQLTokenizationJSON();
         } catch (BraintreeException e) {
             callback.onResult(null, e);
             return;
         }
 
-        braintreeClient.sendGraphQLPOST(payload, new HttpResponseCallback() {
+        braintreeClient.sendGraphQLPOST(payload.toString(), new HttpResponseCallback() {
             @Override
             public void success(String responseBody) {
                 try {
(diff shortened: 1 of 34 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/TokenizationClient.java
```diff
@@ -25,56 +22,14 @@
...
-        braintreeClient.getConfiguration(new ConfigurationCallback() {
-            @Override
-            public void onResult(@Nullable Configuration configuration, @Nullable Exception error) {
-                if (configuration != null) {
-                    if (paymentMethod instanceof GraphQLTokenizable &&
-                            configuration.isGraphQLFeatureEnabled(Features.TOKENIZE_CREDIT_CARDS)) {
-                        tokenizeGraphQL(braintreeClient, (GraphQLTokenizable) paymentMethod, callback);
-                    } else {
-                        tokenizeRest(braintreeClient, paymentMethod, callback);
-                    }
```
sshropshire, line 59: We decided to move this logic to `CardClient` since `Card` is technically the only feature using GraphQL at the moment. This also allowed us to make the `buildJSON` method `package-private`.

BraintreeModels/src/main/java/com/braintreepayments/api/PaymentMethod.java
```diff
@@ -46,120 +43,49 @@ void setSource(String source) {
...
     }
 
-    protected abstract void buildJSON(JSONObject base, JSONObject paymentMethodNonceJson) throws JSONException;
-
-    protected abstract void buildGraphQL(JSONObject base, JSONObject input) throws BraintreeException,
-            JSONException;
-
-    public abstract String getApiPath();
-
-    public abstract String getResponsePaymentMethodType();
```
sshropshire, line 164: This method was not being used.

PayPal/src/main/java/com/braintreepayments/api/PayPalAccount.java
```diff
@@ -36,6 +63,16 @@ void setClientMetadataId(String clientMetadataId) {
...
 
+    /**
+     * @param shouldValidate Flag to denote when the associated {@link PaymentMethodNonce}
+     *                       will be validated. When set to {@code true}, the {@link PaymentMethodNonce}
+     *                       will be validated immediately. When {@code false}, the {@link PaymentMethodNonce}
+     *                       will be validated when used by a server side library for a Braintree gateway action.
+     */
+    public void setShouldValidate(boolean shouldValidate) {
+        this.shouldValidate = shouldValidate;
+    }
```
sshropshire, line 74: Maybe we could remove these for all classes except `Card`?

We kept `shouldValidate` for non Credit Card types, however, we aren't sure if these are needed. In iOS, the `shouldValidate` property is only defined for `BTCard` and it's omitted for the other payment method types.

sshropshire: It is being set to `false` for PayPalCheckoutRequests internally to prevent vaulting. This may be another solution to prevent the merchant from having to call `shouldValidate`.

Venmo/src/main/java/com/braintreepayments/api/VenmoAccount.java
```diff
@@ -9,29 +9,40 @@
...
-    protected void buildGraphQL(JSONObject base, JSONObject input) {}
+    /**
+     * @param shouldValidate Flag to denote when the associated {@link PaymentMethodNonce}
+     *                       will be validated. When set to {@code true}, the {@link PaymentMethodNonce}
+     *                       will be validated immediately. When {@code false}, the {@link PaymentMethodNonce}
+     *                       will be validated when used by a server side library for a Braintree gateway action.
+     */
+    public void setShouldValidate(boolean shouldValidate) {
+        this.shouldValidate = shouldValidate;
+    }
```
sshropshire, line 38: Same here. We don't pass `validate = true` when [tokenizing Venmo](https://github.com/braintree/braintree_ios/blob/7ccda0f255230b8d427258766cb8041ba5ead5a4/Sources/BraintreeVenmo/BTVenmoDriver.m#L178) on iOS.

Card/src/main/java/com/braintreepayments/api/Card.java
```diff
@@ -70,6 +86,16 @@ public void setMerchantAccountId(String id) {
         merchantAccountId = TextUtils.isEmpty(id) ? null : id;
     }
 
+    /**
+     * @param shouldValidate Flag to denote when the associated {@link PaymentMethodNonce}
+     *                       will be validated. When set to {@code true}, the {@link PaymentMethodNonce}
+     *                       will be validated immediately. When {@code false}, the {@link PaymentMethodNonce}
+     *                       will be validated when used by a server side library for a Braintree gateway action. Defaults to false.
+     */
```
scannillo, line 94: I think can align our docs comments for `shouldValidate` across SDKs better (see [iOS](https://github.com/braintree/braintree_ios/blob/354475775cc745c9e571cd154437e9812cac5266/Sources/BraintreeCard/Public/BraintreeCard/BTCard.h#L110-L115)).

In general I think the `validate` concept is confusing and I'm not sure which docstring is more accurate - but wanted to raise the concern.

sshropshire: Good callout 🙏 . Added an additional note in the docs similar to iOS.

CHANGELOG.md
```diff
@@ -35,6 +38,18 @@
   * Remove `PaymentMethodNoncesCallback`
   * Remove `PaymentMethodNonce#getDescription()` method
   * `BraintreeClient` constructor no longer throws `InvalidArgumentException`
+  * Make protected static member variables `OPTIONS_KEY`, `OPERATION_NAME_KEY` on `PaymentMethod` package-private
+  * Make `PaymentMethod` constructor package-private
+  * Remove `setValidate` method from `PaymentMethod` base class and all subclasses
```
sestevens, line 43: Would it make sense to combine these lines:
```
* Add `Card#setShouldValidate()` method
* Add `PayPalAccount#setShouldValidate()` method
* Add `VenmoAccount#setShouldValidate()` method
```
with this one?

For someone trying to understand how this change impacts a merchant integration, it might help to keep all of this together. Something like:

```
* Rename `setValidate` to `setShouldValidate` and move it from `PaymentMethod` base class to `Card`, `PayPalAccount` and `VenmoAccount` subclasses
```

BraintreeModels/src/main/java/com/braintreepayments/api/PaymentMethod.java
sestevens: Is it something that merchants might have used? Or maybe Drop-in?

sarahkoop: It's not used by Drop-in, and it's not currently documented anywhere in the Braintree Dev Docs. We figured that it was safe to make package-private for now (and later potentially remove it without a breaking change), and it could be added back if any merchants do need it.

CHANGELOG.md
sarahkoop: Good idea - updated the CHANGELOG with that single entry

PayPal/src/main/java/com/braintreepayments/api/PayPalAccount.java
sarahkoop: We went ahead and removed these `setShouldValidate` methods and aligned the behavior with iOS. @demerino does this seem correct?

demerino (approved):
Makes sense, does the `setValidate` -> `setShouldValidate` need to go into the migration guide?

sestevens (approved):

PR #374, opened 2021-04-30
Braintree client instantiate without throws

sarahkoop:
### Summary of changes

 - `BraintreeClient` constructor no longer throws `InvalidArgumentException`
 - Create `InvalidToken` class to handle invalid authorizations
 - Throw/callback errors from http clients when `InvalidToken` passed in

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire 
- @sarahkoop


BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
```diff
@@ -50,7 +50,7 @@ private static BraintreeClientParams createDefaultParams(Context context, String
      * @param authorization The tokenization key or client token to use.
      * @throws InvalidArgumentException If the tokenization key or client token is not valid, or cannot be parsed.
```
scannillo, line 51: Docstring needs updating

scannillo:
Some questions:

I was thinking - how come this PR is creating a new `InvalidToken` class and checking if the auth is an instance of it instead of just having an `Authorization.isValid` property?

But then I was curious more broadly - what is the motivation for our removing the `throws InvalidArgumentException` entirely from the `BraintreeClient` creation?

sshropshire:
> Some questions:
> 
> I was thinking - how come this PR is creating a new `InvalidToken` class and checking if the auth is an instance of it instead of just having an `Authorization.isValid` property?
> 
> But then I was curious more broadly - what is the motivation for our removing the `throws InvalidArgumentException` entirely from the `BraintreeClient` creation?

@scannillo good questions! Mainly to reduce friction when instantiating a new `BraintreeClient`. The `try/catch` logic is cumbersome in Java. We were looking at it from a docs perspective and have wanted to remove `throws` from the constructor for a while. The merchant will still be notified of an invalid token via `onResult` callbacks.

This also aligns more with Kotlin since there is no try/catch necessary in that case. We looked at `isValid` as a possibility but the `instanceof` pattern was already present when checking if auth is `TokenizationKey` or `ClientToken`. Even when `invalid = false` we would have to cast to `InvalidToken` to access the error message. The `instanceof`/cast pattern felt more natural in this case.

sestevens:
Another option would be to make the `BraintreeClient` initializer nullable and log out a message if the credentials are invalid, which is [what the iOS SDK does](https://github.com/braintree/braintree_ios/blob/354475775cc745c9e571cd154437e9812cac5266/Sources/BraintreeCore/BTAPIClient.m#L41). But it doesn't look like we typically log error messages in the Android SDK, so maybe that doesn't make as much sense here.

BraintreeCore/src/test/java/com/braintreepayments/api/BraintreeClientUnitTest.java
```diff
@@ -62,42 +62,20 @@ public void beforeEach() {
 
     @Test
     public void constructor_whenAuthStringIsValidTokenizationKey_shouldNotThrow() {
```
sestevens, line 64: I don't really get what these three tests are doing any more. The `BraintreeClient` constructor never throws or returns null, right? So maybe these can be removed? Or changed to do something more useful?

BraintreeCore/src/test/java/com/braintreepayments/api/BraintreeGraphQLHttpClientUnitTest.java
```diff
@@ -96,4 +96,46 @@ public void post_withPathAndDataAndConfiguration_sendsHttpRequest() throws Excep
...
+
+        ArgumentCaptor<BraintreeException> captor = ArgumentCaptor.forClass(BraintreeException.class);
+        verify(httpResponseCallback).failure(captor.capture());
+
+        BraintreeException exception = captor.getValue();
+        assertEquals("token invalid", exception.getMessage());
+    }
+
+    @Test
+    public void post_withPathAndDataAndConfiguration_withInvalidToken_throwsBraintreeException() throws Exception {
```
sestevens, line 129:
```suggestion
    public void post_withPathAndDataAndConfiguration_withInvalidToken_throwsBraintreeException() {
```
It looks like that might not be needed since we're catching the exception in the test?

BraintreeModels/src/main/java/com/braintreepayments/api/Authorization.java
```diff
@@ -1,15 +1,13 @@
...
-import android.os.Parcel;
-import android.os.Parcelable;
 import android.text.TextUtils;
 
 import androidx.annotation.Nullable;
 
 /**
  * Generic base class for Braintree authorization
  */
-abstract class Authorization implements Parcelable {
```
scannillo, line 12: Is removing `Parcelable` part of the removing the `throws` from the `BraintreeClient` constructor? Or just an additional thing we wanted to do that is in this PR?

BraintreeCore/src/test/java/com/braintreepayments/api/BraintreeClientUnitTest.java
sarahkoop: Good call - Removed these tests since the valid auth construction is already covered in `AuthorizationUnitTest`

BraintreeModels/src/main/java/com/braintreepayments/api/InvalidToken.java
```diff
@@ -0,0 +1,25 @@
+package com.braintreepayments.api;
+
+class InvalidToken extends Authorization {
```
scannillo, line 3:
```suggestion
class InvalidAuthorization extends Authorization {
```

Take it or leave it comment - renaming this since we don't really use `token` to mean `auth` consistently.

BraintreeCore/src/test/java/com/braintreepayments/api/BraintreeGraphQLHttpClientUnitTest.java
sarahkoop: Good catch - Removed the unnecessary throws in this test file 👍

BraintreeModels/src/main/java/com/braintreepayments/api/InvalidToken.java
sarahkoop: We renamed this to `InvalidAuthorization` 👍

BraintreeModels/src/main/java/com/braintreepayments/api/Authorization.java
sarahkoop: When we created the new `InvalidAuthorization` class that extends from `Authorization` we realized that `Parcelable` wasn't needed since these classes are no longer merchant-facing. Instead of adding it unnecessarily to the new class, we decided to remove it from `Authorization` all together. We can add a note to the PR description about this for clarity.

scannillo: Gotcha - thanks for clarifying!

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
```diff
@@ -48,9 +48,8 @@ private static BraintreeClientParams createDefaultParams(Context context, String
      *
      * @param context       Android Context
      * @param authorization The tokenization key or client token to use.
-     * @throws InvalidArgumentException If the tokenization key or client token is not valid, or cannot be parsed.
```
scannillo, line 51: I think it is still useful for this docstring to call out how an error will surface if an invalid credential is provided.

Something like:

```
     * @param authorization The tokenization key or client token to use. If an invalid key is provided, a `BraintreeException` will be returned in the `HttpResponseCallback`.
```

BraintreeModels/src/main/java/com/braintreepayments/api/Authorization.java
```diff
@@ -22,18 +20,21 @@
      *
      * @param authorizationString Given string to transform into an {@link Authorization}.
      * @return {@link Authorization}
-     * @throws InvalidArgumentException This method will throw this exception type if the string
```
scannillo, line 25: Same comment here - can we provide some info in this docstring for how an invalid string will surface as an error?

sarahkoop: Added more detail to the description of this method 👍

scannillo (approved):
🏁

sestevens (approved):

PR #375, opened 2021-05-03
Bundle Cardinal aar from Filesystem

sarahkoop:
### Summary of changes

 -  Remove Bintray dependency for Cardinal SDK

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire 
- @sarahkoop


sestevens:
I like it. We should probably also add a note about this to the migration guide.

sarahkoop:
> I like it. We should probably also add a note about this to the migration guide.

Good call, we added a note that these credentials can be removed.  We are also planning to make this change in a `v3` release, but still think it's noteworthy to include in the migration guide.

scannillo (approved):

scannillo:
Might be helpful if CHANGELOG refers to the GH issue #

sarahkoop:
We are reverting this in v3, since local repository does not propagate MPI aar file. Closing this PR.

PR #378, opened 2021-05-04
GooglePayCardNonce isThreeDSecureEligible Convenience Method

sshropshire:
### Summary of changes

 - Adding a convenience method to check if a `GooglePayCardNonce` is 3D Secure eligible. This prevents merchants from having to know what network tokenized means.

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop
- @sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Add `PaymentMethodType` enum
 * Add `PaymentMethodNonce#getType()` method
 * Add wallet enabled metadata tag to `AndroidManifest.xml` in `google-pay` module 
+* Add `GooglePayCardNonce#isThreeDSecureEligible()` method
 * Breaking Changes
   * Rename `DownForMaintenanceException` to `ServiceUnavailableException`
   * Remove `GoogleApiClientException`
--- a/GooglePay/src/main/java/com/braintreepayments/api/GooglePayCardNonce.java
+++ b/GooglePay/src/main/java/com/braintreepayments/api/GooglePayCardNonce.java
@@ -176,6 +176,11 @@ public class GooglePayCardNonce extends PaymentMethodNonce {
     }
 
     /**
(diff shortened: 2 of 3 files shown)


scannillo (approved):
Changes look good. I am curious though if this change came from a merchants request/ confusion?

We do clarify in our docs what [network tokenized](https://developers.braintreepayments.com/guides/3d-secure/client-side/android/v3#using-3d-secure-with-google-pay) means. An alternative to this PR would be to add more information to the docstring for `isNetworkTokenized`.

Whichever route you settle on works for me, just wanted to ask!

sestevens:
I have the same question as Sammy! Since "network tokenized" is used in the docs, switching up the language in the SDK might make things more confusing. `is_network_tokenized` is also used in the [Server SDKs](https://developers.braintreepayments.com/reference/response/google-pay-card/ruby#is_network_tokenized), so I'd lean towards keeping things consistent.

Clarifying the doc string instead seems like a good idea to me!

sshropshire:
@scannillo @sestevens awesome feedback! We came up with this through docs-driven development. It made the docs a bit cleaner by introducing a boolean property that aligned with checking if a `GooglePayCardNonce` is eligible for 3D Secure.

We thought we could have both. I think it will be useful for merchants whose client / server teams are distinct. Definitely up for adding additional documentation to educate merchants, but I think this will decrease some onboarding friction and decrease the probability of a merchant performing 3DS verification on a network tokenized Google Pay nonce.

sshropshire:
Overall we're indifferent but after looking at `v3` docs we naturally were looking for small wins to improve the SDK. A lot of improvements like this we can make post GA after more discussion.

sshropshire:
Closing this for now. We agree with aligning with the server for this one. Better to have a larger overall strategy for convenience methods and investigating whether this would actually be helpful.

PR #379, opened 2021-05-07
Spike -- Adding PayPal Checkout SDK To Braintree SDK (in Development)

jcnoriega:
### Summary of changes

This PR is the continuation of #365 as I will be taking ownership of this integration.

Updated PayPalClient to launch the Checkout SDK for single transactions while retaining the current functionality for billing agreements. Looking forward to feedback to understand what is left to formalize this integration 🙏

- Updated PayPalClient to launch Checkout experience.
- Updated PayPalClient to tokenize the request similar to how the web experience works today.

 ### Checklist

 - [ ] Added a changelog entry

### Authors

- @CodyEngel
- @jcnoriega

### DIFF
--- a/PayPal/build.gradle
+++ b/PayPal/build.gradle
@@ -32,6 +32,10 @@ dependencies {
     implementation 'androidx.appcompat:appcompat:1.2.0'
     implementation project(':BraintreeDataCollector')
 
+    implementation('com.paypal.checkout:android-sdk:0.1.0') {
+        exclude group: 'com.paypal.android.sdk'
+    }
+
     testImplementation 'org.robolectric:robolectric:4.1'
     testImplementation 'org.powermock:powermock-module-junit4:1.6.6'
     testImplementation 'org.powermock:powermock-module-junit4-rule:1.6.6'
--- a/PayPal/build.gradle
+++ b/PayPal/build.gradle
@@ -33,7 +33,7 @@ dependencies {
     implementation project(':BraintreeDataCollector')
 
     implementation('com.paypal.checkout:android-sdk:0.1.0') {
-        exclude group: 'com.paypal.android.sdk'
+        exclude module: 'data-collector'
     }
 
     testImplementation 'org.robolectric:robolectric:4.1'
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
@@ -13,10 +13,13 @@ import androidx.fragment.app.FragmentActivity;
 import com.paypal.checkout.PayPalCheckout;
 import com.paypal.checkout.approve.Approval;
 import com.paypal.checkout.approve.OnApprove;
+import com.paypal.checkout.cancel.OnCancel;
 import com.paypal.checkout.config.CheckoutConfig;
 import com.paypal.checkout.config.Environment;
 import com.paypal.checkout.createorder.CreateOrder;
 import com.paypal.checkout.createorder.CreateOrderActions;
+import com.paypal.checkout.error.ErrorInfo;
+import com.paypal.checkout.error.OnError;
 
 import org.json.JSONException;
 import org.json.JSONObject;
@@ -126,7 +129,14 @@ public class PayPalClient {
                     callback.onResult(manifestInvalidError);
                     return;
                 }
-                String clientId = configuration.getPayPalClientId() != null ? configuration.getPayPalClientId() : "Client ID Is Null";
+                String clientId = configuration.getPayPalClientId();
+                if (clientId == null) {
+                    //In the future module, we should default this to our new callback function.
+                    //Is this the correct way to send an analytic?
+                    braintreeClient.sendAnalyticsEvent("paypal.native.client-id-null");
+                    callback.onResult(new BraintreeException("Client Id is null"));
+                    return;
+                }
                 String redirectUrl = getPayPalReturnUrl(activity);
                 Environment environment = Environment.SANDBOX;
                 if ("production".equalsIgnoreCase(configuration.getEnvironment())) {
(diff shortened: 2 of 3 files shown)


PayPal/build.gradle
```diff
@@ -32,6 +32,10 @@ dependencies {
     implementation 'androidx.appcompat:appcompat:1.2.0'
     implementation project(':BraintreeDataCollector')
 
+    implementation('com.paypal.checkout:android-sdk:0.1.0') {
+        exclude module: 'data-collector'
+    }
```
jcnoriega, line 37: This was fixed from the previous PR. This module was causing a duplicate class error.

PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
```diff
@@ -113,12 +129,41 @@ public void onResult(@Nullable final Configuration configuration, @Nullable Exce
...
                     return;
                 }
+                String clientId = configuration.getPayPalClientId();
+                if (clientId == null) {
+                    //In the future module, we should default this to our new callback function.
+                    //Is this the correct way to send an analytic?
+                    braintreeClient.sendAnalyticsEvent("paypal.native.client-id-null");
+                    callback.onResult(new BraintreeException("Client Id is null"));
+                    return;
+                }
```
jcnoriega, line 139: When I started the checkout with Client Id = null, i got an error. Is this the best way to handle it? As it is probably some problem that comes from the client.
Also, is that the correct way to send an analytic?

PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
```diff
@@ -113,12 +129,41 @@ public void onResult(@Nullable final Configuration configuration, @Nullable Exce
...
         });
 
     }
 
+    private String getPayPalReturnUrl(Context context) {
+        //Shouldnt the configuration include the return URL of the client? why are generating this?
+        if (context != null) {
+            return String.format("%s://paypalpay", context.getPackageName().toLowerCase(Locale.ROOT)).replace("_", "");
+        }
+        return null;
```
jcnoriega, line 164: I'm deleting underscores to follow Braintree's convention. Question though, why is the return URL generated here? should the return URL come from the `Braintree Configuration`?

PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
```diff
@@ -181,12 +226,65 @@ private void startBrowserSwitch(FragmentActivity activity, PayPalResponse payPal
...
+                            //is this the correct analytic?
+                            braintreeClient.sendAnalyticsEvent("paypal.native.client_cancel");
+                        }
+                    },
+                    new OnError() {
+                        @Override
+                        public void onError(@NonNull ErrorInfo errorInfo) {
+                            //do we send analytic here? if so we should specify error
+                            braintreeClient.sendAnalyticsEvent("paypal.native.error");
+                        }
```
jcnoriega, line 284: `PayPalFlowStarted callback` is not suitable for this scenario, nor is the `BrowserSwitch` functionality as this is a native experience, so what is the correct call back that the client should use in this case? 
I'm guessing that the error handling will depend more upon the implementation of the `PayPalCheckoutNative` module. 
Also, is the analytics ok?

jcnoriega:
@sarahkoop @CodyEngel @sshropshire This is a continuation of #379. I have a couple of concerns.

If there is a new `PayPalCheckoutNative` module coming up, it's my understanding that most of our concerns will be resolved when that module is ready and we can integrate to it. Im referring (amongst other things) to PayPalCheckout SDK `onError` and `onCancel` callbacks. In the new module those errors should be delegated to the Client via the callbacks that the new module exposes. Im guessing that the new module won't support the `BrowserSwitch` functionality as it is a `native` experience, right?
I couldn't find any analytics convention or documentation, so i just added what seemed right (hehe), i left some comments in the code.

PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
CodyEngel: We should validate this with a sample application that uses underscores as there could be additional work required to make this work. [Here's an issue we're currently working through with the Checkout SDK](https://github.com/paypal/android-checkout-sdk/issues/8) related to underscores in the application id.

PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
sshropshire: I think it makes sense to callback an error. I'd guess the root issue is an improper BT <-> PayPal Merchant account linkage? Maybe we could provide a URL to `PayPalNative` docs in the error message.

PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
sshropshire: If the URL is in the Braintree Configuration that would be the best strategy here, but historically we've always done it this way. Might be able to improve our onboarding flow if that return url is present in the config!

PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
sshropshire: Yeah we'll probably do a second pass on the analytics since we need to align them with iOS. The callback pattern is subject to change whatever is best for `PayPalNative` is what we'll go with. We have a shell module [here](https://github.com/braintree/braintree_android/tree/paypal_native) to start ironing out the interface.

sshropshire:
@jcnoriega all good questions! We feel like the callback flow and overall addition to the SDK should be collaborative. We're open to all types of new ideas. I agree with this being a native module, we can leverage `BrowserSwitch` internals without actually having to use browser switch. We want to make sure that deep linking back into the Merchant application is done efficiently and effectively.

jcnoriega:
SPKIE no longer relevant

PR #384, opened 2021-05-12
PayPalNativeModule: Added fallback to billing agreements web experience

jcnoriega:
### Summary of changes

 - If the merchant chooses Billing Agreement in the native expirience, it will default to the Web as we have right now in PayPalClient. Modified the demo app to choose Billing Argeements inside the native module

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/PayPalNativeFragment.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/PayPalNativeFragment.java
@@ -26,24 +26,24 @@ public class PayPalNativeFragment extends BaseFragment {
     private DataCollector dataCollector;
 
 
-    private Button billingVaultButton;
-    private Button singlePaymentButton;
+    private Button vaultButton;
+    private Button checkoutButton;
 
     @Override
     public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
         final View view = inflater.inflate(R.layout.fragment_paypal_native, container, false);
-        billingVaultButton = view.findViewById(R.id.paypal_native_vault_button);
-        singlePaymentButton = view.findViewById(R.id.paypal_native_single_payment_button);
+        vaultButton = view.findViewById(R.id.paypal_native_vault_button);
+        checkoutButton = view.findViewById(R.id.paypal_native_checkout_button);
 
-        billingVaultButton.setOnClickListener(this::launchVault);
-        singlePaymentButton.setOnClickListener(this::launchSinglePayment);
+        vaultButton.setOnClickListener(this::launchVault);
+        checkoutButton.setOnClickListener(this::launchCheckout);
 
         DemoViewModel viewModel = new ViewModelProvider(getActivity()).get(DemoViewModel.class);
         viewModel.getPayPalBrowserSwitchResult().observe(getViewLifecycleOwner(), this::handlePayPalBrowserSwitchResult);
         return view;
     }
 
-    private void launchSinglePayment(View v) {
+    private void launchCheckout(View v) {
         getBraintreeClient(braintreeClient -> {
             payPalNativeClient = new PayPalNativeClient(braintreeClient);
             startPayPalNative();
--- a/Demo/src/main/java/com/braintreepayments/demo/PayPalNativeFragment.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/PayPalNativeFragment.java
@@ -12,12 +12,10 @@ import androidx.navigation.fragment.NavHostFragment;
 
 import com.braintreepayments.api.BrowserSwitchResult;
 import com.braintreepayments.api.DataCollector;
-import com.braintreepayments.api.PayPalClient;
 import com.braintreepayments.api.PayPalNativeCheckoutRequest;
 import com.braintreepayments.api.PayPalNativeClient;
 import com.braintreepayments.api.PaymentMethodNonce;
 
-import static com.braintreepayments.demo.PayPalRequestFactory.createPayPalCheckoutRequest;
 import static com.braintreepayments.demo.PayPalRequestFactory.createPayPalNativeVaultRequest;
 
 public class PayPalNativeFragment extends BaseFragment {
@@ -28,16 +26,16 @@ public class PayPalNativeFragment extends BaseFragment {
     private DataCollector dataCollector;
 
 
-    private Button billingAgreementButton;
+    private Button billingVaultButton;
     private Button singlePaymentButton;
 
     @Override
     public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
         final View view = inflater.inflate(R.layout.fragment_paypal_native, container, false);
-        billingAgreementButton = view.findViewById(R.id.paypal_native_billing_agreement_button);
+        billingVaultButton = view.findViewById(R.id.paypal_native_vault_button);
         singlePaymentButton = view.findViewById(R.id.paypal_native_single_payment_button);
 
-        billingAgreementButton.setOnClickListener(this::launchBillingAgreement);
+        billingVaultButton.setOnClickListener(this::launchVault);
         singlePaymentButton.setOnClickListener(this::launchSinglePayment);
 
         DemoViewModel viewModel = new ViewModelProvider(getActivity()).get(DemoViewModel.class);
@@ -45,14 +43,14 @@ public class PayPalNativeFragment extends BaseFragment {
         return view;
     }
 
-    public void launchSinglePayment(View v) {
+    private void launchSinglePayment(View v) {
         getBraintreeClient(braintreeClient -> {
             payPalNativeClient = new PayPalNativeClient(braintreeClient);
             startPayPalNative();
         });
     }
 
-    public void launchBillingAgreement(View v) {
+    private void launchVault(View v) {
         FragmentActivity activity = getActivity();
         activity.setProgressBarIndeterminateVisibility(true);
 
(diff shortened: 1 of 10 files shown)


Demo/src/main/java/com/braintreepayments/demo/PayPalNativeFragment.java
```diff
@@ -4,29 +4,99 @@
...
+            super.onPaymentMethodNonceCreated(paymentMethodNonce);
+
+            PayPalNativeFragmentDirections.ActionPayPalNativeFragmentToDisplayNonceFragment action =
+                    PayPalNativeFragmentDirections.actionPayPalNativeFragmentToDisplayNonceFragment(paymentMethodNonce);
+            action.setDeviceData(deviceData);
+
+            NavHostFragment.findNavController(this).navigate(action);
+        }
+    }
+
```
jcnoriega, line 97: Most of the modifications inside `PayPalNativeFragment` are a copy of whats inside `PayPalFragment`. The idea is to replicate the same behaviour for vault in both `PayPal` and `PayPalNative`

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
```diff
@@ -10,13 +11,22 @@
...
     }
 
-    public void tokenizePayPalAccount(final FragmentActivity activity, final PayPalNativeCheckoutRequest request, final PayPalNativeFlowStartedCallback callback) {
+    public void tokenizePayPalAccount(final FragmentActivity activity, final PayPalNativeRequest request, final PayPalNativeFlowStartedCallback callback) {
+        if (request instanceof PayPalNativeCheckoutRequest) {
+            sendNativeCheckoutRequest(activity, (PayPalNativeCheckoutRequest) request, callback);
+        } else if (request instanceof PayPalNativeVaultRequest) {
+            sendVaultRequest(activity, (PayPalNativeVaultRequest) request, callback);
+        }
+    }
```
jcnoriega, line 27: Same as `PayPayClient`, we select the flow depending on the request

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
```diff
@@ -38,6 +48,17 @@ public void onResult(Configuration configuration, Exception error) {
...
+        //this one should default to the one we already have, once PayPalNative supports billing agreements, this should just default to native.
+        payPalClient = new PayPalClient(braintreeClient);
+        payPalClient.tokenizePayPalAccount(activity, payPalVaultRequest, new PayPalFlowStartedCallback() {
+            @Override
+            public void onResult(@Nullable Exception error) {
+                callback.onResult(error);
+            }
+        });
+    }
+
```
jcnoriega, line 61: In order to default to the same vault experience we have now, I just call the PayPalClient and wrap the callbacks.

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
```diff
@@ -60,7 +81,15 @@ private static Exception createPayPalNotEnabledError() {
...
+    public void onActivityResumed(final BrowserSwitchResult browserSwitchResult, final PayPalNativeOnActivityResumedCallback callback) {
+        if (payPalClient != null) {
+            //it means it was a vault request
+            payPalClient.onBrowserSwitchResult(browserSwitchResult, new PayPalBrowserSwitchResultCallback() {
+                @Override
+                public void onResult(@Nullable PayPalAccountNonce payPalAccountNonce, @Nullable Exception error) {
+                    callback.onResult(payPalAccountNonce, error);
+                }
+            });
+        }
```
jcnoriega, line 93: Same. If `payPalClient` is not null, it means it was a vault request, so we default to the current behaviour in `PayPalClient`

Demo/src/main/java/com/braintreepayments/demo/PayPalNativeFragment.java
```diff
@@ -4,29 +4,99 @@
...
 public class PayPalNativeFragment extends BaseFragment {
 
-    PayPalNativeClient payPalNativeClient;
+    private PayPalNativeClient payPalNativeClient;
+
+    private String deviceData;
+    private DataCollector dataCollector;
+
+
+    private Button billingAgreementButton;
```
sarahkoop, line 31: I know we haven't updated the `PayPalFragment` in the demo app for this, but let's go with `vault` and `checkout` for the variable/method naming throughout to align with the request types

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeRequest.java
```diff
@@ -0,0 +1,4 @@
+package com.braintreepayments.api;
+
+public interface PayPalNativeRequest {
```
sarahkoop, line 3: Currently we have a base class with subclasses for `PayPalRequest`, is there a need for this to be an interface instead, or can we follow the same pattern here?

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
sarahkoop: I think there could be a more reliable assertion here - I believe there are differences in the deep link URL of the `BrowswerSwitchResult` based on the request type, would it make more sense to assert on that?

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeRequest.java
jcnoriega: It depends really. Do we want `PayPalNative` to accept `PayPalRequest`? Really dont know if the request are different if its for web or native to be honest. The way I thought of this was that `PayPalNative` can only accept `PayPalNativeRequest`, but for `PayPalNativeVaultRequest` and `PayPalNativeCheckoutRequest` we need the implementation of `PayPalRequest` base class and subclasses (for now). If I made that `PayPalNative` accepts `PayPalRequest`, then the client could send to our module requests of type `PayPalVaultRequest` and `PayPalCheckoutRequest` (from `PayPal` module), and im not sure if we want that. If we need to differentiate requests between those two modules, then this implementation allows for that. If not, then we can just dispose of any request in `PayPalNative` and use the ones in `PayPal`. Does any of this make sense?

Demo/src/main/java/com/braintreepayments/demo/PayPalNativeFragment.java
```diff
@@ -4,29 +4,99 @@
...
+                    PayPalNativeFragmentDirections.actionPayPalNativeFragmentToDisplayNonceFragment(paymentMethodNonce);
+            action.setDeviceData(deviceData);
+
+            NavHostFragment.findNavController(this).navigate(action);
+        }
+    }
+
     private void startPayPalNative() {
-        PayPalNativeCheckoutRequest nativeCheckoutRequest = new PayPalNativeCheckoutRequest();
+        PayPalNativeCheckoutRequest nativeCheckoutRequest = new PayPalNativeCheckoutRequest("1.0");
```
sarahkoop, line 99:
```suggestion
        PayPalNativeCheckoutRequest nativeCheckoutRequest = new PayPalNativeCheckoutRequest("1.00");
```

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeRequest.java
sshropshire: Makes sense. We're discussing this internally with iOS as well. It's more about SDK alignment what we do here may be influenced by them and vice versa since we strive to keep the SDKs as similar as possible with the exception of a few edge cases.

sshropshire: `v4` was also basically a complete re-write for us. We have an established set of conventions in general. It's hard to say what the final decision will be on the types, but I think it's safe to say we can easily change this as we get closer to GA.

sshropshire: Maybe we could use the `PayPalNativeCheckoutRequest` concrete type for now and circle back after native XO is fully integrated?

Demo/src/main/java/com/braintreepayments/demo/PayPalNativeFragment.java
```diff
@@ -4,29 +4,97 @@
...
 public class PayPalNativeFragment extends BaseFragment {
 
-    PayPalNativeClient payPalNativeClient;
+    private PayPalNativeClient payPalNativeClient;
+
+    private String deviceData;
+    private DataCollector dataCollector;
+
+
+    private Button billingVaultButton;
```
sarahkoop, line 29:
```suggestion
    private Button vaultButton;
```

Demo/src/main/java/com/braintreepayments/demo/PayPalNativeFragment.java
```diff
@@ -4,29 +4,97 @@
...
+    private Button singlePaymentButton;
```
sarahkoop, line 30:
```suggestion
    private Button checkoutButton;
```

Demo/src/main/java/com/braintreepayments/demo/PayPalNativeFragment.java
```diff
@@ -4,29 +4,97 @@
...
+
+        DemoViewModel viewModel = new ViewModelProvider(getActivity()).get(DemoViewModel.class);
+        viewModel.getPayPalBrowserSwitchResult().observe(getViewLifecycleOwner(), this::handlePayPalBrowserSwitchResult);
+        return view;
     }
 
-    @Override
-    public void onCreate(Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState);
+    private void launchSinglePayment(View v) {
```
sarahkoop, line 46:
```suggestion
    private void launchCheckout(View v) {
```

sarahkoop (approved):

sshropshire (approved):

PR #385, opened 2021-05-13
Make PayPalRequest Parcelable

sarahkoop:
### Summary of changes

 - Make `PayPalRequest` and subclasses `Parcelable`. We previously removed this, but realized it needs to be parcelable for Drop in, so adding it back.
 - Remove missed hungarian notation from `PayPalRequest`.

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop 
- @sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,7 @@
 
 ## unreleased
 
-* Breaking Changes
-  * Make `PayPalRequest` and subclasses `Parcelable`
+* Make `PayPalRequest` and subclasses `Parcelable`
 
 ## 4.0.0-beta3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* Breaking Changes
+  * Make `PayPalRequest` and subclasses `Parcelable`
+
 ## 4.0.0-beta3
 
 * Add `PaymentMethodType` enum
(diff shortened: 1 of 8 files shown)


PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
```diff
@@ -36,6 +39,7 @@
     static final String MERCHANT_ACCOUNT_ID = "merchant_account_id";
     static final String LINE_ITEMS_KEY = "line_items";
 
+
```
sestevens, line 42:
```suggestion
```

sestevens (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* Breaking Changes
+  * Make `PayPalRequest` and subclasses `Parcelable`
```
scannillo, line 6: Updating `PayPalCheckoutRequest` to implement `Parcelable` shouldn't break existing integrations with `PayPalCheckoutRequest`, right? I wouldn't think this is a breaking change.

We should also call out the full list of classes that got updated in this PR

sarahkoop: The concern here was anyone subclassing any of these classes would have to implement `Parcelable` on their subclass  now too.  There shouldn't be a need to subclass this class in order to integrate, but should it still be considered a breaking change because of that?

scannillo: Hm - I would think it is unlikely that a merchant would be subclassing the requests nor is it our intended use-case, so I would opt to say this isn't a breaking change. That is my vote, but trust whatever y'all decide!

sshropshire: Bumped it up to a regular change!

scannillo (approved):

PR #388, opened 2021-05-19
Remove PaymentMethodClient

sarahkoop:
### Summary of changes

 - Add back `description` and `typeLabel` to `PaymentMethodNonce` and subclasses to be used in Drop-in
 - Make `getType` on `PaymentMethodNonce` public

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire 
- @sarahkoop


sestevens:
2 thoughts:

1. I'm not sure if Android Drop-in is modifying `description` before displaying it, but the descriptions we're returning here aren't localized, so they can't be displayed to customers as-is. Specifically, "with email address ..." for PayPal and "ending in ..." for card aren't localized. On iOS Drop-in, we just use the PayPal email address and the last four digits of the card instead. Should we do the same on Android?
2. On iOS, we moved `description` from the base SDK into Drop-in, since it's only be used there. Would it make sense to do the same on Android?

Here's a link to iOS Drop-in: https://github.com/braintree/braintree-ios-drop-in/blob/743bda4fe045dab76cadacf741bf56e4fbbc89d6/Sources/BraintreeDropIn/BTPaymentMethodNonce%2BDropIn.m#L23

sarahkoop:
> 2 thoughts:
> 
> 1. I'm not sure if Android Drop-in is modifying `description` before displaying it, but the descriptions we're returning here aren't localized, so they can't be displayed to customers as-is. Specifically, "ending in ..." for card isn't localized. On iOS Drop-in, we use the last four digits of the card instead of the `description` that comes back from the gateway. Should we do the same on Android?
> 2. On iOS, we moved `description` from the base SDK into Drop-in, since it's only be used there. Would it make sense to do the same on Android?
> 
> Here's a link to iOS Drop-in: https://github.com/braintree/braintree-ios-drop-in/blob/743bda4fe045dab76cadacf741bf56e4fbbc89d6/Sources/BraintreeDropIn/BTPaymentMethodNonce%2BDropIn.m#L23

1. I agree we should change it to the last four digits to align with iOS for cards
2. We thought about doing that, but currently the `PaymentMethodClient` in in the core SDK returns a list of `PaymentMethodNonce` objects rather than specific typed nonce. This change was necessary to move the nonce type classes out of the `BraintreeCore` module into their respective modules.  We had a TODO to investigate leveraging `compileOnly` dependencies in the core SDK to allow `PaymentMethodClient` to return typed nonces, but that gets us back into the circular dependency issues that we have resolved in v4.

BraintreeCore/src/test/java/com/braintreepayments/api/PaymentMethodNonceUnitTest.java
```diff
@@ -56,6 +64,8 @@ public void constructor_parsesLocalPaymentPlainJSONObject() throws JSONException
         assertFalse(nonce.isDefault());
         assertEquals(PaymentMethodType.PAYPAL, nonce.getType());
         assertEquals("e11c9c39-d6a4-0305-791d-bfe680ef2d5d", nonce.getString());
+        assertEquals("PayPal", nonce.getTypeLabel());
```
scannillo, line 67: Do we need both of these (description and typeLabel)? They look to be the same in most cases. Which ones does Drop In use?

sarahkoop: They are used in Drop-in in different ways currently - the description is used to display the PayPal email address or Venmo account name on vaulted payment methods and the typeLabel is used to get the display icons (and getType alone won't work for that because it needs to be card brand specific or PayPal or Venmo etc). We are still working on the Drop-in refactoring though, so it's possible we will find a better way to consolidate these, but since we are adding them back as package-private we can always remove in the future if no longer needed.

BraintreeCore/src/main/java/com/braintreepayments/api/PaymentMethodNonce.java
```diff
@@ -16,26 +15,61 @@
...
-    private @PaymentMethodType final int mType;
+    private @PaymentMethodType final int type;
 
     static PaymentMethodNonce fromJSON(JSONObject inputJson) throws JSONException {
         String typeString = inputJson.getString(PAYMENT_METHOD_TYPE_KEY);
         int type = paymentMethodTypeFromString(typeString);
 
         String nonce = inputJson.getString(PAYMENT_METHOD_NONCE_KEY);
         boolean isDefault = inputJson.optBoolean(PAYMENT_METHOD_DEFAULT_KEY, false);
+        String description = inputJson.optString(DESCRIPTION_KEY);
```
sestevens, line 33: Just for the sake of clarity and consistency with iOS, I wonder if we should stop using the `description` field from the Gateway altogether and just set the description based on the actual property being used (i.e., email address for PayPal, username for Venmo, etc.).

It feels a little odd to me that right now we're using the Gateway's `description` for some payment method types but not others.

BraintreeCore/src/test/java/com/braintreepayments/api/PaymentMethodNonceUnitTest.java
```diff
@@ -29,6 +31,8 @@ public void constructor_parsesPayPalFromPlainJSONObject() throws JSONException {
         assertFalse(nonce.isDefault());
         assertEquals(PaymentMethodType.PAYPAL, nonce.getType());
         assertEquals("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee", nonce.getString());
+        assertEquals("PayPal", nonce.getTypeLabel());
+        assertEquals("with email paypalaccount@example.com", nonce.getDescription());
```
sestevens, line 35: I know these are just unit tests, but it's confusing to me that we say "with email paypalaccount@example.com" here. Can we just make the description "paypalaccount@example.com"?

BraintreeCore/src/test/java/com/braintreepayments/api/PaymentMethodNonceUnitTest.java
```diff
@@ -47,6 +53,8 @@ public void constructor_parsesVisaCheckoutFromPlainJSONObject() throws JSONExcep
         assertFalse(nonce.isDefault());
         assertEquals(PaymentMethodType.VISA_CHECKOUT, nonce.getType());
         assertEquals("123456-12345-12345-a-adfa", nonce.getString());
+        assertEquals("Visa Checkout", nonce.getTypeLabel());
+        assertEquals("ending in ••11", nonce.getDescription());
```
sestevens, line 57: Same thing here - can we change this to "1111" instead of "ending in ••11"?

VisaCheckout/src/test/java/com/braintreepayments/api/VisaCheckoutNonceUnitTest.java
```diff
@@ -26,6 +26,8 @@ public void fromJson_createsVisaCheckoutNonce() throws JSONException {
                 Fixtures.PAYMENT_METHODS_VISA_CHECKOUT_RESPONSE));
 
         assertEquals(PaymentMethodType.VISA_CHECKOUT, visaCheckoutNonce.getType());
+        assertEquals("Visa Checkout", visaCheckoutNonce.getTypeLabel());
+        assertEquals("ending in ••11", visaCheckoutNonce.getDescription());
```
sestevens, line 30: It seems like this should also be "1111".

BraintreeCore/src/main/java/com/braintreepayments/api/PaymentMethodNonce.java
sarahkoop: Yes good call - we aligned with iOS and now return email for PayPal, username for Venmo, and last four for Card. Since Google Pay can be a card or PayPal account we decided to just return "Google Pay" for description, and since Visa Checkout seems to only have last two in the response we decided to return "Visa Checkout".

sestevens: Sounds good! So we don't need to pull the description from the Gateway response anymore, right?
```suggestion
        String description;
```

Card/src/main/java/com/braintreepayments/api/CardNonce.java
```diff
@@ -37,6 +37,7 @@
     private static final String EXPIRATION_MONTH_KEY = "expirationMonth";
     private static final String EXPIRATION_YEAR_KEY = "expirationYear";
     private static final String CARDHOLDER_NAME_KEY = "cardholderName";
+    private static final String DESCRIPTION_KEY = "description";
```
sestevens, line 40: Is this still needed?

PayPal/src/main/java/com/braintreepayments/api/PayPalAccountNonce.java
```diff
@@ -35,6 +34,7 @@
     private static final String PHONE_KEY = "phone";
     private static final String PAYER_ID_KEY = "payerId";
     private static final String CLIENT_METADATA_ID_KEY = "correlationId";
+    private static final String DESCRIPTION_KEY = "description";
```
sestevens, line 37: Can this be removed?

VisaCheckout/src/main/java/com/braintreepayments/api/VisaCheckoutNonce.java
```diff
@@ -27,6 +26,7 @@
     private static final String SHIPPING_ADDRESS_KEY = "shippingAddress";
     private static final String USER_DATA_KEY = "userData";
     private static final String CALL_ID_KEY = "callId";
+    private static final String DESCRIPTION_KEY = "description";
```
sestevens, line 29: Remove?

BraintreeCore/src/test/java/com/braintreepayments/api/PaymentMethodNonceUnitTest.java
sshropshire: The test assertions should be good now!

BraintreeCore/src/test/java/com/braintreepayments/api/PaymentMethodNonceUnitTest.java
sshropshire: There's a lot of coupling between the core sdk and DropIn in Android. It seems like it naturally evolved that way. We could take a final look to make sure these methods actually provide value to merchants. But we know at least if we leave some stuff out of `v4` we can always add it back in a minor release so we have some options.

sshropshire:
Closed in favor of #396.

PR #389, opened 2021-05-19
Client SDK deprecation policy

scannillo:
### Summary of changes

 - There was an old JIRA in our backlog to add the Client SDK Deprecation policy to the Android README. I formatted it similarly to[ the README in iOS](https://github.com/braintree/braintree_ios#versions).
 - I called out that v2.7.3 and below are unsupported since those use expired SSL certs.

 ### Checklist

### Authors
@scannillo

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ This SDK abides by our Client SDK Deprecation Policy. For more information on th
 | -------------------- | ------ | -------- | ---------- | ----------- |
 | 4.x.x | Beta | March 2021 | TBA | TBA |
 | 3.x.x | Active | February 2019 | TBA | TBA |
-| 2.x.x | Inactive | November 2015 | March 2020 | March 2021 |
+| 2.x.x | Unsupported | November 2015 | March 2020 | March 2021 |
 
 Versions 2.7.3 and below use outdated SSL certificates and are unsupported.
 
@@ -64,4 +64,3 @@ for personal support at any phase of integration
 ## License
 
 The Braintree Android SDK is open source and available under the MIT license. See the [LICENSE](LICENSE) file for more info.
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,6 +32,19 @@ Start with [**'Hello, Client!'**](https://developers.braintreepayments.com/start
...
+## Versions
+
+This SDK abides by our Client SDK Deprecation Policy. For more information on the potential statuses of an SDK check our [developer docs](http://developers.braintreepayments.com/guides/client-sdk/deprecation-policy).
+
+<!-- TODO: Update TBA dates when v4 GA releases. -->
+| Major version number | Status | Released | Deprecated | Unsupported |
+| -------------------- | ------ | -------- | ---------- | ----------- |
+| 4.x.x | Beta | March 2021 | TBA | TBA |
+| 3.x.x | Active | February 2019 | TBA | TBA |
+| 2.x.x | Inactive | November 2015 | March 2020 | March 2021 |
```
sarahkoop, line 44: Should the deprecated and unsupported months for 2.x.x be Feb since 3.0.0 was released Feb 2019? But also since v2 is unsupported as of March this year, can we remove it from this table all-together?

scannillo: Yeh I went back and forth on this. I was hesitant to make the months for v2 February, since our deprecation policy technically wasn't released yet when v3 came out 🤔 .

But If v2 is definitely unsupported anyways, we can just update it to February. I don't see the harm in leaving it in the table, though. I know people are still using Android v2.

sarahkoop: Ah ok yeah that makes sense - I don't feel strongly either way about the dates then. We did remove public docs for it, so I'd say let's list it as unsupported in the table.

scannillo, line 44:
```suggestion
| 2.x.x | Unsupported | November 2015 | March 2020 | March 2021 |
```

scannillo: Done!

sestevens (approved):

sarahkoop (approved):

PR #390, opened 2021-05-20
PayPal Checkout Integration to PayPalNative Module

jcnoriega:
### Summary of changes

Added Unit tests as well:
Screen Shot 2021-05-27 at 11 03 46
Screen Shot 2021-05-27 at 11 04 09
Screen Shot 2021-06-30 at 09 25 16
Screen Shot 2021-06-30 at 09 30 15
Screen Shot 2021-05-27 at 11 10 05
Screen Shot 2021-05-27 at 11 10 31


PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
```diff
@@ -13,33 +31,44 @@
...
     }
 
-    private void sendNativeCheckoutRequest(final FragmentActivity activity, final PayPalNativeCheckoutRequest request, final PayPalNativeFlowStartedCallback callback) {
+    private void sendNativeCheckoutRequest(final FragmentActivity activity, final PayPalNativeCheckoutRequest request, final PayPalNativeFlowCallback callback) {
+        //This is copied from PayPal module.
+        braintreeClient.sendAnalyticsEvent("paypal.native.single-payment.selected");
+        if (request.getShouldOfferPayLater()) {
+            braintreeClient.sendAnalyticsEvent("paypal.native.single-payment.paylater.offered");
+        }
+
```
jcnoriega, line 58: Copied this from `PayPal` module, is `single-payment` the correct analytic?

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
```diff
@@ -50,33 +79,130 @@ public void onResult(Configuration configuration, Exception error) {
...
+        if ("production".equalsIgnoreCase(configuration.getEnvironment())) {
+            environment = Environment.LIVE;
+        }
+
+        CheckoutConfig checkoutConfig = new CheckoutConfig(
+                activity.getApplication(),
+                clientId,
+                environment,
+                redirectUrl
+        );
```
jcnoriega, line 118: Once the `PayPalRequest` is successfull, then I configure the Checkout SDK.

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
```diff
@@ -50,33 +79,130 @@ public void onResult(Configuration configuration, Exception error) {
...
+                    public void onCancel() {
+                        braintreeClient.sendAnalyticsEvent("paypal.native.client_cancel");
+                        callback.onResult(null, new BraintreeException("User Cancelled"));
+                    }
+                },
+                new OnError() {
+                    @Override
+                    public void onError(final ErrorInfo errorInfo) {
+                        braintreeClient.sendAnalyticsEvent("paypal.native.error");
+                        callback.onResult(null, new BraintreeException(errorInfo.getReason()));
```
jcnoriega, line 145: I delegate both `OnCancel` and `OnError` to the callback. I have doubts about `OnCancel` though, as it is technically not an error, but should we add another parameter to `PayPalNativeFlowCallback`?

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
```diff
@@ -50,33 +79,130 @@ public void onResult(Configuration configuration, Exception error) {
...
+                            callback.onResult(null, e);
+                        }
+                    } else {
+                        callback.onResult(null, exception);
+                    }
+                }
+            });
+        } catch (final JSONException | BraintreeException e) {
+            callback.onResult(null, e);
+        }
```
jcnoriega, line 194: I took all of this from `PayPal` module. I wonder if its completely necessary or if it even applies here.
It was the only way to go from `Approval` object that I get from the SDK, to `PaymentMethodNonce`. Perhaps there is a simpler. Do we need to pass through `tokenizationClient.tokenizeREST()`?

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
```diff
@@ -50,33 +79,130 @@ public void onResult(Configuration configuration, Exception error) {
...
+                approval.getData().getPaymentId(),
+                approval.getData().getOrderId(),
+                approval.getData().getPayerId()));
+        try {
+            JSONObject urlResponseData = parseUrlResponseData(deepLinkUri, payPalResponse.getSuccessUrl(), payPalResponse.getApprovalUrl(), "token");
+            PayPalAccount payPalAccount = new PayPalAccount();
+            payPalAccount.setClientMetadataId(payPalResponse.getClientMetadataId());
+            payPalAccount.setSource("paypal-browser"); // paypal native throws error
+            payPalAccount.setUrlResponseData(urlResponseData);
+            payPalAccount.setPaymentType("single-payment"); // is this the correct way?
```
jcnoriega, line 165: If I pass `paypal-native` to `setSource()`, `tokenizeREST` returns an error. Where can we configure that to accept `paypal-native` (or maybe something else?). Is also 'single-payment` the correct payment type?

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeFlowCallback.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+//TODO: i changed the name because i think one call back is enough. The clients wants to know whether it succeeded or failed
+public interface PayPalNativeFlowCallback {
+
+    /**
+     * @param success if the paypal experience is successful.
+     * @param error an exception that occurred while initiating PayPal Native
+     */
+    void onResult(PaymentMethodNonce success, Exception error);
+}
```
jcnoriega, line 17: Changed name from `PayPalNativeStartedFlowCallback` to `PayPalNativeFlowCallback`. My guess is that the client doesn't "care" if the flow fails before the SDK is called or after, it cares only if fails or succeedes. I think this is cleaner, we only have one callback for a successful flow and errors, and thats it. Thoughts?

BraintreeModels/src/main/java/com/braintreepayments/api/PaymentMethodType.java
```diff
@@ -16,7 +16,8 @@
         PaymentMethodType.VENMO,
         PaymentMethodType.VISA_CHECKOUT,
         PaymentMethodType.GOOGLE_PAY,
-        PaymentMethodType.LOCAL_PAYMENT
+        PaymentMethodType.LOCAL_PAYMENT,
+        PaymentMethodType.PAYPAL_NATIVE
```
sarahkoop, line 20: I'm not sure that this is necessary. These types are used to differentiate PaymentMethodNonce` types, but it looks like we are returning a `PayPalAccountNonce` at the end of this flow (the same nonce type as the legacy PayPal flow), so I don't see a need to differentiate types here.

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
sarahkoop: Yeah these analytics event strings haven't been updated to align with the new method naming, so I think it's ok to match them to the events in `PayPalClient` for now.

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
sarahkoop: We have documented that a `BraintreeException` with a message indicating user cancellation will be returned on user cancellation for all the payment flows in the SDK, so I think it's good to go with that approach here as well. It also allows us to maintain the two-parameter callback pattern.

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeFlowCallback.java
sarahkoop: Agreed one callback here makes sense. Maybe `PayPalNativeTokenizeCallback` would make sense as the name?

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeFlowCallback.java
```diff
@@ -0,0 +1,17 @@
...
+ * {@link PayPalNativeClient#tokenizePayPalAccount(FragmentActivity, PayPalRequest, PayPalNativeFlowCallback)}
+ */
+//TODO: i changed the name because i think one call back is enough. The clients wants to know whether it succeeded or failed
+public interface PayPalNativeFlowCallback {
+
+    /**
+     * @param success if the paypal experience is successful.
+     * @param error an exception that occurred while initiating PayPal Native
+     */
+    void onResult(PaymentMethodNonce success, Exception error);
```
sarahkoop, line 16:
```suggestion
    void onResult(PayPalAccount nonce payPalAccountNonce, Exception error);
```

This callback should only be returning PayPalAccountNonces

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeFlowCallback.java
```diff
@@ -0,0 +1,17 @@
...
+ * Callback for receiving success and error when calling
+ * {@link PayPalNativeClient#tokenizePayPalAccount(FragmentActivity, PayPalRequest, PayPalNativeFlowCallback)}
+ */
+//TODO: i changed the name because i think one call back is enough. The clients wants to know whether it succeeded or failed
+public interface PayPalNativeFlowCallback {
+
+    /**
+     * @param success if the paypal experience is successful.
+     * @param error an exception that occurred while initiating PayPal Native
+     */
```
sarahkoop, line 15:
```suggestion
    /**
     * @param payPalAccountNonce {@link PayPalAccountNonce}
     * @param error an exception that occurred during the PayPal Native flow
     */
```
Clarifying the this handles exceptions that occur after initiating too.

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
```diff
@@ -13,33 +31,44 @@
...
+                if (payPalConfigInvalid(configuration) || error != null) {
+                    Exception configInvalidError = createPayPalNotEnabledError();
+                    callback.onResult(null, configInvalidError);
                     return;
                 }
 
                 String payPalClientId = configuration.getPayPalClientId();
                 if (payPalClientId == null) {
-                    callback.onResult(new BraintreeException("PayPal Client ID is null. TODO: write error message explaining why PayPal Native couldn't be started"));
+                    callback.onResult(null, new BraintreeException("PayPal Client ID is null. TODO: write error message explaining why PayPal Native couldn't be started"));
```
sarahkoop, line 71: Just calling out that we should clarify this exception message and remove the TODO

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
```diff
@@ -50,33 +79,130 @@ public void onResult(Configuration configuration, Exception error) {
...
+                    public void onError(final ErrorInfo errorInfo) {
+                        braintreeClient.sendAnalyticsEvent("paypal.native.error");
+                        callback.onResult(null, new BraintreeException(errorInfo.getReason()));
+                    }
+                }
+        );
+    }
+
+    //TODO: I copied all of this from PayPal client and adapted it. Is all of this necessary?
+    private void approved(final Approval approval, final PayPalResponse payPalResponse, final PayPalNativeFlowCallback callback) {
```
sarahkoop, line 152:
```suggestion
    private void handlePayPalNativeCheckoutApprovale(final Approval approval, final PayPalResponse payPalResponse, final PayPalNativeFlowCallback callback) {
```

sarahkoop, line 152:
```suggestion
    private void handleNativeCheckoutApproval(final Approval approval, final PayPalResponse payPalResponse, final PayPalNativeFlowCallback callback) {
```
I'm open to other method names here, but something a little more descriptive than `approved` would be helpful

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
sarahkoop: I believe these fields are outside of the SDK domain, as they are used for tokenizing the PayPal account. We should look into which are required/can be modified to align with this new flow.

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
```diff
@@ -50,33 +79,130 @@ public void onResult(Configuration configuration, Exception error) {
...
+                    public void onApprove(final Approval approval) {
+                        braintreeClient.sendAnalyticsEvent("paypal.native.approved");
+                        approved(approval, payPalResponse, callback);
+                    }
+                },
+                new OnCancel() {
+                    @Override
+                    public void onCancel() {
+                        braintreeClient.sendAnalyticsEvent("paypal.native.client_cancel");
+                        callback.onResult(null, new BraintreeException("User Cancelled"));
```
sarahkoop, line 138:
```suggestion
                        callback.onResult(null, new BraintreeException("User Canceled PayPal"));
```
Nitpick - but aligning this with the legacy cancellation message

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
```diff
@@ -102,4 +228,38 @@ public void onResult(@Nullable PayPalAccountNonce payPalAccountNonce, @Nullable
...
+            response.put("webURL", uri.toString());
+            urlResponseData.put("response", response);
+
+            urlResponseData.put("response_type", "web");
+
+            return urlResponseData;
+        } else {
+            throw new BraintreeException("The response contained inconsistent data.");
+        }
+    }
```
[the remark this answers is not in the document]
sarahkoop: I'm fine with creating a utility class for this or leaving it here.

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
sarahkoop: I think we need to be returning a `PaymentMethodNonce` to the merchant so that they can send it to their servers to create transactions.  Using the Approval object info to create a tokenization request which returns a nonce seems like the correct flow to me. We can also look into how this is being done on iOS, as it should probably align between the two.

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
```diff
@@ -50,33 +79,130 @@ public void onResult(Configuration configuration, Exception error) {
...
+                }
+            });
+        } catch (final JSONException | BraintreeException e) {
+            callback.onResult(null, e);
+        }
+
+    }
+
+    private String getPayPalReturnUrl(Context context) {
+        //Shouldnt the configuration include the return URL of the client? why are generating this?
```
sarahkoop, line 199: I don't think this return URL is currently included in the fetched configuration response, so I think we do need to generate this here for now.

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
jcnoriega: do you know where i can take a look at which are the required fields?

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
jcnoriega: I took i look at the repo and it does, they create the tokenization request with the approval information

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
```diff
@@ -1,49 +1,265 @@
...
+                    public void onApprove(final Approval approval) {
+                        braintreeClient.sendAnalyticsEvent("paypal.native.approved");
+                        handleNativeCheckoutApproval(approval, payPalResponse, callback);
+                    }
+                },
+                new OnCancel() {
+                    @Override
+                    public void onCancel() {
+                        braintreeClient.sendAnalyticsEvent("paypal.native.client_cancel");
+                        callback.onResult(null, new BraintreeException("Canceled"));
```
sarahkoop, line 147: Since this PR was last reviewed, we updated the exception type for cancellations to `UserCanceledException`, so that should be returned here now instead.

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
```diff
@@ -1,49 +1,265 @@
...
+                    approval.getData().getPaymentId(),
+                    approval.getData().getOrderId(),
+                    approval.getData().getPayerId()));
+            tokenKey = "token";
+        }
+        try {
+            JSONObject urlResponseData = parseUrlResponseData(deepLinkUri, payPalResponse.getSuccessUrl(), payPalResponse.getApprovalUrl(), tokenKey);
+            PayPalAccount payPalAccount = new PayPalAccount();
+            payPalAccount.setClientMetadataId(payPalResponse.getClientMetadataId());
+            payPalAccount.setSource("paypal-browser"); //TODO: check for valid sources
```
sarahkoop, line 183: It looks like [`BTPayPalNativeTokenizationRequest`](https://github.com/braintree/braintree_ios/blob/pp-checkout-module/Sources/BraintreePayPalNative/Public/BraintreePayPalNative/BTPayPalNativeTokenizationRequest.swift) in the iOS SDK aligns with `PayPalAccount` here, so we might be able to take `source` from the `clientMetadataId` here.

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeVaultRequest.java
```diff
@@ -0,0 +1,6 @@
+package com.braintreepayments.api;
+
+//TODO: once that native checkout supports vault, this class should be eliminated,
```
sarahkoop, line 3: I think we still need this class even with the vault fallback. We can probably remove this TODO and add doc strings from `PayPalVaultRequest` though.

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutRequest.java
```diff
@@ -1,7 +1,18 @@
...
- * Represents the parameters that are needed to start the PayPal Native Checkout flow
- */
-public class PayPalNativeCheckoutRequest {
+public class PayPalNativeCheckoutRequest extends PayPalCheckoutRequest {
+
+    /**
+     * @param amount The transaction amount in currency units (as * determined by setCurrencyCode).
+     *               For example, "1.20" corresponds to one dollar and twenty cents. Amount must be a non-negative
+     *               number, may optionally contain exactly 2 decimal places separated by '.', optional
+     *               thousands separator ',', limited to 7 digits before the decimal point.
```
sarahkoop, line 9:
```suggestion
     *               number, may optionally contain exactly 2 decimal places separated by '.'
     *               and is limited to 7 digits before the decimal point.
```
This doc string has been updated slightly on `PayPalCheckoutRequest` and the change should apply here also

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutRequest.java
```diff
@@ -1,7 +1,18 @@
 package com.braintreepayments.api;
 
-/**
- * Represents the parameters that are needed to start the PayPal Native Checkout flow
```
sarahkoop, line 4: Let's keep this doc string for the class

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeOnActivityResumedCallback.java
```diff
@@ -0,0 +1,6 @@
+package com.braintreepayments.api;
+
+public interface PayPalNativeOnActivityResumedCallback {
```
sarahkoop, line 3: Let's add a doc string for this (see the doc string pattern for any other OnActivityResumed callback)

PayPalNative/src/test/java/com/braintreepayments/api/PayPalNativeClientTest.java
```diff
@@ -0,0 +1,759 @@
...
+import static org.mockito.ArgumentMatchers.isNull;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@PrepareForTest({PayPalCheckout.class, Uri.class})
+@RunWith(RobolectricTestRunner.class)
+public class PayPalNativeClientTest {
```
sarahkoop, line 47: Let's rename this to `PayPalNativeClientUnitTest` to follow our test naming convention

PayPalNative/src/test/java/com/braintreepayments/api/PayPalNativeCheckoutRequestTest.java
```diff
@@ -0,0 +1,62 @@
...
+import org.robolectric.RobolectricTestRunner;
+
+import static junit.framework.Assert.assertEquals;
+import static junit.framework.Assert.assertFalse;
+import static junit.framework.Assert.assertNull;
+import static junit.framework.Assert.assertTrue;
+import static org.junit.Assert.assertNotNull;
+
+@RunWith(RobolectricTestRunner.class)
+public class PayPalNativeCheckoutRequestTest {
```
sarahkoop, line 14: Let's rename this to `PayPalNativeCheckoutRequestUnitTest`

PayPalNative/src/test/java/com/braintreepayments/api/PayPalNativeClientTest.java
```diff
@@ -0,0 +1,759 @@
...
+    private Configuration payPalDisabledConfig;
+    private Configuration payPalNoClientIdConfig;
+
+    private PayPalNativeOnActivityResumedCallback payPalNativeOnActivityResumedCallback;
+    private PayPalNativeTokenizeCallback payPalNativeTokenizeCallback;
+    private PayPalClient payPalClient;
+
+    @Rule
+    public PowerMockRule rule = new PowerMockRule();
+
```
sarahkoop, line 61: Nitpick but can we remove the duplicate new lines throughout this file?

PayPalNative/src/test/java/com/braintreepayments/api/PayPalNativeClientTest.java
```diff
@@ -0,0 +1,759 @@
...
+        doAnswer(new Answer<Void>() {
+            @Override
+            public Void answer(InvocationOnMock invocation) throws Throwable {
+                PayPalFlowStartedCallback callback = invocation.getArgument(2, PayPalFlowStartedCallback.class);
+                callback.onResult(new BraintreeException(errorMessage));
+
+                ArgumentCaptor<Exception> errorCaptor = ArgumentCaptor.forClass(Exception.class);
+                verify(payPalNativeTokenizeCallback).onResult(isNull(), errorCaptor.capture());
+                assertTrue(errorCaptor.getValue() instanceof BraintreeException);
+                assertEquals(errorMessage, errorCaptor.getValue().getMessage());
```
sarahkoop, line 124: For all of these tests lets move the test assertions outside of the mocking. I think these can all go after `sut.tokenizePayPalAccount` (this comment applies to all the tests in this file)

PayPalNative/src/test/java/com/braintreepayments/api/PayPalNativeVaultRequestTest.java
```diff
@@ -0,0 +1,47 @@
...
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+
+import static junit.framework.Assert.assertEquals;
+import static junit.framework.Assert.assertFalse;
+import static junit.framework.Assert.assertNull;
+import static junit.framework.Assert.assertTrue;
+
+@RunWith(RobolectricTestRunner.class)
+public class PayPalNativeVaultRequestTest {
```
sarahkoop, line 13: `PayPalNativeRequestUnitTest`

PayPalNative/src/androidTest/java/com/braintreepayments/api/ExampleInstrumentedTest.java
```diff
@@ -15,7 +15,7 @@
  *
  * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
  */
-@RunWith(AndroidJUnit4.class)
+@RunWith(RobolectricTestRunner.class)
 public class ExampleInstrumentedTest {
```
sarahkoop, line 19: I think this entire file can be deleted (it looks like we accidentally included it when creating this new module).

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeOnActivityResumedCallback.java
jcnoriega: this class is no longer needed, i forgot to delete it

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
jcnoriega: @sarahkoop sorry i dont understand, how do i get the `source` from the `clientMetadataId`?

PayPalNative/src/test/java/com/braintreepayments/api/PayPalNativeClientTest.java
jcnoriega: Im going to apply it to other tests, but this one im going to delete it because it doesnt apply now

Demo/src/main/java/com/braintreepayments/demo/PayPalNativeFragment.java
```diff
@@ -4,35 +4,76 @@
...
+    private DataCollector dataCollector;
+
+
+    private Button vaultButton;
+    private Button checkoutButton;
 
     @Override
     public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
-        return inflater.inflate(R.layout.fragment_paypal_native, container, false);
+        final View view = inflater.inflate(R.layout.fragment_paypal_native, container, false);
```
sshropshire, line 39:
```suggestion
       View view = inflater.inflate(R.layout.fragment_paypal_native, container, false);
```

sshropshire (approved):
🌊

PayPalNative/src/main/java/com/braintreepayments/api/PayPalNativeClient.java
sarahkoop: Ah I think I was mistaken. We can leave this as a TODO for now.

sarahkoop (approved):

PR #391, opened 2021-05-21
Make Additional Methods Package-Private

sarahkoop:
### Summary of changes

- Make additional methods only used internally or by Drop-in package-private

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/Authorization.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/Authorization.java
@@ -59,15 +59,15 @@ abstract class Authorization {
         return rawValue;
     }
 
-    private static boolean isTokenizationKey(String tokenizationKey) {
+    static boolean isTokenizationKey(String tokenizationKey) {
         return !TextUtils.isEmpty(tokenizationKey) && tokenizationKey.matches(TokenizationKey.MATCHER);
     }
 
-    private static boolean isPayPalUAT(String payPalUAT) {
+    static boolean isPayPalUAT(String payPalUAT) {
         return !TextUtils.isEmpty(payPalUAT) && payPalUAT.matches(PayPalUAT.MATCHER);
     }
 
-    private static boolean isClientToken(String clientToken) {
+    static boolean isClientToken(String clientToken) {
         return !TextUtils.isEmpty(clientToken) && clientToken.matches(ClientToken.BASE_64_MATCHER);
     }
 }
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/PaymentMethodClient.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/PaymentMethodClient.java
@@ -16,18 +16,18 @@ import java.util.List;
 /**
  * Class used to retrieve a customer's payment methods.
  */
-class PaymentMethodClient {
+public class PaymentMethodClient {
 
     private static final String PAYMENT_METHOD_NONCE_COLLECTION_KEY = "paymentMethods";
 
-    protected static final String SINGLE_USE_TOKEN_ID = "singleUseTokenId";
-    protected static final String VARIABLES = "variables";
-    protected static final String INPUT = "input";
-    protected static final String CLIENT_SDK_META_DATA = "clientSdkMetadata";
+    private static final String SINGLE_USE_TOKEN_ID = "singleUseTokenId";
+    private static final String VARIABLES = "variables";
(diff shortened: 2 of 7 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/Authorization.java
```diff
@@ -59,15 +59,15 @@ public String toString() {
         return rawValue;
     }
 
-    private static boolean isTokenizationKey(String tokenizationKey) {
```
sarahkoop, line 62: Drop-in uses these methods so making them package-private

BraintreeCore/src/main/java/com/braintreepayments/api/PaymentMethodClient.java
```diff
@@ -16,7 +16,7 @@
 /**
  * Class used to retrieve a customer's payment methods.
  */
-public class PaymentMethodClient {
```
sarahkoop, line 19: I believe this class should only be used by Drop-in

GooglePay/src/main/java/com/braintreepayments/api/BraintreeWalletConstants.java
```diff
@@ -11,5 +11,5 @@
     /**
      * Card network Elo.
      */
-    public static final int CARD_NETWORK_ELO = WalletConstants.CARD_NETWORK_OTHER + 1;
+    static final int CARD_NETWORK_ELO = WalletConstants.CARD_NETWORK_OTHER + 1;
```
sarahkoop, line 14: This is used internally by `GooglePayCilent` - I don't see where a merchant would need this.

BraintreeCore/src/main/java/com/braintreepayments/api/PaymentMethodClient.java
```diff
@@ -70,7 +70,7 @@ public PaymentMethodClient(BraintreeClient braintreeClient) {
      *        payment methods will be ordered by most recently added.
      * @param callback {@link GetPaymentMethodNoncesCallback}
      */
-    public void getPaymentMethodNonces(boolean defaultFirst, final GetPaymentMethodNoncesCallback callback) {
+    void getPaymentMethodNonces(boolean defaultFirst, final GetPaymentMethodNoncesCallback callback) {
```
sestevens, line 73: This seems like a method that would be useful to merchants

BraintreeCore/src/main/java/com/braintreepayments/api/Authorization.java
scannillo: My understand of package-private is that it can only be seen and used by the package in which it was declared. So doesn't this mean drop-in can't use it then, and it would need to be public? (Or protected if something in Drop-In subclasses `Authorization`)?

sarahkoop: We refactoring Drop-in to follow the same package structure as core so that we can use the package-private methods there as well.

BraintreeCore/src/main/java/com/braintreepayments/api/PaymentMethodClient.java
sarahkoop: Ok I wasn't sure on this - I couldn't find any documentation in our dev docs on any of the `PaymentMethodClient` methods, and our only usage of them was in Drop-in.  If merchants do need these methods I can put them back to public.

BraintreeCore/src/main/java/com/braintreepayments/api/Authorization.java
scannillo: Ah - I didn't remember that. Thanks!

BraintreeCore/src/main/java/com/braintreepayments/api/PaymentMethodClient.java
scannillo: We have a [similar public method on iOS](https://github.com/braintree/braintree_ios/blob/8b56f308f69ca4500a3c699fd1d2f55f2d6891f3/Sources/BraintreeCore/Public/BraintreeCore/BTAPIClient.h#L85-L94) `fetchPaymentMethodNonces`. 

May as well keep things consistent and have it be public on both (or hidden on both). I can see this being useful to merchants.

scannillo (approved):

sestevens (approved):

PR #393, opened 2021-05-26
[DO NOT MERGE] v4 Audit

demerino:
Making notes of possible changes and topics for discussion before GA

- @demerino


BraintreeDataCollector/src/main/java/com/braintreepayments/api/DataCollector.java
```diff
@@ -136,6 +139,7 @@ public String getPayPalClientMetadataId(Context context) {
         return "";
     }
 
+    // this method can be removed, we won't be using it
```
sarahkoop, line 142: This method submits risk data to Magnes - do we no longer need to be doing that?

PR #398, opened 2021-06-01
Add Getters to Data Classes

sshropshire:
### Summary of changes

 - Add getters to data classes to improve support for Kotlin synthesized properties
 - Replace Boxed types with primitives where possible

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop
- @sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
   * Remove `PaymentMethodDeleteException`
   * Remove `GetPaymentMethodNoncesCallback`
   * Remove `DeletePaymentMethodNonceCallback`
+  * Update `GooglePayIsReadyToPayCallback` to expect a boolean primitive instead of a Boxed type
 
 ## 4.0.0-beta3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## unreleased
 
 * Make `PayPalRequest` and subclasses `Parcelable`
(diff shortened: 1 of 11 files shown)


Card/src/main/java/com/braintreepayments/api/BaseCard.java
```diff
@@ -220,6 +220,104 @@ public void setExtendedAddress(String extendedAddress) {
...
+
+    /**
+     * @return The expiration year of the card.
+     */
+    public String getExpirationYear() {
+        return expirationYear;
+    }
+
+    /**
+     * @return The extended Address address of the card.
```
sestevens, line 273:
```suggestion
     * @return The extended address of the card.
```

CHANGELOG.md
```diff
@@ -19,6 +20,7 @@
   * Remove `PaymentMethodDeleteException`
   * Remove `GetPaymentMethodNoncesCallback`
   * Remove `DeletePaymentMethodNonceCallback`
+  * Update `GooglePayIsReadyToPayCallback` to expect a boolean primitive instead of a Boxed type
```
sestevens, line 23: I noticed that there's a few other places that we've updated to use primitives instead of boxed types. Any reason we're calling out this one and not the others?

sestevens (approved):

scannillo (approved):
Code changes look good. +1 to Susan's question about the CHANGELOG.

I think we either include all individual methods that changed to primitive types, or have a blanket statement for updating all boxed to primitives.

PR #400, opened 2021-06-02
Add displayName to VenmoContext creation

sshropshire:
### Summary of changes

 - Add `displayName` property to `VenmoRequest`

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop
- @sshropshire

### DIFF
--- a/Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
+++ b/Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
@@ -106,6 +106,9 @@ public class VenmoClient {
                     input.put("merchantProfileId", venmoProfileId);
                     input.put("customerClient", "MOBILE_APP");
                     input.put("intent", "CONTINUE");
+
+                    input.putOpt("displayName", request.getDisplayName());
+
                     JSONObject variables = new JSONObject();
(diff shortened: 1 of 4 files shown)


demerino (approved):

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -106,6 +106,9 @@ public void onResult(@Nullable final Configuration configuration, @Nullable Exce
                     input.put("merchantProfileId", venmoProfileId);
                     input.put("customerClient", "MOBILE_APP");
                     input.put("intent", "CONTINUE");
+
+                    input.putOpt("displayName", request.getDisplayName());
```
scannillo, line 110: In our iOS PR - we aren't adding the "displayName" if it isn't set. It probably doesn't matter / the graphQL api should be able to handle either, but figured I'd bring it up.

sshropshire: Definitely yeah good callout. In this case if `getDisplayName()` returns null it should no-op.

sestevens (approved):

PR #402, opened 2021-06-03
Add Kotlin Unit Tests For Card and UnionPayCard

sarahkoop:
### Summary of changes

 - Add Kotlin unit tests for `Card` and `UnionPayCard`
 - Fix bugs in Kotlin properties uncovered by Kotlin unit tests

 ### Checklist

- @sarahkoop


Card/src/main/java/com/braintreepayments/api/BaseCard.java
```diff
@@ -32,20 +32,21 @@
     static final String REGION_KEY = "region";
     static final String STREET_ADDRESS_KEY = "streetAddress";
 
-    protected String cardholderName;
```
sarahkoop, line 35: When these properties are `protected` and set in Kotlin using the properties like `card.cardHolderName = "some name"`, the underlying setter method is not invoked for some reason, and the properties are set directly. This means that the conversion from empty string to `null` doesn't happen. This led to test failures in the tests around not adding emptying strings to the `json` objects. Updating these to be `private` and using the getters to access them in `Card` and `UnionPayCard` subclasses resolved the problem.

Card/src/main/java/com/braintreepayments/api/BaseCard.java
```diff
@@ -101,6 +102,7 @@ public void setExpirationDate(@Nullable String expirationDate) {
         if (TextUtils.isEmpty(expirationDate)) {
             expirationMonth = null;
             expirationYear = null;
+            this.expirationDate = null;
```
sarahkoop, line 105: We had a setExpirationDate method, but no getter so the property wasn't synthesized in Kotlin. There wasn't a member variable for expirationDate since it was directly being converted to month and year in the setter, so I added a variable to allow for a getter and the property in Kotlin. Open to other ideas on how to fix this property.

scannillo (approved):

Card/src/main/java/com/braintreepayments/api/BaseCard.java
sestevens: Another option, instead of adding the member variable, would be to have the expiration date setter reconstruct the expiration date from the month and year member variables.

sestevens (approved):

Card/src/main/java/com/braintreepayments/api/BaseCard.java
sarahkoop: Good call - we decided to remove the local variable and reconstruct it 👍

PR #405, opened 2021-06-08
Update Cardinal Credentials and Bump to 2.2.5-3

sarahkoop:
### Summary of changes

 - Bump Cardinal version to `2.2.5-3`
 - Update Cardinal credentials in demo project and migration guide

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,10 +54,10 @@ nexusStaging {
 subprojects {
     repositories {
         maven {
-            url  "https://cardinalcommerce.bintray.com/android"
+            url "https://cardinalcommerceprod.jfrog.io/artifactory/android"
             credentials {
-                username 'braintree-team-sdk@cardinalcommerce'
-                password '220cc9476025679c4e5c843666c27d97cfb0f951'
+                username 'braintree_team_sdk'
+                password 'AKCp8jQcoDy2hxSWhDAUQKXLDPDx6NYRkqrgFLRc3qDrayg6rrCbJpsKKyMwaykVL8FWusJpp'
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -54,10 +54,10 @@ nexusStaging {
 subprojects {
     repositories {
         maven {
-            url  "https://cardinalcommerce.bintray.com/android"
+            url "https://cardinalcommerceprod.jfrog.io/artifactory/android"
             credentials {
-                username 'braintree-team-sdk@cardinalcommerce'
-                password '220cc9476025679c4e5c843666c27d97cfb0f951'
+                username 'braintree_team_sdk'
```
sarahkoop, line 59: Should there be a CHANGELOG entry for the credentials change? It's more of an integration change on the merchant side so not sure

sestevens: Hmm... I think it would be helpful to call it out in the changelog. It could be a note underneath the changelog entries for the next version that calls out that merchants using 3DS will need to update their credentials.

sestevens (approved):
I like the idea of adding a note about the new credentials to the changelog, but other than that, this looks good!

sarahkoop:
> Looks good! We should also update our docs to mention this change b/w Android versions

I've opened a Docs PR for this change! Upgrading to this Cardinal version requires the new credentials, but even merchants using v3 should update their credentials because there is no guarantee that the Bintray read-only repository will remain up.

demerino (approved):

PR #407, opened 2021-06-15
Rename TokenizationClient to ApiClient

sshropshire:
### Summary of changes

- Rename TokenizationClient to ApiClient
- Update HttpResponseCallback interface to contain a single method

 ### Checklist

### Authors

- @sarahkoop 
- @sshropshire


BraintreeCore/src/androidTest/java/com/braintreepayments/api/BraintreeHTTPClientTest.java
```diff
@@ -24,18 +24,14 @@ public void setup() {
...
-        braintreeHttpClient.get("https://api.sandbox.braintreegateway.com/", null, new HttpResponseCallback() {
-            @Override
-            public void success(String responseBody) {
-                fail("Request should fail");
-            }
+        braintreeHTTPClient.get("https://api.sandbox.braintreegateway.com/", null, new HTTPResponseCallback() {
 
             @Override
-            public void failure(Exception exception) {
+            public void onResult(String responseBody, Exception httpError) {
```
sarahkoop, line 32: Should we assert that `responseBody` is null for these tests? Or an equivalent assertion to cover the previous fail("Request should fail") assertion on success?

PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
```diff
@@ -50,9 +50,15 @@ public void onResult(@Nullable Configuration configuration, @Nullable Exception
 
                     String requestBody = payPalRequest.createRequestBody(configuration, braintreeClient.getAuthorization(), successUrl, cancelUrl);
 
-                    braintreeClient.sendPOST(url, requestBody, new HttpResponseCallback() {
+                    braintreeClient.sendPOST(url, requestBody, new HTTPResponseCallback() {
+
                         @Override
-                        public void success(String responseBody) {
+                        public void onResult(String responseBody, Exception httpError) {
+                            if (httpError != null) {
```
sarahkoop, line 57: Nitpick - seems like all the other HttpResponseCallback null checks in this PR use `if (responseBody != null) { } else { } ` maybe we should follow the same pattern in this class?

SharedUtils/src/main/java/com/braintreepayments/api/HTTPClientException.java
```diff
@@ -3,9 +3,9 @@
 /**
  * Exception thrown when an HTTP request fails.
  */
-public class HttpClientException extends Exception {
+public class HTTPClientException extends Exception {
```
sarahkoop, line 6: Changing this public exception class name would be a breaking change right? Might have to leave this as is 😞

sarahkoop (approved):
Other than reverting the `HttpClientException` name change - looks good to me!

SharedUtils/src/main/java/com/braintreepayments/api/HTTPClientException.java
sshropshire: Ah true 🤦  I'll revert this one and maybe put up a next major version comment.

PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
sshropshire: 👍  most def will change.

BraintreeCore/src/androidTest/java/com/braintreepayments/api/BraintreeHTTPClientTest.java
sshropshire: I think the assertions will fail properly. There were originally two code paths so the `fail()` would prevent a false positive where no assertions were invoked. Now that both result/error are in the same callback it should be cool.

BraintreeCore/src/main/java/com/braintreepayments/api/APIClient.java
```diff
@@ -7,18 +7,18 @@
 
 import java.lang.ref.WeakReference;
 
-class TokenizationClient {
+class APIClient {
```
scannillo, line 10: Why are we opting to rename this?

SharedUtils/src/main/java/com/braintreepayments/api/HttpNoResponse.java
```diff
@@ -1,12 +0,0 @@
-package com.braintreepayments.api;
-
-class HttpNoResponse implements HttpResponseCallback {
-
-    @Override
-    public void success(String responseBody) {
```
scannillo, line 6: Why we opting now for `onResult` instead of `success` and `failure` callbacks?

BraintreeCore/src/main/java/com/braintreepayments/api/APIClient.java
sshropshire: Mostly to better align with iOS. The discussion started [here](https://github.com/braintree/braintree_android/pull/364#discussion_r641822294) on another PR.

sshropshire: We did talk about renaming tokenizeRest and tokenizeGraphQL to `vault()` as well if that would make more sense. I forgot to add that change.

SharedUtils/src/main/java/com/braintreepayments/api/HttpNoResponse.java
sshropshire: We use the [Single Abstract Method](https://www.javacodemonk.com/single-abstract-method-sam-and-functional-interface-in-java-44f22fc3#:~:text=Interface%20that%20has%20single%20abstract,has%20exactly%20one%20abstract%20method.) pattern throughout the sdk. We deferred this change for a while since it was internal. SAM is really good for Kotlin interop though, if we end up converting more classes to Kotlin in the future this will make it a lot easier.

BraintreeCore/src/main/java/com/braintreepayments/api/APIClient.java
sshropshire: There is also a really peculiar arch issue with tokenizeGraphQL and tokenizeREST. This class used to have a circular dependency on the `Card` module. To eliminate, we made `tokenizeGraphQL` take a `JSONObject` payload, and `tokenizeREST` still takes a `PaymentMethod` to get the correct `POST` route for a particular payment method.

It might make sense to go with `vaultUsingGraphQL` and `vaultUsingREST` here, but there's still something to figure out internally before we can make it a single method. I believe Susan was saying that tokenization doesn't always happen.

SharedUtils/src/main/java/com/braintreepayments/api/HttpNoResponse.java
scannillo: Awesome - thank you for this context!

BraintreeCore/src/main/java/com/braintreepayments/api/APIClient.java
sarahkoop: The vault naming still seems a little confusing to me. For the Venmo case it makes sense because we only call `tokenizeREST` for vaulting, but it seems like we call `tokenizeREST` in `PayPalClient#onBrowserSwitchResult` for all PayPal browser switch results. 

It looks like iOS has separate [tokenize](https://github.com/braintree/braintree_ios/blob/68ca8ee70daa279ebfd5bbea62e52212f948d6d7/Sources/BraintreeVenmo/BTVenmoDriver.m#L160) and [vault](https://github.com/braintree/braintree_ios/blob/68ca8ee70daa279ebfd5bbea62e52212f948d6d7/Sources/BraintreeVenmo/BTVenmoDriver.m#L208) methods in the payment method drivers that call the underlying `BTAPIClient` `POST`  methods. That seems to make sense, but I'm not sure it applies the same way here unless we move the `POST` and `GET` methods from the `HTTPClient` to `APIClient` and call them in a similar way from the payment method clients.

SharedUtils/src/main/java/com/braintreepayments/api/HttpNoResponse.java
sshropshire: 🙏

BraintreeCore/src/main/java/com/braintreepayments/api/APIClient.java
sshropshire: Yeah the naming confuses me a bit also. I think we could legitimately defer the `vault()` naming change since `APIClient` in iOS doesn't abstract HTTP, and this `APIClient` keeps the HTTP communication under the hood. There isn't a direct 1-to-1 correlation between the two classes, but we could one day make it a point to align APIClient for both iOS and Android (and maybe JS 🤷‍♂️ ) since they're internal classes.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
```diff
@@ -48,7 +48,7 @@ private static BraintreeClientParams createDefaultParams(Context context, String
      * Create a new instance of {@link BraintreeClient} using a tokenization key or client token.
      *
      * @param context       Android Context
-     * @param authorization The tokenization key or client token to use. If an invalid authorization is provided, a {@link BraintreeException} will be returned in the {@link HttpResponseCallback}.
+     * @param authorization The tokenization key or client token to use. If an invalid authorization is provided, a {@link BraintreeException} will be returned in the {@link HTTPResponseCallback}.
```
sestevens, line 51: It's a bit confusing to me that we're referencing `HTTPResponseCallback` in this doc string. That's not a public interface, is it?

sshropshire: Done!

PR #411, opened 2021-07-06
Samsung Pay

sshropshire:
### Summary of changes

 - This PR adds Samsung Pay to the repo.

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sshropshire


SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayClient.java
```diff
@@ -0,0 +1,235 @@
...
+     */
+    public void goToUpdatePage() {
+        getInternalClient(new GetSamsungPayInternalClientCallback() {
+            @Override
+            public void onResult(@Nullable SamsungPayInternalClient internalClient, @Nullable Exception error) {
+                if (internalClient != null) {
+                    internalClient.goToSamsungPayUpdatePage();
+                    braintreeClient.sendAnalyticsEvent("samsung-pay.goto-update-page");
+                } else {
+                    // TODO: determine if we should notify an error here
```
sarahkoop, line 44: I'd say we callback the error here since it would likely be a configuration error from the internal client right? That would probably be helpful for the merchant

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayInternalClient.java
```diff
@@ -0,0 +1,164 @@
...
+                    samsungPayError = new SamsungPayException(error);
+                }
+
+                callback.onResult(statusCode, samsungPayError);
+            }
+
+            @Override
+            public void onFail(int errorCode, Bundle bundle) {
+                SamsungPayException exception = new SamsungPayException(errorCode);
+                callback.onResult(null, exception);
```
sarahkoop, line 105: Why not call back the error code instead of null here?

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayClient.java
```diff
@@ -0,0 +1,235 @@
...
+                if (internalClient != null) {
+                    internalClient.getSamsungPayStatus(callback);
+                } else {
+                    callback.onResult(null, error);
+                }
+            }
+        });
+    }
+
+    private void getBraintreeSupportedSamsungPayCards(final GetAcceptedCardBrandsCallback callback) {
```
sarahkoop, line 136: Can we make this method name align with the internal client method and callback name?

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayInternalClient.java
```diff
@@ -0,0 +1,164 @@
...
+                                break;
+                            case ERROR_SPAY_SETUP_NOT_COMPLETED:
+                                error = SamsungPayError.SAMSUNG_PAY_SETUP_NOT_COMPLETED;
+                                break;
+                        }
+                    } else {
+                        if (statusCode == SPAY_NOT_READY) {
+                            error = SamsungPayError.SAMSUNG_PAY_NOT_READY;
+                        } else {
+                            // SPAY_NOT_SUPPORTED
```
sarahkoop, line 92: Couldn't this also be an unknown error? Might be safer to assert on the actual error code here too

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayInternalClient.java
```diff
@@ -0,0 +1,164 @@
...
+
+            @Override
+            public void onFailure(int errorCode, Bundle bundle) {
+                Exception error = new SamsungPayException(errorCode);
+                callback.onResult(null, error);
+            }
+        });
+    }
+
+    void startSamsungPay(CustomSheetPaymentInfo customSheetPaymentInfo, final SamsungPayStartListener listener) {
```
sarahkoop, line 133: Would a more generic name like `SamsungPayListener` make sense here? It's a little misleading since the listener is also being called when the card info is updated, not just on start.

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayInternalClient.java
```diff
@@ -0,0 +1,164 @@
...
+                    listener.onSamsungPayStartSuccess(samsungPayNonce, customSheetPaymentInfo);
+                } catch (JSONException e) {
+                    listener.onSamsungPayStartError(e);
+                }
+            }
+
+            @Override
+            public void onFailure(int errorCode, Bundle bundle) {
+                if (errorCode == SpaySdk.ERROR_USER_CANCELED) {
+                    UserCanceledException userCanceledError = new UserCanceledException("User Canceled");
```
sarahkoop, line 155:
```suggestion
                    UserCanceledException userCanceledError = new UserCanceledException("User canceled Samsung Pay.");
```

nitpick just to match the other payment flows.

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayClient.java
sshropshire: Yeah it's just a bit wack since the method is asynchronous for the Braintree SDK, but for SamsungPay itself it's a synchronous operation. I'm down to add a callback.

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayInternalClient.java
sshropshire: Yeah that makes sense. Technically there are 3 possible values for get status, but it'll be good to be defensive here just in case.

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayInternalClient.java
sshropshire: Yeah the assumption I went with here was a `null` value communicates via callback that an Error occurred. Also not sure if there could be a potential collision between SamsungPay status codes and SamsungPay error codes.

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayInternalClient.java
sshropshire: I agree yes. We had to add a `Listener` style callback here because of the `cardInfoUpdated` method.

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayClient.java
```diff
@@ -0,0 +1,235 @@
...
+     * the listener will be called back with a value of false. If the Samsung Pay callback returns an error,
+     * see {@link SamsungPayError} for a list of possible error codes
+     *
+     * @param callback {@link SamsungPayIsReadyToPayCallback}
+     */
+    public void isReadyToPay(final SamsungPayIsReadyToPayCallback callback) {
+        getSamsungPayStatus(new GetSamsungPayStatusCallback() {
+            @Override
+            public void onResult(@Nullable Integer status, @Nullable Exception samsungPayError) {
+                if (status != null) {
```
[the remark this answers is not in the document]
sshropshire: I think we could explore that possibility yeah. The `SamsungPayError` type exists to bring the error codes into our domain. We could do some mapping to make sure we've covered all the bases. But mainly, wanted to have an `isReadyToPay` method similar to `GooglePay` and `VenmoPay`. Form-fitting SamsungPay to this pattern wasn't super straightforward. I'm up for taking another look.

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayClient.java
sshropshire: Most def.

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayPartnerInfoBuilder.java
```diff
@@ -0,0 +1,71 @@
...
+import static com.samsung.android.sdk.samsungpay.v2.SpaySdk.PARTNER_SERVICE_TYPE;
+import static com.samsung.android.sdk.samsungpay.v2.SpaySdk.ServiceType.INAPP_PAYMENT;
+import static com.samsung.android.sdk.samsungpay.v2.payment.PaymentManager.EXTRA_KEY_TEST_MODE;
+
+class SamsungPayPartnerInfoBuilder {
+
+    private static final String API_VERSION_KEY = "braintreeTokenizationApiVersion";
+    private static final String CLIENT_SDK_METADATA_KEY = "clientSdkMetadata";
+
+    private static final String API_VERSION = "2018-10-01";
```
sarahkoop, line 19: Is this accurate? Is there a newer api version we should be upgrading to?

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayStartListener.java
```diff
@@ -0,0 +1,34 @@
...
+ * {@link SamsungPayClient#startSamsungPay(CustomSheetPaymentInfo, SamsungPayStartListener)}.
+ */
+public interface SamsungPayStartListener {
+
+    /**
+     * Called when Samsung Pay card info has been updated. {@link SamsungPayClient#updateCustomSheet(CustomSheet)}
+     * @param cardInfo SamsungPay [CardInfo]
+     * @param customSheet SamsungPay [CustomSheet]
+     */
+    void onSamsungPayCardInfoUpdated(CardInfo cardInfo, CustomSheet customSheet);
```
sarahkoop, line 20: Are these @NonNull?

SamsungPay/src/test/java/com/braintreepayments/api/MockPaymentManagerBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+                if (cardInfo != null && customSheet != null) {
+                    callback.onCardInfoUpdated(cardInfo, customSheet);
+                } else if (startInAppPayWithCustomSheetPaymentInfo != null && startInAppPayWithCustomSheetSuccess != null) {
+                    callback.onSuccess(startInAppPayWithCustomSheetPaymentInfo, startInAppPayWithCustomSheetSuccess, new Bundle());
+                } else if (startInAppPayWithCustomSheetError != null) {
+                    callback.onFailure(startInAppPayWithCustomSheetError, new Bundle());
+                }
+                return null;
+            }
+        }).when(paymentManager).startInAppPayWithCustomSheet(any(CustomSheetPaymentInfo.class), any (PaymentManager.CustomSheetTransactionInfoListener.class));
```
sarahkoop, line 91:
```suggestion
        }).when(paymentManager).startInAppPayWithCustomSheet(any(CustomSheetPaymentInfo.class), any(PaymentManager.CustomSheetTransactionInfoListener.class));
```

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayPartnerInfoBuilder.java
sshropshire: Yeah but we can rename to `BRAINTREE_API_VERSION` I think to be more clear.

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayStartListener.java
sshropshire: Good call, will have to check.

v4_MIGRATION_GUIDE.md
```diff
@@ -536,7 +536,99 @@ However, `requestOneTimePayment` and `requestBillingAgreement` have been depreca
...
+  protected void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    braintreeClient = new BraintreeClient(this, "TOKENIZATION_KEY_OR_CLIENT_TOKEN");
+    samsungPayClient = new SamsungPayClient(braintreeClient);
+  }
+
+  private void checkIfSamsungPayIsAvailable() {
+    samsungPayClient.isReadyToPay((isReadyToPay, error) -> {
+      if (isReadyToPay) {
+        // Samsung Pay is available
```
sarahkoop, line 567: The merchant should call `launchSamsungPay()` here right?

v4_MIGRATION_GUIDE.md
```diff
@@ -536,7 +536,99 @@ However, `requestOneTimePayment` and `requestBillingAgreement` have been depreca
...
+    });
+  }
+
+  private CustomSheet getCustomSheet() {
+    CustomSheet sheet = new CustomSheet();
+
+    final AddressControl billingAddressControl = new AddressControl("billingAddressId", SheetItemType.BILLING_ADDRESS);
+    billingAddressControl.setAddressTitle("Billing Address");
+    billingAddressControl.setSheetUpdatedListener((controlId, customSheet) -> {
+      Log.d("billing sheet updated", controlId);
```
sarahkoop, line 595: Let's take out these log statements we don't have them in any other examples

v4_MIGRATION_GUIDE.md
```diff
@@ -536,7 +536,99 @@ However, `requestOneTimePayment` and `requestBillingAgreement` have been depreca
...
+  }
+
+  @Override
+  public void onSamsungPayStartError(@NonNull Exception error) {
+    // handle samsung pay start error
+  }
+
+  @Override
+  public void onSamsungPayStartSuccess(@NonNull SamsungPayNonce samsungPayNonce, CustomSheetPaymentInfo paymentInfo) {
+    // send samsungPayNonce to server to create a transaction
```
sarahkoop, line 622:
```suggestion
    // send samsungPayNonce.getString() to server to create a transaction
```

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayClient.java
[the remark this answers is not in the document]
sarahkoop: Actually - the above comment about why not to return the status code instead of null for error status codes makes sense, so leaving this as nullable is probably the right call.

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayClient.java
sshropshire: Added callbacks!

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayClient.java
sshropshire: Done!

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayInternalClient.java
sshropshire: Done!

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayInternalClient.java
sshropshire: Done!

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayStartListener.java
sshropshire: I think they're all non-null here since it's more listener style than callback / result-error style. I updated the annotations in the listener.

v4_MIGRATION_GUIDE.md
sshropshire: Good call yeah I think they may want to display the button here. We may want to update other docs Google Pay says the same thing.

v4_MIGRATION_GUIDE.md
sshropshire: Good catch! Done!

sarahkoop (approved):
Pending testing the flow on a physical device - looks good to me 🌟

PR #416, opened 2021-07-09
Upgrade Cardinal Version to 2.2.5-4

sarahkoop:
### Summary of changes

 - Bump Cardinal version to `2.2.5-4`

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop 
- @sshropshire

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.java
@@ -151,10 +151,7 @@ class AnalyticsDatabase extends SQLiteOpenHelper {
                         request.id = Integer.valueOf(ids[i]);
                         request.event = events[i];
                         request.timestamp = Long.valueOf(timestamps[i]);
-                        int columnIndex = cursor.getColumnIndex(META_JSON);
-                        if (columnIndex >= 0) {
-                            request.metadata = new JSONObject(cursor.getString(columnIndex));
-                        }
+                        request.metadata = new JSONObject(cursor.getString(cursor.getColumnIndex(META_JSON)));
                         innerList.add(request);
                     } catch (JSONException ignored) {}
                 }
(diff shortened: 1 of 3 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.java
```diff
@@ -151,7 +151,10 @@ public void run() {
                         request.id = Integer.valueOf(ids[i]);
                         request.event = events[i];
                         request.timestamp = Long.valueOf(timestamps[i]);
-                        request.metadata = new JSONObject(cursor.getString(cursor.getColumnIndex(META_JSON)));
+                        int columnIndex = cursor.getColumnIndex(META_JSON);
+                        if (columnIndex >= 0) {
```
sarahkoop, line 155: The build was failing locally for me on this index potentially being 0... I'm not sure what changed to cause this error, but adding a check here resolved it.

crookedneighbor: can I get some more details on this bit? I'm a little confused about it. What is this doing exactly?

crookedneighbor (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.java
sarahkoop: I think this lint build error was related to the beta Android Studio version I have... We will probably need to make this change in the future but I'm going to revert it for now since it's not related to this PR.

kate-paypal (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.java
crookedneighbor: :+1:

PR #420, opened 2021-07-14
Add `cardAdd` to `ThreeDSecureRequest`

sarahkoop:
### Summary of changes

 -  Add `cardAdd` to `ThreeDSecureRequest`

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @kate-paypal 
- @sarahkoop

### DIFF
--- a/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
+++ b/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
@@ -317,6 +317,7 @@ public class ThreeDSecureRequest implements Parcelable {
         dest.writeByte(challengeRequested ? (byte) 1 : 0);
         dest.writeByte(dataOnlyRequested ? (byte) 1 : 0);
         dest.writeByte(exemptionRequested ? (byte) 1 : 0);
+        dest.writeByte(cardAdd ? (byte) 1 : 0);
         dest.writeParcelable(v2UiCustomization, flags);
         dest.writeParcelable(v1UiCustomization, flags);
         dest.writeString(accountType);
@@ -334,6 +335,7 @@ public class ThreeDSecureRequest implements Parcelable {
         challengeRequested = in.readByte() > 0;
         dataOnlyRequested = in.readByte() > 0;
         exemptionRequested = in.readByte() > 0;
+        cardAdd = in.readByte() > 0;
         v2UiCustomization = in.readParcelable(ThreeDSecureV2UiCustomization.class.getClassLoader());
         v1UiCustomization = in.readParcelable(ThreeDSecureV1UiCustomization.class.getClassLoader());
         accountType = in.readString();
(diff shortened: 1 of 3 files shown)


crookedneighbor (approved):

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
```diff
@@ -162,6 +163,17 @@ public void setExemptionRequested(boolean exemptionRequested) {
...
     }
 
+    /**
+     * Optional. An authentication created using this flag should only be used for card add operations
+     * (creation of customers' credit cards or payment methods) and not for creating transactions.
+     *
+     * @param cardAdd If set to true, card-add challenge will be requested from the issuer to confirm
+     *                adding new card to the merchant's vault.
+     */
+    public void setCardAdd(boolean cardAdd) {
```
scannillo, line 173: Can we update the name of this property to match the others on this class?

```suggestion
    public void setCardAddChallengeRequested(boolean cardAddChallengeRequested) {
```

Also - does setting this param to true result in an additional UI challenge? I don't understand totally what this param does?

sarahkoop: So this change has already been made in JS with the parameter named `cardAdd` - do we want to deviate from that for iOS and Android? 

I believe this parameter is to allow merchants to request the usual 3DS UI challenge when adding a card with a $0 amount (so no additional UI challenge).

scannillo: > So this change has already been made in JS with the parameter named cardAdd - do we want to deviate from that for iOS and Android?

I think it is worth renaming the one in JS too, if we can come up with a more clarifying name. `ThreeDSecureRequest.cardAdd` is really confusing / vague to me.

scannillo: > I believe this parameter is to allow merchants to request the usual 3DS UI challenge when adding a card with a $0 amount (so no additional UI challenge).

That makes sense, thanks for clarifying!

scannillo: Also, in general `card add operations` and `card-add challenge` mentioned in the docstring I think can be further clarified. I'm not sure a merchant knows what this means.

sarahkoop: > I think it is worth renaming the one in JS too, if we can come up with a more clarifying name. `ThreeDSecureRequest.cardAdd` is really confusing / vague to me.

i agree the method name is very confusing. It looks like the JS changes have already been released... does it make more sense to differ from JS until next major version and go with `cardAddChallengeRequested` here or keep the naming aligned?

sarahkoop: > Also, in general `card add operations` and `card-add challenge` mentioned in the docstring I think can be further clarified. I'm not sure a merchant knows what this means.

I updated these docstrings - let me know if they seem clear enough

scannillo: > does it make more sense to differ from JS until next major version and go with cardAddChallengeRequested here or keep the naming aligned?

I'm in favor of naming it how we think will be most easily-understood on mobile now, it also follows our existing convention / naming pattern in Android as-is.

sarahkoop: Ok agreed - updated the name to be `cardAddChallengeRequested` for clarity 👍

sshropshire (approved):

PR #423, opened 2021-07-20
Defer adding SamsungPay to CHANGELOG

sshropshire:
### Summary of changes

 - Defer adding SamsungPay to CHANGELOG until it can be properly tested.

 ### Checklist

 - [x] ~Added~ Removed a changelog entry

### Authors

- @sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,14 +8,6 @@
   * Make `ThreeDSecureResult#getLookup()` public
   * Bump Cardinal version to `2.2.5-4`
   * Add `cardAddChallengeRequested` to `ThreeDSecureRequest`
-* SamsungPay
-  * Add `BuildCustomSheetPaymentInfoCallback`
-  * Add `SamsungPayClient`
-  * Add `SamsungPayException`
-  * Add `SamsungPayNonce`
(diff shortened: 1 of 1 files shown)


sarahkoop (approved):

scannillo:
Shouldn't we also remove the actual SamsungPay code as well as the CHANGELOG entry?

sshropshire:
> Shouldn't we also remove the actual SamsungPay code as well as the CHANGELOG entry?

I think keeping the code will make it easier to release SamsungPay once we're able to verify it working. We technically won't be releasing it until we add the module to our `Rakefile`. It's almost like a `package-private` module now.

scannillo:
> I think keeping the code will make it easier to release SamsungPay once we're able to verify it working. We technically won't be releasing it until we add the module to our Rakefile. It's almost like a package-private module now.

Gotcha, so merchants can't access it in it's current state then? This is good to know!

sshropshire:
> Gotcha, so merchants can't access it in it's current state then? This is good to know!

Yup! The SamsungPay artifact won't get published to Sonatype until we explicitly add it to our release script.

PR #424, opened 2021-07-20
Update `cardAddChallengeRequested` Param to Pass False

sarahkoop:
### Summary of changes

 - Update `ThreeDSecureRequest` to include `cardAddChallengeRequested` param when set to true or false, and not include when null.

 ### Checklist

- @kate-paypal 
- @sarahkoop

### DIFF
--- a/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
+++ b/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
@@ -40,7 +40,7 @@ public class ThreeDSecureRequest implements Parcelable {
     private boolean challengeRequested = false;
     private boolean dataOnlyRequested = false;
     private boolean exemptionRequested = false;
-    private boolean cardAddChallengeRequested = false;
+    private Boolean cardAddChallengeRequested;
     private ThreeDSecureV2UiCustomization v2UiCustomization;
     private ThreeDSecureV1UiCustomization v1UiCustomization;
 
@@ -167,8 +167,12 @@ public class ThreeDSecureRequest implements Parcelable {
      * Optional. An authentication created using this flag should only be used for adding a
      * payment method to the merchant's vault and not for creating transactions.
      *
-     * @param cardAddChallengeRequested If set to true, the authentication challenge will be requested from the issuer
-     *                to confirm adding new card to the merchant's vault.
+     * @param cardAddChallengeRequested If set to true, the authentication challenge will be requested
+     *                                  from the issuer to confirm adding new card to the merchant's
+     *                                  vault. If not set and amount is 0, the authentication challenge
+     *                                  will be presented to the user. If set to false, when the amount
+     *                                  is 0, the authentication challenge will not be presented to the user.
+     *
      */
     public void setCardAddChallengeRequested(boolean cardAddChallengeRequested) {
         this.cardAddChallengeRequested = cardAddChallengeRequested;
--- a/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
+++ b/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
@@ -174,7 +174,7 @@ public class ThreeDSecureRequest implements Parcelable {
      *                                  is 0, the authentication challenge will not be presented to the user.
      *
      */
-    public void setCardAddChallengeRequested(boolean cardAddChallengeRequested) {
+    public void setCardAddChallengeRequested(@Nullable Boolean cardAddChallengeRequested) {
         this.cardAddChallengeRequested = cardAddChallengeRequested;
     }
 
@@ -290,9 +290,10 @@ public class ThreeDSecureRequest implements Parcelable {
 
     /**
      * @return If the authentication challenge will be requested from the issuer to confirm adding
-     * new card to the merchant's vault.
+     * new card to the merchant's vault or null if the value has not been set.
      */
-    public boolean isCardAddChallengeRequested() {
+    @Nullable
+    public Boolean isCardAddChallengeRequested() {
         return cardAddChallengeRequested;
     }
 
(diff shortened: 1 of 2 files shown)


ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
```diff
@@ -167,10 +167,14 @@ public void setExemptionRequested(boolean exemptionRequested) {
...
-     *                to confirm adding new card to the merchant's vault.
+     * @param cardAddChallengeRequested If set to true, the authentication challenge will be requested
+     *                                  from the issuer to confirm adding new card to the merchant's
+     *                                  vault. If not set and amount is 0, the authentication challenge
+     *                                  will be presented to the user. If set to false, when the amount
+     *                                  is 0, the authentication challenge will not be presented to the user.
+     *
      */
-    public void setCardAddChallengeRequested(boolean cardAddChallengeRequested) {
+    public void setCardAddChallengeRequested(@Nullable Boolean cardAddChallengeRequested) {
```
sshropshire, line 177: I can't remember exactly with the whole Kotlin conversion, but it'd be nice if we could still use `boolean` in the merchant facing API and store the variable in a `Boolean` internally.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
```diff
@@ -286,9 +290,10 @@ public boolean isExemptionRequested() {
 
     /**
      * @return If the authentication challenge will be requested from the issuer to confirm adding
-     * new card to the merchant's vault.
+     * new card to the merchant's vault or null if the value has not been set.
      */
-    public boolean isCardAddChallengeRequested() {
+    @Nullable
+    public Boolean isCardAddChallengeRequested() {
```
sshropshire, line 296: But on the other hand this makes perfect sense. Returns `null` if not set.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
sarahkoop: My thought was the getter has to be able to return null because the param can be null and doesn't the setter need to align with the getter for synthesized properties. Is there another way around that?

sshropshire: Yeah you're right I agree. The possibility of a collision with Kotlin property synthesis makes this our best option.

scannillo: `If not set and amount is 0, the authentication challenge will be presented to the user.`

Isn't this a breaking change?

sshropshire: It sounds like default behavior to me but I'm not sure. I wonder if this still holds true with the current version since this property not being set is equivalent to the property not existing.

scannillo: Thinking again about this, it's pretty confusing to me. It seems like a limitation of the GW API - I take this as if we pass `"card_add"` to the GW without a bool value and the amount is 0, it automatically induces a challenge? I would consider this a breaking change.

We could either talk to the GW team and get them to fix this before we put this in the SDKs .... OR .... simply not include the `"card_add"` param in the request to the GW unless the merchant has set it as true or false. This is assuming my understanding of how this is working, is correct! Let me know!

scannillo: @sshropshire - agreed, the behavior of this property not being set should be equivalent to it not existing in previous SDK versions.

sshropshire: Yeah tbh I'm not 100% clear on the gateway behavior either. On our side we do make `null` the default value by using a boxed type `Boolean` instead of a primitive. Then we do a null check before adding the `card_add` parameter to the payload.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
```diff
@@ -381,7 +386,7 @@ public String build(String dfReferenceId) {
             base.put("additional_info", additionalInfo);
             base.putOpt("account_type", accountType);
 
-            if (cardAddChallengeRequested) {
+            if (cardAddChallengeRequested != null) {
```
sshropshire, line 389: @scannillo this was the one change we had to add to make that possible, otherwise the parameter would always be set whether `true` or `false`. Now when this values `null`, the request is the same as the current version, and the behavior is up to the gateway.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
sarahkoop: Ok we are going to follow up with 3DS to confirm the GW behavior on this and make sure it's not a breaking change. Let's hold off on merging this / releasing the previous changes until we get that confirmation.

scannillo: > Ok we are going to follow up with 3DS to confirm the GW behavior on this and make sure it's not a breaking change. Let's hold off on merging this / releasing the previous changes until we get that confirmation.

Thanks for looking into this. Is this something that can be tested manually (by setting a 0 amount) in the SDK itself and going through a 3DS flow?

sarahkoop: Not sure about manual testing - when I set amount to 0 I get a nonce but can't create a transaction (as expected because this behavior is for creating payment method not transaction). 

But looks like we have confirmation that this GW behavior was introduced over a year ago so we are just adding the true/false option to give merchants more control over existing behavior.

scannillo: Rad - thanks for confirming this, Sarah!

scannillo (approved):
🕺

PR #426, opened 2021-08-05
Fix Samsung Pay

sshropshire:
### Summary of changes

 - Add `com.samsung.android.spay` to the SamsungPay `AndroidManifest.xml` `<queries>` block. 
 - Add missing values to the SamsungPay `AndroidManifest.xml` that are required for SamsungPay to work.

 ### Checklist

### Authors

- @sshropshire

### DIFF
--- a/SamsungPay/src/main/AndroidManifest.xml
+++ b/SamsungPay/src/main/AndroidManifest.xml
@@ -2,4 +2,12 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.braintreepayments.api.samsungpay">
 
+    <application>
+        <meta-data android:name="spay_sdk_api_level" android:value="2.4" />
+        <meta-data android:name="debug_mode" android:value="N" />
+    </application>
+
+    <queries>
+        <package android:name="com.samsung.android.spay" />
+    </queries>
 </manifest>


SamsungPay/src/main/AndroidManifest.xml
```diff
@@ -2,4 +2,12 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.braintreepayments.api.samsungpay">
 
+    <application>
+        <meta-data android:name="spay_sdk_api_level" android:value="2.4" />
```
sarahkoop, line 6: What does the 2.4 mean here? It looks like we use version 2.5 of the Samsung Pay SDK.

sshropshire: I'm not entirely sure. There's a bit of a description [here](https://d3sfvyfh4b9elq.cloudfront.net/ptr/guide/techguide/v1.4/guide/05_00_understanding_debug_api_key_api_level.html). I think it has something to do with backward compatability with older versions of the SamsungPay app? I kept `2.4` because it's what we use in the current implementation.

sarahkoop: Ah ok that seems good to me then 👍

sarahkoop (approved):
I'm not able to manually test this - but assuming you were able to successfully complete the flow on a device with this change looks good to me 🌟

scannillo (approved):

PR #432, opened 2021-08-19
BraintreeClient Custom URL Scheme Constructor

sshropshire:
### Summary of changes

 - Create a new `BraintreeClient` instructor that allows for a custom `returnUrlScheme` for browser and app switching
 - Cleanup random lint errors

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sshropshire

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
@@ -66,7 +66,9 @@ public class BraintreeClient {
     }
 
     /**
-     * Create a new instance of {@link BraintreeClient} using a tokenization key or client token.
+     * Create a new instance of {@link BraintreeClient} using a tokenization key or client token and a custom url scheme.
+     *
+     * This constructor should only be used for applications with multiple activities and multiple supported return url schemes.
      *
      * @param context         Android Context
      * @param authorization   The tokenization key or client token to use. If an invalid authorization is provided, a {@link BraintreeException} will be returned via callback.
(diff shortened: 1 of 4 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
```diff
@@ -55,6 +65,17 @@ public BraintreeClient(@NonNull Context context, @NonNull String authorization)
         this(createDefaultParams(context, authorization));
     }
 
+    /**
+     * Create a new instance of {@link BraintreeClient} using a tokenization key or client token.
```
sarahkoop, line 69: Let's add a note in the description that this constructor should only be used for applications with multiple activities that need to handle deep link returns (or some wording like that)

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
```diff
@@ -183,7 +201,7 @@ boolean canPerformBrowserSwitch(FragmentActivity activity, @BraintreeRequestCode
         return manifestValidator.isUrlSchemeDeclaredInAndroidManifest(applicationContext, urlScheme, klass);
     }
 
-    ActivityInfo getManifestActivityInfo(Class klass) {
+    <T> ActivityInfo getManifestActivityInfo(Class<T> klass) {
```
sarahkoop, line 204: Curious - why does this need to be generic now?

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
sshropshire: Yeah that's a good idea I'm down to add that in.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
sshropshire: This one was a lint error. I think it doesn't like when `Class` has no associated type since it's generic. The one above has a similar kind of syntax figured it was at least reducing the number of lint errors.

sarahkoop (approved):
Looks good to me 🌟

scannillo (approved):

PR #434, opened 2021-08-23
Android 12 Updates v4

sarahkoop:
### Summary of changes

 - Upgrade `targetSdkVersion` and `compileSdkVersion` to API 31
 - Bump `browser-switch` version to `2.1.0` 

**Note**: Tested all flows on an Android 12 device manually. Transacting with a local payment nonce is getting rejected (same behavior on main), and the visa checkout flow is not working (not released in v4 GA).

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop


BraintreeCore/build.gradle
```diff
@@ -45,26 +45,26 @@ android {
 def playServicesWallet = "com.google.android.gms:play-services-wallet:${rootProject.playServicesWalletVersion}"
 
 dependencies {
-    implementation 'androidx.appcompat:appcompat:1.2.0'
-    implementation 'androidx.work:work-runtime:2.4.0'
+    implementation 'androidx.appcompat:appcompat:1.3.1'
+    implementation 'androidx.work:work-runtime:2.7.0-alpha05'
```
sarahkoop, line 49: Using a 2.7.0-alpha version of this library is required to fix an [Android 12 PendingIntent bug](https://developer.android.com/jetpack/androidx/releases/work#2.7.0-alpha02)

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.java
```diff
@@ -151,7 +151,10 @@ public void run() {
                         request.id = Integer.valueOf(ids[i]);
                         request.event = events[i];
                         request.timestamp = Long.valueOf(timestamps[i]);
-                        request.metadata = new JSONObject(cursor.getString(cursor.getColumnIndex(META_JSON)));
+                        int index = cursor.getColumnIndex(META_JSON);
+                        if (index >= 0) {
+                            request.metadata = new JSONObject(cursor.getString(index));
+                        }
```
sarahkoop, line 157: This fixes a compiler lint error when running the tests from Android Studio Arctic Fox (required for Android 12 emulators)

Demo/src/main/java/com/braintreepayments/demo/GooglePayFragment.java
```diff
@@ -115,7 +115,11 @@ private void handleGooglePayActivityResult(ActivityResult activityResult) {
         int resultCode = activityResult.getResultCode();
         Intent data = activityResult.getData();
         googlePayClient.onActivityResult(resultCode, data, (paymentMethodNonce, error) -> {
-            handleGooglePayActivityResult(paymentMethodNonce);
+            if (error != null) {
```
sarahkoop, line 118: We weren't previously handling errors in some payments flows in the demo app. I added them now to make manual testing easier (avoid app crashes)

sshropshire (approved):

v4_MIGRATION_GUIDE.md
```diff
@@ -48,6 +49,8 @@ In the `AndroidManifest.xml`, migrate the `intent-filter` from your v3 integrati
 </activity>
 ``` 
 
+Note: `android:exported` is required if your app compile SDK version is API 31 (Android 12) or later.
+
```
scannillo, line 53: Should this go in the README instead because it also affects merchants onboarding directly to v4?

sarahkoop: We don't have anything in the README currently showing how the deep-link intent-filter should be configured in the `AndroidManifest`. We have a JIRA ticket open to update the dev docs with the new flag. Do you think we should also add a section to the README about Android 12?

scannillo: As long as we have a JIRA to make this more explicit, that is OK. Ideally it is called out elsewhere than just the migration guide.

scannillo (approved):

PR #442, opened 2021-09-01
Update SignatureVerification to Use MessageDigest

sarahkoop:
### Summary of changes

 - Update SignatureVerification to use `MessageDigest` and not rely on `publicKey.hashCode()` to verify signatures match
 - Remove ignored integration tests related to old PayPal app switching signature verification, and add unit tests for signature verification

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop 
- @sshropshire


SharedUtils/src/main/java/com/braintreepayments/api/SignatureVerification.java
```diff
@@ -47,38 +43,28 @@ static boolean isSignatureValid(Context context, String packageName,
...
+                    MessageDigest md = MessageDigest.getInstance("SHA-256");
+                    md.update(signature.toByteArray());
+                    currentSignature = Base64.encodeToString(md.digest(), Base64.DEFAULT);
+                } catch (NoSuchAlgorithmException e) {
+                    return false;
+                }
 
+                boolean validated = base64EncodedSignature.equals(currentSignature);
                 if (!validated) {
                     return false;
```
sarahkoop, line 63: The legacy approach for signature verification was to return false if there are any signatures that don't match (i.e. the app as more than one signature). Does that seem correct still?

scannillo: Sounds right to me

SharedUtils/src/main/java/com/braintreepayments/api/SignatureVerification.java
```diff
@@ -47,38 +43,28 @@ static boolean isSignatureValid(Context context, String packageName,
...
-                        certificateIssuer.equals(issuer) &&
-                        publicKeyHashCode == actualPublicKeyHashCode);
+        for (Signature signature : signatures) {
+            if (signature != null) {
+                String currentSignature;
+                try {
+                    MessageDigest md = MessageDigest.getInstance("SHA-256");
+                    md.update(signature.toByteArray());
+                    currentSignature = Base64.encodeToString(md.digest(), Base64.DEFAULT);
+                } catch (NoSuchAlgorithmException e) {
```
scannillo, line 57: Will this prevent us from having issues like this in the future? Just to summarize - the problem was that the hashCode() functionality changed b/w Android 11 and 12, and this issue wasn't specific to Venmo?

sarahkoop: Yes there were no guarantees about the underlying implementation of the `hashCode` function, but this new approach relies on the SHA-256 algorithm and base 64 encoding. I'm not a cryptography expert, but as far as I know the algorithm should always produce the same output for a given input.

scannillo (approved):
I'm also not a cryptography expert, but looks good to me!

sshropshire (approved):
🌊

sarahkoop:
@scannillo @sshropshire Can you all give this one last look? I made some changes based on feedback from Venmo team. I wasn't able to get a unit test working for `CertificateHelper`, but the remainder of the code is unit tested (and provides more test coverage than we had previously). Let me know if this is still good to merge.

sshropshire (approved):
🌊 &nbsp;&nbsp;LGTM!

PR #445, opened 2021-09-08
[V3] Update Signature Verification

sarahkoop:
### Summary of changes

 - Fix issue where Venmo app is not detected on Android 12 devices
 - Update signature verification encoded signatures for PayPal - app switch doesn't happen anymore, but updating for any potential future use cases

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop

### DIFF
--- a/Braintree/src/main/java/com/braintreepayments/api/Venmo.java
+++ b/Braintree/src/main/java/com/braintreepayments/api/Venmo.java
@@ -44,8 +44,6 @@ public class Venmo {
 
     static final String PACKAGE_NAME = "com.venmo";
     static final String APP_SWITCH_ACTIVITY = "controller.SetupMerchantActivity";
-    static final String CERTIFICATE_SUBJECT = "CN=Andrew Kortina,OU=Engineering,O=Venmo,L=Philadelphia,ST=PA,C=US";
-    static final String CERTIFICATE_ISSUER = "CN=Andrew Kortina,OU=Engineering,O=Venmo,L=Philadelphia,ST=PA,C=US";
     static final int PUBLIC_KEY_HASH_CODE = -129711843;
 
     private static final String VENMO_BASE_64_ENCODED_SIGNATURE = "x34mMawEUcCG8l95riWCOK+kAJYejVmdt44l6tzcyUc=\n";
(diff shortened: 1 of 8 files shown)


Core/src/main/java/com/braintreepayments/api/internal/SignatureVerification.java
```diff
@@ -31,6 +36,7 @@
      * @param publicKeyHashCode the hash code of the app's public key.
      * @return true is signature is valid or signature verification has been disabled.
      */
+    @Deprecated
```
scannillo, line 39: Can you provide a deprecation note in Android (I know you can in iOS). Something to signal that this won't work in Android 12?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Upgrade `targetSdkVersion` and `compileSdkVersion` to API 31
 * Bump `browser-switch` version to `1.2.0`
+* Fix issue where Venmo app is not detected on Android 12 devices
```
scannillo, line 7: Do we think it is worth calling out that this change also includes the fix for PayPal? _Maybe_ a merchant is using our signature verification functionality to detect if PayPal is installed 🤷🏻

Core/src/main/java/com/braintreepayments/api/internal/SignatureVerification.java
sarahkoop: Just checked and unfortunately this isn't possible with the deprecated tag in Android

CHANGELOG.md
sarahkoop: Yeah in case anyone is using that method I updated the CHANGELOG 👍

Braintree/src/main/java/com/braintreepayments/api/Venmo.java
```diff
@@ -48,14 +48,15 @@
     static final String CERTIFICATE_ISSUER = "CN=Andrew Kortina,OU=Engineering,O=Venmo,L=Philadelphia,ST=PA,C=US";
```
sshropshire, line 48: Some of these static constants could probably be removed if we want to do some `v3` style due diligence.

sarahkoop: Good call yeah removed those and some old tests

scannillo (approved):

PR #449, opened 2021-09-21
Make VenmoRequest Parcelable

sarahkoop:
### Summary of changes

 - As part of Drop-in updates, we want to allow merchants to set a `VenmoRequest` on their `DropInRequest` - similar to what they do for PayPal and Google Pay. To make that change in Drop-in, `VenmoRequest` needs to be parcelable.

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## unreleased
 
+* Venmo
+  * Make `VenmoRequest` parcelable
 * ThreeDSecure
   * Make `pareq` optional on `ThreeDSecureLookup`
 
--- a/Venmo/src/main/java/com/braintreepayments/api/VenmoRequest.java
+++ b/Venmo/src/main/java/com/braintreepayments/api/VenmoRequest.java
@@ -1,11 +1,14 @@
 package com.braintreepayments.api;
 
+import android.os.Parcel;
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## unreleased
 
+* Venmo
+  * Make `VenmoRequest` parcelable
```
sarahkoop, line 6: Question - is this ok as a non-breaking change? The only issue I could see is if a merchant is subclassing `VenmoRequest`, their subclass will now also have to implement `Parcelable`. There is no need for a merchant to subclass this, so not sure if we think this is ok.

sshropshire: This is needed because `VenmoRequest` is a property on `DropInRequest` in `v6`, which is also `Parcelable`. I think it makes sense to treat this breaking change as `non-breaking`. 

The number of merchants subclassing this type _should_ be low. I think the value added to the `DropIn` API outweighs any risk of a breaking change occurring for a non-standard integration.

scannillo: Agreed, I think we're good to count this as non breaking!

scannillo (approved):
🎉

PR #452, opened 2021-09-27
[BLOCKED] PwV - Enrich Consumer Data

demerino:
### Summary of changes
- Add email, externalId, firstName, lastName, phoneNumber to VenmoAccountNonce

These fields will only be available in certain situations.

**Blocked for e2e testing**

### Checklist

- [ ] Added a changelog entry

### Authors
- @demerino

### DIFF
--- a/Venmo/src/main/java/com/braintreepayments/api/VenmoAccountNonce.java
+++ b/Venmo/src/main/java/com/braintreepayments/api/VenmoAccountNonce.java
@@ -23,9 +23,20 @@ public class VenmoAccountNonce extends PaymentMethodNonce {
     private static final String VENMO_USERNAME_KEY = "username";
 
     private static final String VENMO_PAYMENT_METHOD_ID_KEY = "paymentMethodId";
+    private static final String VENMO_PAYER_INFO_KEY = "payerInfo";
+    private static final String VENMO_EMAIL_KEY = "email";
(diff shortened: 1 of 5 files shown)


Venmo/src/main/java/com/braintreepayments/api/VenmoAccountNonce.java
```diff
@@ -51,14 +62,70 @@ static VenmoAccountNonce fromJSON(JSONObject inputJson) throws JSONException {
...
-        return new VenmoAccountNonce(nonce, username, isDefault);
+        return new VenmoAccountNonce(nonce, username, isDefault, json);
+    }
+
+    VenmoAccountNonce(String nonce, String username, boolean isDefault, JSONObject json) throws JSONException {
+        super(nonce, isDefault);
+        this.username = username;
+
+        if (json.has(VENMO_PAYMENT_METHOD_ID_KEY)) {
+            if (json.has(VENMO_PAYER_INFO_KEY)) {
```
sarahkoop, line 73: Based on the fixture added, it looks like these two keys are on the same level in the JSON - do we need both checks?  If we do - can it be `json.has(VENMO_PAYMENT_METHOD_ID_KEY) && json.has(VENMO_PAYER_INFO_KEY)` rather than a nested if statement?

PR #469, opened 2021-11-02
Update Migration Guide with Conflicting Intent Scenario 

sshropshire:
### Summary of changes

 - Mention conflicting intent scenario in v4_MIGRATION_GUIDE (Addresses [braintree-android-drop-in #296](https://github.com/braintree/braintree-android-drop-in/issues/296))

 ### Checklist

### DIFF
--- a/v4_MIGRATION_GUIDE.md
+++ b/v4_MIGRATION_GUIDE.md
@@ -51,7 +51,7 @@ In the `AndroidManifest.xml`, migrate the `intent-filter` from your v3 integrati
 </activity>
 ``` 
 
-Additionally, apps that use both DropIn and BraintreeClient should specify a custom url scheme, since `DropInActivity` already uses the `${applicationId}.braintree` url intent filter.
+Additionally, apps that use both Drop-in and BraintreeClient should specify a custom url scheme, since `DropInActivity` already uses the `${applicationId}.braintree` url intent filter.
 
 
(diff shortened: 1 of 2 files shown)


v4_MIGRATION_GUIDE.md
```diff
@@ -49,7 +51,8 @@ In the `AndroidManifest.xml`, migrate the `intent-filter` from your v3 integrati
 </activity>
 ``` 
 
-Note: `android:exported` is required if your app compile SDK version is API 31 (Android 12) or later.
+Additionally, apps that use both DropIn and BraintreeClient should specify a custom url scheme, since `DropInActivity` already uses the `${applicationId}.braintree` url intent filter.
```
sarahkoop, line 54:
```suggestion
Additionally, apps that use both Drop-in and BraintreeClient should specify a custom url scheme, since `DropInActivity` already uses the `${applicationId}.braintree` url intent filter.
```

sarahkoop:
Should we change the doc string for that `BraintreeClient` constructor to note it should be used if using core and drop-in?

sshropshire:
> Should we change the doc string for that `BraintreeClient` constructor to note it should be used if using core and drop-in?

Yeah that's a good call. Just made an update.

sarahkoop (approved):

demerino (approved):

PR #476, opened 2021-11-17
Revert Work Manager and Room Dependencies

sshropshire:
### Summary of changes

 - Revert Work Manager and Room dependencies to keep Java 7 support.

 ### Checklist

 - [x] Removed changelog entry

### Authors

- @sarahkoop
- @sshropshire

### DIFF
--- a/BraintreeCore/build.gradle
+++ b/BraintreeCore/build.gradle
@@ -46,10 +46,10 @@ def playServicesWallet = "com.google.android.gms:play-services-wallet:${rootProj
 
 dependencies {
     implementation 'androidx.appcompat:appcompat:1.3.1'
-    implementation 'androidx.work:work-runtime:2.7.0'
+    implementation 'androidx.work:work-runtime:2.7.0-alpha05'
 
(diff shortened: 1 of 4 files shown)


scannillo (approved):
I wonder if it would be helpful to our merchants (it would be helpful to me) if we called out in our README what versions of Java the SDK supports. Also, might help for future version bumps if we make that explicit in our README / if we have plans to drop Java 7 soon.

jplukarski (approved):
Seems pretty straightforward! 🚢

sarahkoop:
> I wonder if it would be helpful to our merchants (it would be helpful to me) if we called out in our README what versions of Java the SDK supports. Also, might help for future version bumps if we make that explicit in our README / if we have plans to drop Java 7 soon.

Yeah we haven't been super explicit about which Java versions we support in the past, but we want to make it clear in the next major version that we only support Java 8 and above. I think it would make sense to put that in the change log with the next major version (similar to what we did for the Drop-in [CHANGELOG](https://github.com/braintree/braintree-android-drop-in/blob/master/CHANGELOG.md#unreleased))

Issue #446, opened 2021-09-08
Google Pay - Update Wallet Version


nate-eisner:
### General information

* SDK/Library version: 4.5.0
* Environment: Both
* Android Version and Device: All
* Braintree dependencies:
```
  com.braintreepayments.api:paypal
  com.braintreepayments.api:google-pay
  com.braintreepayments.api:card
```

### Issue description

I am working through the 3.x migration to 4.x and was wondering why we are required to use such an old version of Google wallet library? The latest version is com.google.android.gms:play-services-wallet:18.1.3 and you have it set for [16.0.1](https://github.com/braintree/braintree_android/blob/master/GooglePay/build.gradle#L38)

sarahkoop:
Hi @nate-eisner - we can look into upgrading the wallet version, but in the meantime, feel free to open a PR with the requested change.

sarahkoop:
This was released in `4.8.1`

PR #477, opened 2021-11-17
Update Google Pay SDK

sarahkoop:
### Summary of changes

 - Update `play-services-wallet` dependency to `18.1.3` (fixes: #446)
 - Updating to this version of the play services wallet SDK causes issues with our unit tests because the Google `Wallet` class becomes a static final class (difficult for mocking), and breaks our tests that use Powermock (to mock static classes), and Robolectric (for the Android system classes - like Activity). The google pay integration requires both Activity and Wallet, so I added wrapper classes to separate out as much of the testable logic into their own classes to allow us to maintain test coverage while upgrading this library. 

 ### Checklist

 - [x] Added a changelog entry

- @sarahkoop

### DIFF
--- a/GooglePay/src/main/java/com/braintreepayments/api/PaymentsClientWrapper.java
+++ b/GooglePay/src/main/java/com/braintreepayments/api/PaymentsClientWrapper.java
@@ -10,11 +10,7 @@ class PaymentsClientWrapper {
     private final WalletOptionsWrapper walletOptionsWrapper;
 
     PaymentsClientWrapper() {
-        this(new WalletOptionsWrapper());
-    }
-
-    PaymentsClientWrapper(WalletOptionsWrapper walletOptionsWrapper) {
-        this.walletOptionsWrapper = walletOptionsWrapper;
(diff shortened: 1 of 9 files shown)


GooglePay/src/main/java/com/braintreepayments/api/PaymentsClientWrapper.java
```diff
@@ -0,0 +1,19 @@
+package com.braintreepayments.api;
+
+import androidx.fragment.app.FragmentActivity;
+
+import com.google.android.gms.wallet.PaymentsClient;
+import com.google.android.gms.wallet.Wallet;
+
+class PaymentsClientWrapper {
```
sarahkoop, line 8: This is the one class that can't be tested with the updated version of the play services wallet library. The static final `Wallet` classes causes errors when tested with Robolectric (required to pass a `FragmentActivity` to the `getPaymentsClient` method, and Powermock (required to mock the static Wallet class).

sarahkoop:
Closing in favor of #487

PR #482, opened 2021-12-08
Make Google Pay Merchant ID Explicitly Optional

sarahkoop:
### Summary of changes

 - Update doc string to note that `googleMerchantId` is optional
 - Remove setting merchant ID from demo app

 ### Checklist

- @sarahkoop

### DIFF
--- a/Card/build.gradle
+++ b/Card/build.gradle
@@ -41,7 +41,7 @@ dependencies {
     testImplementation 'org.skyscreamer:jsonassert:1.4.0'
     testImplementation project(':TestUtils')
     testImplementation project(':ThreeDSecure')
-    testImplementation "androidx.core:core-ktx:+"
+    testImplementation "androidx.core:core-ktx:1.6.0"
     testImplementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:1.5.10"
     testImplementation "org.jetbrains.kotlin:kotlin-test-junit:1.5.10"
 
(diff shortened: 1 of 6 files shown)


Card/build.gradle
```diff
@@ -41,7 +41,7 @@ dependencies {
     testImplementation 'org.skyscreamer:jsonassert:1.4.0'
     testImplementation project(':TestUtils')
     testImplementation project(':ThreeDSecure')
-    testImplementation "androidx.core:core-ktx:+"
+    testImplementation "androidx.core:core-ktx:1.6.0"
```
sarahkoop, line 44: Updating the `core-ktx` dependencies to specific versions is unrelated to this PR, but using `+` is causing tests to fail, and we should be pinning to a specific version.

sshropshire (approved):
🌊

GooglePay/src/main/java/com/braintreepayments/api/GooglePayRequest.java
```diff
@@ -172,13 +175,17 @@ public void setAllowedCardNetworks(@NonNull String paymentMethodType, @NonNull J
     }
 
     /**
+     * Optional.
```
Epreuve, line 178: can we add `@Deprecated` on the method and `@deprecated` on the jdoc here just to enforce they'll be going away in the next major version?

Epreuve (approved):
Small doc nit, but overall 🍰

GooglePay/src/main/java/com/braintreepayments/api/GooglePayRequest.java
sarahkoop: Good call - added the deprecation tags and an CHANGELOG entry to make it explicit

PR #484, opened 2021-12-13
Standardize Gradle Dependencies

sarahkoop:
### Summary of changes

 - Move dependency libraries and versions to top-level `build.gradle` and standardize versions across modules (where possible)
 - This should make it easier to keep dependency versions the same across modules, and to make updates to versions in the future.
 - With this update, we lose the Android Studio highlighting/auto-update keyboard shortcut when a newer library is available, but this suggested dependency updates can still be found in Android Studio under `File` > `Project Structure` > `Suggestions`. 
 - Overall, the ability to keep dependency versions aligned and ensure that they are all updated together seems higher priority than the keyboard shortcut, but we will need to make it a regular practice to check for suggestions. 

**Note**: This PR organizes the dependencies into a single file, but does not update the versions (unless required for alignment). Another PR to follow will handle upgrading dependencies to the newest versions where possible. 

 ### Checklist

- @sarahkoop

### DIFF
--- a/Card/src/test/java/com/braintreepayments/api/CardClientUnitTest.java
+++ b/Card/src/test/java/com/braintreepayments/api/CardClientUnitTest.java
@@ -36,7 +36,7 @@ public class CardClientUnitTest {
     @Before
     public void beforeEach() throws JSONException {
         context = mock(Context.class);
-        card = mock(Card.class);
+        card = new Card();
         cardTokenizeCallback = mock(CardTokenizeCallback.class);
 
         dataCollector = mock(DataCollector.class);
@@ -75,7 +75,6 @@ public class CardClientUnitTest {
 
         CardClient sut = new CardClient(braintreeClient, apiClient);
 
-        Card card = new Card();
         sut.tokenize(card, cardTokenizeCallback);
 
         ArgumentCaptor<CardNonce> captor = ArgumentCaptor.forClass(CardNonce.class);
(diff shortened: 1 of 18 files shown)


sshropshire (approved):
🔥

Card/src/test/java/com/braintreepayments/api/CardClientUnitTest.java
```diff
@@ -36,7 +36,7 @@
     @Before
     public void beforeEach() throws JSONException {
         context = mock(Context.class);
-        card = mock(Card.class);
+        card = new Card();
```
Epreuve, line 39: 👀  Just cleanup while we're here?

Epreuve (approved):
Looks good, I'm assuming some of the changes were just related to some minor dep adjustments.

Overall I don't think we're pushing to stay on the most bleeding edge of versions where the loss of the shortcut will sting much, if at all.  I'd rather take the gain of having a centralized dependency version that allows us to avoid weird mismatches, build errors, and help us spot problems early with bumps.

Card/src/test/java/com/braintreepayments/api/CardClientUnitTest.java
sarahkoop: No this change resolved some test failures related to not mocking all the underlying setters on card. Figured just using an actual Card object made more sense then mocking everything out.

PR #493, opened 2022-01-14
Authorization Loader

sshropshire:
### Notes

By taking control of when Client Tokens are fetched, registering and responding to app / browser switches can be further encapsulated by the SDK. This allows us to improve the developer experience for 3DS, PayPal, Venmo, Google Pay, and Local Payments.

### Summary of changes

 - Add `AuthorizationProvider` interface to allow the merchant to define a strategy for asynchronously fetching a client token from their server
 - Add `ClientTokenCallback` interface
 - Add `BraintreeClient` constructors that accept a `ClientTokenProvider`

 ### Checklist

 - [x] Added a changelog entry

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsBaseWorker.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsBaseWorker.java
@@ -14,6 +14,7 @@ import androidx.work.WorkerParameters;
  * This class is used internally by the SDK and should not be used directly.
  * It is not subject to semantic versioning and may change at any time.
  */
+// NEXT-MAJOR-VERSION: remove this class, it may no longer be needed
 public abstract class AnalyticsBaseWorker extends Worker {
 
     public AnalyticsBaseWorker(@NonNull Context context, @NonNull WorkerParameters workerParams) {
@@ -21,9 +22,6 @@ public abstract class AnalyticsBaseWorker extends Worker {
     }
 
     protected AnalyticsClient createAnalyticsClientFromInputData() {
-        Data inputData = getInputData();
-        String authString = inputData.getString(WORK_INPUT_KEY_AUTHORIZATION);
-        Authorization authorization = Authorization.fromString(authString);
         return new AnalyticsClient(getApplicationContext());
     }
 }
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsBaseWorker.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsBaseWorker.java
@@ -24,6 +24,6 @@ public abstract class AnalyticsBaseWorker extends Worker {
         Data inputData = getInputData();
         String authString = inputData.getString(WORK_INPUT_KEY_AUTHORIZATION);
         Authorization authorization = Authorization.fromString(authString);
-        return new AnalyticsClient(getApplicationContext(), authorization);
+        return new AnalyticsClient(getApplicationContext());
     }
 }
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsBaseWorker.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsBaseWorker.java
@@ -14,7 +14,7 @@ import androidx.work.WorkerParameters;
  * This class is used internally by the SDK and should not be used directly.
  * It is not subject to semantic versioning and may change at any time.
  */
-// NEXT-MAJOR-VERSION: remove this class, it may no longer be needed
+// NEXT_MAJOR_VERSION: remove this class, it may no longer be needed
 public abstract class AnalyticsBaseWorker extends Worker {
 
     public AnalyticsBaseWorker(@NonNull Context context, @NonNull WorkerParameters workerParams) {
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.java
@@ -135,9 +135,9 @@ class AnalyticsClient {
         String sessionId = inputData.getString(WORK_INPUT_KEY_SESSION_ID);
         String integration = inputData.getString(WORK_INPUT_KEY_INTEGRATION);
 
-        boolean shouldFail =
+        boolean isMissingInputData =
             Arrays.asList(configuration, authorization, sessionId, integration).contains(null);
-        if (shouldFail) {
+        if (isMissingInputData) {
             return ListenableWorker.Result.failure();
         }
 
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.java
@@ -48,7 +48,7 @@ class AnalyticsClient {
 
     AnalyticsClient(Context context, Authorization authorization) {
         this(
-                new BraintreeHttpClient(authorization),
+                new BraintreeHttpClient(),
                 AnalyticsDatabase.getInstance(context.getApplicationContext()),
                 WorkManager.getInstance(context.getApplicationContext()),
                 new DeviceInspector()
@@ -147,7 +147,7 @@ class AnalyticsClient {
                 JSONObject analyticsRequest = serializeEvents(httpClient.getAuthorization(), events, metadata);
 
                 String analyticsUrl = configuration.getAnalyticsUrl();
-                httpClient.post(analyticsUrl, analyticsRequest.toString(), configuration);
+                httpClient.post(analyticsUrl, analyticsRequest.toString(), configuration, );
                 analyticsEventDao.deleteEvents(events);
             }
             return ListenableWorker.Result.success();
@@ -171,7 +171,7 @@ class AnalyticsClient {
         List<AnalyticsEvent> events = Collections.singletonList(event);
         try {
             JSONObject analyticsRequest = serializeEvents(httpClient.getAuthorization(), events, metadata);
-            httpClient.post(lastKnownAnalyticsUrl, analyticsRequest.toString(), null, new HttpNoResponse());
+            httpClient.post(lastKnownAnalyticsUrl, analyticsRequest.toString(), null, , new HttpNoResponse(), );
         } catch (JSONException e) { /* ignored */ }
     }
 
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.java
@@ -46,7 +46,7 @@ class AnalyticsClient {
 
     private String lastKnownAnalyticsUrl;
 
-    AnalyticsClient(Context context, Authorization authorization) {
+    AnalyticsClient(Context context) {
         this(
                 new BraintreeHttpClient(),
                 AnalyticsDatabase.getInstance(context.getApplicationContext()),
@@ -63,22 +63,21 @@ class AnalyticsClient {
         this.analyticsDatabase = analyticsDatabase;
     }
 
-    void sendEvent(Configuration configuration, String eventName, String sessionId, String integration) {
+    void sendEvent(Configuration configuration, String eventName, String sessionId, String integration, Authorization authorization) {
         long timestamp = System.currentTimeMillis();
-        sendEvent(configuration, eventName, sessionId, integration, timestamp);
+        sendEvent(configuration, eventName, sessionId, integration, timestamp, authorization);
     }
 
     @VisibleForTesting
-    UUID sendEvent(Configuration configuration, String eventName, String sessionId, String integration, long timestamp) {
+    UUID sendEvent(Configuration configuration, String eventName, String sessionId, String integration, long timestamp, Authorization authorization) {
         lastKnownAnalyticsUrl = configuration.getAnalyticsUrl();
 
         String fullEventName = String.format("android.%s", eventName);
-        scheduleAnalyticsWrite(fullEventName, timestamp);
-        return scheduleAnalyticsUpload(configuration, httpClient.getAuthorization(), sessionId, integration);
+        scheduleAnalyticsWrite(fullEventName, timestamp, authorization);
+        return scheduleAnalyticsUpload(configuration, authorization, sessionId, integration);
     }
 
-    private void scheduleAnalyticsWrite(String eventName, long timestamp) {
-        Authorization authorization = httpClient.getAuthorization();
+    private void scheduleAnalyticsWrite(String eventName, long timestamp, Authorization authorization) {
         Data inputData = new Data.Builder()
                 .putString(WORK_INPUT_KEY_AUTHORIZATION, authorization.toString())
                 .putString(WORK_INPUT_KEY_EVENT_NAME, eventName)
@@ -130,6 +129,8 @@ class AnalyticsClient {
 
     ListenableWorker.Result uploadAnalytics(Context context, Data inputData) {
         Configuration configuration = getConfigurationFromData(inputData);
+        Authorization authorization = getAuthorizationFromData(inputData);
+
         String sessionId = inputData.getString(WORK_INPUT_KEY_SESSION_ID);
         String integration = inputData.getString(WORK_INPUT_KEY_INTEGRATION);
 
@@ -144,10 +145,10 @@ class AnalyticsClient {
             boolean shouldUploadAnalytics = !events.isEmpty();
             if (shouldUploadAnalytics) {
                 DeviceMetadata metadata = deviceInspector.getDeviceMetadata(context, sessionId, integration);
-                JSONObject analyticsRequest = serializeEvents(httpClient.getAuthorization(), events, metadata);
+                JSONObject analyticsRequest = serializeEvents(authorization, events, metadata);
 
                 String analyticsUrl = configuration.getAnalyticsUrl();
-                httpClient.post(analyticsUrl, analyticsRequest.toString(), configuration, );
+                httpClient.post(analyticsUrl, analyticsRequest.toString(), configuration, authorization);
                 analyticsEventDao.deleteEvents(events);
             }
             return ListenableWorker.Result.success();
@@ -156,13 +157,13 @@ class AnalyticsClient {
         }
     }
 
-    void reportCrash(Context context, String sessionId, String integration) {
-        reportCrash(context, sessionId, integration, System.currentTimeMillis());
+    void reportCrash(Context context, String sessionId, String integration, Authorization authorization) {
+        reportCrash(context, sessionId, integration, System.currentTimeMillis(), authorization);
     }
 
     @VisibleForTesting
-    void reportCrash(Context context, String sessionId, String integration, long timestamp) {
-        if (lastKnownAnalyticsUrl == null) {
+    void reportCrash(Context context, String sessionId, String integration, long timestamp, Authorization authorization) {
+        if (lastKnownAnalyticsUrl == null || authorization == null) {
             return;
         }
 
@@ -170,8 +171,8 @@ class AnalyticsClient {
         AnalyticsEvent event = new AnalyticsEvent("android.crash", timestamp);
         List<AnalyticsEvent> events = Collections.singletonList(event);
         try {
-            JSONObject analyticsRequest = serializeEvents(httpClient.getAuthorization(), events, metadata);
-            httpClient.post(lastKnownAnalyticsUrl, analyticsRequest.toString(), null, , new HttpNoResponse(), );
+            JSONObject analyticsRequest = serializeEvents(authorization, events, metadata);
+            httpClient.post(lastKnownAnalyticsUrl, analyticsRequest.toString(), null, authorization, new HttpNoResponse());
         } catch (JSONException e) { /* ignored */ }
     }
 
@@ -198,6 +199,13 @@ class AnalyticsClient {
         return requestObject;
     }
 
+    private static Authorization getAuthorizationFromData(Data inputData) {
+        if (inputData != null) {
+            String authString = inputData.getString(WORK_INPUT_KEY_AUTHORIZATION);
+            return Authorization.fromString(authString);
+        }
+        return null;
+    }
     private static Configuration getConfigurationFromData(Data inputData) {
         if (inputData != null) {
             String configJson = inputData.getString(WORK_INPUT_KEY_CONFIGURATION);
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.java
@@ -13,6 +13,7 @@ import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
@@ -134,7 +135,9 @@ class AnalyticsClient {
         String sessionId = inputData.getString(WORK_INPUT_KEY_SESSION_ID);
         String integration = inputData.getString(WORK_INPUT_KEY_INTEGRATION);
 
-        if (configuration == null || sessionId == null || integration == null) {
+        boolean shouldFail =
+            Arrays.asList(configuration, authorization, sessionId, integration).contains(null);
+        if (shouldFail) {
             return ListenableWorker.Result.failure();
         }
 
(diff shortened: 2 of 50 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
```diff
@@ -126,42 +136,84 @@ public BraintreeClient(@NonNull Context context, @NonNull String authorization,
      *
      * @param callback {@link ConfigurationCallback}
      */
-    public void getConfiguration(@NonNull ConfigurationCallback callback) {
-        configurationLoader.loadConfiguration(applicationContext, authorization, callback);
+    public void getConfiguration(@NonNull final ConfigurationCallback callback) {
```
sshropshire, line 139: I don't think this is a breaking change but we could capture the callback as `final` internally if needed.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
```diff
@@ -232,17 +293,14 @@ boolean canPerformBrowserSwitch(FragmentActivity activity, @BraintreeRequestCode
...
-        analyticsClient.reportCrash(applicationContext, sessionId, integrationType);
+        Authorization authorization = authorizationLoader.getAuthorizationFromCache();
+        analyticsClient.reportCrash(applicationContext, sessionId, integrationType, authorization);
     }
 
     static boolean isAnalyticsEnabled(Configuration configuration) {
         return configuration != null && configuration.isAnalyticsEnabled();
     }
 
-    Authorization getAuthorization() {
```
sshropshire, line 242: Replaced this with an asynchronous variant.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
```diff
@@ -126,42 +136,84 @@ public BraintreeClient(@NonNull Context context, @NonNull String authorization,
...
+        });
+    }
+
+    void getAuthorization(@NonNull final AuthorizationCallback callback) {
+        authorizationLoader.loadAuthorization(callback);
+    }
+
+    AuthorizationType getAuthorizationType() {
+        return authorizationLoader.getAuthorizationType();
     }
```
sshropshire, line 158: This allows the `AuthorizationType` to be inferred synchronously without triggering a client token fetch e.g. TokenizationKey or ClientToken.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsBaseWorker.java
```diff
@@ -14,16 +14,14 @@
  * This class is used internally by the SDK and should not be used directly.
  * It is not subject to semantic versioning and may change at any time.
  */
+// NEXT-MAJOR-VERSION: remove this class, it may no longer be needed
```
sarahkoop, line 17: Nitpick since I think we usually search for underscores
```suggestion
// NEXT_MAJOR_VERSION: remove this class, it may no longer be needed
```

BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationLoader.java
```diff
@@ -0,0 +1,50 @@
...
+
+    AuthorizationLoader(@Nullable String initialAuthString, @Nullable ClientTokenProvider clientTokenProvider) {
+        this.clientTokenProvider = clientTokenProvider;
+        if (initialAuthString != null) {
+            this.authorization = Authorization.fromString(initialAuthString);
+        }
+    }
+
+    void loadAuthorization(@NonNull final AuthorizationCallback callback) {
+        if (authorization != null) {
```
sarahkoop, line 19: Are we planning to keep tuple single method callbacks for internal or do we want this to align more with the `ClientTokenCallback` pattern? Either way seems ok to me just curious about the decision

BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationLoader.java
```diff
@@ -0,0 +1,50 @@
...
+                    callback.onAuthorizationResult(authorization, null);
+                }
+
+                @Override
+                public void onFailure(@NonNull Exception exception) {
+                    callback.onAuthorizationResult(null, exception);
+                }
+            });
+        } else {
+            callback.onAuthorizationResult(null, new BraintreeException("unable to fetch client token"));
```
sarahkoop, line 35: Nitpick and we should probably also callback a developer error if `ClientTokenProvider` is null
```suggestion
            callback.onAuthorizationResult(null, new BraintreeException("Unable to fetch client token"));
```

BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationLoader.java
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+
+    @Nullable
+    Authorization getAuthorizationFromCache() {
+        return authorization;
+    }
+
+    AuthorizationType getAuthorizationType() {
```
sarahkoop, line 44: Do we need to handle the `InvalidAuthorization` type in this method too?

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
sarahkoop: Agreed - I think this `final` here only impacts our ability to alter the callback param within this method.

BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationLoader.java
sshropshire: Yeah for here, I was following that one piece of feedback we got for dead code paths. I'm open to either I guess this depends on the overall callback / listener balance in the API. I think the tuple still makes sense for SAM but we could go with listeners everywhere it might not be too bad.

BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationLoader.java
sshropshire: I forgot about that use case. It is possible with an invalid initial auth string. Still have to write unit tests for `AuthorizationLoader` I can add that test case in.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.java
```diff
@@ -130,6 +129,8 @@ private UUID scheduleAnalyticsUpload(Configuration configuration, Authorization
 
     ListenableWorker.Result uploadAnalytics(Context context, Data inputData) {
         Configuration configuration = getConfigurationFromData(inputData);
+        Authorization authorization = getAuthorizationFromData(inputData);
```
sarahkoop, line 132: Do we need to add a null check for this on line 137?

BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationLoader.java
```diff
@@ -0,0 +1,59 @@
...
+    AuthorizationLoader(@Nullable String initialAuthString, @Nullable ClientTokenProvider clientTokenProvider) {
+        this.clientTokenProvider = clientTokenProvider;
+        if (initialAuthString != null) {
+            this.authorization = Authorization.fromString(initialAuthString);
+        }
+    }
+
+    void loadAuthorization(@NonNull final AuthorizationCallback callback) {
+        if (authorization != null) {
+            callback.onAuthorizationResult(authorization, null);
```
sarahkoop, line 20: Do we need to do anything about client token expiration or is that up to the merchant? What happens here if we have an authorization but it's no longer valid?

BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationLoader.java
```diff
@@ -0,0 +1,59 @@
...
+    AuthorizationType getAuthorizationType() {
+        if (authorization instanceof InvalidAuthorization) {
+            return AuthorizationType.INVALID;
+        }
+
+        if (authorization instanceof TokenizationKey) {
+            return AuthorizationType.TOKENIZATION_KEY;
+        }
+
+        if (clientTokenProvider != null) {
```
sarahkoop, line 53: Do we not need to also confirm that it's instance of `ClientToken`? Might be good to check in case there is a different auth type in the future

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
```diff
@@ -76,7 +78,11 @@ private static BraintreeClientParams createDefaultParams(Context context, String
      * @param authorization The tokenization key or client token to use. If an invalid authorization is provided, a {@link BraintreeException} will be returned via callback.
      */
     public BraintreeClient(@NonNull Context context, @NonNull String authorization) {
-        this(createDefaultParams(context, authorization));
+        this(createDefaultParams(context, authorization, null));
+    }
+
+    public BraintreeClient(@NonNull Context context, @NonNull ClientTokenProvider clientTokenProvider) {
```
sarahkoop, line 84: I think it's fine to handle deprecation of old methods in another PR, but let's add doc strings for these new ones now so we don't miss any that are added

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
```diff
@@ -91,18 +97,22 @@ public BraintreeClient(@NonNull Context context, @NonNull String authorization)
      * @param returnUrlScheme A custom return url to use for browser and app switching
      */
     public BraintreeClient(@NonNull Context context, @NonNull String authorization, @NonNull String returnUrlScheme) {
-        this(createDefaultParams(context, authorization, returnUrlScheme));
+        this(createDefaultParams(context, authorization, null, returnUrlScheme));
+    }
+
+    public BraintreeClient(@NonNull Context context, @NonNull ClientTokenProvider clientTokenProvider, @NonNull String returnUrlScheme) {
```
sarahkoop, line 103: Doc string

BraintreeCore/src/main/java/com/braintreepayments/api/ClientTokenCallback.java
```diff
@@ -0,0 +1,8 @@
+package com.braintreepayments.api;
+
+import androidx.annotation.NonNull;
+
+public interface ClientTokenCallback {
```
sarahkoop, line 5: Doc string

sarahkoop: Also wondering about the naming on this - Do we want to be consistent with `Authorization`? Seems like sometimes we use authorization and sometimes we use client token

BraintreeCore/src/main/java/com/braintreepayments/api/ClientTokenProvider.java
```diff
@@ -0,0 +1,7 @@
+package com.braintreepayments.api;
+
+import androidx.annotation.NonNull;
+
+public interface ClientTokenProvider {
```
sarahkoop, line 5: doc string and same comment on naming

BraintreeCore/src/test/java/com/braintreepayments/api/AnalyticsClientUnitTest.java
```diff
@@ -319,14 +313,29 @@ public void reportCrash_whenLastKnownAnalyticsUrlExists_sendsCrashAnalyticsEvent
...
     @Test
-    public void reportCrash_whenLastKnownAnalyticsUrlMissing_doesNothing() {
+    public void reportCrash_whenLastKnownAnalyticsUrlMissing_doesNothing() throws JSONException {
+        Configuration configuration = Configuration.fromJson(Fixtures.CONFIGURATION_WITH_ANALYTICS);
+
+        DeviceMetadata metadata = createSampleDeviceMetadata();
+        when(deviceInspector.getDeviceMetadata(context, sessionId, integration)).thenReturn(metadata);
+
+        AnalyticsClient sut = new AnalyticsClient(httpClient, analyticsDatabase, workManager, deviceInspector);
+        sut.sendEvent(configuration, eventName, sessionId, integration, authorization);
```
sarahkoop, line 323: Doesn't `sendEvent` set the `lastKnownAnalyticsUrl`? I'm confused about this test

BraintreeCore/src/test/java/com/braintreepayments/api/AuthorizationLoaderUnitTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+        AuthorizationCallback callback = mock(AuthorizationCallback.class);
+        sut.loadAuthorization(callback);
+
+        ArgumentCaptor<Exception> captor = ArgumentCaptor.forClass(Exception.class);
+        verify(callback).onAuthorizationResult((Authorization)isNull(), captor.capture());
+
+        Exception error = captor.getValue();
+        assertTrue(error instanceof BraintreeException);
+        assertEquals("Unable to fetch client token", error.getMessage());
```
sarahkoop, line 96: Maybe this error message should be more specific in this case? Should we have the "Unable to fetch" as the default and return a specific error message about requiring authorization string or auth provider if both params are null?

Demo/src/main/java/com/braintreepayments/demo/PreferredPaymentMethodsFragment.java
```diff
@@ -51,68 +44,41 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
         getActivity().setProgressBarIndeterminateVisibility(true);
 
-        initializeFeatureClients(initError -> {
-            PayPalCheckoutRequest payPalRequest = createPayPalCheckoutRequest(getActivity(), "1.00");
-            payPalClient.tokenizePayPalAccount(getActivity(), payPalRequest, requestError -> {
-                if (requestError != null) {
-                    handleError(requestError);
-                }
-            });
-        });
```
sarahkoop, line 88: Do we not need to actually launch these anymore?

GooglePay/src/main/java/com/braintreepayments/api/GooglePayClient.java
```diff
@@ -190,33 +199,41 @@ public void requestPayment(@NonNull final FragmentActivity activity, @NonNull fi
...
+
+                            setGooglePayRequestDefaults(configuration, authorization, request);
+                            braintreeClient.sendAnalyticsEvent("google-payment.started");
+
+                            PaymentDataRequest paymentDataRequest = PaymentDataRequest.fromJson(request.toJson());
+                            Intent intent = new Intent(activity, GooglePayActivity.class)
+                                    .putExtra(EXTRA_ENVIRONMENT, getGooglePayEnvironment(configuration))
+                                    .putExtra(EXTRA_PAYMENT_DATA_REQUEST, paymentDataRequest);
+
+                            activity.startActivityForResult(intent, BraintreeRequestCodes.GOOGLE_PAY);
```
sarahkoop, line 228:
```suggestion
                            // TODO: Use activity result API to launch this activity
                            activity.startActivityForResult(intent, BraintreeRequestCodes.GOOGLE_PAY);
```

Just noticed maybe we should add a TODO to update this internal activity launching to the new API as well

PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
```diff
@@ -35,66 +35,75 @@
...
+                                callback.onResult(null, configError);
+                                return;
+                            }
+                            try {
+                                final boolean isBillingAgreement = payPalRequest instanceof PayPalVaultRequest;
+                                String endpoint = isBillingAgreement
+                                        ? SETUP_BILLING_AGREEMENT_ENDPOINT : CREATE_SINGLE_PAYMENT_ENDPOINT;
+                                String url = String.format("/v1/%s", endpoint);
+
+                                // TODO: call async getAuthorization here
```
sarahkoop, line 55: Is this for a future PR?

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
```diff
@@ -144,45 +144,55 @@ public void onResult(@Nullable Configuration configuration, @Nullable Exception
...
-                if (configuration.getCardinalAuthenticationJwt() == null) {
-                    Exception authError = new BraintreeException("Merchant is not configured for 3DS 2.0. " +
-                            "Please contact Braintree Support for assistance.");
-                    callback.onResult(null, null, authError);
-                    return;
-                }
+                    final JSONObject lookupJSON = new JSONObject();
+                    try {
+                        lookupJSON
+                                // TODO: call async getAuthorization here
```
sarahkoop, line 155: Is this for this PR or future?

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
```diff
@@ -144,45 +144,55 @@ public void onResult(@Nullable Configuration configuration, @Nullable Exception
...
+                        @Override
+                        public void onResult(@Nullable Configuration configuration, @Nullable Exception configError) {
+                            if (configuration == null) {
+                                callback.onResult(null, null, configError);
+                                return;
+                            }
+                            if (configuration.getCardinalAuthenticationJwt() == null) {
+                                Exception authError = new BraintreeException("Merchant is not configured for 3DS 2.0. " +
+                                        "Please contact Braintree Support for assistance.");
+                                callback.onResult(null, null, authError);
```
sarahkoop, line 175:
```suggestion
                                Exception configError = new BraintreeException("Merchant is not configured for 3DS 2.0. " +
                                        "Please contact Braintree Support for assistance.");
                                callback.onResult(null, null, configError);
```

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
sarahkoop: We do use this in Drop-in now to set the authorization in the launch intent extras. I think we might want to keep this until we confirm the use case in Drop-in

sarahkoop:
I think something is not quite right with the `ThreeSecureClient`. The `ThreeDSecureVerificationTests` in the demo are failing currently

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.java
sshropshire: I guess technically we could in order to make the linter happy, but we do produce the `inputData` being sent in here.

sshropshire: It is a good call here to defensively check for `null`. I'll add some tests in for that too.

BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationLoader.java
sshropshire: I believe in `v3` and `v4` we assume the merchant has recently fetched a client token and we keep a reference for as long as the SDK client / BraintreeFragment is in memory. This mimics that behavior since the auth provider is owned by the `BraintreeClient` instance, they'll be de-allocated at the same time.

sshropshire: Expired tokens will result in an error callback and that should be at parity with how it works now. AFAIK we haven't considered caching client tokens but this opens up the possibility. We can gather more information about client token caching we can make this a feature in another minor version.

BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationLoader.java
sshropshire: For here we're assuming if the merchant has implemented a `ClientTokenProvider`, we can infer that they will fetch a `ClientToken` without having to fetch the token before it's needed. That is a good point for future auth types, I think there may be a way to remove the need for this method.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.java
sshropshire: Added the `null` checks!

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
sshropshire: Done!

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
sshropshire: Done!

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
sshropshire: Good call this needs to be removed.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
sshropshire: This is technically still an `authError` with cardinal. This block of code didn't change in this PR it just got moved.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
sshropshire: Drop-in should be able to use the async version too. This would initially be `null` when a `ClientTokenProvider` is used and would invite race conditions in DropIn if the sync version is relied on.

BraintreeCore/src/main/java/com/braintreepayments/api/ClientTokenCallback.java
sshropshire: I went with ClientToken prefixes to encourage merchants to fetch a client token, although you can technically callback any valid form of Authorization with a `ClientTokenProvider`. I can add doc strings but maybe a second pass on the naming once everything's in place?

BraintreeCore/src/test/java/com/braintreepayments/api/AnalyticsClientUnitTest.java
sshropshire: Good call this test is identical to the one below. It's a copy-paste error I'll revert.

BraintreeCore/src/test/java/com/braintreepayments/api/AuthorizationLoaderUnitTest.java
sshropshire: Yeah definitely should be improved. I added a dev docs URL. This happens when `null` is passed in for either an auth string or a `ClientTokenProvider`. Our API says `@NonNull` for either parameter but it's still technically possible.

Demo/src/main/java/com/braintreepayments/demo/PreferredPaymentMethodsFragment.java
sshropshire: Yeah good call. I added those snippets back in. I used the refactoring tool the IDE clobbered some of the code.

Demo/src/main/java/com/braintreepayments/demo/DemoActivity.java
```diff
@@ -51,15 +45,13 @@ protected void onCreate(Bundle savedInstanceState) {
...
-            authorization = savedInstanceState.getString(EXTRA_AUTHORIZATION);
-            customerId = savedInstanceState.getString(EXTRA_CUSTOMER_ID);
-        }
-
         setupActionBar();
         setProgressBarIndeterminateVisibility(true);
 
+        // perform reset on preference change
+        Settings.getPreferences(this)
+                .registerOnSharedPreferenceChangeListener((sharedPreferences, s) -> performReset());
```
sshropshire, line 53: This is a bit of a hack but we should be able to remove it once each fragment owns the creation of its own `BraintreeClient`. In the future, the concept of a "reset" may no longer be necessary in the Demo app.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
```diff
@@ -91,18 +103,33 @@ public BraintreeClient(@NonNull Context context, @NonNull String authorization)
      * @param returnUrlScheme A custom return url to use for browser and app switching
      */
     public BraintreeClient(@NonNull Context context, @NonNull String authorization, @NonNull String returnUrlScheme) {
-        this(createDefaultParams(context, authorization, returnUrlScheme));
+        this(createDefaultParams(context, authorization, null, returnUrlScheme));
+    }
+
+    /**
+     * Create a new instance of {@link BraintreeClient} using a tokenization key or client token and a custom url scheme.
```
sarahkoop, line 110:
```suggestion
     * Create a new instance of {@link BraintreeClient} using a {@link ClientTokenProvider} and a custom url scheme.
```

sarahkoop (approved):
Tests look good now, this is a great change 🔥

cgdibble (approved):

PR #508, opened 2022-01-28
3DS Encapsulate App Switching

sshropshire:
### Summary of changes

  * Add `ThreeDSecureListener` to receive results from the 3DS flow
  * Deprecate methods requiring a callback in favor of listener pattern
  
**Note**: This is one of several PRs that will update the payments methods that require app/browser switching to leverage the newer Activity Result API and encapsulate the app/browser switching logic.

 ### Checklist

 - [X] Added a changelog entry

- @cgdibble 
- @andrespelaezp 
- @sshropshire 
- @sarahkoop


ThreeDSecure/src/main/java/com/braintreepayments/api/CardinalResult.java
```diff
@@ -0,0 +1,42 @@
+package com.braintreepayments.api;
+
+import com.cardinalcommerce.cardinalmobilesdk.models.ValidateResponse;
+
+class CardinalResult {
```
sarahkoop, line 5: This class is needed to support a single result object required by the `ActivityResultContract`

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureAPI.java
```diff
@@ -0,0 +1,87 @@
+package com.braintreepayments.api;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+class ThreeDSecureAPI {
```
sarahkoop, line 6: This class was created to extract out some the API logic and clean up testing.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
```diff
@@ -32,18 +34,77 @@
...
+
+    /**
+     * Create a new instance of {@link ThreeDSecureClient} using a {@link BraintreeClient}.
+     *
+     * Deprecated. Use {@link ThreeDSecureClient(Fragment, BraintreeClient)} or
+     * {@link ThreeDSecureClient(FragmentActivity, BraintreeClient)} instead.
+     *
+     * @param braintreeClient a {@link BraintreeClient}
+     */
+    @Deprecated
```
sarahkoop, line 71: I ended up deprecating the old integration pattern to make it more clear which methods support the new pattern. We can revisit whether or not we want to deprecate these methods yet, though. I just wasn't sure how to make it clear which methods go together without deprecating.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
```diff
@@ -117,25 +179,6 @@ public void onResult(String consumerSessionId, Exception error) {
...
-     * Continues the 3DS verification. Should be called from {@link ThreeDSecureResultCallback#onResult(ThreeDSecureResult, Exception)}
-     *
-     * @param activity Android FragmentActivity
-     * @param request  the {@link ThreeDSecureRequest} with information used for authentication.
-     * @param result   the {@link ThreeDSecureResult} returned for this request.
-     *                 Contains information about the 3DS verification request that will
-     *                 be invoked in this method.
-     * @param callback {@link ThreeDSecureResultCallback}
-     */
-    public void continuePerformVerification(@NonNull final FragmentActivity activity, @NonNull final ThreeDSecureRequest request, @NonNull final ThreeDSecureResult result, @NonNull final ThreeDSecureResultCallback callback) {
```
sarahkoop, line 130: This method was just moved to the //Deprecated region

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
```diff
@@ -196,25 +239,152 @@ public void onResult(String consumerSessionId, Exception error) {
...
+//     * which points to the original payment method, as well as the 3D Secure verification.
+//     * Transactions created with this nonce will be 3D Secure, and benefit from the appropriate
+//     * liability shift if authentication is successful or fail with a 3D Secure failure.
+//     *
+//     * The result of this verification will be returned to your {@link ThreeDSecureListener}.
+//     *
+//     * @param activity Android FragmentActivity
+//     * @param request  the {@link ThreeDSecureRequest} with information used for authentication.
+//     */
+//    public void performVerification(@NonNull final FragmentActivity activity, @NonNull final ThreeDSecureRequest request) {
```
sarahkoop, line 262: We updated this method to follow the listener result pattern, but realized that if the merchant invoked this method, and receives a result in their listener `onSuccess` method, they then nee to call `continuePerformVerification` from that listener `onSuccess` method, and would receive another result back in the same listener method. This seemed weird. Instead, we opted to leave the callback pattern for the `performVerification` method, and use the listener pattern for the methods that actually invoke a browser/app switch. This is documented in a code snipped in the `RESULT_API_MIGRATION.md`

sarahkoop: Left it commented out in case anyone thinks we should offer this with the listener instead. Otherwise will remove.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
```diff
@@ -196,25 +239,152 @@ public void onResult(String consumerSessionId, Exception error) {
...
+    /**
+     * Continues the 3DS verification. Should be called from {@link ThreeDSecureResultCallback#onResult(ThreeDSecureResult, Exception)}.
+     * The result of this verification will be returned to your {@link ThreeDSecureListener}.
+     *
+     * @param activity an Android FragmentActivity
+     * @param request the {@link ThreeDSecureRequest} with information used for authentication
+     * @param result a {@link ThreeDSecureResult} that contains information about the 3DS verification request.
+     */
+    public void continuePerformVerification(@NonNull final FragmentActivity activity, @NonNull final ThreeDSecureRequest request, @NonNull final ThreeDSecureResult result) {
+        continuePerformVerification(activity, request, result, new ThreeDSecureResultCallback() {
```
sarahkoop, line 284: We opted to just invoke the existing deprecated methods from within the new methods and return the results to the listener to avoid code/test duplication.

ThreeDSecure/src/test/java/com/braintreepayments/api/ThreeDSecureClientUnitTest.java
```diff
@@ -317,50 +348,285 @@ public void onBrowserSwitchResult_whenFailure_postsErrorWithResponse() throws Ex
...
+        verify(listener).onThreeDSecureFailure(captor.capture());
+        verify(listener, never()).onThreeDSecureSuccess(any(ThreeDSecureResult.class));
 
         Exception exception = captor.getValue();
         assertTrue(exception instanceof UserCanceledException);
         assertEquals("User canceled 3DS.", exception.getMessage());
     }
+
+    @Test
+    public void onBrowserSwitchResult_whenListenerNull_doesNothing() {
```
sarahkoop, line 408: Marking a note here to call out our questions noted in this test.

Demo/src/main/java/com/braintreepayments/demo/DemoActivity.java
```diff
@@ -80,13 +80,15 @@ public BraintreeClient getBraintreeClient() {
         return braintreeClient;
     }
 
+    // TODO: Remove this method completely when all payment methods have been migrated to
```
sshropshire, line 83: We could delete this code since we're merging into `next-minor-version`.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureAPI.java
```diff
@@ -0,0 +1,87 @@
...
+                    callback.onResult(null, httpError);
+                }
+            }
+        });
+    }
+
+    private void notify3DSComplete(ThreeDSecureResult result, ThreeDSecureResultCallback callback) {
+        ThreeDSecureInfo info = result.getTokenizedCard().getThreeDSecureInfo();
+
+        braintreeClient.sendAnalyticsEvent(String.format("three-d-secure.verification-flow.liability-shifted.%b", info.isLiabilityShifted()));
```
sshropshire, line 82: nit: I do feel like analytics should live in the client and API would purely be responsible for interacting with the gateway.

ThreeDSecure/src/test/java/com/braintreepayments/api/ThreeDSecureClientUnitTest.java
sshropshire: We could no-op when the listener is `null` and then defer delivery until a listener is attached.

Demo/src/main/java/com/braintreepayments/demo/DemoActivity.java
sarahkoop: 👍

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
sshropshire: I guess it depends if we want to add another `onLookupComplete` listener to the interface? We can be flexible, it doesn't have to exactly match the old way. There may be an opportunity here to enhance the 3DS developer experience.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureAPI.java
sarahkoop: Ok yeah that's fair we can update this

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
sshropshire: That's legit. The tests could use some general clean-up anyway as well. If it's easier to duplicate code we could circle back and refine everything later.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureAPI.java
sarahkoop: Done 👍

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
sarahkoop: Yeah that might be nice for merchants who want to only use the listener pattern. I don't mind keeping listeners for browser/app switches and callbacks for API though. Also if we add that additional method to the listener interface, merchants using server-side 3DS would have to implement that method even though they don't need it. Can go either way - thoughts?

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
```diff
@@ -32,18 +34,77 @@
...
         this.browserSwitchHelper = browserSwitchHelper;
+        this.api = threeDSecureAPI;
+        if (activity != null && lifecycle != null) {
+            addObserver(activity, lifecycle);
+        }
+    }
+
+    @VisibleForTesting
+    void addObserver(@NonNull FragmentActivity activity, @NonNull Lifecycle lifecycle) {
+        if (observer == null) {
```
sshropshire, line 89: nit: won't this always be null?

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
sshropshire: qq: Do we need to uncomment this method?

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
```diff
@@ -196,25 +238,152 @@ public void onResult(String consumerSessionId, Exception error) {
...
      *
+     * Deprecated. Use {@link ThreeDSecureClient#initializeChallengeWithLookupResponse(FragmentActivity, String)} instead.
+     *
      * @param activity Android FragmentActivity
      * @param lookupResponse The lookup response from the server side call to lookup the 3D Secure information.
      * @param callback {@link ThreeDSecureResultCallback}
      */
+    @Deprecated
     public void initializeChallengeWithLookupResponse(@NonNull FragmentActivity activity, @NonNull String lookupResponse, @NonNull ThreeDSecureResultCallback callback) {
         initializeChallengeWithLookupResponse(activity, null, lookupResponse, callback);
```
sshropshire, line 373: random q: what happens here when `ThreeDSecureRequest` is null? I know it's a legacy thing but I can't remember if we do something by default.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
```diff
@@ -357,8 +485,8 @@ public void onBrowserSwitchResult(@NonNull BrowserSwitchResult browserSwitchResu
                         if (result.hasError()) {
                             callback.onResult(null, new ErrorWithResponse(422, authResponse));
                         } else {
-                            notify3DSComplete(result, callback);
-                        }
+                            sendLiabilityShiftedAnalytics(result);
```
sshropshire, line 488: 🔥

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureLifecycleObserver.java
```diff
@@ -0,0 +1,60 @@
...
+                });
+            case ON_RESUME:
+                FragmentActivity activity = null;
+                if (lifecycleOwner instanceof FragmentActivity) {
+                    activity = (FragmentActivity) lifecycleOwner;
+                } else if (lifecycleOwner instanceof Fragment) {
+                    activity = ((Fragment) lifecycleOwner).getActivity();
+                }
+
+                if (activity != null) {
```
sshropshire, line 55: Guarding the browser switch for only 3DS here would take some of the burden off of the client. We should no-op if it isn't 3DS.

ThreeDSecure/src/test/java/com/braintreepayments/api/ThreeDSecureClientUnitTest.java
sshropshire: Ah that's a good point. For now i'd say let's assume the listener is set. It's a two-liner integration pattern there's bound to be some weirdness somewhere lol. We can re-evaluate but I'd bank on that listener being set to keep the scope of this PR narrow and maybe place a TODO here and circle back before GA?

sshropshire (commented):
This is awesome! I left a few comments but overall LGTM. I can't approve because I opened the initial draft PR. @sarahkoop @cgdibble and @andrespelaezp took it to completion and I 👍 !

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
sarahkoop: This method was added in case we wanted to follow the listener pattern for all methods. Since we are ok with callback only for non app/browser switch methods we can remove. I'll do that 👍

LISTENER_MIGRATION.md
```diff
@@ -0,0 +1,69 @@
...
+    override fun onThreeDSecureFailure(error: Exception) {
+        // handle error
+    }
+}
+
+// MerchantFragment.kt
+class MerchantFragment: Fragment(), ThreeDSecureListener {
+    
+    lateinit var braintreeClient: BraintreeClient
+    lateinit var paypalClient: PayPalClient
```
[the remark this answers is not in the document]
sarahkoop: Yep that was a copy/paste error - good catch!

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
sarahkoop: Looks like the only thing request is used for here is V1 UI customizations (in ThreeDSecureV1BrowserSwitchHelper), and it just doesn't pass any UI customization params if null.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
sarahkoop: Yes good call - removed.

andrespelaezp (approved):
Looking nice!

cgdibble (approved):

PR #510, opened 2022-02-01
Fix Venmo Issue and Hide Authorization Loader

sshropshire:
### Summary of changes

 - Fix issue where null value causes VenmoAccountNonce#fromJSON() to throw.
 - Delay release of `AuthorizationProvider` until we finalize the interface (see [discussion](https://github.com/braintree/braintree_android/discussions/496))

 ### Checklist

 - [x] Added a changelog entry

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationProvider.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationProvider.java
@@ -6,7 +6,7 @@ import androidx.annotation.NonNull;
  * Implement this interface to provide an asynchronous way for {@link BraintreeClient} to fetch
  * a client token from your server when it is needed.
  */
-public interface AuthorizationProvider {
+interface AuthorizationProvider {
 
     /**
      * Method used by {@link BraintreeClient} to fetch a client token.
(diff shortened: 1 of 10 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationProvider.java
```diff
@@ -6,7 +6,7 @@
  * Implement this interface to provide an asynchronous way for {@link BraintreeClient} to fetch
  * a client token from your server when it is needed.
  */
-public interface AuthorizationProvider {
+interface AuthorizationProvider {
```
crookedneighbor, line 9: Is this ok that it's no longer public?

crookedneighbor (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationProvider.java
sshropshire: Yeah we're still back and forth on the naming. Going to add back soon 🤞

sarahkoop (approved):

Venmo/src/main/java/com/braintreepayments/api/VenmoAccountNonce.java
```diff
@@ -65,12 +65,12 @@ static VenmoAccountNonce fromJSON(JSONObject inputJson) throws JSONException {
         return new VenmoAccountNonce(nonce, username, isDefault, json);
     }
 
-    VenmoAccountNonce(String nonce, String username, boolean isDefault, JSONObject json) throws JSONException {
```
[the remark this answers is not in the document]
sshropshire: In this case the `VenmoAccountNonce#fromJSON()` method still uses `has` internally.

We should migrate our codebase to use `opt` methods where it makes sense, but I guess when certain info is missing we would have to return `null` instead of throwing.

PR #513, opened 2022-02-14
PayPal - Encapsulate Browser Switch Logic

sarahkoop:
### Summary of changes

- Add `PayPalListener` to receive results from the PayPal flow
- Deprecate methods requiring a callback in favor of listener pattern
- Move API logic to `PayPalInternalClient`

 ### Checklist

 - [x] Added a changelog entry

### DIFF
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
@@ -43,6 +43,10 @@ public class PayPalClient {
     PayPalClient(FragmentActivity activity, Lifecycle lifecycle, BraintreeClient braintreeClient, PayPalInternalClient internalPayPalClient) {
         this.braintreeClient = braintreeClient;
         this.internalPayPalClient = internalPayPalClient;
+        if (activity != null && lifecycle != null) {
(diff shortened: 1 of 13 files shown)


PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
```diff
@@ -18,19 +20,68 @@
...
 
     private final BraintreeClient braintreeClient;
-    private final ApiClient apiClient;
-
     private final PayPalInternalClient internalPayPalClient;
 
+    private PayPalListener listener;
+
+    @VisibleForTesting
+    BrowserSwitchResult pendingBrowserSwitchResult;
```
sshropshire, line 28: nit: we may be able to leverage shared prefs as a source of truth instead of member variables when notifying listeners of a pending browser switch result after `setListener()` is called.

sshropshire (approved):

PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
sshropshire: Actually talking it through with Sarah she reminded me about us needing to have access to the `Intent`. This is the best approach.

cgdibble (approved):

PR #514, opened 2022-02-14
Update docs links to point to PayPal domain

Epreuve:
### Summary of changes

 - Update docs links/strings to point to PayPal Braintree docs domain
 - Fix broken docs to point to most appropriate replacement docs

 ### Checklist

- @Epreuve @jaxdesmarais

### DIFF
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
@@ -41,16 +41,20 @@ public class PayPalClient {
         return !braintreeClient.canPerformBrowserSwitch(activity, BraintreeRequestCodes.PAYPAL);
     }
 
+    //TODO NEXT_MAJOR_VERSION
+    //Change link to docs to https://developer.paypal.com/braintree/docs/guides/paypal/overview/android/v4
     private static Exception createPayPalError() {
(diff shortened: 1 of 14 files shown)


PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
```diff
@@ -43,14 +43,14 @@ private boolean browserSwitchNotPossible(FragmentActivity activity) {
 
     private static Exception createPayPalError() {
         return new BraintreeException("PayPal is not enabled. " +
-                "See https://developers.braintreepayments.com/guides/paypal/overview/android/ " +
+                "See https://developer.paypal.com/braintree/docs/guides/paypal/overview/android/v4 " +
```
sarahkoop, line 46: I think in the past we've had discussions around whether or not changing an error message is a breaking change - I think it's better to change than to have broken links, but wanted to double check.

sshropshire (approved):

PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
Epreuve: We reverted the changes here and in the 3DS client and added next major version notes to update them (at least).

sarahkoop (approved):

PR #515, opened 2022-02-15
Local Payment - Encapsulate Browser Switch Logic

sarahkoop:
### Summary of changes

 - Add `ThreeDSecureListener` to receive results from the 3DS flow
 - Deprecate methods requiring a callback in favor of listener pattern
 - Extract API calls to `LocalPaymentApi`
 
 ### Checklist

 - [x] Added a changelog entry

- @andrespelaezp 
- @sarahkoop

### DIFF
--- a/LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentClient.java
+++ b/LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentClient.java
@@ -25,7 +25,9 @@ public class LocalPaymentClient {
     private final PayPalDataCollector payPalDataCollector;
     private final LocalPaymentApi localPaymentApi;
     private LocalPaymentListener listener;
-    private BrowserSwitchResult pendingBrowserSwitchResult;
+
+    @VisibleForTesting
+    BrowserSwitchResult pendingBrowserSwitchResult;
 
     /**
      * Create a new instance of {@link LocalPaymentClient} from within an Activity using a {@link BraintreeClient}.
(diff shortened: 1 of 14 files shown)


LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentClient.java
```diff
@@ -16,20 +18,74 @@
...
+
+    /**
+     * Add a {@link LocalPaymentListener} to your client to receive results or errors from the Local Payment flow.
+     * This method must be invoked on a {@link LocalPaymentClient(Fragment, BraintreeClient)} or
+     * {@link LocalPaymentClient(FragmentActivity, BraintreeClient)} in order to receive results.
+     *
+     * @param context an Android Context
+     * @param listener a {@link LocalPaymentListener}
+     */
+    public void setListener(Context context, LocalPaymentListener listener) {
```
sarahkoop, line 84: This method currently requires context, because the `onBrowserSwitchResult` method requires context to call `PayPalDataCollector#getClientMetadataId(Context context)`. Looking at the docs for that method, it doesn't seem like we can pre-call this method, or cache the value.

Requiring context on a listener setter doesn't feel like a great merchant experience. This is only needed so that we don't fail silently when a merchant doesn't set their listener before trying to handle a browser switch result. We think in this case it would be better to remove `Context` from this method, and not try to deliver a pending browser switch result when the listener is set. We left `Context` in for now, so that you can see what the experience would look like. Open to other thoughts on the best approach for this.

sshropshire: `BraintreeClient` has an accessor for `applicationContext` we could use that internally here.

sshropshire (approved):

LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentClient.java
sarahkoop: Ah of course - good call!

jaxdesmarais (approved):
⛵

PR #523, opened 2022-03-04
Encapsulate App and Browser Switching

sarahkoop:
### Summary of changes

 - This PR is to merge in `next-minor-version`, which contains changes from multiple PRs (one for each app/browser switch payment method). We'd really like extra review on the public/private methods, merchant facing method names and doc strings, and migration guide.
 - Add `AuthorizationProvider` to allow `BraintreeClient` to asynchronously fetch authorization as needed
 - Encapsulate app/browser switch flows for each payment method and return results to a listener interface
 - Add v4.9.0+ migration guide with latest merchant integration story.

These changes originated from [this discussion](https://github.com/braintree/braintree_android/discussions/496).

 ### Checklist

 - [x] Added a changelog entry

- This PR is the culmination of a large group effort!


LISTENER_MIGRATION.md
```diff
@@ -0,0 +1,338 @@
+# Migration Guide
```
sshropshire, line 1: random: I wonder if we should merge this into the v4 migration guide instead of keeping them separate?

LISTENER_MIGRATION.md
```diff
@@ -0,0 +1,338 @@
...
+          BraintreeClient(this, MerchantClientTokenProvider())
+          
+        threeDSecureClient = ThreeDSecureClient(this, braintreeClient)
+        threeDSecureClient.listener = this
+    }
+    
+    private fun launch3DS() {
+        ...
+        threeDSecureClient.performVerification(this, threeDSecureRequest) { threeDSecureLookupResult, lookupError ->
+          threeDSecureClient.continuePerformVerification(this@MerchantActivity, threeDSecureRequest, threeDSecureLookupResult)
```
sshropshire, line 175: future: I wonder if we can convert this perform verification callback into additional listener methods?

sshropshire: Could be next major version too when we make this the primary integration method.

PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
```diff
@@ -198,10 +272,39 @@ private static String getAnalyticsEventPrefix(PayPalRequest request) {
         return request instanceof PayPalVaultRequest ? "paypal.billing-agreement" : "paypal.single-payment";
     }
 
+    void onBrowserSwitchResult(FragmentActivity activity) {
+        this.pendingBrowserSwitchResult = braintreeClient.deliverBrowserSwitchResult(activity);
+
+        if (pendingBrowserSwitchResult != null && listener != null) {
```
sshropshire, line 278: I wonder if we should no-op when there's no listener. It seems like some merchants like to have the option to handle the result in a separate activity / fragment. It may be for another PR honestly, but that has come up a few times.

LISTENER_MIGRATION.md
sarahkoop: Yeah I think we did the `performVerification` with a callback because of the server side 3DS option - if we added another listener method, merchants not needing it would still have to implement it.

LISTENER_MIGRATION.md
sarahkoop: Agreed. Will update.

LISTENER_MIGRATION.md
sshropshire: Yeah that's true I forgot about that part of it.

LISTENER_MIGRATION.md
sarahkoop: Realizing now that might be a little confusing - having multiple code snippets for each flow that aren't compatible. Do you think we should replace the current migration guide code snippets/documentation with this new implementation pattern? Otherwise we could rename this guide to v4.9.0_MIGRATION_GUIDE and include more description about how/why to migrate to this pattern?

sarahkoop: I think what I'll do is move all of the still relevant portions of the migration guide, and the new listener pattern snippets to a `v4.9_MIGRATION_GUIDE` and then link to that new guide from the README and the old migration guide (I don't want to lost the old snippets for merchants stuck in the middle)

sshropshire: That's a really good idea actually `v4.9.0_MIGRATION_GUIDE` I like that it's explicit about the affected version. We'll have an opportunity to clean up the migration guides with the next major version too.

BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationProvider.java
```diff
@@ -6,7 +6,7 @@
  * Implement this interface to provide an asynchronous way for {@link BraintreeClient} to fetch
  * a client token from your server when it is needed.
  */
-interface AuthorizationProvider {
+public interface AuthorizationProvider {
```
sarahkoop, line 9: We are calling this AuthorizationProvider but the method name is `getClientToken` - should that be `getAuthorization`?

v4.9.0+_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,1019 @@
+# Braintree Android v4.9.0+ Migration Guide
```
sarahkoop, line 1: This copies the entire contents of v4 migration guide, the main updates are for the payment method code snippets, the addition of the Authorization Provider and Fragment Architecture sections

v4.9.0+_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,1019 @@
...
+      public void failure(RetrofitError error) {
+        // forward error to the SDK
+        callback.onFailure(error);
+      }
+    });
+  }
+}
+```
+
+Then in an Activity or Fragment, create an instance of `BraintreeClient`:
```
sarahkoop, line 167: Can this be done in a Fragment? Originally said Activity, but in our example code snippets we were doing this in a fragment previously

BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationProvider.java
sshropshire: Yeah that's a weird part since the `Authorization` type is no longer merchant-facing. Initially `ClientTokenProvider` / `getClientToken` went well together, but technically a merchant could provide a tokenization key here.

v4.9.0+_MIGRATION_GUIDE.md
sshropshire: It is possible but we never actually made overloads for `Fragment`. I wonder if we should to simplify the integration pattern?

BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationProvider.java
sshropshire: This naming was also brought up in the DX GitHub [discussion](https://github.com/braintree/braintree_android/discussions/496#discussioncomment-2062004).

sshropshire (approved):
🌊 &nbsp;🔥

cgdibble (approved):

PR #524, opened 2022-03-07
Update GraphQL pinned Certificates

Epreuve:
### Summary of changes

 - Remove expired Digicert leaf certificates
 - Add new Digicert Intermediate OV and root cert
 
 The functionality can be tested using the `update-graphql-certs-tests` branch.  There is some setup required to correctly test and the logic/hacks there are not meant to be retained and should be deleted as they are unnecessary once this PR is merged.

 ### Checklist

 - [x] Added a changelog entry

- @Epreuve @saralvasquez


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree Android SDK Release Notes
 
+## Unreleased
+* BraintreeCore
+  * Update pinned certificates used by `BraintreeGraphQLClient`
```
sarahkoop, line 5: `BraintreeGraphQLClient` is internal so not sure if this is confusing for merchants. Do we want a more generic update pinned certificates statement or is this helpful for our future selves?

sarahkoop (approved):
Ran the integration tests and all looks good. Happy to do additional manual testing if needed, otherwise LGTM

sshropshire (approved):

CHANGELOG.md
Epreuve: I went back and forth on this and am definitely open to the best way to surface this.  

I wouldn't normally put an internal change like this in the changelog, but since it's an extremely important change (since it will eventually impact tokenization for version prior) that we might want/need to point to directly and make discoverable, I figured this was worth the specific call out.

PR #525, opened 2022-03-08
Implement SEPA Mandate Browser Switch

sarahkoop:
### Summary of changes

 - Implement browser switch to display SEPA Debit mandate 
 - Implement SEPADebitLifecycleObserver to encapsulate browser switching logic for the merchant
 - Parse success/error deep link URLs to prep for `tokenize` implementation (future PR)

 ### Checklist

- @jaxdesmarais 
- @sarahkoop


BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeRequestCodes.java
```diff
@@ -26,4 +27,5 @@
     int GOOGLE_PAY = 13593;
     int SAMSUNG_PAY = 13595;
     int LOCAL_PAYMENT = 13596;
+    int SEPA = 13597;
```
sarahkoop, line 30: We just incremented the previous request code, but is there any info on how these numbers were determined or what this new one should be?

Demo/src/main/java/com/braintreepayments/demo/DemoActivity.java
```diff
@@ -63,7 +63,7 @@ protected void onResume() {
                 != PackageManager.PERMISSION_GRANTED) {
             ActivityCompat.requestPermissions(this, new String[]{WRITE_EXTERNAL_STORAGE}, 1);
         }
-        handleBrowserSwitchResultIfNecessary();
+//        handleBrowserSwitchResultIfNecessary();
```
sarahkoop, line 66: This is necessary until #523 is merged

Demo/src/main/java/com/braintreepayments/demo/SEPADebitFragment.java
```diff
@@ -46,15 +49,23 @@ public void launchSEPADebit(View view) {
 
         SEPADebitRequest request = new SEPADebitRequest();
         request.setAccountHolderName("John Doe");
-        request.setCustomerId("a-customer-id");
-        request.setIban("FR7618106000321234566666610");
+        request.setCustomerId(generateRandomCustomerId());
+        request.setIban(generateRandomIban());
```
sarahkoop, line 53: A specific IBAN/customer ID mandate can only be approved once, so randomly generating for continued testing.

SEPADebit/src/main/java/com/braintreepayments/api/SEPADebitRequest.java
```diff
@@ -39,6 +40,7 @@ public String getIban() {
     }
 
     /**
+     * Required.
      * @param iban The full IBAN.
      */
     public void setIban(@NonNull String iban) {
```
sarahkoop, line 46: We wen't with the empty constructor/setter pattern to match most of our other request objects, but all params except merchant account ID are required. Should we add the required params to the constructor instead? Since they are required but technically could be null, should all have `@Nullable` instead of `@NonNull` annotation?

SEPADebit/src/main/java/com/braintreepayments/api/SEPADebitClient.java
```diff
@@ -47,7 +63,8 @@ public SEPADebitClient(@NonNull Fragment fragment, @NonNull BraintreeClient brai
      * @param listener a {@link SEPADebitListener}
      */
     public void setListener(SEPADebitListener listener) {
-
+        // TODO: handle delivering pending browser switch result when listener is set
```
sarahkoop, line 66: This TODO was to align with the other client `setListener` methods in #523 - but this case is a little different, since we will call `tokenize` after browser switching. In this case a merchant will have to set the listener to receive results from tokenize - does that seem like the right approach?

SEPADebit/src/main/java/com/braintreepayments/api/SEPADebitRequest.java
sshropshire: I'm up for either, but honestly I do like the empty constructor pattern. Do we callback an exception when a param is missing?

sshropshire: I'd say whichever choice makes the API most consistent is the best path to take.

sshropshire (approved):
🌊

cgdibble (approved):

SEPADebit/src/main/java/com/braintreepayments/api/SEPADebitRequest.java
sarahkoop: Ok will update - I agree leave the empty constructor, I think all the params should be nullable since they _could_ be null, and we'll add checks in the tokenize method to return error if any aren't set 👍

sarahkoop: Oh wait forgot the API would handle this and return a descriptive error so we left off the null check

PR #533, opened 2022-03-30
Update Cardinal Library for AndroidX Support

sarahkoop:
### Summary of changes

 - Update Cardinal library to 2.2.6-1 (adds AndroidX support)
 - Fix NPE in 3DS v2 base UI customization class (and deprecate class)
 - Simplify parcelable implementation for UI customization classes
 - Remove Jetifier
 - Add additional dependency to Samsung Pay for AndroidX support

 ### Checklist

 - [x] Added a changelog entry

- @jaxdesmarais 
- @sarahkoop

### DIFF
--- a/SamsungPay/build.gradle
+++ b/SamsungPay/build.gradle
@@ -37,6 +37,7 @@ dependencies {
     api project(':Card')
 
     api deps.samsungPay
+    implementation deps.annotation
 
     testImplementation deps.robolectric
     testImplementation deps.junitTest
--- a/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureV2ButtonCustomization.java
+++ b/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureV2ButtonCustomization.java
@@ -12,11 +12,6 @@ import com.cardinalcommerce.shared.userinterfaces.ButtonCustomization;
  */
 public class ThreeDSecureV2ButtonCustomization extends ThreeDSecureV2BaseCustomization implements Parcelable {
 
-    private String backgroundColor;
-    private int cornerRadius;
-    private String textFontName;
-    private String textColor;
-    private int textFontSize;
     private final ButtonCustomization cardinalButtonCustomization = new ButtonCustomization();
 
     public ThreeDSecureV2ButtonCustomization() {
@@ -26,7 +21,6 @@ public class ThreeDSecureV2ButtonCustomization extends ThreeDSecureV2BaseCustomi
      * @param textFontName Font type for the UI element.
      */
     public void setTextFontName(@Nullable String textFontName) {
-        this.textFontName = textFontName;
         cardinalButtonCustomization.setTextFontName(textFontName);
     }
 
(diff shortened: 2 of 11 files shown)


SamsungPay/build.gradle
```diff
@@ -37,6 +37,7 @@ dependencies {
     api project(':Card')
 
     api deps.samsungPay
+    implementation deps.annotation
```
sarahkoop, line 40: This was required to remove jetifier for Samsung Pay

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureV2ButtonCustomization.java
```diff
@@ -12,27 +12,66 @@
  */
 public class ThreeDSecureV2ButtonCustomization extends ThreeDSecureV2BaseCustomization implements Parcelable {
 
-    private String backgroundColor;
```
sarahkoop, line 15: We opted to remove this internal member variables across all the customization classes in favor of getting/setting the actual customization features on the cardinal customization classes.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureV2ButtonCustomization.java
```diff
@@ -62,15 +111,35 @@ public int describeContents() {
...
 
     protected ThreeDSecureV2ButtonCustomization(Parcel in) {
-        super(in);
+        textFontName = in.readString();
+        textColor = in.readString();
+        textFontSize = in.readInt();
         backgroundColor = in.readString();
         cornerRadius = in.readInt();
+
+        if (textFontName != null) {
```
sarahkoop, line 128: Our parcelable implementation was setting member variables in these classes, but not setting the customizations on the actual cardinal customization class. Updated the implementations to set the cardinal customizations too.

gradle.properties
```diff
@@ -1,4 +1,3 @@
-android.enableJetifier=true
```
sshropshire, line 1: 🔥

sshropshire (approved):

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureV2TextBoxCustomization.java
```diff
@@ -12,13 +12,62 @@
  */
 public class ThreeDSecureV2TextBoxCustomization extends ThreeDSecureV2BaseCustomization implements Parcelable {
 
+    private String textFontName;
+    private String textColor;
+    private int textFontSize;
```
jcnoriega, line 17: Whats the point of these 3 variables in all of the customizations? On the `setters` for each, they get set in the `customization` variable, an on the `getters` it returns the value directly from the `customization`. Can we dispose of these?

sarahkoop: Yeah we tried to remove these but they were needed for the parcelable implementation ☹️

sarahkoop: Ah good catch! The were needed in the parceling method but not as member variables, just updated to remove those 👍

jcnoriega (approved):

PR #542, opened 2022-04-19
Remove mention of setGoogleMerchantId in MIGRATION guides.

sshropshire:
### Summary of changes

 - This PR removes mention of `setGoogleMerchantId` for setting up a `GooglePayRequest`. This is no longer required for Google Pay integrations.

 ### Checklist


jaxdesmarais (approved):
⛵

sshropshire:
> should we un-deprecate these methods?

@sarahkoop got it, yeah maybe so. We should definitely add it back if needed.

PR #546, opened 2022-05-03
Method on client token provider for caching authorization

### LINKED ISSUE #544: No ability to invalidate cached credentials
### General information

* SDK/Library version: 4.9.0
* Environment: Production
* Android Version and Device: N/A
* Braintree dependencies:

com.braintreepayments.api:card:4.9.0
com.braintreepayments.api:data-collector:4.9.0
com.braintreepayments.api:google-pay:4.9.0
com.braintreepayments.api:paypal:4.9.0
com.braintreepayments.api:three-d-secure:4.9.0
com.braintreepayments.api:venmo:4.9.0

### Issue description

There is currently no way to invalidate the credentials once they are read once from a `ClientTokenProvider` or set via `initialAuthString`. Because of the instantiation strategy for 3rd party clients (google pay, venmo, etc.) on android requiring the instantiation to be prior to the onCreate lifecycle, there is no way to change the braintree client or invalidate the cached credentials once a token has been read.

We have a few places where we are being provided a braintree client token to use for authentication, and the permissions tied to that token may vary from screen to screen. Because we use a single activity application, the clients are shared across all screens. Once the authentication token is read once (either from the `initialAuthString` or `clientTokenProvider`), that authentication value persists for the entirety of the application. This behavior has downstream effects for us if, for example, 3D Secure is not enabled for the first token received, but is for a subsequent token. 

From the contract of the `ClientTokenProvider`, it appears from the onset that this will act like an interceptor and will read the client token value any time authorization is requested. Instead, the `ClientTokenProvider` class is essentially setting the `initialAuthString` in a `by lazy` way.

There are several solutions to this, which I'd be happy to make a PR for, but I'm not sure what is preferable. One of the main problems from an architecture perspective is that external consumers of the braintree api don't have any concept of the `AuthorizationLoader` or its caching strategy.

This is our interim solution, which uses reflection and is very undesirable:

```
private fun forceAuthorizationRefresh() = try {
        // Find the authorization loader field that has the cached auth credentials
        val authorizationLoaderField = braintreeClient.javaClass
            .getDeclaredField("authorizationLoader")
            .apply {
                isAccessible = true
            }

        val authorizationField = authorizationLoaderField
            .type
            .getDeclaredField("authorization")

        // Get the current value of the authorizationLoader field so that we can overwrite it
        val authorizationLoader = authorizationLoaderField.get(braintreeClient)

        authorizationField
            .apply {
                isAccessible = true
                // Actually apply the nullification here so the credential is refetched on next getAuthorization request
                set(authorizationLoader, null)
            }
    } catch (e: Exception) {
        e.printStackTrace()
    }
```

I think the most straightforward solution to this would be to add a method to the interface for `ClientTokenProvider` along the lines of a `boolean fetchNewToken()` (name pending!) and is then used in the `AuthorizationLoader` code as such:

```
void loadAuthorization(@NonNull final AuthorizationCallback callback) {
        final boolean shouldForceClientTokenProviderUsage = clientTokenProvider != null && clientTokenProvider.fetchNewToken(); 
        if (authorization != null && !shouldForceClientTokenProviderUsage) {
            callback.onAuthorizationResult(authorization, null);
        } else if (clientTokenProvider != null) {
            clientTokenProvider.getClientToken(new ClientTokenCallback() {
                @Override
                public void onSuccess(@NonNull String clientToken) {
                    authorization = Authorization.fromString(clientToken);
                    callback.onAuthorizationResult(authorization, null);
                }

                @Override
                public void onFailure(@NonNull Exception error) {
                    callback.onAuthorizationResult(null, error);
                }
            });
        } else {
            String clientSDKSetupURL
                = "https://developer.paypal.com/braintree/docs/guides/client-sdk/setup/android/v4#initialization";
            String message = String.format("Authorization required. See %s for more info.", clientSDKSetupURL);
            callback.onAuthorizationResult(null, new BraintreeException(message));
        }
    }
```

Another option would be to just never use the cached value of `authorization` if `clientTokenProvider` is non-null, but that is probably a breaking change.

If the `BraintreeClient` could be swapped for the 3rd party clients, that is also an option (albeit more dangerous since there are so many touchpoints for those variables).

A final option is probably to expose a method / param on the braintree client that disables the cached authentication.

josephyanks:

Resolves #544 

There is another option that adds the method `invalidateClientToken` on `BraintreeClient` seen in this PR: #547

I'll add tests and a changelog entry if / when we decide between the two PR's

### Summary of changes

 - Adds a `shouldUseCachedToken` method to the `ClientTokenProvider` that will be checked during the `loadAuthorization` call. If `true`, the cached authorization field will be set to null. 

 ### Checklist

 - [ ] Added a changelog entry

- @josephyanks

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationLoader.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationLoader.java
@@ -6,16 +6,22 @@ import androidx.annotation.Nullable;
 class AuthorizationLoader {
 
     private Authorization authorization;
+
     private final ClientTokenProvider clientTokenProvider;
 
     AuthorizationLoader(@Nullable String initialAuthString, @Nullable ClientTokenProvider clientTokenProvider) {
         this.clientTokenProvider = clientTokenProvider;
         if (initialAuthString != null) {
-            this.authorization = Authorization.fromString(initialAuthString);
+            authorization = Authorization.fromString(initialAuthString);
         }
     }
 
     void loadAuthorization(@NonNull final AuthorizationCallback callback) {
+        // Clear the token prior to making the request if the token shouldn't be cached any longer
+        if (clientTokenProvider != null && clientTokenProvider.shouldUseCachedToken() && authorization != null) {
+            authorization = null;
+        }
+
         if (authorization != null) {
             callback.onAuthorizationResult(authorization, null);
         } else if (clientTokenProvider != null) {
@@ -33,7 +39,7 @@ class AuthorizationLoader {
             });
         } else {
             String clientSDKSetupURL
-                = "https://developer.paypal.com/braintree/docs/guides/client-sdk/setup/android/v4#initialization";
+                    = "https://developer.paypal.com/braintree/docs/guides/client-sdk/setup/android/v4#initialization";
             String message = String.format("Authorization required. See %s for more info.", clientSDKSetupURL);
             callback.onAuthorizationResult(null, new BraintreeException(message));
         }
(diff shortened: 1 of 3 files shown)


josephyanks:
Withdrawing in favor of #547

PR #547, opened 2022-05-03
Method on braintree client for invalidating cached auth

### LINKED ISSUE #544: No ability to invalidate cached credentials

josephyanks:

Resolves #544 

I think that this is the most straightforward solution, although it involves adding a method onto the `BraintreeClient`, rather than in the `ClientTokenProvider`. There is another option that adds a `shouldUseCachedToken` method to the `ClientTokenProvider` seen in this PR: #546

I'll add tests and a changelog entry if / when we decide between the two PR's

### Summary of changes

 - Adds method `invalidateClientToken` on `BraintreeClient` that will invalidate the cached authorization credentials for the next `loadAuthorization`. Note that we can only invalidate this valid if a `ClientTokenProvider` was provided.

 ### Checklist

 - [x] Added a changelog entry

- @josephyanks


sshropshire:
@josephyanks this is awesome! Thanks for taking the time to contribute 🙏

I have a few thoughts:
1. I like the idea of having an `invalidateClientToken()` method on `BraintreeClient` – it's the public-facing entry point of the SDK created by the merchant. Allowing merchants to clear the Client Token cache feels like a good feature to add.
2. I think we can remove the `invalidateNextAuthorizationRequest` boolean flag in favor of setting the [authorization property](https://github.com/braintree/braintree_android/blob/c4bf27878725edaff705617ac6a617f98530cf8a/BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationLoader.java#L8) to `null`. `AuthorizationLoader` has logic in place to [re-fetch a client token](https://github.com/braintree/braintree_android/blob/c4bf27878725edaff705617ac6a617f98530cf8a/BraintreeCore/src/main/java/com/braintreepayments/api/AuthorizationLoader.java#L22) when needed.

josephyanks:
Just two follow up questions and I'll update the PR.

1. The reason I made the `invalidateNextAuthorizationRequest` was in case we crashed before the next `loadAuthorization` call. I see that braintree will pull the last known authorization object out for logging in the case of a crash. Do we want to create a second authorization field that retains the last authorization value (and ignores clears)?
2. I still think that we need to check for the presence of the `clientTokenProvider`, otherwise if we allow the `invalidateClientToken` to null out the `authorization` field, and the `clientTokenProvider` is null, then the next requests will fail. That hidden implementation detail is the cause of this reported issue in the first place. From the `AuthorizationLoader` code, it looks like it takes both the initial value as well as the `clientTokenProvider` and uses both but, in reality, it's only using one. If this was rewritten in Kotlin, I'd assume these two variables would be replaced with a `sealed class`, where the hardcoded value and `clientTokenProvider` are two implementations. Let me know if you think we should still be nulling out the `authorization` value if there is a null `clientTokenProvider`

sshropshire:
Oh I understand. I wouldn't worry about that particular edge case since the crash log will be uploaded when the app is relaunched. And as a backup, Google Play has its own crash analytics available for merchants to use when diagnosing production issues.

For No. 2, that is a good point. We would only want to clear the cache for `ClientTokenProvider` integrations. That's a good piece of acceptance criteria for this use case. We can mention in the docs that the `invalidateClientToken()` method does nothing otherwise.

josephyanks:
Updated, let me know what else is required. Thanks!

sshropshire:
> Updated, let me know what else is required. Thanks!

Looks great! I have a couple updates [we can add](https://github.com/josephyanks/braintree_android/pull/1), then I think it'll be good for a formal review by the SDK team.

josephyanks:
Just merged them into my branch, thank you

sarahkoop (approved):
Looks good to me!

josephyanks:
Merged the conflicting `CHANGELOG.md`

sshropshire (approved):
:fire:

josephyanks:
Thank you all! @sshropshire do you have an idea of when the next release is schedule for? We'd love to take out the reflection nullification hack from our codebase :)

PR #548, opened 2022-05-04
Add `isExplicitCancellation` to `UserCanceledException`

sarahkoop:
### Summary of changes

 - Add `isExplicitCancellation` parameter to `UserCanceledException`

 ### Checklist

 - [x] Added a changelog entry

- @jaxdesmarais 
- @mattwylder 
- @sarahkoop

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/UserCanceledException.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/UserCanceledException.java
@@ -5,7 +5,27 @@ package com.braintreepayments.api;
  */
 public class UserCanceledException extends BraintreeException {
 
+    private boolean isExplicitCancellation;
+
     UserCanceledException(String message) {
         super(message);
+    }
+
+    UserCanceledException(String message, boolean isExplicitCancellation) {
+        super(message);
+        this.isExplicitCancellation = isExplicitCancellation;
+    }
+
+    /**
+     * @return whether or not the user explicitly canceled the payment flow.
+     *
+     * {@link true} if the user confirms cancellation of the payment flow.
+     *
+     * {@link false} if the user returns to the app without completing the payment flow and without
+     * performing an explicit cancellation action (i.e. presses the back button or closes the
+     * browser tab)
+     */
+    public boolean isExplicitCancellation() {
+        return isExplicitCancellation;
     }
 }
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/UserCanceledException.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/UserCanceledException.java
@@ -19,11 +19,12 @@ public class UserCanceledException extends BraintreeException {
     /**
      * @return whether or not the user explicitly canceled the payment flow.
      *
-     * {@link true} if the user confirms cancellation of the payment flow.
+     * {@link true} if the user manually confirms cancellation of the payment flow.
      *
-     * {@link false} if the user returns to the app without completing the payment flow and without
-     * performing an explicit cancellation action (i.e. presses the back button or closes the
-     * browser tab)
+     * {@link false} if the user returns to the app without completing the payment flow and the
+     * action performed to return to the app is unknown. For browser switching flows, this could
+     * mean the user returned to the app through multi-tasking without completing the flow, the user
+     * closed the browser tab, or the user pressed the back button.
      */
     public boolean isExplicitCancellation() {
         return isExplicitCancellation;
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/UserCanceledException.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/UserCanceledException.java
@@ -5,15 +5,15 @@ package com.braintreepayments.api;
  */
 public class UserCanceledException extends BraintreeException {
 
-    private boolean isExplicitCancellation;
+    private boolean isExplicitCancelation;
 
     UserCanceledException(String message) {
         super(message);
     }
 
-    UserCanceledException(String message, boolean isExplicitCancellation) {
+    UserCanceledException(String message, boolean isExplicitCancelation) {
         super(message);
-        this.isExplicitCancellation = isExplicitCancellation;
+        this.isExplicitCancelation = isExplicitCancelation;
     }
 
     /**
@@ -26,7 +26,7 @@ public class UserCanceledException extends BraintreeException {
      * mean the user returned to the app through multi-tasking without completing the flow, the user
      * closed the browser tab, or the user pressed the back button.
      */
-    public boolean isExplicitCancellation() {
-        return isExplicitCancellation;
+    public boolean isExplicitCancelation() {
+        return isExplicitCancelation;
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Braintree Android SDK Release Notes
 
 ## unreleased
-* Add `isExplicitCancellation` parameter to `UserCanceledException`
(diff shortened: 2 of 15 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/UserCanceledException.java
```diff
@@ -5,7 +5,28 @@
...
+     * @return whether or not the user explicitly canceled the payment flow.
+     *
+     * {@link true} if the user manually confirms cancellation of the payment flow.
+     *
+     * {@link false} if the user returns to the app without completing the payment flow and the
+     * action performed to return to the app is unknown. For browser switching flows, this could
+     * mean the user returned to the app through multi-tasking without completing the flow, the user
+     * closed the browser tab, or the user pressed the back button.
+     */
+    public boolean isExplicitCancellation() {
```
sarahkoop, line 29: For browser switching flows, we can't distinguish between a user returning to the app through multi tasking and not completing the payment flow, or actually closing the browser (x or back button). We tried to express that this boolean will be true only if we _know_ that the user explicitly cancelled. 

Open to suggestions on how to make this as clear as possible

GooglePay/src/main/java/com/braintreepayments/api/GooglePayActivityResultContract.java
```diff
@@ -33,7 +33,7 @@ public GooglePayResult parseResult(int resultCode, @Nullable Intent intent) {
                 return new GooglePayResult(PaymentData.getFromIntent(intent), null);
             }
         } else if (resultCode == RESULT_CANCELED) {
-            return new GooglePayResult(null, new UserCanceledException("User canceled Google Pay."));
+            return new GooglePayResult(null, new UserCanceledException("User canceled Google Pay.", true));
```
sarahkoop, line 36: We set this as true because there isn't a way to multi-task out of the Google Pay sheet. This activity result is set to canceled when the user manually clicks the X out of the pay sheet.

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayInternalClient.java
```diff
@@ -151,7 +151,7 @@ public void onSuccess(CustomSheetPaymentInfo customSheetPaymentInfo, String s, B
             @Override
             public void onFailure(int errorCode, Bundle bundle) {
                 if (errorCode == SpaySdk.ERROR_USER_CANCELED) {
-                    UserCanceledException userCanceledError = new UserCanceledException("User canceled Samsung Pay.");
+                    UserCanceledException userCanceledError = new UserCanceledException("User canceled Samsung Pay.", true);
```
sarahkoop, line 154: This seems like an explicit cancellation from the SDK, but we don't have a Samsung device to confirm this on - can anyone assist here?

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
```diff
@@ -522,7 +522,7 @@ public void onResult(@Nullable ThreeDSecureResult threeDSecureResult, @Nullable
                 braintreeClient.sendAnalyticsEvent("three-d-secure.verification-flow.failed");
                 break;
             case CANCEL:
-                callback.onResult(null, new UserCanceledException("User canceled 3DS."));
+                callback.onResult(null, new UserCanceledException("User canceled 3DS.", true));
```
sarahkoop, line 525: This occurs when the Cardinal SDK returns cancellation in the validate response (explicit cancellation).

BraintreeCore/src/main/java/com/braintreepayments/api/UserCanceledException.java
```diff
@@ -5,7 +5,28 @@
  */
 public class UserCanceledException extends BraintreeException {
 
+    private boolean isExplicitCancellation;
```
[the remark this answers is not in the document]
sshropshire:
```suggestion
    private boolean isExplicitCancelation;
```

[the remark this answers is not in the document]
sshropshire: I know this one requires using the refactor tool, but we should use the single `l` here for consistency.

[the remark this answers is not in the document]
sarahkoop: Good call - will update 👍

BraintreeCore/src/main/java/com/braintreepayments/api/UserCanceledException.java
sshropshire: I think that's the crux of the issue. Chrome Custom Tabs doesn't make that information available. We only know that the user has returned to the screen.

sshropshire: In my opinion we shouldn't have a canceled event for when the user returns back to the screen. It's a bit premature to say the User has "canceled" the flow. They could be multi-tasking while completing a purchase.

PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java
```diff
@@ -378,7 +378,7 @@ private JSONObject parseUrlResponseData(Uri uri, String successUrl, String appro
         String status = uri.getLastPathSegment();
 
         if (!Uri.parse(successUrl).getLastPathSegment().equals(status)) {
-            throw new UserCanceledException("User canceled PayPal.");
+            throw new UserCanceledException("User canceled PayPal.", true);
```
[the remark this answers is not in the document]
sshropshire: This specific use case may be the most important one since PayPal does notify us of an explicit cancelation by using the `cancel_url`.

SamsungPay/src/main/java/com/braintreepayments/api/SamsungPayInternalClient.java
sshropshire: Yeah I can take a look I have a Samsung.

sshropshire: This lines up! There's a "Cancel" button and the system back button – both result in the same `SpaySdk.ERROR_USER_CANCELED` error code on a Galaxy S21.

sshropshire (approved):
🌊  LGTM! Minor 🧶  on the "canceled" naming scheme.

BraintreeCore/src/main/java/com/braintreepayments/api/UserCanceledException.java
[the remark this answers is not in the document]
mattwylder: We made this change but Android Studio is marking it as a typo. Both spellings are technically correct, but according to the internet "cancellation" is far more common. This is the first instance of the word "cancellation" in our code so we have an opportunity to define the style. In my opinion the Android Studio typo underline is distracting and makes our code appear less clean. What do others think?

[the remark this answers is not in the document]
sshropshire: We decided on one `l` a while back actually. I think there's a BT style guide somewhere that mentions this spelling as preferred. We use this spelling in a number of places and it would be a breaking change if we went the two `ll` route.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+* Add `isExplicitCancellation` parameter to `UserCanceledException`
```
sarahkoop, line 4:
```suggestion
* Add `isExplicitCancelation` parameter to `UserCanceledException`
```

PR #549, opened 2022-05-04
Simplify Demo App

sshropshire:
### Summary of changes

 - This PR cleans up the Demo app and updates the UX to make it more scalable for future features down the road.

 ### Checklist


andrespelaezp (approved):
LGTM

Demo/src/main/java/com/braintreepayments/demo/DemoFeature.java
```diff
@@ -0,0 +1,27 @@
+package com.braintreepayments.demo;
+
+public enum DemoFeature {
+    CREDIT_OR_DEBIT_CARDS(0),
+    PAYPAL(1),
+    VENMO(2),
+    GOOGLE_PAY(3),
+    SAMSUNG_PAY(4),
+    VISA_CHECKOUT(5),
```
sarahkoop, line 9: Should we remove Visa from this demo since we don't support it for v4? (it's also not androidx compatible so crashing without jetifier now)

Demo/src/main/java/com/braintreepayments/demo/GooglePayFragment.java
```diff
@@ -52,38 +57,28 @@ public void onResume() {
...
-        GooglePayFragmentDirections.ActionGooglePayFragmentToDisplayNonceFragment action =
-            GooglePayFragmentDirections.actionGooglePayFragmentToDisplayNonceFragment(paymentMethodNonce);
-
-        NavHostFragment.findNavController(this).navigate(action);
-    }
-
     public void launchGooglePay(View v) {
-        FragmentActivity activity = getActivity();
-        activity.setProgressBarIndeterminateVisibility(true);
+        FragmentActivity activity = requireActivity();
```
sarahkoop, line 81: Can we remove this and just pass `getActivity()` to the `requestPayment` method?

Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
```diff
@@ -55,9 +60,7 @@ public void launchBillingAgreement(View v) {
     }
 
     private void launchPayPal(boolean isBillingAgreement) {
-        FragmentActivity activity = getActivity();
-        activity.setProgressBarIndeterminateVisibility(true);
-
+        FragmentActivity activity = requireActivity();
```
sarahkoop, line 63: Same here can we pass `requireActivity()` every time instead?

build.gradle
```diff
@@ -92,7 +92,7 @@ allprojects {
         }
         mavenLocal()
         google()
-        jcenter()
+        mavenCentral()
```
sarahkoop, line 95: I was getting a weird cert error from this change trying to access the new kotlin plugin in this PR, I think we might have proxy settings blocking `mavenCentral`...

Demo/src/main/java/com/braintreepayments/demo/DemoFeature.java
sshropshire: Removing Visa Checkout from demo feature list.

Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
sshropshire: Reduced the number of locally scoped activity references. Adding a next major version node to refactor `PayPalClient` to allow tokenization without the need for an activity reference.

build.gradle
sshropshire: This seems to have been resolved upstream.

PR #553, opened 2022-05-17
Create NXO Data Models

sshropshire:
### Summary of changes

- Add `tokenize()` method to `BTPayPalNativeCheckoutClient`
- Create `BTPayPalNativeCheckoutNonce`
- Create `BTPayPalNativeCheckoutRequest`

### Checklist

### Authors

- @mattwylder
- @sarahkoop 
- @sshropshire

### DIFF
--- a/PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutRequest.java
+++ b/PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutRequest.java
@@ -1,0 +1,17 @@
+package com.braintreepayments.api;
+
+import androidx.annotation.NonNull;
+
+public class PayPalNativeCheckoutRequest {
+
+    private final String amount;
+
+    PayPalNativeCheckoutRequest(@NonNull String amount) {
+        this.amount = amount;
+    }
+
+    @NonNull
+    public String getAmount() {
+        return amount;
+    }
+}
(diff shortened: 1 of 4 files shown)


PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutRequest.java
```diff
@@ -0,0 +1,17 @@
+package com.braintreepayments.api;
+
+import androidx.annotation.NonNull;
+
+public class PayPalNativeCheckoutRequest {
```
jaxdesmarais, line 5: Do we not need a return URL for Android? 👀

jaxdesmarais (approved):

PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutRequest.java
sshropshire: Good callout. We _do_ but it's technically in the `AndroidManifest.xml`. We may end up adding one down the line, but for now we can get it from `BraintreeClient`.

jcnoriega (approved):

PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutRequest.java
jaxdesmarais: Ah, gotcha gotcha, that makes sense!

PR #559, opened 2022-06-02
Move Release Process to GitHub Actions

sarahkoop:
### Summary of changes

 - Move release process to GitHub Actions
**Note**: This process is set up for current major versions only, and will need to be updated for next-major-version beta versions 

 ### Checklist

- @mattwylder 
- @sarahkoop

### DIFF
--- a/.github/actions/publish_module/action.yml
+++ b/.github/actions/publish_module/action.yml
@@ -28,5 +28,11 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - run: echo "published module"
+    - run: ./gradlew --stacktrace clean :${{ inputs.module }}:publishToSonatype
       shell: bash
+      env:
+        SONATYPE_USERNAME: ${{ inputs.sonatype_usr }}
+        SONATYPE_PASSWORD: ${{ inputs.sonatype_pwd }}
+        SIGNING_KEY_ID: ${{ inputs.signing_key_id }}
+        SIGNING_KEY_PASSWORD: ${{ inputs.signing_key_pwd }}
+        SIGNING_KEY_FILE: ${{ inputs.signing_key_file }}
--- a/.github/actions/publish_module/action.yml
+++ b/.github/actions/publish_module/action.yml
@@ -1,0 +1,27 @@
+name: 'Publish Module'
+description: 'Publishes module'
+inputs:
+  module:
+    description: 'Module'
+    required: true
+    default: ''
+  sonatype_usr:
+    description: 'Sonatype user'
+    required: true
+    default: ''
+  sonatype_pwd:
+    description: 'Sonatype password'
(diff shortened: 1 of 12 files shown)


.github/actions/publish_module/action.yml
```diff
@@ -0,0 +1,46 @@
...
+  steps:
+    - run: ./gradlew --stacktrace clean :${{ inputs.module }}:publishToSonatype
+      shell: bash
+      env:
+        SONATYPE_USERNAME: ${{ inputs.sonatype_usr }}
+        SONATYPE_PASSWORD: ${{ inputs.sonatype_pwd }}
+        SIGNING_KEY_ID: ${{ inputs.signing_key_id }}
+        SIGNING_KEY_PASSWORD: ${{ inputs.signing_key_pwd }}
+        SIGNING_KEY_FILE: ${{ inputs.signing_key_file }}
+    - run: ./gradlew --stacktrace closeAndReleaseRepository
```
sarahkoop, line 39: This step is untested since it would actually release the SDK

.github/workflows/release.yml
```diff
@@ -5,3 +5,557 @@ on:
...
+          ./gradlew -PversionParam=${{ github.event.inputs.version }} updateCHANGELOGVersion
+          git commit -am 'Release ${{ github.event.inputs.version }}'
+          git tag ${{ github.event.inputs.version }} -a -m 'Release ${{ github.event.inputs.version }}'
+
+          ./gradlew -PversionParam=${{ github.event.inputs.version }} incrementSNAPSHOTVersion
+          ./gradlew incrementVersionCode
+          git commit -am 'Prepare for development'
+          git push origin master ${{ github.event.inputs.version }}
+
+  create_github_release:
```
sarahkoop, line 540: This step was tested in Drop-in but not here, since it creates a GitHub release and notifies all repo watchers

build.gradle
```diff
@@ -125,3 +129,61 @@ subprojects {
...
+
+task changeREADMEVersion {
+    doLast {
+        def readmeFile = new File('README.md')
+        def readmeFileText = readmeFile.text.replaceFirst(":card:\\d+\\.\\d+\\.\\d+(-.*)?'", ":card:" + versionParam + "'")
+        readmeFile.write(readmeFileText)
+    }
+}
+
+task changeMigrationGuideVersion {
```
sarahkoop, line 149: These migration guide update tasks look for any versions of the format :#.#.# (like card:4.10.1) and change them to the new version. This is a little different than how we did it in the Rakefile previously, because Groovy regex parsing works slightly differently. Tested and this works for all current migration guide dependency statements. This would need to be updated for future migration guide versions.

jaxdesmarais (approved):
⛵

PR #568, opened 2022-07-08
Lazily Reference Kount

sshropshire:
### Summary of changes

 - There have been some issues reported with custom Kount integrations that result in duplicate symbol errors.
 - This PR references Kount only when needed, which should prevent the JVM from loading it if it isn't used by the SDK.

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire


jaxdesmarais (approved):
This looks great and is a great solution to allow merchants to use their own reference for Kount! 🚀

kgangineni (approved):

mattwylder (approved):

PR #573, opened 2022-07-28
Add Support for PayPal Risk Correlation Id

sshropshire:
### Summary of changes

- This PR adds support for `riskCorrelationId` when collecting PayPal only risk data via `DataCollector`
- Add `collectPayPalDeviceData` method to `DataCollector`
- ~Add `CollectPayPalDeviceDataCallback` interface~ 

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire


BraintreeDataCollector/src/main/java/com/braintreepayments/api/DataCollector.java
```diff
@@ -106,7 +106,7 @@ public void onResult(@Nullable String kountSessionId, @Nullable Exception error)
     /**
      * Collect device information for fraud identification purposes from PayPal only.
      *
-     * @param context Android Context
+     * @param context       Android Context
```
jaxdesmarais, line 109: nit: should we update the other inline docs below this with this same formatting?

BraintreeDataCollector/src/main/java/com/braintreepayments/api/DataCollector.java
```diff
@@ -117,4 +117,32 @@ private String getPayPalClientMetadataId(Context context, Configuration configur
...
     }
+
+    /**
+     * Collects device information for fraud identification purposes from PayPal only.
+     *
+     * @param context           Android Context
+     * @param riskCorrelationId Correlation id to associate with data collection
+     * @param callback          {@link DataCollectorCallback}
+     */
+    public void collectPayPalDeviceData(@NonNull final Context context, @NonNull final String riskCorrelationId, @NonNull final CollectPayPalDeviceDataCallback callback) {
```
jaxdesmarais, line 128: Just a comment, but we should finalize what we want to call this for next major version on iOS and Android. We call it `pairingID` but later it's called `clientMetadataID` on iOS and here we are using `riskCorrelationId`. Would be nice to align the function signatures a bit more.

BraintreeDataCollector/src/main/java/com/braintreepayments/api/DataCollector.java
```diff
@@ -117,4 +117,32 @@ private String getPayPalClientMetadataId(Context context, Configuration configur
...
         return "";
     }
+
+    /**
+     * Collects device information for fraud identification purposes from PayPal only.
+     *
+     * @param context           Android Context
+     * @param riskCorrelationId Correlation id to associate with data collection
+     * @param callback          {@link DataCollectorCallback}
+     */
```
jaxdesmarais, line 127: On iOS we have this note for this function since we recommend using `collectDeviceData`:
```
This returns a raw client metadata ID, which is not the correct format for device data when creating a transaction. Instead, it is recommended to use `collectDeviceData`.
```
Worth calling that out here too?

jaxdesmarais (approved):
⛵ couple of questions but they aren't blocking

BraintreeDataCollector/src/main/java/com/braintreepayments/api/DataCollector.java
sshropshire: When it renders on the docs site it shouldn't have an effect. This was android studio auto formatting at work. It might make sense to consider some automation around auto-formatting though in the future.

BraintreeDataCollector/src/main/java/com/braintreepayments/api/DataCollector.java
sshropshire: Definitely yeah. I'll add this in.

BraintreeDataCollector/src/main/java/com/braintreepayments/api/DataCollector.java
sshropshire: Yeah on iOS I noticed `collectDeviceData` isn't aware of the Braintree configuration. On Android, we need to fetch config asynchronously before we can return a `clientMetadataId`.

I wonder if there's an opportunity to make it env aware on iOS?

demerino: +1 that we really need to align on the name of this ID in the next major versions!

`riskCorrelationId` seems to be the clearest and approved as a public term by the risk team.

jaxdesmarais: @sshropshire - passing in APIClient to the data collector is in next major version, so we should be aligned there as well. 😄 

@demerino will add a ticket for updating the name on iOS to `riskCorrelationId` in next major version.

demerino:
Is this PR no longer needed?

sshropshire:
@demerino technically yeah since merchant's can use `:PayPalDataCollector` module directly now. I'll close we can always re-open.

Issue #569, opened 2022-07-11
Update kount-data-collector 


spinnerx1:
Currently library uses outdated [kount-data-collector](https://github.com/braintree/braintree_android/blob/master/BraintreeDataCollector/libs/kount-data-collector-3.2.jar) 

I'm kindly ask you to update it to use fresh version published in Jitpack:
https://developer.kount.com/hc/en-us/articles/4777351032084-How-to-Integrate-the-Device-Data-Collector-into-Mobile-Apps-using-Android-Gradle

In our project we already used this lib and now because of the different versions we struggle with a duplicated classes of this lib...

sshropshire:
@spinnerx1 thanks for using the Braintree SDK for Android. We're considering removing Kount as a dependency in our next major version.

On the other hand, we have recently changed how Kount is referenced in our latest [4.12.1-SNAPSHOT](https://github.com/braintree/braintree_android#adding-it-to-your-project) build. Could you try out this SNAPSHOT version to see if your issue is resolved? A release will follow shortly once we have confirmation that lazy loading Kount resolves the issue.

sshropshire:
@spinnerx1 oh got it. We are planning to remove Kount in our next major version. I don't have a timeline for when that will be, but we are prioritizing its removal.

spinnerx1:
@sshropshire Remove completelly??? But we need it for our backend, to collect device data (recieve correlation_id field)
According to the docs:
https://developer.paypal.com/braintree/docs/guides/premium-fraud-management-tools/client-side/android/v4

sshropshire:
@spinnerx1 you'll still receive a `correlation_id` via our SDK data collector. Kount is a 3rd party tool for the Braintree SDK and we are gradually phasing it out in favor of 1st party data collection.

sshropshire:
Hey @spinnerx1 in version [4.15.0](https://github.com/braintree/braintree_android/blob/master/CHANGELOG.md#4140) we have a new `:PayPalDataCollector` module. It has the same interface as `:BraintreeDataCollector`, and removes Kount as a dependency internally.

If you replace a `DataCollector` instance with a `PayPalDataCollector` instance, is this issue resolved?

spinnerx1:
@sshropshire yes, it should help! Thank you!

sshropshire:
@spinnerx1 awesome! Closing this issue for now. Feel free to re-open, or start a new ticket if an issue arises.

PR #574, opened 2022-08-01
Paypal Data Collector Module

sshropshire:
### Summary of changes

 - Create new module to allow for device data collection without Kount.
 - This should help scenarios like #569 where the SDK version of Kount conflicts with merchant versions.

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire


PayPal/build.gradle
```diff
@@ -35,7 +35,7 @@ dependencies {
     api project(':BraintreeCore')
 
     implementation deps.appCompat
-    implementation project(':BraintreeDataCollector')
+    implementation project(':PayPalDataCollector')
```
hollabaq86, line 38: is this a breaking change?

LocalPayment/build.gradle
```diff
@@ -35,7 +35,7 @@ dependencies {
     implementation deps.appCompat
 
     api project(':BraintreeCore')
-    implementation project(':BraintreeDataCollector')
+    implementation project(':PayPalDataCollector')
```
hollabaq86, line 38: same q, re:breaking change

Card/build.gradle
```diff
@@ -34,7 +34,6 @@ android {
 
 dependencies {
     api project(':BraintreeCore')
-    implementation project(':BraintreeDataCollector')
```
hollabaq86, line 37: How is it that we can safely remove this data collector without needing a replacement? If this is something I'm overlooking I blame that I don't have enough caffeine this morning 🤣

hollabaq86 (commented):
Some Qs, great work!

PayPal/build.gradle
sshropshire: Fortunately it should be good in a minor release! If it was marked as `api` then it would definitely break but these modules were including `:BraintreeDataCollector` only to use the package private `PayPalDataCollector` class which up until now was invisible to merchants.

LocalPayment/build.gradle
sshropshire: We should be good. This main reason we switch to `:PayPalDataCollector` in these other modules is to remove an implicit dependency on Kount.

Card/build.gradle
sshropshire: I really appreciate you bringing this up actually. In a strict sense for a merchant that has Kount enabled I guess it actually would result in a behavioral breaking change – only the API is unchanged. I wonder if maybe we should defer the change in that case for `Card`, `LocalPayment` and `PayPal` modules?

sshropshire: Actually i'm mistaken in this case. The `:Card` module wasn't actually performing data collection so it made sense to remove it entirely.

Since this dependency is marked as `implementation` it was never added to the merchant app's Java class path, and if a merchant is minifying their release build it should be optimized out since we aren't using it within the SDK.

PayPal/build.gradle
sshropshire: Here in [PayPalInternalClient](https://github.com/braintree/braintree_android/blob/0059a909b99ae492eb99542aaa5a523c5a740f2c/PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java#L23) and [LocalPaymenetClient](https://github.com/braintree/braintree_android/blob/0059a909b99ae492eb99542aaa5a523c5a740f2c/LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentClient.java#L25), `PayPalDataCollector` is the only class referenced from the `:BraintreeDataCollector` module.

At the moment, we're pulling in a public module to reference a hidden internal class. This is the first time we're putting PayPal specific data collection (without Kount) in GA.

jaxdesmarais (approved):
⛵

demerino (approved):
🍿 looks great!

PayPalDataCollector/src/main/java/com/braintreepayments/api/PayPalDataCollector.java
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @param context  Android Context
+     * @param callback {@link PayPalDataCollectorCallback}
+     */
+    public void collectDeviceData(@NonNull final Context context, @NonNull final PayPalDataCollectorCallback callback) {
+        collectDeviceData(context, null, callback);
+    }
+
+    /**
+     * Collects device data based on your merchant configuration.
```
demerino, line 92: Clarification here is that collection isn't based on merchant config.

Maybe `Collects device data for PayPal APIs`

PR #578, opened 2022-08-09
Refactor NXO Constructor

sshropshire:
### Summary

NXO can be constructed with less parameters since it doesn't need to app switch.

### Changes

 - Add new `PayPalNativeCheckoutClient` that requires only a `BraintreeClient` instance.
 - Add new `PayPalNativeCheckoutClient#launchNativeCheckout` method that launches native checkout without throwing.
 - Deprecate `PayPalNativeCheckoutClient` constructor that requires both `Fragment` and `BraintreeClient` instances.
 - Deprecate `PayPalNativeCheckoutClient#tokenizePayPalAccount` method that throws an exception.

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutClient.java
+++ b/PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutClient.java
@@ -74,6 +74,9 @@ public class PayPalNativeCheckoutClient {
      */
     @Deprecated
     public void tokenizePayPalAccount(@NonNull final FragmentActivity activity, @NonNull final PayPalNativeRequest payPalRequest) throws Exception {
+        // NEXT_MAJOR_VERSION: remove tokenizePayPalAccount method and refactor tests to center
+        // around launchNativeCheckout in the future. Keeping the tests as they are for now allows
+        // us to maintain test coverage across both the tokenizePayPalAccount and launchNativeCheckout methods
         boolean isCheckoutRequest = payPalRequest instanceof PayPalNativeCheckoutRequest;
         boolean isVaultRequest = payPalRequest instanceof PayPalNativeCheckoutVaultRequest;
         if (isCheckoutRequest || isVaultRequest) {
--- a/PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutClient.java
+++ b/PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutClient.java
@@ -74,10 +74,10 @@ public class PayPalNativeCheckoutClient {
      */
     @Deprecated
     public void tokenizePayPalAccount(@NonNull final FragmentActivity activity, @NonNull final PayPalNativeRequest payPalRequest) throws Exception {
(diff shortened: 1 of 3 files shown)


PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutClient.java
```diff
@@ -52,6 +62,30 @@ public void setListener(PayPalNativeCheckoutListener listener) {
...
+    public void tokenizePayPalAccount(@NonNull final FragmentActivity activity, @NonNull final PayPalNativeRequest payPalRequest) throws Exception {
+        // NEXT_MAJOR_VERSION: remove tokenizePayPalAccount method and refactor tests to center
+        // around launchNativeCheckout in the future. Keeping the tests as they are for now allows
+        // us to maintain test coverage across both the tokenizePayPalAccount and launchNativeCheckout methods
+        boolean isCheckoutRequest = payPalRequest instanceof PayPalNativeCheckoutRequest;
+        boolean isVaultRequest = payPalRequest instanceof PayPalNativeCheckoutVaultRequest;
+        if (isCheckoutRequest || isVaultRequest) {
+            launchNativeCheckout(activity, payPalRequest);
+        } else {
+            throw new Exception("Unsupported request type");
```
jaxdesmarais, line 85: Should we make this message similar to the one we are using below to make it more explicit to a merchant what they are doing wrong? So something like "Unsupported request type. Please use either a PayPalNativeCheckoutRequest or a PayPalNativeCheckoutVaultRequest."?

PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutClient.java
```diff
@@ -79,9 +115,9 @@ private void sendCheckoutRequest(final FragmentActivity activity, final PayPalNa
 
         braintreeClient.getConfiguration((configuration, error) -> {
             sendPayPalRequest(
-                activity,
-                payPalCheckoutRequest,
-                configuration
+                    activity,
```
jaxdesmarais, line 118: Is the spacing here and below for the rest of the diff on this file correct? It looks really far in now but that may just be my Android n00b eyes.

jaxdesmarais (approved):
⛵  just a couple small comments/questions

PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutClient.java
sshropshire: We could yeah. I wasn't as concerned with this one since it's deprecated, but we could add the same message.

PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutClient.java
sshropshire: The linter wilds out sometimes lol. I was able to reduce this to one line.

andrespelaezp (approved):

PR #581, opened 2022-08-10
Rename `clientMetadataId` parameter to `riskCorrelationId`

sshropshire:
### Summary of changes

 - This PR updates the `clientMetadataId` to `riskCorrelationId` on `PayPalDataCollector` to make it clearer.
 - This shouldn't be considered a breaking change since Java / Kotlin interop [doesn't generate named parameters](https://discuss.kotlinlang.org/t/what-is-the-reason-for-not-allowing-named-arguments-for-java-interop/4571) for Java methods. (also verified this in a sample project)

 ### Checklist

- @sshropshire

### DIFF
--- a/PayPalDataCollector/src/main/java/com/braintreepayments/api/PayPalDataCollector.java
+++ b/PayPalDataCollector/src/main/java/com/braintreepayments/api/PayPalDataCollector.java
@@ -45,8 +45,8 @@ public class PayPalDataCollector {
      * "future payment") from a mobile device. Pass the result to your server, to include in the
      * payment request sent to PayPal. Do not otherwise cache or store this value.
      *
-     * @param context          Android Context
-     * @param configuration    the merchant configuration
+     * @param context       Android Context
+     * @param configuration the merchant configuration
      */
     @MainThread
     String getClientMetadataId(Context context, Configuration configuration) {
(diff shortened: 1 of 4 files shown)


PayPalDataCollector/src/main/java/com/braintreepayments/api/PayPalDataCollector.java
```diff
@@ -45,8 +45,8 @@ String getPayPalInstallationGUID(Context context) {
      * "future payment") from a mobile device. Pass the result to your server, to include in the
      * payment request sent to PayPal. Do not otherwise cache or store this value.
      *
-     * @param context          Android Context
-     * @param configuration    the merchant configuration
+     * @param context       Android Context
+     * @param configuration the merchant configuration
```
jaxdesmarais, line 49:
```suggestion
     * @param configuration The merchant configuration
```

PayPalDataCollector/src/main/java/com/braintreepayments/api/PayPalDataCollector.java
```diff
@@ -109,12 +109,12 @@ public void onResult(@Nullable Configuration configuration, @Nullable Exception
                     try {
                         PayPalDataCollectorRequest request = new PayPalDataCollectorRequest()
                                 .setApplicationGuid(getPayPalInstallationGUID(context));
-                        if (clientMetadataId != null) {
-                            request.setClientMetadataId(clientMetadataId);
+                        if (riskCorrelationId != null) {
+                            request.setClientMetadataId(riskCorrelationId);
```
jaxdesmarais, line 113: Are we able to rename this method to `request.setRiskCorrelationId` as well?

jaxdesmarais (approved):
⛵

PayPalDataCollector/src/main/java/com/braintreepayments/api/PayPalDataCollector.java
sshropshire: Yeah for sure 👍

kaylagalway (approved):
:green_check:

mattwylder (commented):
🚢

sshropshire:
> the standard here is Id not ID?

@mattwylder on BT we haven't made this standard. We did on Northstar mainly for feature parity since `id` is reserved in Obj-C (and Swift? 🤷).

Definitely something to consider for Android next major version though good call. Taking note of it. We have a few things that we could do to better align with iOS.

PR #596, opened 2022-09-28
Add Analytics for PayPal Native Checkout

jaxdesmarais:
### Summary of changes

- add analytics events for PayPal Native Checkout
- add tests for analytics
- These events match the [iOS analytics PR as closely as possible](https://github.com/braintree/braintree_ios/pull/862) - missing events are due to differences in how iOS vs Android handles the implementation

 ### Checklist

- @jaxdesmarais

### DIFF
--- a/PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutClient.java
+++ b/PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutClient.java
@@ -144,8 +144,8 @@ public class PayPalNativeCheckoutClient {
     ) {
         internalPayPalClient.sendRequest(activity, payPalRequest, (payPalResponse, error) -> {
             if (payPalResponse != null) {
-                String analyticsPrefix = getAnalyticsEventPrefix(payPalRequest);
-                braintreeClient.sendAnalyticsEvent(String.format("%s.started", analyticsPrefix));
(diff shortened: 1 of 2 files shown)


PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutClient.java
```diff
@@ -139,7 +145,7 @@ private void sendPayPalRequest(
         internalPayPalClient.sendRequest(activity, payPalRequest, (payPalResponse, error) -> {
             if (payPalResponse != null) {
                 String analyticsPrefix = getAnalyticsEventPrefix(payPalRequest);
-                braintreeClient.sendAnalyticsEvent(String.format("%s.app-switch.started", analyticsPrefix));
+                braintreeClient.sendAnalyticsEvent(String.format("%s.started", analyticsPrefix));
```
scannillo, line 148: Can we remove this `getAnalyticsEventPrefix()` method and just use raw strings like the other events? Looks like we're only using it once, here.

jaxdesmarais: Yep, we can simplify this and just do the switch on the line above so we can remove the `getAnalyticsEventPrefix` method entirely.

scannillo (approved):

mattwylder (approved):

Issue #604, opened 2022-10-06
Crash SharedPrefKeysetReader


DmitriiBroli:
### General information

* SDK/Library version: 4.16.0
* Environment: Sandbox
* Android Version and Device: Any
* Compile SDK: 33
* Targer SDK: 33
* Braintree dependencies:

    com.braintreepayments.api:card:4.16.0
    com.braintreepayments.api:three-d-secure:4.16.0

### Issue description

Crashlytics crash

```
java.io.FileNotFoundException: can't read keyset; the pref value __androidx_security_crypto_encrypted_prefs_value_keyset__ does not exist
        at xl6.b(SharedPrefKeysetReader.java:33)
        at xl6.a(SharedPrefKeysetReader.java:1)
        at ee3.j(KeysetHandle.java:1)
        at gd$b.e(AndroidKeysetManager.java:7)
        at gd$b.f(AndroidKeysetManager.java:1)
        at gd$b.d(AndroidKeysetManager.java:12)
        at gt1.b(EncryptedSharedPreferences.java:97)
        at gt1.a(EncryptedSharedPreferences.java:5)
        at e30.d(BraintreeSharedPreferences.java:22)
        at e30.a(BraintreeSharedPreferences.java:1)
        at pr0.b(ConfigurationCache.java:20)
        at pr0.a(ConfigurationCache.java:5)
        at tr0.c(ConfigurationLoader.java:7)
        at tr0.d(ConfigurationLoader.java:49)
        at n20$a.a(BraintreeClient.java:17)
        at qo$a.b(AuthorizationLoader.java:19)
        at kv.resumeWith(ContinuationImpl.kt:12)
        at sk1.run(DispatchedTask.kt:119)
        at android.os.Handler.handleCallback(Handler.java:938)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:223)
        at android.app.ActivityThread.main(ActivityThread.java:7656)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
```

scannillo:
👋 Hello @DmitriiBroli - do you have steps to replicate?

DmitriiBroli:
@scannillo Hi, no steps unfortunately, this is a Crashlytics crash.

[...]

PR #612, opened 2022-10-20
Add Explicit Master Key Alias for Encrypted Shared Prefs

sshropshire:
### Summary of changes

 - This PR adds an explicit key alias for encrypted shared prefs (potential fix for #604)

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/SharedUtils/src/main/java/com/braintreepayments/api/BraintreeSharedPreferences.java
+++ b/SharedUtils/src/main/java/com/braintreepayments/api/BraintreeSharedPreferences.java
@@ -9,6 +9,7 @@ import androidx.security.crypto.MasterKey;
 class BraintreeSharedPreferences {
 
     private static volatile BraintreeSharedPreferences INSTANCE;
+    private static final String BRAINTREE_KEY_ALIAS = "com.braintreepayments.api.masterkey";
     private static final String BRAINTREE_SHARED_PREFS_FILENAME = "BraintreeApi";
 
     static BraintreeSharedPreferences getInstance() {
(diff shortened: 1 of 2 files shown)


jaxdesmarais (approved):
👻

SharedUtils/src/main/java/com/braintreepayments/api/BraintreeSharedPreferences.java
```diff
@@ -27,7 +28,7 @@ private BraintreeSharedPreferences() {}
 
     static SharedPreferences getSharedPreferences(Context context) {
         try {
-            MasterKey masterKey = new MasterKey.Builder(context)
+            MasterKey masterKey = new MasterKey.Builder(context, BRAINTREE_KEY_ALIAS)
```
scannillo, line 31: So basically this change switches from using the [default key alias](https://developer.android.com/reference/androidx/security/crypto/MasterKey#DEFAULT_MASTER_KEY_ALIAS()), to using `com.braintreepayments.api.masterkey`.

I know we use `BraintreeSharedPreferences` for config caching, but is there anything else it is used for that is merchant-facing, where changing this key value could break a merchant integration? Could they be relying on it for anything?

sshropshire: Not that I can think of. The storage is mostly ephemeral, even config is cached for 5 minutes at a time I believe.

jcnoriega (approved):

Issue #619, opened 2022-10-26
App Crashes in BraintreeSharedPreferences.java


tanmayyy:
### **General information**
Environment: Production
Android Version and Device: Android 8.1.0, Huawei P20
Braintree dependencies:

```
implementation "com.braintreepayments.api:paypal:4.12.0
```

### Issue description
The app is crashing for this one particular user when he switches to the BrainTree component to complete the payment.

```
Fatal Exception: java.lang.SecurityException: Could not decrypt value. decryption failed
       at androidx.security.crypto.EncryptedSharedPreferences.getDecryptedObject(EncryptedSharedPreferences.java:579)
       at androidx.security.crypto.EncryptedSharedPreferences.getLong(EncryptedSharedPreferences.java:436)
       at com.braintreepayments.api.BraintreeSharedPreferences.getLong(BraintreeSharedPreferences.java:95)
       at com.braintreepayments.api.ConfigurationCache.getConfiguration(ConfigurationCache.java:41)
       at com.braintreepayments.api.ConfigurationCache.getConfiguration(ConfigurationCache.java:34)
       at com.braintreepayments.api.ConfigurationLoader.getCachedConfiguration(ConfigurationLoader.java:75)
       at com.braintreepayments.api.ConfigurationLoader.loadConfiguration(ConfigurationLoader.java:39)
       at com.braintreepayments.api.BraintreeClient$1.onAuthorizationResult(BraintreeClient.java:165)
       at com.braintreepayments.api.AuthorizationLoader.loadAuthorization(AuthorizationLoader.java:20)
       at com.braintreepayments.api.BraintreeClient.getAuthorization(BraintreeClient.java:174)
       at com.braintreepayments.api.BraintreeClient.getConfiguration(BraintreeClient.java:161)
       at com.braintreepayments.api.BraintreeClient$4.onAuthorizationResult(BraintreeClient.java:222)
       at com.braintreepayments.api.AuthorizationLoader.loadAuthorization(AuthorizationLoader.java:20)
       at com.braintreepayments.api.BraintreeClient.getAuthorization(BraintreeClient.java:174)
       at com.braintreepayments.api.BraintreeClient.sendPOST(BraintreeClient.java:218)
       at com.braintreepayments.api.ApiClient.tokenizeREST(ApiClient.java:64)
       at com.braintreepayments.api.PayPalInternalClient.tokenize(PayPalInternalClient.java:114)
[...]

PR #620, opened 2022-10-26
Add Defensive Guards to Prevent Crashes Related to Jetpack Security Crypto

sshropshire:
### Summary of changes

 - Allow `BraintreeSharedPreferences` to gracefully degrade when `EncryptedSharedPreferences` fails (fix for #619)

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/ConfigurationLoader.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/ConfigurationLoader.java
@@ -38,10 +38,10 @@ class ConfigurationLoader {
 
 
         Configuration cachedConfig = null;
-        UnexpectedException loadFromCacheException = null;
+        BraintreeSharedPreferencesException loadFromCacheException = null;
         try {
             cachedConfig = getCachedConfiguration(authorization, configUrl);
-        } catch (UnexpectedException e) {
+        } catch (BraintreeSharedPreferencesException e) {
             loadFromCacheException = e;
         }
 
@@ -50,7 +50,7 @@ class ConfigurationLoader {
             callback.onResult(resultFromCache, null);
         } else {
 
-            final UnexpectedException finalLoadFromCacheException = loadFromCacheException;
+            final BraintreeSharedPreferencesException finalLoadFromCacheException = loadFromCacheException;
             httpClient.get(configUrl, null, authorization, HttpClient.RETRY_MAX_3_TIMES, new HttpResponseCallback() {
 
                 @Override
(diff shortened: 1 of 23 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/ConfigurationLoader.java
```diff
@@ -36,20 +36,39 @@ void loadConfiguration(final Context context, final Authorization authorization,
...
-        Configuration cachedConfig = getCachedConfiguration(context, authorization, configUrl);
+
+        Configuration cachedConfig = null;
+        BraintreeSharedPreferencesException loadFromCacheException = null;
+        try {
+            cachedConfig = getCachedConfiguration(authorization, configUrl);
+        } catch (BraintreeSharedPreferencesException e) {
+            loadFromCacheException = e;
+        }
+
```
jcnoriega, line 47: Instead of declaring variables and exceptions and then check which one is null, can we simply execute the succesfull flow inside the `try` and the error flow inside `catch`? Something like:
```java
        try {
            Configuration cachedConfig = getCachedConfiguration(authorization, configUrl);
            // return config to callback
        } catch (BraintreeSharedPreferencesException e) {
            BraintreeSharedPreferencesException loadFromCacheException = e;
            // execute getting configuration remotley
        }
```

SharedUtils/src/main/java/com/braintreepayments/api/BraintreeSharedPreferences.java
```diff
@@ -39,84 +50,105 @@ static SharedPreferences getSharedPreferences(Context context) {
...
-            return sharedPreferences.getString(key, fallback);
+    BraintreeSharedPreferences(SharedPreferences sharedPreferences) {
+        this.sharedPreferences = sharedPreferences;
+        this.createException = null;
+    }
+
+    BraintreeSharedPreferences(BraintreeSharedPreferencesException createException) {
+        this.sharedPreferences = null;
+        this.createException = createException;
+    }
```
jcnoriega, line 66: Why can we throw the exception directly on `getInstance`? Why do we store it and just move the error further? wouldnt be best to directly "inform" that shared preferences were not created?

jcnoriega (approved):

SharedUtils/src/main/java/com/braintreepayments/api/BraintreeSharedPreferences.java
jaxdesmarais: My guess is that there are some instances where we are required to have shared prefs like configuration and maybe the Venmo flow (my recollection could be off here). And in those instances we want to surface the error, but in other places we may not care about shared prefs as much so can just no-op when there is an error instead of surface the error. Not sure if this tracks because my android/shared prefs knowledge is rocky at best.

jaxdesmarais (approved):
👻

PR #636, opened 2022-12-19
Random IBAN Generation for the Demo App

mattwylder:
### Summary of changes

 - Add helper functions to `SEPADirectDebitFragment` to generate random IBANs since they expire after a certain number of uses.
 - Resources: [Gateay helper functions for IBAN generation](https://github.braintreeps.com/braintree/gateway/pull/24498/files) and a [article about calculating values for IBAN generation](https://community.appway.com/screen/kb/article/generating-and-validating-an-iban-1683400256881)
 - [iOS Version](https://github.com/braintree/braintree_ios/pull/910)

 ### Checklist

- @mattwylder @jaxdesmarais

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
   * Update exception documentation links to point to valid PayPal Braintree documentation URL
 * Braintree Core
   * Update pinned certificates used by `BraintreeGraphQLClient` and `BraintreeHttpClient`
+* BraintreeSEPADirectDebit
+    * Add BTSEPADirectDebitTestHelper.generateValidSandboxIBAN() method to generate valid IBANs for testing in Sandbox
 
 ## 4.21.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,6 @@
   * Update exception documentation links to point to valid PayPal Braintree documentation URL
 * Braintree Core
   * Update pinned certificates used by `BraintreeGraphQLClient` and `BraintreeHttpClient`
-* BraintreeSEPADirectDebit
-    * Add BTSEPADirectDebitTestHelper.generateValidSandboxIBAN() method to generate valid IBANs for testing in Sandbox
 
 ## 4.21.0
 
--- a/SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitTestHelper.java
+++ b/SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitTestHelper.java
@@ -1,0 +1,56 @@
+package com.braintreepayments.api;
+
+import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ThreadLocalRandom;
+
+/**
+ * Provides helper functions to test the SEPA Direct Debit flow in sandbox.
+ */
+public class SEPADirectDebitTestHelper {
+    /**
+     * Generates a valid 27-digit IBAN (International Bank Account Number) for testing SEPA Direct Debit flows.
+     * @return a valid IBAN
+     */
+    public static String generateSandboxIBAN() {
+        String countryCode = "FR";
+        long bankCode = 30006L;
+        String branchCode = "00001";
+        long accountNumber = ThreadLocalRandom.current().nextLong(10_000_000_000l, 100_000_000_000l);
+        String accountNumberWithChecksum = accountNumberWithChecksum(bankCode, Long.parseLong(branchCode), accountNumber);
+        String checksum = checksum(bankCode, branchCode, accountNumberWithChecksum);
+        String result = countryCode + checksum + String.valueOf(bankCode) + branchCode + accountNumberWithChecksum;
+        return result;
+    }
+
+    private static String accountNumberWithChecksum(long bankCode, long branchCode, long accountNumber) {
+        long sum = 89 * bankCode + 15 * branchCode + 3 * accountNumber;
+        long checksum = 97 - calculateMod97(sum);
+        return String.valueOf(accountNumber) + String.valueOf(checksum);
+    }
+
+    private static long calculateMod97(long accountNumber) {
+        long result = 0;
+        List<Long> arr = new ArrayList<>();
+        while(accountNumber > 0) {
+            Long lastDigit = accountNumber % 10;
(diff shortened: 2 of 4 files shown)


sshropshire (approved):
🌊

SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitTestHelper.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Provides helper functions to test the SEPA Direct Debit flow in sandbox.
+ */
+public class SEPADirectDebitTestHelper {
+
+    /**
+     * Generates a valid 27-digit IBAN (International Bank Account Number) for testing SEPA Direct Debit flows.
+     * @return a valid IBAN
+     */
+    public static String generateSandboxIBAN() {
```
scannillo, line 17: Nitpick - should we add semver note like in this commit on iOS? https://github.com/braintree/braintree_ios/commit/6fe0c36815e5aa7cc1ab531c0f0d57a650151664

scannillo (approved):
One small nit, otherwise LGTM!

scannillo:
Another thing to match with [iOS - we should also include this in the CHANGELOG.](https://github.com/braintree/braintree_ios/blob/5.x/CHANGELOG.md#unreleased)

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
   * Update exception documentation links to point to valid PayPal Braintree documentation URL
 * Braintree Core
   * Update pinned certificates used by `BraintreeGraphQLClient` and `BraintreeHttpClient`
+* BraintreeSEPADirectDebit
+    * Add BTSEPADirectDebitTestHelper.generateValidSandboxIBAN() method to generate valid IBANs for testing in Sandbox
```
sshropshire, line 12: 🧶 : Can we move this into the Demo project?

sshropshire: Only reason I'd say it's better in the Demo project is because it's test code.

mattwylder: Now that I think about it -- since we added that note to the documentation that the code will not be effected by semver, should we remove mention of it from the CHANGELOG, which is essentially a list of changes in semantic versions?

sshropshire: Yeah if it's just for the integration tests I'd go a step further and put it in the Demo project. I'm sure it'll be helpful for merchants but there are some other resources out their for IBAN generation.

[This sketchy site](http://randomiban.com/?country=Netherlands) is one example lol. Plus there are some node modules out there if merchants need test account numbers.

sshropshire: The demo is more of an example integration. It's one of the few modules that we don't deploy, so we keep it out of the CHANGELOG.

scannillo: > Now that I think about it -- since we added that note to the documentation that the code will not be effected by semver, should we remove mention of it from the CHANGELOG, which is essentially a list of changes in semantic versions?

Yeh, I get what you mean. Maybe the dev docs site is better for us to call out they can use this method to test instead, too. I don't feel strongly - either work. Was moreso just trying to prioritize parity b/w iOS and Android.

mattwylder: I guess both of these discussions are related. On iOS we put the IBAN generation code into the SEPA module to facilitate sandbox testing. By putting the IBAN generation into the Demo it makes it more difficult for merchants to use it. In iOS I'm not aware of any precedents for this kind of testing tool in an existing module, so it's sort of a new idea.

sshropshire: Yeah it'd be nice if we had test IBANs on the doc site similar to test card numbers. My concern with putting it in the merchant facing SDK is that we run the risk of a test helper leaking into production apps.

mattwylder: > My concern with putting it in the merchant facing SDK is that we run the risk of a test helper leaking into production apps.

Very good point. @jaxdesmarais what do you think?

jaxdesmarais: 🤔 so on iOS we _did_ already add it to the module instead of the demo (conversation here: https://github.com/braintree/braintree_ios/pull/910#issuecomment-1353462567). This version has not yet been released, so if we don't want to go with that approach we will need to align iOS to what is decided here before releasing a new version from the 5.x branch. PR here where this was merged into 5.x: https://github.com/braintree/braintree_ios/pull/913

> Yeah it'd be nice if we had test IBANs on the doc site similar to test card numbers.
This has been a series of ongoing conversations but so far has been a pretty hard no to be prioritized from the PayPal side. The PayPal backend is ultimately what would need to create these static values and allow us to pass them in sandbox.

While I agree that adding a TestHelper class to the module directly is unconventional, it would otherwise be quite difficult for merchants to test SEPA at all. IBANs have a short lifespan due to some validation PayPal does in Sandbox and they do not plan to change this. I think the _advantages_ of adding these to the SDK directly allow us to do a few things that potentially outweigh the con of it being an unconventional pattern:
1. Add integration tests - we don't have any currently
2. Add something to our docs to point merchants to this testing class, here we can add notes about _not_ using it in Production - this can potentially help with adoption of the feature as the current testing strategy is nothing + us basically saying "trust us it works"

I think the semver note also can allow us the ability to rip this out entirely if the PayPal team prioritizes static testing IBANs in Sandbox down the road. Thoughts?

sshropshire: Could we keep them in the Demo app until we have integration tests and re-evaluate afterward? That should buy us some time to collect feedback from merchants on SEPA testing.

jaxdesmarais: We would need to duplicate this logic entirely in our integration tests in order to even create the integration tests. Curious to hear @scannillo thoughts since she's the one that initially recommended moving it into the module. 

I personally don't think there is much more risk to a merchant using a method called `BTSEPADirectDebitTestHelper. generateValidSandboxIBAN()` in Production any more than shipping a card integration with `4111 1111 1111 1111` hardcoded as the card number.

scannillo: > My concern with putting it in the merchant facing SDK is that we run the risk of a test helper leaking into production apps.

I agree with Jax that since the name of the method & class + the docstring + the dev doc notes we can add that this is only for testing should be enough to prevent merchants from using it in Prod.

Our docstring can be even more explicit "Do not use this in production. Only use for sandbox testing".

PR #645, opened 2023-01-05
M1 Mac Build Script Fix

sshropshire:
### Summary of changes

 - Update build script to update Jetpack Room dependency to an M1 Mac friendly variant
 - Fallback to Java 7 friendly Room version during CI build

 ### Checklist

- @sshropshire

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ buildscript {
     // On M1 Macs, switch room version to 2.4.0 to fix java.lang.ExceptionInInitializerError
     def m1MacSafeRoomVersion = "2.4.0"
 
-    // The 'CI' build variable will equal 'true' when build script is executed via GitHub Actions
+    // The 'CI' build variable will evaluate to 'true' when build script is executed via GitHub Actions
     // Ref: https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
     def roomVersion = System.getenv('CI') ? java7SafeRoomVersion : m1MacSafeRoomVersion
 
--- a/build.gradle
+++ b/build.gradle
@@ -8,8 +8,8 @@ buildscript {
         jcenter()
     }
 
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -8,14 +8,24 @@ buildscript {
...
+    // We are restricted to room version 2.2.6 since we still offer Java 7 support. Upgrading to
+    // newer versions of room that require Java 8 will introduce a breaking change to our SDK
+    def java7SafeRoomVersion = "2.2.6"
+
+    // On M1 Macs, switch room version to 2.4.0 to fix java.lang.ExceptionInInitializerError
+    def m1MacSafeRoomVersion = "2.4.0"
+
+    // The 'CI' build variable will evaluate to 'true' when build script is executed via GitHub Actions
+    // Ref: https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables
+    def roomVersion = System.getenv('CI') ? java7SafeRoomVersion : m1MacSafeRoomVersion
```
jaxdesmarais, line 20: Would this also use the `m1MacSafeRoomVersion` if we are building on an intel machine and would that cause any issues?

sshropshire: This is a good callout. I can double check on the Intel mac I have.

sshropshire: Confirmed Room `2.2.6` and `2.4.0` both compile correctly on the Intel Mac 🙌 . The error message on the M1 Mac is `Invoke-customs are only supported starting with Android O (--min-api 26)`. It must be some kind of implicit dependency on Java 8.

The issue is similar to this [Stack Overflow post ](https://stackoverflow.com/a/50198499) we saw a while back that encouraged us to pin the Room version. We should be able to remove the workaround once we update the SDK to require Java 11 and fully align with NXO.

scannillo (approved):
LGTM - thanks for adding those helpful comments!

PR #646, opened 2023-01-06
Refactor `Configuration` Class into Kotlin

sshropshire:
### Summary of changes

 - Refactor `Configuration` class to Kotlin

 ### Checklist

- @sshropshire
- @jcnoriega


jcnoriega (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
```diff
@@ -0,0 +1,410 @@
...
+import androidx.annotation.RestrictTo
+import org.json.JSONException
+import org.json.JSONObject
+
+/**
+ * Contains the remote configuration for the Braintree Android SDK.
+ */
+// NEXT MAJOR VERSION: remove 'open' modifiers, Java classes/methods are open by default
+// Ref: https://kotlin-quick-reference.com/102c-R-open-final-classes.html
+open class Configuration internal constructor(configurationString: String?) {
```
jaxdesmarais, line 12: General question, can the `configurationString` be a required string instead of optional?

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
```diff
@@ -0,0 +1,410 @@
...
+
+    private val payPalConfiguration: PayPalConfiguration
+    private val googlePayConfiguration: GooglePayConfiguration
+
+    private val venmoConfiguration: VenmoConfiguration
+    private val kountConfiguration: KountConfiguration
+    private val unionPayConfiguration: UnionPayConfiguration
+    private val visaCheckoutConfiguration: VisaCheckoutConfiguration
+    private val graphQLConfiguration: GraphQLConfiguration
+    private val samsungPayConfiguration: SamsungPayConfiguration
```
jaxdesmarais, line 65: Is there a specific reason for the groupings here? It is not immediately clear why some of these properties have spaces and some don't.

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
```diff
@@ -0,0 +1,410 @@
...
+
+    /**
+     * @return `true` if Local Payment is enabled for the merchant account; `false` otherwise.
+     */
+    open val isLocalPaymentEnabled = isPayPalEnabled // Local Payments are enabled when PayPal is enabled
+
+    /**
+     * @return `true` if Kount is enabled for the merchant account; `false` otherwise.
+     */
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
jaxdesmarais, line 201: This is neat!

BraintreeCore/src/test/java/com/braintreepayments/api/ConfigurationUnitTest.kt
```diff
@@ -0,0 +1,479 @@
...
+import org.json.JSONException
+import org.junit.Assert.*
+import org.junit.Ignore
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import java.util.*
+
+@RunWith(RobolectricTestRunner::class)
+class ConfigurationUnitTest {
```
jaxdesmarais, line 13: Maybe a silly question but more for my own knowledge. Does Kotlin require us to convert the tests alongside the files?

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
mattwylder: What is `LIBRARY_GROUP`? How broad of a restriction is that?

mattwylder (approved):

BraintreeCore/src/test/java/com/braintreepayments/api/ConfigurationUnitTest.kt
jcnoriega: No it doesnt. Kotlin and java files can live together, you can have a mix of them in whatever form you like.
We are changing the tests becuase well, Kotlin is the better language hehe but also, because this can allows us to use [Mockk](https://mockk.io/), which is a very powerfull library for mocking kotlin classes

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
scannillo, line 12: I'm confused why we need to mark this as `open` when converting to Kotin? In [java](https://github.com/braintree/braintree_android/blob/4ffc342d2dd2386f5c1628fa2e599faa88d0573e/BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.java#L18), it's currently `public`.

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
```diff
@@ -0,0 +1,410 @@
...
+
+    /**
+     * @return `true` if Venmo is enabled for the merchant account; `false` otherwise.
+     */
+    open val isVenmoEnabled = venmoConfiguration.isAccessTokenValid
+
+    /**
+     * @return the Access Token used by the Venmo app to tokenize on behalf of the merchant.
+     */
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
scannillo, line 172: This [`RestrictTo`](https://developer.android.com/reference/androidx/annotation/RestrictTo.Scope#LIBRARY_GROUP) is pretty neat - it basically lets us have public access level, but restrict to exposing the documentation to within our SDK.

On iOS, our `nodoc` usage isn't great, since it is not respected if a merchant builds the docs on their own through Xcode. Good news though,[ Swift 5.8 will release a similar feature](https://forums.swift.org/t/nodoc-attribute-for-hiding-symbols-from-the-symbol-graph/59227) to `nodoc` in DocC.

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
scannillo: Could we alphabetize some of these too? Would make it easier to navigate all of the config types.

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
scannillo: https://developer.android.com/reference/androidx/annotation/RestrictTo.Scope

I wonder if this is just a docs annotation to hide the value in the reference docs, or if it actually prevents the value from being accessed outside the "library group". I beleive the later.

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
```diff
@@ -0,0 +1,410 @@
...
+    /**
+     * @return the GraphQL url.
+     */
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    val graphQLUrl = graphQLConfiguration.url
+
+    /**
+     * @return `true` if Samsung Pay is enabled; `false` otherwise.
+     */
+    open val isSamsungPayEnabled = samsungPayConfiguration.isEnabled
```
scannillo, line 348: On iOS, we've gotten into the practice of organizing Swift files by scope, so basically grouping methods and properties by access level. For example,[ these types of MARKs](https://github.com/braintree/braintree_ios/blob/c449e0b81c78fcae47bb02cfb8f987aed2a19996/Sources/BraintreeCore/BTAPIClient.swift#L20). We outline it in our [style guide too](https://hackmd.io/kIyOnIFIRrKY4b5wQ0dUpw?both#Code-Organization).

This grouping will probably help readability on Android too.

BraintreeCore/src/test/java/com/braintreepayments/api/ConfigurationUnitTest.kt
```diff
@@ -0,0 +1,479 @@
...
+        assertEquals(expected, sut.googlePaySupportedNetworks)
+    }
+
+    @Test
+    fun googlePayPayPalClientId_forwardsValuesFromConfiguration() {
+        val sut = Configuration.fromJson(Fixtures.CONFIGURATION_WITH_GOOGLE_PAY)
+        assertEquals("pay-pal-client-id", sut.googlePayPayPalClientId)
+    }
+
+    @Ignore("Update Samsung Pay Configuration to only check configuration instead of checking for `braintree-android-samsung-pay` classes to be present on Java classpath.")
```
scannillo, line 375: Are we set to remove Samsung pay in V5?

PayPalNativeCheckout/src/test/java/com/braintreepayments/api/PayPalNativeCheckoutClientUnitTest.java
```diff
@@ -25,17 +25,24 @@
...
 
-@RunWith(PowerMockRunner.class)
-@PrepareForTest( { PayPalCheckout.class, PLog.class })
+@RunWith(RobolectricTestRunner.class)
+@PowerMockIgnore({"org.powermock.*", "org.mockito.*", "org.robolectric.*", "android.*", "androidx.*"})
+@PrepareForTest({PayPalCheckout.class, PLog.class})
 public class PayPalNativeCheckoutClientUnitTest {
 
+    @Rule
+    public PowerMockRule powerMockRule = new PowerMockRule();
```
scannillo, line 44: Why was this added to PayPal Native tests? Doesn't look like any of the tests were updated to use this new powerMock rule?

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
sshropshire: In Kotlin everything defaults to the Java equivalent of `final`. We technically need to mark public facing types as `open` to prevent introducing a breaking change.

BraintreeCore/src/test/java/com/braintreepayments/api/ConfigurationUnitTest.kt
sshropshire: Not sure about this. We do have a merchant asking for DropIn support for it though so I wouldn't be surprised if we keep it.

PayPalNativeCheckout/src/test/java/com/braintreepayments/api/PayPalNativeCheckoutClientUnitTest.java
sshropshire: We added it so we could run Robolectric and PowerMock at the same time. There's a slight difference in the internals that caused Android code to be called and this test blew up because it wasn't mocked.

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
sshropshire: Yeah we should definitely align our style guide with y'alls at some point. We may even end up migrating feature related configuration code into extensions like in iOS since Kotlin has a similar feature.

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
sshropshire: In the future yeah, but in the `ConfigurationLoader` class it is possible that `null` gets passed in. It broke one of the tests so for now decided to keep it optional.

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
```diff
@@ -0,0 +1,410 @@
...
+        this.configurationString = configurationString
+        val json = JSONObject(configurationString)
+        assetsUrl = Json.optString(json, ASSETS_URL_KEY, "")
+        clientApiUrl = json.getString(CLIENT_API_URL_KEY)
+
+        // parse json challenges
+        json.optJSONArray(CHALLENGES_KEY)?.let { challengesArray ->
+            for (i in 0 until challengesArray.length()) {
+                challenges.add(challengesArray.optString(i, ""))
+            }
```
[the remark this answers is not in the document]
sshropshire: It turns out we need to keep this because `JSONArray` doesn't have the same interface as a standard Kotlin array. It's awesome that we have a default JSON parser that ships with Android, but its interface is a bit dated.

In the future, we may add some extensions to it like we did in Northstar to make it easier to work with.

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
jaxdesmarais: Gotcha, thanks for the explanation - do we want to add a next major version for that as well? I know for iOS it's been nice to have a list of some todo's that we want to revisit in a breaking change.

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
sshropshire: It's also restricts access to libraries with the group id `com.braintreepayments.com` so basically any of the libraries we develop have access e.g `com.braintreepayments.com:card` can talk to `com.braintreepayments.com:braintree-core`, etc.

BraintreeCore/src/test/java/com/braintreepayments/api/ConfigurationUnitTest.kt
sshropshire: Moving to Mockk improves code quality as well. Powermock and Robolectric don't work well together, and we've wanted to get rid of Mockito for a really long time.

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
scannillo: Got it. Kotlin's default is public / final.

scannillo (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
sshropshire: Def yeah. Actually I'll add a `TODO` to fix it when we migrate `ConfigurationLoader` over to Kotlin. This constructor is internal so it won't break anything.

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
sshropshire: Added a ticket for this: No. 2211

PR #650, opened 2023-01-13
Generate Documentation With Dokka

sshropshire:
### Summary of changes

 - Replace Javadoc with [Dokka](https://kotlin.github.io/dokka) to generate docs for both Java and Kotlin
 - Add `./gradlew dokkaHtmlMultiModule` to the GitHub actions release workflow
 - Output docs in the `./docs` folder so we can host docs via GitHub Pages  

 ### Checklist

- @sshropshire

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
@@ -63,7 +63,7 @@ import org.json.JSONObject
  */
 open class Configuration internal constructor(configurationString: String?) {
 
-    companion object {
+    internal companion object {
         private const val ANALYTICS_KEY = "analytics"
         private const val ASSETS_URL_KEY = "assetsUrl"
         private const val BRAINTREE_API_KEY = "braintreeApi"
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
@@ -267,6 +267,7 @@ open class Configuration internal constructor(configurationString: String?) {
      *
      * @param feature The feature to check.
      * @return `true` if GraphQL is enabled and the feature is enabled, `false` otherwise.
+     * @suppress
      */
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
     fun isGraphQLFeatureEnabled(feature: String) = graphQLConfiguration.isFeatureEnabled(feature)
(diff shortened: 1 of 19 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
```diff
@@ -10,61 +10,29 @@ import org.json.JSONObject
...
- * @property venmoEnvironment the Venmo environment used to handle this payment.
- * @property venmoMerchantId the Venmo merchant id used by the Venmo app to authorize payment.
- * @property visaCheckoutApiKey the Visa Checkout API key configured in the Braintree Control Panel.
- * @property visaCheckoutExternalClientId the Visa Checkout External Client ID configured in the Braintree Control Panel.
- * @property visaCheckoutSupportedNetworks the Visa Checkout supported networks enabled for the merchant account.
  */
 open class Configuration internal constructor(configurationString: String?) {
 
-    companion object {
+    internal companion object {
```
sshropshire, line 35: Realized when looking at the docs that this should be internal.

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
```diff
@@ -117,37 +85,197 @@ open class Configuration internal constructor(configurationString: String?) {
     // endregion
 
     // region Internal Properties
+
+    /**
+     * @return [String] url of the Braintree analytics service.
+     * @suppress
+     */
```
sshropshire, line 92: Made all restricted property annotations inline to allow them to be hidden in the docs.

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
sshropshire: Also broke the build lol. Reverted.

jcnoriega (approved):
@sshropshire do we need to add the `@suppress` annotation to the docs in our `internal` classes?

sshropshire:
> @sshropshire do we need to add the `@suppress` annotation to the docs in our `internal` classes?

@jcnoriega we should be good here. Dokka only generates docs for public facing API by default.

jaxdesmarais (approved):
This looks good! Will you remind me, do we need to do something after this is merged in to update all the old versions to use `dokka` or is that something that's going to happen auto-magically?

sshropshire:
> Will you remind me, do we need to do something after this is merged in to update all the old versions to use dokka or is that something that's going to happen auto-magically?

😅 still working on that part lol. There's a Dokka CLI but it was difficult to build out versioned docs. We may need to spend some extra time in a separate ticket to address that issue. Theoretically though we'll have a script that we point at the tag for each previous version to retroactively generate docs for older versions.

I made a ticket for this: No. 2218.

PR #654, opened 2023-01-19
Refactor `Authorization`, `InvalidAuthorization`, `TokenizationKey` and `ClientToken` to Kotlin

jcnoriega:

### Summary of changes

 - Refactor `Authorization`, `InvalidAuthorization`, `TokenizationKey` , `ClientToken` and `InvalidArgumentException` to Kotlin
 - Refactor Unit tests
 ### Checklist

 - [ ] Added a changelog entry

- @jcnoriega 
- @sshropshire


BraintreeCore/src/main/java/com/braintreepayments/api/ClientToken.kt
```diff
@@ -0,0 +1,73 @@
...
+     * @return The authorizationFingerprint for the current session
+     */
+    var authorizationFingerprint: String? = null
+
+    /**
+     * @return The customer ID in the authorizationFingerprint if it is present
+     */
+    val customerId: String?
+        get() {
+            val result = authorizationFingerprint?.let { fingerPrint ->
```
sshropshire, line 31: 🧶 : Could we put this logic in the initializer now?

BraintreeCore/src/main/java/com/braintreepayments/api/InvalidAuthorization.kt
```diff
@@ -0,0 +1,13 @@
+package com.braintreepayments.api
+
+internal class InvalidAuthorization(rawValue: String?, val errorMessage: String) :
+    Authorization(rawValue) {
+    
+    public override fun getConfigUrl(): String? {
+        return null
```
sshropshire, line 7: 🧶 : Can these be properties?

BraintreeCore/src/main/java/com/braintreepayments/api/InvalidArgumentException.kt
```diff
@@ -0,0 +1,9 @@
+package com.braintreepayments.api
+
+import java.lang.Exception
+
+/**
+ * Error thrown when arguments provided to a method are invalid.
+ * @param message the error message
+ */
+class InvalidArgumentException internal constructor(message: String?) : Exception(message)
```
sshropshire, line 9:
```suggestion
open class InvalidArgumentException internal constructor(message: String?) : Exception(message)
```

sshropshire: Yeah technically though to avoid a breaking change it's necessary. If a merchant can do it now they need to be able to subclass.

This will be true for all public facing API since we never marked anything `final` on the Java side.

sshropshire: I wish there was a better way. We'll have to next major version remove `open` in a lot of places once we get to `v5`.

jcnoriega: i havent done java in a while, but if the default access modifier is `package-private`, then merchants never had access to this class. So we shouldnt be introducing any breaking change if we left out `open`. Or am i missing something?

sshropshire: Kotlin classes are `public final` by default when used in Java. Since we marked [InvalidArgumentException.java](https://github.com/braintree/braintree_android/blob/master/BraintreeCore/src/main/java/com/braintreepayments/api/InvalidArgumentException.java) public in Java, we'll need to make it `open` on the Kotlin side because it isn't marked `final`.

Historically we've stayed away from finalizers in the Android SDK, mostly because mockito blows up unless you configure `mock-maker-inline` and all that stuff. Even then, when you add in powermock and Robolectric the unit tests become flakey and unpredictable.

sshropshire (approved):
🌊

jaxdesmarais (approved):

PR #655, opened 2023-01-19
Suppress Preferred Payment Methods and Other Classes

sshropshire:
### Summary of changes

 - Add Dokka HTML `@hide` support for Java classes
 - Suppress documentation for `PreferredPaymentMethod` classes
 - Suppress documentation for `visa-checkout` module

 ### Checklist

- @sshropshire

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/PreferredPaymentMethodsResult.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/PreferredPaymentMethodsResult.java
@@ -5,7 +5,7 @@ import org.json.JSONObject;
 
 /**
  * Contains information about which payment methods are preferred on the device.
- * This class is currently in beta and may change in future releases.
+ * This class is currently in beta and may be removed in future releases.
  * @hide
  */
 public class PreferredPaymentMethodsResult {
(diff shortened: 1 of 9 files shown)


jcnoriega (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/PreferredPaymentMethodsResult.java
```diff
@@ -4,8 +4,9 @@
 import org.json.JSONObject;
 
 /**
- *  Contains information about which payment methods are preferred on the device.
- *  This class is currently in beta and may change in future releases.
+ * Contains information about which payment methods are preferred on the device.
+ * This class is currently in beta and may change in future releases.
```
jaxdesmarais, line 8: Wonder if this should be updated to "may change or be removed" since it seems like the latter is the path that we may go down

jaxdesmarais (approved):
👻

BraintreeCore/src/main/java/com/braintreepayments/api/PreferredPaymentMethodsResult.java
sshropshire: Good call. I replaced "may change" with "may be removed."

PR #658, opened 2023-01-20
Create `BraintreeOptions` Class

sshropshire:
### Summary of changes

 - Create `BraintreeOptions` class to allow more flexibility when adding new constructor parameters
 - All changes are `package-private`

 ### Checklist

- @sshropshire

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
@@ -32,62 +32,6 @@ public class BraintreeClient {
 
     private boolean launchesBrowserSwitchAsNewTask;
 
-    private static BraintreeClientParams createDefaultParams(Context context, String authString, ClientTokenProvider clientTokenProvider) {
-        String returnUrlScheme = context
-                .getApplicationContext()
-                .getPackageName()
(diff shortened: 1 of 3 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.java
```diff
@@ -135,39 +89,61 @@ public BraintreeClient(@NonNull Context context, @NonNull String authorization,
...
+        this(
+                new BraintreeOptions(context, integrationType)
+                        .authorization(authorization)
+                        .sessionId(sessionId)
+        );
+    }
+
+    // NEXT MAJOR VERSION: Make BraintreeClient(BraintreeOptions) the primary constructor to
+    // allow additional allow parameters to grow over time without having to support multiple
+    // different constructors
```
jcnoriega, line 117: do we think this is going to be necessary once we move this to `Kotlin`? Optional parameters would fix all of this

sshropshire: Yeah you're right. This was some cleanup related to a DropIn issue, but it's internal. We can probably close this PR.

sshropshire:
Closing in favor of Kotlin refactor. Default parameters should be used instead.

PR #660, opened 2023-01-23
Add `setRequestedExemptionType` to `ThreeDSecuresRequest`

jaxdesmarais:
### Summary of changes

 - Add `setRequestedExemptionType` to `ThreeDSecuresRequest`

 ### Checklist

 - [x] Added a changelog entry

- @mattwylder @jaxdesmarais


ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
```diff
@@ -414,6 +444,7 @@ public String build(String dfReferenceId) {
             base.put("challenge_requested", challengeRequested);
             base.put("data_only_requested", dataOnlyRequested);
             base.put("exemption_requested", exemptionRequested);
+            base.put("requested_exemption_type", requestedExemptionType);
```
sshropshire, line 447: There should be a test for producing this JSON somewhere. It'll be good to get an assertion for sending this parameter.

sshropshire (approved):

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
sshropshire: [ThreeDSecureClientUnitTest](https://github.com/braintree/braintree_android/blob/master/ThreeDSecure/src/test/java/com/braintreepayments/api/ThreeDSecureClientUnitTest.java) definitely needs to be refactored, but we can add an assertion to any of the test methods that call `performLookup()`.

jaxdesmarais: Ah! Good call. Updated in this commit: https://github.com/braintree/braintree_android/pull/660/commits/a8a58b0b78592a585a2dfe1a91bc86fdf1ec9a82

jcnoriega (approved):

PR #662, opened 2023-01-25
Fixing shipping address override in nxo flow

kelay858:

### Summary of changes

 - Updating the native version and including return url
 - Fixing a bug with address override not being set

 ### Checklist

 - [x] Added a changelog entry

-


PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutVaultRequest.java
```diff
@@ -87,7 +88,7 @@ public int describeContents() {
         return 0;
     }
 
-    public static final Creator<PayPalNativeCheckoutVaultRequest> CREATOR = new Creator<PayPalNativeCheckoutVaultRequest>() {
+    public static final Creator<PayPalNativeCheckoutVaultRequest> CREATOR = new Creator<>() {
```
jaxdesmarais, line 91: Is this a breaking change? Java is far from my forte so not sure if removing `PayPalNativeCheckoutVaultRequest` from the `Creator` is considered a breaking change in java. If it is, we should call it out in the changelog.

PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeRequest.java
```diff
@@ -78,6 +79,15 @@ public void setShippingAddressEditable(boolean shippingAddressEditable) {
         this.shippingAddressEditable = shippingAddressEditable;
     }
 
+    /**
+     * Grabs the shipping address editable field
```
jaxdesmarais, line 83: I think we can use a similar docstring here to what we use in standard PayPal request: https://github.com/braintree/braintree_android/blob/b2ced4e6bc983a10a5ae7d8eebddcb8808f97c7d/PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java#L89-L96

jaxdesmarais: Ah, but that string is for the setter. Do we have a setter for this? In any case I think we can mimic the docstrings there for all of these properties (same with the ones below)

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 * BraintreeCore
   * Allow uppercase characters in default return url scheme
 
+* PayPalNativeCheckout (BETA)
+  * Bump native-checkout version to `0.8.8`
+  * Adding in a requirement for `returnUrl` inside of `PayPalNativeRequest` this url is needed to redirect after authentication
```
jaxdesmarais, line 10: General question, is this a different return URL than the one that was removed before.

PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutVaultRequest.java
kelay858: nah java will cast to the type declaration

CHANGELOG.md
kelay858: no its the same one, its a requirement on oath 2.0 to have this in so we had to add it back to native

PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeRequest.java
jaxdesmarais: Ah, no I was confused - I didn't expand things and see the setter above with the same docstring as the docs I linked. 🤦

PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutVaultRequest.java
jaxdesmarais: Nice! That's dope.

CHANGELOG.md
jaxdesmarais: Dang, that's a bummer to have to add it back in. Do you know if we should expect a similar PR for iOS? We should also add this to the docs. Just curious what the experience will be if a merchant doesn't include this? Is this auth change already live or waiting until we do a release to make updates?

kelay858: Yup we need to make a small change to the dev docs to add this in! I can work on doing that, is there a process for that? I remember there being a git somewhere. IOS will not need this change, its an android only requirement since their SVC can just close without context. There is actually no change on auth here, its all client side with the connect flow. If a merchant does not include this, they will be stuck after auth and will not be redirected to their application, which is expected as thats the current behavior on nxo

jaxdesmarais: Yeah the [docs repo is here](https://github.braintreeps.com/braintree/docs) - you'd need to open up a PR to it with this change, once it's merged it gets auto released by the docs team weekly.

And thanks for clarifying that it's only needed on Android!

jaxdesmarais, line 10:
```suggestion
  * Breaking change
    *`PayPalNativeRequest` requires a `returnUrl` to redirect correctly after authentication
```
Thinking we should maybe call this out as a breaking change

jaxdesmarais (approved):

CHANGELOG.md
sshropshire: +1 on the breaking change callout.

sshropshire (approved):

CHANGELOG.md
kelay858: thank you! Will open a PR now

PR #663, opened 2023-01-26
Refactor Analytics to Kotlin

sshropshire:
### Summary of changes

- Refactoring the following classes to Kotlin:
  - `AnalyticsBaseWorker`
  - `AnalyticsClient`
  - `AnalyticsDatabase`
  - `AnalyticsEvent`
  - `AnalyticsEventDao`
  - `AnalyticsUploadWorker`
  - `AnalyticsWriteToDbWorker`
  - `HttpResponseCallback`
- All classes are intended for internal use only

 ### Checklist

- @sshropshire


Demo/src/main/java/com/braintreepayments/demo/DemoActivity.java
```diff
@@ -48,16 +48,6 @@ protected void onCreate(Bundle savedInstanceState) {
...
-    @Override
-    protected void onResume() {
-        super.onResume();
-
-        if (BuildConfig.DEBUG && ContextCompat.checkSelfPermission(this, WRITE_EXTERNAL_STORAGE)
-                != PackageManager.PERMISSION_GRANTED) {
-            ActivityCompat.requestPermissions(this, new String[]{WRITE_EXTERNAL_STORAGE}, 1);
-        }
-    }
-
```
sshropshire, line 60: This code causes an infinite `onResume()` loop on my test device. The write permission isn't a part of the SDK we don't need storage permissions.

SharedUtils/src/main/java/com/braintreepayments/api/HttpResponseCallback.kt
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * @suppress
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+interface HttpResponseCallback {
+
+    @MainThread
+    fun onResult(responseBody: String?, httpError: Exception?)
+}
```
sshropshire, line 14: Had to convert this to Kotlin for the tests to pass because of a [bug in mockk](https://github.com/mockk/mockk/issues/865).

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEvent.kt
```diff
@@ -0,0 +1,22 @@
+package com.braintreepayments.api
+
+import androidx.room.PrimaryKey
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+
+// NEXT MAJOR VERSION: Convert to data class, we're unable to do so now because the
```
jaxdesmarais, line 7: [Data classes in Kotlin](https://kotlinlang.org/docs/data-classes.html) seem super neat. I don't believe Swift has anything comparable, so was cool to read about.

jaxdesmarais (approved):
👻

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEvent.kt
sshropshire: Yeah I'd say `struct` is close. It's a good cross-platform analogy for the most part.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
```diff
@@ -0,0 +1,220 @@
...
+            listOf(configuration, authorization, sessionId, integration).contains(null)
+        return if (isMissingInputData) {
+            ListenableWorker.Result.failure()
+        } else try {
+            val analyticsEventDao = analyticsDatabase.analyticsEventDao()
+            val events = analyticsEventDao.getAllEvents()
+            if (events.isNotEmpty()) {
+                val metadata = deviceInspector.getDeviceMetadata(context, sessionId, integration)
+                val analyticsRequest = serializeEvents(authorization, events, metadata)
+                val analyticsUrl = configuration!!.analyticsUrl
```
scannillo, line 121: `!!` could throw a NPE right?

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
```diff
@@ -0,0 +1,220 @@
...
+        sessionId: String?,
+        integration: String?,
+        timestamp: Long,
+        authorization: Authorization?
+    ) {
+        if (lastKnownAnalyticsUrl == null || authorization == null) {
+            return
+        }
+        val metadata = deviceInspector.getDeviceMetadata(context, sessionId, integration)
+        val event = AnalyticsEvent("android.crash", timestamp)
```
scannillo, line 151: Interesting - I didn't realize we were doing this on Android. We don't have an equivalent on iOS. Is this something we've ever used to troubleshoot merchant issues?

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsBaseWorker.kt
```diff
@@ -0,0 +1,23 @@
...
+ * It is not subject to semantic versioning and may change at any time.
+ * @suppress
+ */
+abstract class AnalyticsBaseWorker(
+    context: Context,
+    workerParams: WorkerParameters
+) : Worker(context, workerParams) {
+
+    internal open fun createAnalyticsClientFromInputData(): AnalyticsClient {
+        return AnalyticsClient(applicationContext)
```
scannillo, line 21: Kotlin newb question - why don't we need to do `context.applicationContext` to access `applicationContext`?

sshropshire: It's a WorkManager base class that has an [applicationContext](https://developer.android.com/reference/androidx/work/ListenableWorker#getApplicationContext()) property. I agree it is confusing here with context being highly visible in the constructor.

I see in one of the [JetPack examples](https://developer.android.com/guide/background/persistent/getting-started/define-work#input_output) they name the constructor parameter `appContext`, which is interesting. My guess is the `applicationContext` property exists for defensive programming to discourage leaking an `Activity` context, since calling `context.applicationContext` on an application context should return a reference to itself.

^ I wish I didn't have to write that last sentence but this is Android lol.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
sshropshire: Yup! Good call definitely a side effect of using the automated Kotlin conversion tool. I'll update.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
sshropshire: I have a feeling it's for tracking. I've never seen it come up in an inbound. It was weird to write when we ported `v3` to `v4` because it relies on the presence of an analytics url, which means it'll only work if it has fetched configuration before. Another guess here, it's probably for tracking purposes so we have some insight into app crashes. But honestly these may be available to us in the Google Play SDK console now. It would be cool to remove it in the next major version if possible.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsBaseWorker.kt
scannillo: Ah interesting, thank you!

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
scannillo: Yeh, I've also never seen it come up or be used. Definitely worth consider removing in V5

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsBaseWorker.kt
```diff
@@ -0,0 +1,23 @@
...
+ * This class is used internally by the SDK and should not be used directly.
+ * It is not subject to semantic versioning and may change at any time.
+ * @suppress
+ */
+abstract class AnalyticsBaseWorker(
+    context: Context,
+    workerParams: WorkerParameters
+) : Worker(context, workerParams) {
+
+    internal open fun createAnalyticsClientFromInputData(): AnalyticsClient {
```
jcnoriega, line 20: this method doesnt need the `open` access modifier. Its is an internal method used only in subclasses `AnalyticsWriteToDbWorker` and `AnalyticsUploadToDbWorker`. I think we are safe to delete `open` modifiers for classes that are for internal use of the SDK (even if they are public), as they are not part of semantic versioning

scannillo (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsBaseWorker.kt
sshropshire: I agree overall, but I would still defer removing `open` until the next major version to follow strict semantic versioning rules. Technically a user can subclass `AnalyticsBaseWorker` currently, and they should be able to subclass this method since it's marked as `protected` on the Java side and Kotlin doesn't have an equivalent to protected.

To be totally clear I don't _want_ to do this lol, but I'm following the rules of semantic versioning and the mantra of "If it can happen, it will happen." If it's possible now, then we should support the 0.000001% of SDK users that do this. Then we can tighten the restrictions in the next major version. It's counter intuitive but it leads to high quality SDKs.

jcnoriega: @sshropshire makes perfect sense!

sshropshire: @jcnoriega actually you're right on this one. I just checked a Java 7 Demo I made. `AnalyticsClient` is marked as internal so a merchant won't be able to declare a subclass.

Some of the Java / Kotlin conversion is gonna be messy for a bit. I want to clean it up at some point, but I do want to continue being thorough with semantic versioning whenever possible.

PR #664, opened 2023-01-31
Kotlin Core Exceptions

sshropshire:
### Summary of changes

- This PR refactors the following classes into Kotlin:
  - `AppSwitchNotAvailable.java`
  - `BraintreeException.java`
  - `ConfigurationException.java`
  - `UserCanceledException.java`
- Deferring `ErrorWithResponse.java` refactor for another PR since it will require additional refactoring

 ### Checklist

- @sshropshire


BraintreeCore/src/main/java/com/braintreepayments/api/UserCanceledException.kt
```diff
@@ -0,0 +1,25 @@
...
+     * @return whether or not the user explicitly canceled the payment flow.
+     *
+     * This value will be true if the user manually confirms cancellation of the payment flow.
+     *
+     * This value will be false if the user returns to the app without completing the payment flow
+     * and the action performed to return to the app is unknown. For browser switching flows, this
+     * could mean the user returned to the app through multi-tasking without completing the flow,
+     * the user closed the browser tab, or the user pressed the back button.
+     */
+    var isExplicitCancelation = false
```
jcnoriega, line 23: i think this can be a `val` with a default value. You are only setting it in the alternative constructor. With `val` we can remove `private set`

sshropshire: Definitely yeah. I'll update.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeException.kt
```diff
@@ -0,0 +1,19 @@
...
+open class BraintreeException : IOException {
+
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    constructor() : super()
+
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    constructor(message: String?) : super(message)
+
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    constructor(message: String?, cause: Throwable?) : super(message, cause)
```
jcnoriega, line 18: maybe im missing a case, but could we do something like this: 
```kotlin
open class BraintreeException @JvmOverloads constructor(message: String? = null, cause: Throwable? = null): IOException(message, cause)
```
I dont know if the `@RestricTo` can go besides `@JvmOverloads` or at the top of the class. But ultimatley, we could do:

```kotlin
open class BraintreeException : IOException {

    @JvmOverloads 
    constructor(message: String? = null, cause: Throwable? = null) : super(message, cause)

}
```

BraintreeCore/src/main/java/com/braintreepayments/api/InvalidArgumentException.kt
```diff
@@ -13,6 +13,7 @@ open class InvalidArgumentException : Exception {
      * @param message the error message
      * @suppress
      */
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
     @Suppress("ConvertSecondaryConstructorToPrimary")
-    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP) constructor(message: String?) : super(message)
+    constructor(message: String?) : super(message)
```
jcnoriega, line 18: why cant this constructor go next to the class name?

BraintreeCore/src/main/java/com/braintreepayments/api/UserCanceledException.kt
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * This value will be false if the user returns to the app without completing the payment flow
+ * and the action performed to return to the app is unknown. For browser switching flows, this
+ * could mean the user returned to the app through multi-tasking without completing the flow,
+ * the user closed the browser tab, or the user pressed the back button.
+ */
+open class UserCanceledException @JvmOverloads @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP) constructor(
+    message: String?,
+    val isExplicitCancelation: Boolean = false
+) : BraintreeException(message)
```
jcnoriega, line 20: yeah this is what i meant with the above comment on `BraintreeException`

BraintreeCore/src/main/java/com/braintreepayments/api/ConfigurationException.kt
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Error class thrown when a configuration value is invalid
+ */
+open class ConfigurationException : BraintreeException {
+
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    constructor(message: String?) : super(message)
+
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    constructor(message: String?, cause: Throwable?) : super(message, cause)
```
jcnoriega, line 14: same as above

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeException.kt
sshropshire: Nope I think that would work. This is a side effect of using the code conversion tool. `@JvmOverloads` is a good call.

sshropshire: Honestly not sure why we have a constructor with no message lol. Like wtf

BraintreeCore/src/main/java/com/braintreepayments/api/InvalidArgumentException.kt
sshropshire: Which part?

sshropshire: This is an exception that we return to merchants sometimes so technically access to the class definition should be public, but we don't want merchants creating these exceptions so we restrict access to the constructor here.

BraintreeCore/src/main/java/com/braintreepayments/api/UserCanceledException.kt
sshropshire: Yeah that makes sense. The other ones were ones were good catches. We should leverage `@JvmOverload` when we can. At the root though we may want to consider an overhaul of error handling in `v5`.

scannillo (approved):
🎸

JVMOverloads is pretty neat - _It instructs the Kotlin compiler to generate overloads for the function or constructor that substitute default parameter values._

jcnoriega (approved):

PR #667, opened 2023-01-31
Refactor ApiClient to Kotlin

sshropshire:
### Summary of changes

 - Refactor `ApiClient` and supporting files to Kotlin
 - `ApiClient` is intended for internal use only

 ### Checklist

- @sshropshire


BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
```diff
@@ -0,0 +1,61 @@
...
+                    callback.onResult(JSONObject(responseBody), null)
+                    braintreeClient.sendAnalyticsEvent("card.graphql.tokenization.success")
+                } catch (exception: JSONException) {
+                    braintreeClient.sendAnalyticsEvent("card.graphql.tokenization.failure")
+                    callback.onResult(null, exception)
+                }
+            } else {
+                braintreeClient.sendAnalyticsEvent("card.graphql.tokenization.failure")
+                callback.onResult(null, httpError)
+            }
```
jcnoriega, line 27: suggestion/question: for simple `if not null { ... } else { ... }` we could replace with `jsonBody?.also { ... } ?: { ... }`. Think its worth it? reference: https://stackoverflow.com/a/45878199

BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
```diff
@@ -0,0 +1,61 @@
...
+            braintreeClient.sendPOST(url, body) { responseBody, httpError ->
+                if (responseBody != null) {
+                    try {
+                        callback.onResult(JSONObject(responseBody), null)
+                    } catch (exception: JSONException) {
+                        callback.onResult(null, exception)
+                    }
+                } else {
+                    callback.onResult(null, httpError)
+                }
```
jcnoriega, line 46: same as above

BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
jcnoriega: i figured as well that when we start with coroutines, some major changes are going to be implemented here. Whatever you feel like!

jcnoriega (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
sshropshire: Would you be up for doing a PR into this branch with the change? This pattern will come up a lot once we start refactoring the clients. It'll be nice to lock in on a Kotlin strategy.

jaxdesmarais (approved):

PR #670, opened 2023-02-02
Refactor ApiClient Internally

jcnoriega:

### Summary of changes

 -  remove `if not null` for `also`
 - refactor to remove repeated logic.

 ### Checklist

 - [ ] Added a changelog entry

- @jcnoriega

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
@@ -13,11 +13,11 @@ class ApiClient constructor(private val braintreeClient: BraintreeClient) {
     fun tokenizeGraphQL(tokenizePayload: JSONObject, callback: TokenizeCallback) {
         braintreeClient.sendAnalyticsEvent("card.graphql.tokenization.started")
         braintreeClient.sendGraphQLPOST(tokenizePayload.toString()) { responseBody, httpError ->
-            parseResponse(
+            parseResponseToJSON(
                 responseBody,
                 httpError,
                 callback,
-                analyticsEventName = "card.graphql.tokenization"
+                isGraphQL = true
             )
         }
     }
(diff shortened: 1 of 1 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
```diff
@@ -35,21 +29,46 @@ class ApiClient constructor(private val braintreeClient: BraintreeClient) {
...
+            analyticsEventName?.also { event ->
+                braintreeClient.sendAnalyticsEvent("$event.failure")
+            }
+            returnError(callback, exception)
         }
     }
 
+    private fun returnSuccess(callback: TokenizeCallback, jsonObject: JSONObject?) {
+        callback.onResult(jsonObject, null)
+    }
```
sshropshire, line 66: Could we inline `callback.onResult(jsonObject, null` instead? I see the appeal for readability but the tradeoff of having `returnSuccess` and `returnError` feels more procedural than OOP since we still have to pass in the callback param.

sshropshire (approved):
LGTM! There may be some tweaks in the final PR worth considering. But I'm down to merge this in now and we can polish up the PR going into the main branch.

BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
```diff
@@ -35,21 +29,46 @@ class ApiClient constructor(private val braintreeClient: BraintreeClient) {
         try {
             val body = paymentMethod.buildJSON().toString()
             braintreeClient.sendPOST(url, body) { responseBody, httpError ->
-                if (responseBody != null) {
```
scannillo, line 38: Should we be sending the equivalent of the 2 graphQL analytic events above, for REST as well?

BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
```diff
@@ -35,21 +29,46 @@ class ApiClient constructor(private val braintreeClient: BraintreeClient) {
...
-            callback.onResult(null, exception)
+            returnError(callback, exception)
+        }
+    }
+
+    private fun parseResponse(
+        responseBody: String?,
+        exception: Exception?,
+        callback: TokenizeCallback,
+        analyticsEventName: String? = null) {
```
scannillo, line 43: Nitpick comment - I feel like sending analytics goes beyond the responsibility of the name of the function here `parseResponse`

scannillo (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
sshropshire: Not a bad idea to see if there are some merchants out there with RESTful card tokenization configs left.

BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
sshropshire: That's a good point. `parseResponseBodyToJSON()` would be a good alternative. And we can send the analytics events inline.

BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
sshropshire: Theoretically though the REST tokenization should be highly unlikely, unless there's some legacy merchant with REST tokenization switched on.

scannillo: That's a great question. I wonder if we need to support REST at all anymore, or if all merchants are enabled for GraphQL.

PR #674, opened 2023-02-06
Revert androidx `security-crypto` dependency to `1.1.0-alpha03`

sshropshire:
### Summary of changes

 - Revert androidx `security-crypto` dependency to `1.1.0-alpha03` (`1.1.0-alpha04` requires a compile target of 33)

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* SharedUtils
+  * Revert androidx `security-crypto` dependency to `1.1.0-alpha03` (`1.1.0-alpha04` requires a compile target of 33)
+
 ## 4.25.0
 
 * SharedUtils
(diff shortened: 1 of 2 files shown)


scannillo (approved):
Is there a build check/step we can add to our PR builds that would catch this? Is our API target on our demo/unit tests too high to have caught it?

sshropshire:
> Is there a build check/step we can add to our PR builds that would catch this? Is our API target on our demo/unit tests too high to have caught it?

Yeah we compile against API 33 so we weren't able to detect it. This crypto library has a lot of issues though. We filed bugs in the past and were told development has just restarted on it. We may not need it, I don't think the data we store is sensitive enough to require encryption.

jcnoriega (approved):

jaxdesmarais (approved):

PR #675, opened 2023-02-07
Kotlin Braintree Core Networking

sshropshire:
### Summary of changes

 - Refactor the following classes to Kotlin:
   - `BraintreeGraphQLClient.java`
   - `BraintreeHttpClient.java`
   - `BraintreeHttpResponseParser.java`
   - `TLSCertificatePinning.java`

 ### Checklist

- @sshropshire


BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
```diff
@@ -145,21 +145,23 @@ internal class AnalyticsClient @VisibleForTesting constructor(
...
         try {
             val analyticsRequest = serializeEvents(authorization, events, metadata)
-            httpClient.post(
-                lastKnownAnalyticsUrl,
-                analyticsRequest.toString(),
-                null,
-                authorization,
-                HttpNoResponse()
-            )
+            lastKnownAnalyticsUrl?.let { analyticsUrl ->
```
jcnoriega, line 156: i would move the `lastKnownAnalyticsUrl?.let` right after you check for `authorization == null`, because you are doing a bunch of operations that dont matter if `lastKnownAnalytics is null`

jcnoriega (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeGraphQLClient.kt
```diff
@@ -0,0 +1,81 @@
+package com.braintreepayments.api
+
+internal class BraintreeGraphQLClient(
+    private val httpClient: HttpClient = createDefaultHttpClient()
+) {
+
+    fun post(
+        path: String?,
```
jaxdesmarais, line 8: Since this is internal should we consider making some of the parameters (like path) required?

jaxdesmarais (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeGraphQLClient.kt
sshropshire: Yeah we will eventually, but Java's lack of support for optionals makes things complicated. We're planning on doing a second pass at removing optionals and improving the Kotlin code quality once everything has been converted.

Issue #630, opened 2022-12-07
Error message for: AndroidManifest.xml is incorrectly configured or another app...


Pingapplepie:
**Integration Details (please complete the following information):**

* SDK/Library version: 4.12.0
* Environment: Production
* Android Version and Device: Android 8.0+
* Braintree dependencies:
implementation com.braintreepayments.api:card:4.12.0
implementation com.braintreepayments.api:paypal:4.12.0
implementation com.braintreepayments.api:data-collector:4.12.0
implementation com.braintreepayments.api:three-d-secure:4.12.0

**Describe the bug**
There is an error message reported on some rare cases from users who installed our android apps,  which says:
```
"AndroidManifest.xml is incorrectly configured or another app defines the same browser switch url as this app. 
See https://developers.braintreepayments.com/guides/client-sdk/android/#browser-switch for the correct configuration". 
```
I checked the source code of braintree SDK, and it seems there are just two cases that will report this error, one is there is no valid Activity declared in the manifest to handle the deep link of Braintree and the other is there is no valid web browser to handle HTTPS scheme. Since our app works for million users and our QA has verified all workflow, it shouldn't be the manifest issue, otherwise, this will happen for all users. So I guess this is caused by web browser, then I have asked all our users to install Chrome as the default web browser, but some of them still say it doesn't work.
Do you have any suggestions on this issue? Could you please send the real error message to apps, instead of the general message? 

![image](https://user-images.githubusercontent.com/42831065/206114699-43197801-1446-441b-9049-61e2979f9abe.png)

sshropshire:
Hi @Pingapplepie thanks for using the Braintree SDK for Android. Do you specify a `customUrlScheme` when instantiating a `BraintreeClient`?

This does sound odd. Do you have more information about the specific users experiencing this issue i.e. do their devices, OS versions have similarities?

bm185178:
We are having same issue

AndroidManifest.xml is incorrectly configured or another app defines the same browser switch url as this app. See https://developers.braintreepayments.com/guides/client-sdk/android/#browser-switch for the correct configuration

```
    api 'com.braintreepayments.api:drop-in:6.5.0'
    implementation 'com.braintreepayments.api:data-collector:4.20.0'
```

Tried 
https://developer.paypal.com/braintree/docs/guides/client-sdk/setup
```
<activity android:name="com.company.MyActivity"
    android:exported="true">
    <intent-filter>
        <action android:name="android.intent.action.VIEW" ></action>
[...]

PR #677, opened 2023-02-07
Append Browser Switch Original Error Message on Failure

sshropshire:
### Summary of changes

 - In rare cases, browser switch fails in production for some users. It's difficult to reproduce this issue.
 - The browser switch library provides more information that may be helpful to diagnose an issue 
 - Appending original browser switch error message to error output will help resolve issues like #630

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 4.25.1
 
 * BraintreeCore
-  * Provide more error information on Browser Switch failures
+  * Provide more detailed information for Browser Switch errors for PayPal, PayPalNativeCheckout, and ThreeDSecure payment flows
 * SharedUtils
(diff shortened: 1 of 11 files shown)


jcnoriega (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 4.25.1
 
+* BraintreeCore
+  * Provide more error information on Browser Switch failures
```
jaxdesmarais, line 6: nit: is it worth calling out the flows this will update specifically? Something like "Browser Switch failures in PayPal, Native Checkout, and Three D Secure flows"

jaxdesmarais (approved):

CHANGELOG.md
sshropshire: Definitely yeah I'll update.

PR #396, opened 2021-05-28
Remove PaymentMethodClient

sshropshire:
### Summary of changes

 - Remove `PaymentMethodClient` and associated classes
 - `PaymentMethodClient` will be [migrated to DropIn](https://github.com/braintree/braintree-android-drop-in/pull/232)

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire 
- @sarahkoop

PR #679, opened 2023-02-08
Make `Authorization` Subclasses Library-Restricted 

sshropshire:
### Summary of changes

 - Fixes for DropIn issue [#396](https://github.com/braintree/braintree-android-drop-in/issues/396)
 - DropIn requires visibility of these classes to change behavior

 ### Checklist

- @sshropshire

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/ClientToken.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/ClientToken.kt
@@ -1,6 +1,7 @@
 package com.braintreepayments.api
 
 import android.util.Base64
+import androidx.annotation.RestrictTo
 import org.json.JSONException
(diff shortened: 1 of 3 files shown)


jcnoriega (approved):

jaxdesmarais (approved):
Does this need a changelog or is at all stuff that should be internal (wasn't sure since it seems to be fixing a bug)

sshropshire:
> Does this need a changelog or is at all stuff that should be internal (wasn't sure since it seems to be fixing a bug)

It's a compatibility issue one merchant discovered with the latest DropIn version. DropIn references a package private method that no longer exists in core.

sshropshire:
@jaxdesmarais we should be good without a CHANGELOG since it's all internal.

PR #680, opened 2023-02-08
Remove Encrypted Shared Preferences

sshropshire:
### Summary of changes

 - Remove Encrypted Shared Preferences library in favor of regular shared prefs
 - The androidx security crypto library is problematic and if we can remove it, we should

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire


jcnoriega (approved):

jaxdesmarais (approved):
Love to see all that 🔴 deleted code - bye shared prefs

PR #682, opened 2023-02-09
Throwing error if nxo response is null

kelay858:

### Summary of changes

 - Throwing error if nxo response is null

 ### Checklist

 - [x] Added a changelog entry

-


PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutClient.java
```diff
@@ -203,6 +203,8 @@ private void sendPayPalRequest(
                         braintreeClient.sendAnalyticsEvent("paypal-native.billing-agreement.succeeded");
                     }
                 });
+            } else {
+                listener.onPayPalFailure(new Exception(error));
```
sshropshire, line 207:
```suggestion
                listener.onPayPalFailure(error);
```
🧶 : could we forward the error here?

kelay858: we could but its marked nullable from the above method so I think its a bit safer to create a new error here instead of just forwarding it to the listener and then making the listeners error nullable instead of non null

kelay858: added test `requestNativeCheckout_returnsErrorFromFailedResponse`

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Braintree Android SDK Release Notes
 
 ## unreleased
+* PayPalNativeCheckout (BETA)
+  * Fixes a bug where an error was not thrown inside `PayPalNativeCheckoutClient`
```
jaxdesmarais, line 5:
```suggestion
* PayPalNativeCheckout (BETA)

  * Fixes a bug where an error was not thrown inside `PayPalNativeCheckoutClient` when no PayPal response was received from the API
```
^ I wonder if we can make this a bit more explicit as to when we weren't handling the error - as it is now ready a bit like we were just never handling errors in the client which I don't believe is the case

jaxdesmarais (approved):

PR #684, opened 2023-02-14
Kotlin refactor of DeviceInspector and DeviceMetadata

jcnoriega:

### Summary of changes

 - Refactor `DeviceInspector` and `DeviceMetadata` to Kotlin

 ### Checklist

 - [x] Added a changelog entry

-  @jcnoriega


BraintreeCore/src/main/java/com/braintreepayments/api/DeviceInspector.kt
```diff
@@ -0,0 +1,156 @@
...
+                Build.FINGERPRINT.contains("generic")
+
+    private fun getAppName(context: Context?): String =
+        context?.let {
+            getApplicationInfo(it)?.let { appInfo ->
+                context.packageManager.getApplicationLabel(appInfo).toString()
+            }
+        } ?: "ApplicationNameUnknown"
+
+    private fun getApplicationInfo(context: Context) =
```
sshropshire, line 88:
```suggestion
    private fun getApplicationInfo(context: Context?) =
```
🧶 : We could clean up other `context?.let {}` invocations with this style. Take or leave. We'd end up with some duplication but it may make the code more readable.

sshropshire (approved):
🌊

BraintreeCore/src/main/java/com/braintreepayments/api/DeviceInspector.kt
```diff
@@ -0,0 +1,154 @@
...
+            get() = Intent().setComponent(
+                ComponentName(
+                    VENMO_APP_PACKAGE,
+                    "$VENMO_APP_PACKAGE.$VENMO_APP_SWITCH_ACTIVITY"
+                )
+            )
+
+        internal fun getDropInVersion(): String? {
+            try {
+                val clazz = Class.forName("com.braintreepayments.api.dropin.BuildConfig")
```
jaxdesmarais, line 143:
```suggestion
                val class = Class.forName("com.braintreepayments.api.dropin.BuildConfig")
```
^ assuming this was a typo?

sshropshire: It's legit `class` is a reserved word in Java and Kotlin.

jaxdesmarais: :actual-lol: maybe we can do `val dropInClass` or something instead then? or is it any form? I guess `class` is also protected for properties in swift by itself unless you escape it.

jaxdesmarais (approved):

PR #685, opened 2023-02-15
Kotlin refactor of StreamHelper, GraphQLQueryHelper, UUIDHelper, MetadataBuilder, IntegrationType and GraphQLConstants

jcnoriega:

### Summary of changes

 - Refactor said classes to kotlin with unit tests as well

 ### Checklist

 - [ ] Added a changelog entry

- @jcnoriega


BraintreeCore/src/main/java/com/braintreepayments/api/GraphQLConstants.kt
```diff
@@ -0,0 +1,48 @@
+package com.braintreepayments.api
+
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+object GraphQLConstants {
```
sshropshire, line 6: This top-level `@RestrictTo` annotation should be the only one we need. We can delete the others.

BraintreeCore/src/main/java/com/braintreepayments/api/IntegrationType.kt
```diff
@@ -0,0 +1,14 @@
+package com.braintreepayments.api
+
+import androidx.annotation.StringDef
+import java.lang.annotation.Retention
+import java.lang.annotation.RetentionPolicy
+
+internal object IntegrationType {
```
sshropshire, line 7: This needs to be public with `@RestrictTo` as well to be accessible from DropIn.

sshropshire (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/GraphQLConstants.kt
sshropshire:
```suggestion
/**
 * @suppress
**/
@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
object GraphQLConstants {
```
We should suppress `@RestrictTo` annotated classes in the docs too.

PR #687, opened 2023-02-16
Kotlin refactor PostalAddress and PostalAddressParser

jcnoriega:

### Summary of changes

 -  Refactor PostalAddress and PostalAddressParser to kotlin. Unit tests included

 ### Checklist

 - [ ] Added a changelog entry

- @jcnoriega


BraintreeCore/src/main/java/com/braintreepayments/api/PostalAddressParser.kt
```diff
@@ -0,0 +1,89 @@
...
+            sortingCode = Json.optString(json, USER_ADDRESS_SORTING_CODE_KEY, "")
+        }
+
+    private fun formatExtendedUserAddress(address: JSONObject): String =
+        """
+            ${Json.optString(address, USER_ADDRESS_ADDRESS_2_KEY, "")}
+            ${Json.optString(address, USER_ADDRESS_ADDRESS_3_KEY, "")}
+            ${Json.optString(address, USER_ADDRESS_ADDRESS_4_KEY, "")}
+            ${Json.optString(address, USER_ADDRESS_ADDRESS_5_KEY, "")}
+        """.trimIndent().trim { it <= ' ' }
```
scannillo, line 88: Does something like this work in Kotlin? The `.trimIndent().trim { it <= ' ' }` is a little tricky to read

```suggestion
        return
            "${Json.optString(address, USER_ADDRESS_ADDRESS_2_KEY, "")} + "\n" +
            "${Json.optString(address, USER_ADDRESS_ADDRESS_3_KEY, "")} + "\n" +
            "${Json.optString(address, USER_ADDRESS_ADDRESS_4_KEY, "")} + "\n" +
            "${Json.optString(address, USER_ADDRESS_ADDRESS_5_KEY, "")}"
```

scannillo (approved):
Some small comments - overall LGTM!

BraintreeCore/src/main/java/com/braintreepayments/api/PostalAddressParser.kt
```diff
@@ -0,0 +1,89 @@
...
+            if (streetAddress == null && Json.optString(
+                    accountAddress,
+                    USER_ADDRESS_NAME_KEY,
+                    null
+                ) != null
+            ) {
+                return@let fromUserAddressJson(accountAddress)
+            }
+
+            PostalAddress().apply {
```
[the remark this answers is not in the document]
jcnoriega: yes. `apply()` is commonly used for object configuration. Without `apply` you would have to first create an empty `PostalAddress`, assign to a variable, then set all the variables and later return it. Here we use less code, because `apply` returns the object.
[Here are the official docs](https://kotlinlang.org/docs/scope-functions.html#apply), for more clarification

BraintreeCore/src/main/java/com/braintreepayments/api/PostalAddress.kt
```diff
@@ -0,0 +1,78 @@
...
+        countryCodeAlpha2 = parcel.readString()
+        recipientName = parcel.readString()
+        phoneNumber = parcel.readString()
+        sortingCode = parcel.readString()
+    }
+
+    companion object {
+        @JvmField
+        val CREATOR: Parcelable.Creator<PostalAddress?> =
+            object : Parcelable.Creator<PostalAddress?> {
```
sshropshire, line 68:
```suggestion
        val CREATOR: Parcelable.Creator<PostalAddress> =
            object : Parcelable.Creator<PostalAddress> {
```

BraintreeCore/src/main/java/com/braintreepayments/api/PostalAddressParser.kt
sshropshire: Yeah is `<=` a Kotlin operator?

sshropshire: Oh wow definitely less than or equal too 🤦 . This looks like it's trimming characters that have a numeric value of at lease the space character. We probably don't need the additional trimming.

BraintreeCore/src/main/java/com/braintreepayments/api/PostalAddress.kt
```diff
@@ -0,0 +1,78 @@
...
+import android.os.Parcelable
+import android.text.TextUtils
+import android.os.Parcel
+
+// NEXT_MAYOR_VERSION: transform to data class and check for @Parcelize annotation
+/**
+ * Java object representing a postal address
+ */
+open class PostalAddress : Parcelable {
+    var recipientName: String? = null
```
sshropshire, line 12: We may need to mark these properties as open for now to avoid a breaking change.
```suggestion
    open var recipientName: String? = null
```

BraintreeCore/src/main/java/com/braintreepayments/api/PostalAddress.kt
```diff
@@ -0,0 +1,78 @@
...
+    var countryCodeAlpha2: String? = null
+
+    constructor()
+
+    /**
+     * A [PostalAddress] is considered empty if it does not have a country code.
+     *
+     * @return `true` if the country code is `null` or empty, `false` otherwise.
+     */
+    val isEmpty: Boolean
```
sshropshire, line 29:
```suggestion
    open val isEmpty: Boolean
```
Same here.

sshropshire (approved):
🌊 LGTM. I'd say it's good to merge once we add those `open` qualifiers.

PR #688, opened 2023-02-16
Kotlin refactor of PreferredPaymentMethods client, callback and result

jcnoriega:

### Summary of changes

 -  Refactor said classes and unit tests

 ### Checklist

 - [ ] Added a changelog entry

- @jcnoriega


BraintreeCore/src/main/java/com/braintreepayments/api/PreferredPaymentMethodsCallback.kt
```diff
@@ -0,0 +1,13 @@
+package com.braintreepayments.api
+
+/**
+ * Callback for receiving result of [PreferredPaymentMethodsClient.fetchPreferredPaymentMethods].
+ * This interface is currently in beta and may be removed in future releases.
```
scannillo, line 5: We removed this in [V6 of BT iOS](https://github.com/braintree/braintree_ios/pull/932). It was never merchant-facing or documented, so I wonder if we have to wait until BT Android V5 to remove it?

jcnoriega: In that PR it says that [BTPreferredPaymentMethods.swift](https://github.com/braintree/braintree_ios/pull/932/files#diff-4ed982cde9458d83f20034f0b8c65d3fa8f4f96a1470d25db35214629a17dbc5) was public. Perhaps it was hidden from the merchant in a diferent way? is that possible in iOS?
Still; yes, we ll have to wait for a next major version to completely delete this class, as it is a breaking change

BraintreeCore/src/main/java/com/braintreepayments/api/PreferredPaymentMethodsResult.kt
```diff
@@ -0,0 +1,73 @@
...
+                print(ignored.message)
+            }
+            return PreferredPaymentMethodsResult()
+                .isPayPalPreferred(payPalPreferred)
+                .isVenmoPreferred(venmoInstalled)
+        }
+
+        @Throws(JSONException::class)
+        private fun getObjectAtKeyPath(obj: JSONObject, keyPath: String): JSONObject {
+            val keys = keyPath.split(".").toTypedArray()
```
sshropshire, line 65:
```suggestion
            val keys = keyPath.split(".")
```

BraintreeCore/src/test/java/com/braintreepayments/api/PreferredPaymentMethodsClientResultUnitTest.kt
```diff
@@ -0,0 +1,60 @@
...
+            //language=json
+            """{
+                  "data": {
+                    "preferredPaymentMethods": {
+                      "paypalPreferred": false
+                    }
+                  }
+            }"""
+            //
+        val result = fromJSON(json, false)
```
sshropshire, line 38:
```suggestion
        val result = PreferredPaymentMethodsResult.fromJSON(json, false)
```

sshropshire: It'll be better for readability since we have a ton of `fromJSON` methods in this repo.

BraintreeCore/src/main/java/com/braintreepayments/api/PreferredPaymentMethodsResult.kt
sshropshire: Kotlin has a much better `String.split()` method than Java. We no longer need to convert to a typed Array.

sshropshire (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/PreferredPaymentMethodsCallback.kt
scannillo: Yeh, it was just hidden from the merchant via our reference documentation. It was technically still public & accessible to merchants - just not documented.

scannillo (approved):

PR #689, opened 2023-02-16
BraintreeClient Kotlin Refactor

sshropshire:
### Summary of changes

 - Refactor the following classes to Kotlin:
   - `BraintreeClient.java`
   - `BraintreeClientParams.java`
 - Add `BraintreeOptions.kt` internal class

 ### Checklist

- @sshropshire


BraintreeCore/src/main/java/com/braintreepayments/api/ConfigurationLoaderCallback.kt
```diff
@@ -1,5 +1,5 @@
 package com.braintreepayments.api
 
-internal interface ConfigurationLoaderCallback {
+internal fun interface ConfigurationLoaderCallback {
```
jcnoriega, line 3: :fire: nice catch on `fun` for interfaces

jcnoriega (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/ConfigurationLoaderCallback.kt
sshropshire: Yeah I recently noticed it's a thing. We should be able to go 100% Kotlin in `v4` with functional interfaces.

kgangineni (approved):

PR #693, opened 2023-02-22
Paypal Client Add Listener `null` Check

sshropshire:
### Summary of changes

 - Fix issue that causes a null pointer exception when `PayPalClient` attempts to notify success or failure when the listener is `null`

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* PayPal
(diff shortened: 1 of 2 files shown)


demerino (approved):
This probably can't be tested, can it?

jcnoriega (approved):

sshropshire:
> This probably can't be tested, can it?

It would be difficult since the `tokenize` tests rely on assertions like `verify(listener).onResult(error)` and the instance needs to be non `null`.

Once we refactor `PayPalClient` to Kotlin we'll have access to optionals and it'll add some compiler protection.

PR #697, opened 2023-03-06
Migrate `ErrorWithResponse`, `BraintreeError`, and `BraintreeGraphQLResponseParser` to Kotlin

sshropshire:
### Summary of changes

 - Migrate the following classes to Kotlin:
   - BraintreeError.java
   - ErrorWithResponse.java
   - BraintreeGraphQLResponseParser.java

 ### Checklist

- @sshropshire


BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeError.kt
```diff
@@ -0,0 +1,179 @@
...
+        return 0
+    }
+
+    override fun writeToParcel(dest: Parcel, flags: Int) {
+        dest.writeString(field)
+        dest.writeString(message)
+        dest.writeTypedList(fieldErrors)
+    }
+
+    protected constructor(`in`: Parcel) {
```
scannillo, line 70: Nitpick - can we organize the constructors to be next to each other? Also, moving the constructors to be below the variable definitions at the top of the file.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeError.kt
```diff
@@ -0,0 +1,179 @@
...
+    /**
+     * Method to extract an error for an individual field, e.g. creditCard, customer, etc.
+     *
+     * @param field Name of the field desired, expected to be in camelCase.
+     * @return [BraintreeError] for the field searched, or `null` if not found.
+     */
+    open fun errorFor(field: String): BraintreeError? {
+        var returnError: BraintreeError?
+        if (fieldErrors != null) {
+            for (error in fieldErrors!!) {
```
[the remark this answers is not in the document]
sshropshire: This is a good catch I missed this one! There are a few ways, we had to do this in `ErrorWithResponse` too I think, i'll make sure they use the same strategy. I do like the Swifty `guard` style suggestion, returning `null` early. I may try to apply that in this PR.

This class has some technical debt associated with it as well. Hoping we'll be able to clean it up and make it more readable in the future.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeError.kt
```diff
@@ -0,0 +1,179 @@
+package com.braintreepayments.api
+
+import android.os.Parcel
+import android.os.Parcelable
+import com.braintreepayments.api.GraphQLConstants.ErrorTypes
+import org.json.JSONArray
+import org.json.JSONException
+import org.json.JSONObject
+
+/**
```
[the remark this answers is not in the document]
sshropshire: Technically yes, we are trying to move to this condensed style though. [GooglePayConfiguration.kt](https://github.com/braintree/braintree_android/blob/master/BraintreeCore/src/main/java/com/braintreepayments/api/GooglePayConfiguration.kt) is a good example of where we're trying to be.

Ultimately, it'd be nice to extract the parsing logic out and have pure `data class` value objects. Then we won't need multiple constructors.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeGraphQLResponseParser.kt
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * @param responseCode the response code returned when the http request was made.
+     * @param connection the connection through which the http request was made.
+     * @return the body of the http response.
+     */
+    @Throws(Exception::class)
+    override fun parse(responseCode: Int, connection: HttpURLConnection): String {
+        val response = baseParser.parse(responseCode, connection)
+        val errors = JSONObject(response).optJSONArray(GraphQLConstants.Keys.ERRORS)
+        if (errors != null) {
```
scannillo, line 25: Similar comment re: if we want to do something "guard-like" here in and return early

BraintreeCore/src/main/java/com/braintreepayments/api/ErrorWithResponse.kt
```diff
@@ -0,0 +1,157 @@
...
+ * @property fieldErrors All the field errors.
+ */
+open class ErrorWithResponse : Exception, Parcelable {
+
+    open var statusCode: Int = 0
+        internal set
+
+    private var _message: String? = null
+    override val message: String?
+        get() = _message
```
scannillo, line 29: Do we really need this backing property for `message`?

scannillo:
Curious why we're also converting the tests to Kotlin in these PRs? Is there value to trying to leave the tests in Java to make sure the Kotlin source code refactors don't break anything Java-related?

sshropshire:
> Curious why we're also converting the tests to Kotlin in these PRs? Is there value to trying to leave the tests in Java to make sure the Kotlin source code refactors don't break anything Java-related?

This is a good point. David brought this up as well. We may reevaluate and do source only migration when we can to speed up the process.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeError.kt
sshropshire: On second thought these make the most sense in order because the parcel order in `writeToParcel` and the `constructor(in: Parcel)` order need to match exactly.

In Kotlin we can use [Parcelize](https://developer.android.com/kotlin/parcelize) in the future to remove some of the boilerplate required for `Parcelable`. We may be able to do it in the current major version, but as a precaution it's probably better to delay that change until the next major version when we make everything a `data` object.

BraintreeCore/src/main/java/com/braintreepayments/api/ErrorWithResponse.kt
sshropshire: Technically yes, there's some weirdness with the Java interop here. The `message` property is an override from the `Exception` class and the `super` constructor is never called in the current implementation. I expect we'll refactor this class sometime after the Kotlin migration.

scannillo (approved):
Would be nice to really align our error details as much as possible on iOS & Android at some point.

jcnoriega (approved):

sshropshire:
> Would be nice to really align our error details as much as possible on iOS & Android at some point.

Yeah definitely. Jax and I refactored the iOS one out of necessity because it broke after Swift conversion. There's some sort of recursion that's unnecessary and the Android implementation is the same as the old iOS implementation.

PR #703, opened 2023-03-09
Make Browser Switch Methods on BraintreeClient Nullable to Fix Crash

sshropshire:
### Summary of changes

 - A few methods in `BraintreeClient` need to be nullable to prevent a crash caused by Java/Kotlin interop.
 - Integration tests are failing because Kotlin asserts non-null return values by default

 ### Checklist

- @sshropshire
- @KunJeongPark

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
@@ -300,7 +300,7 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      * @suppress
      */
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-    fun getBrowserSwitchResult(activity: FragmentActivity): BrowserSwitchResult =
+    fun getBrowserSwitchResult(activity: FragmentActivity): BrowserSwitchResult? =
         browserSwitchClient.getResult(activity)
 
     /**
@@ -311,7 +311,7 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      * @param activity
      * @return [BrowserSwitchResult]
      */
-    fun deliverBrowserSwitchResult(activity: FragmentActivity): BrowserSwitchResult {
+    fun deliverBrowserSwitchResult(activity: FragmentActivity): BrowserSwitchResult? {
         return browserSwitchClient.deliverResult(activity)
     }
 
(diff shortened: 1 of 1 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -300,7 +300,7 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      * @suppress
      */
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-    fun getBrowserSwitchResult(activity: FragmentActivity): BrowserSwitchResult =
+    fun getBrowserSwitchResult(activity: FragmentActivity): BrowserSwitchResult? =
```
jaxdesmarais, line 303: Is this something we can have a Unit Test for at all? Or that we could have caught a different way? I know all of the Kotlin conversion changes are going directly into the `master` branch so more wondering if we need to rethink our strategy to catch these types of errors sooner.

jaxdesmarais (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
sshropshire: There isn't a great way. The integration tests are our last resort. They were failing without the optional declaration we're adding in this PR.

`BrowserSwitchClient` is also missing annotations. After Kotlin interop, it has this interface:

```kotlin
fun getBrowserSwitchResult(activity: FragmentActivity): BrowserSwitchResult!!
```

In Swift, this would be the equivalent:

```swift
func getBrowserSwitchResult(activity: FragmentActivity) -> BrowserSwitchResult!
```

It's a contrived example, but it illustrates how Kotlin doesn't enforce `null` checks in these scenarios. The root cause is those methods should've been annotated as `@Nullable` from the start.

jaxdesmarais: Gotcha - that makes sense. Yeah I know with nullability in the Swift conversions we've had to deal with similar things. Is there a way to check the java interface in Android Studio? In Xcode we can view the generated Obj-C compatibility header which especially helped with early on catching things we missed in the conversions.

An somewhat related question, but I wonder if this is a compelling reason to try to move our integration tests into CI to run on every PR.

scannillo (approved):
+1 to considering running integration tests on CI & also keeping tests and demo in Java

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
sshropshire: > A somewhat related question, but I wonder if this is a compelling reason to try to move our integration tests into CI to run on every PR.

Yeah I'd say so it would be helpful in the long wrong. A Java sample project may be our best bet at catching interop errors like Sammy mentioned. It might make sense to have a Swift and a Java demo app in the long run.

sshropshire:
> +1 to considering running integration tests on CI & also keeping tests and demo in Java

Created an SDK Ticket No. 2516 to move CI tests into GitHub actions.

PR #704, opened 2023-03-09
Fix memory issue with DataCollector class

KunJeongPark:
### Summary of changes

 - fix memory leak in "DataCollector" 

- @KunJeongPark 
- @sshropshire

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
@@ -300,7 +300,7 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      * @suppress
      */
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-    fun getBrowserSwitchResult(activity: FragmentActivity): BrowserSwitchResult =
+    fun getBrowserSwitchResult(activity: FragmentActivity): BrowserSwitchResult? =
         browserSwitchClient.getResult(activity)
 
     /**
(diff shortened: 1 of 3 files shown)


jaxdesmarais (commented):
Is this the same as #703? Also do we want to have the full URL links to our internal Jira on public GitHub?

KunJeongPark:
thanks Jax for the catch. I deleted the reference to the live issue. This is not the same, I just wanted to add the memory leak fix that we worked on together on DataCollector. I didn't mean to include previous changes.

jaxdesmarais:
> thanks Jax for the catch. I deleted the reference to the live issue. This is not the same, I just wanted to add the memory leak fix that we worked on together on DataCollector. I didn't mean to include previous changes.

Gotcha - yeah a little confusing to see the `BraintreeClient.kt` changes in both PRs as it's not clear immediately what is being changed (I see the diff in `DataCollector.java` now). If you merge `fix_browser_switch_crash` into `dataCollectorMemoryLeak` it should remove the diff from this PR to make it clearer!

scannillo:
Can you explain a little about what issue this PR fixes exactly?

sshropshire:
> Can you explain a little about what issue this PR fixes exactly?

There's potential for a leak of an `Activity` reference within `DataCollector`. This PR grabs an `applicationContext` reference immediately to retain in the data collector callback. Originally, the callback was retaining the context passed in by the merchant, which is likely to be an `Activity`. Leaking activities in Android is pretty much the ultimate deal breaker because they tend to take up a lot of memory.

jaxdesmarais (approved):

PR #706, opened 2023-03-21
Target Java 8

sshropshire:
### Summary of changes

 - Update SDK to target Java 8 for source compatibility with Kotlin files

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,7 +59,7 @@
   * Allow uppercase characters in default return url scheme
 * ThreeDSecure
   * Add `setRequestedExemptionType` to `ThreeDSecureRequest`
-* *Please note:* This version is dependent on the Java 8 programming language. Please read [Use Java 8 language features](https://developer.android.com/studio/write/java8-support) in the Android developer guide to learn how to use it in your project.
+* *Please note:* This version is dependent on the Java 8 programming language. Please read [Use Java 8 language features](https://developer.android.com/studio/write/java8-support) in the Android developer guide to learn how to use it in your project. If this causes an issue with your integration, please contact our [support](https://developer.paypal.com/braintree/help) team for further assistance.
   
 ## 4.23.1
 


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
   * Use `applicationContext` in `DataCollector#collectDeviceData()` callback to prevent potential `Activity` leaks
 * GooglePay
   * Fix issue that causes `GooglePayNonce#isNetworkTokenized` to always return `false` after being parceled
+* **Potential Breaking Change**
```
scannillo, line 9: Shouldn't this be in a next major version? The Kotlin changes haven't gone out yet to V4, right?

sshropshire: Yeah I agree that is should be next major version. There are Kotlin code changes out in a few of the latest releases. I'm doing some more investigating though it may be caused by a dependency on work-manager I'm hoping we can roll something back without having to worry about bumping the Java version.

sshropshire: @scannillo This PR could fix it without us having to break: https://github.com/braintree/braintree_android/pull/708

jaxdesmarais (approved):
👻

CHANGELOG.md
scannillo: I feel like the most _SemVer_ route would be revert the existing Kotlin changes to v4, as a "bug fix" to the breaking changes. And then save Kotlin rewrites for v5, dropping Java 7.

Is the route you're thinking to go? Or were you leaning toward continuing to do Kotlin rewrites in V4?

sshropshire: There may be a way. I went down the path of cherry picking the Java commits, but we do have a series of merge commits in the mainline that make reverting to Java 7 more complex.

The default compile target is now Java 8 in Android Studio so we should be good. In the worst case we will cut a separate pure Java 7 branch if we have a merchant that is unable to upgrade.

scannillo: Gotcha - I wonder if we can track Kotlin vs Java and version # in our analytics metadata in the future.

CHANGELOG.md
```diff
@@ -59,6 +59,7 @@
   * Allow uppercase characters in default return url scheme
 * ThreeDSecure
   * Add `setRequestedExemptionType` to `ThreeDSecureRequest`
+* *Please note:* This version is dependent on the Java 8 programming language. Please read [Use Java 8 language features](https://developer.android.com/studio/write/java8-support) in the Android developer guide to learn how to use it in your project. If this causes an issue with your integration, please contact our [support](https://developer.paypal.com/braintree/help) team for further assistance.
```
scannillo, line 62: Question - Are there some stats we can point to - such as EOL for Java 7, something like [this](https://www.oracle.com/java/technologies/java-se-support-roadmap.html) which shows extended support dropping in July 2022?

Basically - is there any Java timelines or Android documentation we can point to that reveals the age of Java 7, making it more easily justifiable for us dropping support in a minor version.

scannillo (approved):
Left one small question in the CHANGELOG

CHANGELOG.md
sshropshire: I wish yeah the best way for us to get this information would be to survey our merchants AFAIK. Since this happens at compile time I don't think we can get access to the JDK version used to build a merchant app. By the time it gets to a consumer device it's all Java bytecode.

CHANGELOG.md
sshropshire: Yeah the best thing I cold find was this [Android Gradle Plugin (AGP) Release Note from March 2021](https://developer.android.com/studio/past-releases/past-agp-releases/agp-4-2-0-release-notes#java-8-default). They made Java 8 the default. After this point we had to explicitly declare Java 7 support, but i'm sure that accelerated Java 8 adoption overall.

PR #709, opened 2023-03-22
Remove `fun` Interfaces

sshropshire:
### Summary of changes

 - `fun interface` types in Kotlin rely on Java 8
 - By removing these functional interfaces, we can keep Java 7 support
 - All of the `fun interface` types in this PR are marked `library-private` or are `internal`
 - To prevent a breaking change, we should avoid refactoring Java callback types to adhere to [SAM](https://kotlinlang.org/docs/fun-interfaces.html)

 ### Checklist

- @sshropshire


jaxdesmarais (approved):
Nice find! Are we able to add any sort of Java 7 smoke test to run on CI to catch this sooner in the future? I could see this being something we easily miss or forget about in a conversion.

sshropshire:
> Nice find! Are we able to add any sort of Java 7 smoke test to run on CI to catch this sooner in the future? I could see this being something we easily miss or forget about in a conversion.

Yeah I think it deserves discussion. We may want to prioritize some safe guards and increase automation before additional Kotlin refactoring.

Issue #642, opened 2022-12-28
TransactionTooLargeException in com.braintreepayments.api.ThreeDSecureActivity


IgorButirsky:
**Integration Details (please complete the following information):**

* SDK/Library version: 4.18.0 (also tried 4.22.0)
* Environment: Production
* Android Version and Device: Pixel 7 Pro with Android 13, emulator of Pixel 3a with Android 12
* Braintree dependencies:
    - com.braintreepayments.api:card:4.18.0
    - com.braintreepayments.api:data-collector:4.18.0
    - com.braintreepayments.api:three-d-secure:4.18.0

**Describe the bug**
The app crashes during the 3D Secure payment with `TransactionTooLargeException`.

***Stacktrace:***
```
Transaction too large, intent: Intent { cmp=com.main.gopuff/com.braintreepayments.api.ThreeDSecureActivity (has extras) }, extras size: 683584, icicle size: 0
Second failure launching com.main.gopuff/com.braintreepayments.api.ThreeDSecureActivity, giving up
                                           android.os.TransactionTooLargeException: data parcel size 686688 bytes.   
    at android.os.BinderProxy.transactNative(Native Method)
    at android.os.BinderProxy.transact(BinderProxy.java:571)
    at android.app.IApplicationThread$Stub$Proxy.scheduleTransaction(IApplicationThread.java:2732)
    at android.app.servertransaction.ClientTransaction.schedule(ClientTransaction.java:136)
    at com.android.server.wm.ClientLifecycleManager.scheduleTransaction(ClientLifecycleManager.java:47)
    at com.android.server.wm.ActivityTaskSupervisor.realStartActivityLocked(ActivityTaskSupervisor.java:868)
    at com.android.server.wm.ActivityTaskSupervisor.startSpecificActivity(ActivityTaskSupervisor.java:986)
    at com.android.server.wm.Task.resumeTopActivityInnerLocked(Task.java:6561)
    at com.android.server.wm.Task.resumeTopActivityUncheckedLocked(Task.java:6132)
    at com.android.server.wm.Task.resumeTopActivityUncheckedLocked(Task.java:6177)
    at com.android.server.wm.RootWindowContainer.resumeFocusedTasksTopActivities(RootWindowContainer.java:2364)
    at com.android.server.wm.RootWindowContainer.resumeFocusedTasksTopActivities(RootWindowContainer.java:2311)
[...]

PR #711, opened 2023-03-27
Fix for 3DS TransactionTooLargeException

KunJeongPark:
### Summary of changes
  
 - fix for #642 
 - In `ThreeDSecureClient`, before launching `ThreeDSecureActivity`, we add check for `TransactionTooLargeException`

 ### Checklist

 - [x] Added a changelog entry

### Authors
@sshropshire @KunJeongPark

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
   * Use `applicationContext` in `DataCollector#collectDeviceData()` callback to prevent potential `Activity` leaks
 * GooglePay
   * Fix issue that causes `GooglePayNonce#isNetworkTokenized` to always return `false` after being parceled
+* ThreeDSecure
+  * Catch `TransactionTooLargeException' to prevent crash on large data (fixes #642)
 
 ## 4.26.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * GooglePay
   * Fix issue that causes `GooglePayNonce#isNetworkTokenized` to always return `false` after being parceled
 * ThreeDSecure
-  * Catch `TransactionTooLargeException' to prevent crash on large data (fixes #642)
+  * Catch `TransactionTooLargeException` to prevent crash on large data (fixes #642)
 
 ## 4.26.1
 
--- a/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
+++ b/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
@@ -421,7 +421,6 @@ public class ThreeDSecureClient {
         braintreeClient.sendAnalyticsEvent("three-d-secure.verification-flow.started");
 
         try {
-
             if (observer != null) {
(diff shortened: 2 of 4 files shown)


ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
```diff
@@ -418,16 +419,29 @@ private void startVerificationFlow(FragmentActivity activity, Configuration conf
...
+                Intent intent = new Intent(activity, ThreeDSecureActivity.class);
+                intent.putExtras(extras);
 
-            activity.startActivityForResult(intent, THREE_D_SECURE);
+                activity.startActivityForResult(intent, THREE_D_SECURE);
+            }
+        } catch (RuntimeException runtimeException) {
+            Throwable exceptionCause = runtimeException.getCause();
+            if (exceptionCause instanceof TransactionTooLargeException) {
+                String errorMessage = "The 3D Secure response returned is too large to continue.";
```
jaxdesmarais, line 438: Thinking out loud here, but I wonder if there are details that can be included in this message that would allow us to pinpoint problematic banks and escalate them to Cardinal. Should we also include the support email here and direct merchants to reach out to us if they encounter this error?

sshropshire: Interesting yeah do we have a similar support message format on iOS we could borrow from?

sshropshire: > I wonder if there are details that can be included in this message that would allow us to pinpoint problematic banks and escalate them to Cardinal

Merchants may be in the best position to report the root cause to support. We could technically dig into the `ThreeDSecureResult` for more info but I wouldn't know where to start.

jaxdesmarais: Oh that was a question... we add "Please contact support." to some of our error message that signals an error worth reaching out to us about.

sshropshire: Ok cool yeah I wanted to make sure I didn't create a new type of support message format, that's legit. There's one in this class I can borrow from I'll add.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
   * Use `applicationContext` in `DataCollector#collectDeviceData()` callback to prevent potential `Activity` leaks
 * GooglePay
   * Fix issue that causes `GooglePayNonce#isNetworkTokenized` to always return `false` after being parceled
+* ThreeDSecure
+  * Catch `TransactionTooLargeException' to prevent crash on large data (fixes #642)
```
scannillo, line 10:
```suggestion
  * Catch `TransactionTooLargeException` to prevent crash on large data (fixes #642)
```

jaxdesmarais (approved):

jcnoriega (approved):

PR #712, opened 2023-03-29
Pre 4.26.2 Kotlin Clean Up

sshropshire:
### Summary of changes

 - Updating some access modifiers to match Java semantics
 - Adding `@suppress` annotations to prevent docs generation for some library protected classes

 ### Checklist

- @sshropshire

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
@@ -335,7 +335,7 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      * @param activity
      * @return [BrowserSwitchResult]
      */
-    fun deliverBrowserSwitchResult(activity: FragmentActivity): BrowserSwitchResult? {
+    open fun deliverBrowserSwitchResult(activity: FragmentActivity): BrowserSwitchResult? {
         return browserSwitchClient.deliverResult(activity)
     }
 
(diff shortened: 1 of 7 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -335,7 +335,7 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      * @param activity
      * @return [BrowserSwitchResult]
      */
-    fun deliverBrowserSwitchResult(activity: FragmentActivity): BrowserSwitchResult? {
+    open fun deliverBrowserSwitchResult(activity: FragmentActivity): BrowserSwitchResult? {
```
KunJeongPark, line 338: Steven, what's the reason for these changes?

sshropshire: Java methods are `open` by default. With the exception of Java 8, we're still trying to adhere to semantic versioning when we can. We'll be able to remove all of the `open` modifiers once in our next major version.

KunJeongPark (approved):

jaxdesmarais (approved):
🚀

sshropshire:
Calling out the source of these changes are from:

1. A local smoke test project I have set up that I'm hoping to create a repo for (just don't know where it should live)
2. A static analysis tool within Android Studio that highlights errors. I'm using this with a local build to make sure everything works in DropIn

Everything should be good now. There were a few gotchas that are now resolved.

PR #713, opened 2023-03-31
Perform Browser Check for 3DS v1 only

sshropshire:
### Summary of changes

 - 3DS v2 doesn't use Browser Switching so we don't need to check if they have a browser available.

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire


ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java
```diff
@@ -145,41 +145,42 @@ public void onResult(@Nullable final Configuration configuration, @Nullable Exce
...
-
-                if (configuration.getCardinalAuthenticationJwt() == null && ThreeDSecureRequest.VERSION_2.equals(request.getVersionRequested())) {
-                    callback.onResult(null, new BraintreeException("Merchant is not configured for 3DS 2.0. " +
-                            "Please contact Braintree Support for assistance."));
-                    return;
-                }
-                braintreeClient.sendAnalyticsEvent("three-d-secure.initialized");
-
                 if (ThreeDSecureRequest.VERSION_1.equals(request.getVersionRequested())) {
+                    try {
```
scannillo, line 149: Should we even bother maintaining this logic? I know we have this ticket (DTBTSDK-2495) to just error out if the merchant opts into 3DS v1.

sshropshire: Yeah that's a good call. I went ahead and updated this PR to throw when 3DS v1 is used.

jaxdesmarais (approved):
Goodbye 3DSv1 👏 👋

PR #717, opened 2023-04-11
Catch Cardinal Null Pointer Exception and General Cleanup

sshropshire:
### Summary of changes

 - Catch Null Pointer Exception that can be thrown by `Cardinal.cca_continue()`
 - Remove `android:usesCleartextTraffic` from `AndroidManifest.xml` files
 - Make `AuthorizationCallback` and `ConfigurationLoaderCallback` functional interfaces
 - Remove unnecessary imports and address linting errors

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/ThreeDSecure/src/main/java/com/braintreepayments/api/CardinalClient.java
+++ b/ThreeDSecure/src/main/java/com/braintreepayments/api/CardinalClient.java
@@ -44,7 +44,7 @@ class CardinalClient {
         try {
             Cardinal.getInstance().cca_continue(transactionId, paReq, activity, validateReceiver);
         } catch (NullPointerException e) {
-            throw new BraintreeException("Cardinal SDK Error");
+            throw new BraintreeException("Cardinal SDK Error.", e);
         }
     }
 
--- a/ThreeDSecure/src/main/java/com/braintreepayments/api/CardinalClient.java
+++ b/ThreeDSecure/src/main/java/com/braintreepayments/api/CardinalClient.java
@@ -18,7 +18,7 @@ class CardinalClient {
     CardinalClient () {}
 
     void initialize(Context context, Configuration configuration, final ThreeDSecureRequest request, final CardinalInitializeCallback callback) {
-        configurationCardinal(context, configuration, request);
+        configureCardinal(context, configuration, request);
         Cardinal.getInstance().init(configuration.getCardinalAuthenticationJwt(), new CardinalInitService() {
             @Override
             public void onSetupCompleted(String sessionId) {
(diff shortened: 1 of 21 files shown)


ThreeDSecure/src/main/java/com/braintreepayments/api/CardinalClient.java
```diff
@@ -37,14 +37,18 @@ public void onValidated(ValidateResponse validateResponse, String serverJWT) {
...
-    void continueLookup(FragmentActivity activity, ThreeDSecureResult threeDSecureResult, CardinalValidateReceiver validateReceiver) {
+    void continueLookup(FragmentActivity activity, ThreeDSecureResult threeDSecureResult, CardinalValidateReceiver validateReceiver) throws BraintreeException {
         ThreeDSecureLookup lookup = threeDSecureResult.getLookup();
         String transactionId = lookup.getTransactionId();
         String paReq = lookup.getPareq();
-        Cardinal.getInstance().cca_continue(transactionId, paReq, activity, validateReceiver);
+        try {
+            Cardinal.getInstance().cca_continue(transactionId, paReq, activity, validateReceiver);
+        } catch (NullPointerException e) {
+            throw new BraintreeException("Cardinal SDK Error.", e);
```
demerino, line 47: Do we want to send an analytics event on this specific error? Or do we have observability into this already?

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureActivity.java
```diff
@@ -40,15 +40,23 @@ void onCreateInternal(CardinalClient cardinalClient) {
 
         ThreeDSecureResult threeDSecureResult = extras.getParcelable(EXTRA_THREE_D_SECURE_RESULT);
         if (threeDSecureResult != null) {
-            cardinalClient.continueLookup(this, threeDSecureResult, this);
+            try {
+                cardinalClient.continueLookup(this, threeDSecureResult, this);
+            } catch (BraintreeException e) {
```
KunJeongPark, line 45: Which errors are thrown by Cardinal.getInstance().cca_continue function? Can it throw RuntimeException?

ThreeDSecure/src/main/java/com/braintreepayments/api/CardinalClient.java
KunJeongPark, line 47: Are other RuntimeExceptions possible here?

sshropshire: We actually don't since the `CardinalClient` doesn't depend on `BraintreeClient`. We basically wrap Cardinal's API to encapsulate the dependency. We could potentially forward the client token to `ThreeDSecureActivity` to send a error events.

sshropshire: Yes it is definitely possible. Runtime exceptions in Java [are unchecked](https://www.baeldung.com/java-checked-unchecked-exceptions#unchecked) making it hard to detect Cardinal errors like this one without the GitHub issue we recieved.

sshropshire: My gut says to limit the scope of the exception to a NPE, but I did Google a bit to survey the internet.

- This [answer](https://stackoverflow.com/a/18266490) on Stack Overflow breaks down some Java held best practices that say they shouldn't be caught at all, although in this case the NPE is thrown from within the Cardinal SDK, so we do need to catch it to prevent a crash
- This [medium post](https://matam-kirankumar.medium.com/fortify-issues-poor-error-handling-overly-broad-catch-c1409cde84d6) breaks down how having an overly broad catch block can obscure the intent of the error handling code, but it may provide extra protection against Cardinal breaking the SDK
- Lastly, the docs for [RuntimeException](https://docs.oracle.com/javase/7/docs/api/java/lang/RuntimeException.html) show all known subclasses that could be thrown in that case. If we catch `RuntimeException` this catch block will be relevant for 49 different types of runtime errors

sshropshire: @KunJeongPark all this to say I'm indifferent lol. I could see pros and cons to either, although the best solution is for Cardinal to provide more stability to their SDK. What do you think?

demerino (approved):

KunJeongPark (approved):

PR #731, opened 2023-05-03
Remove Powermock

sshropshire:
### Summary of changes

 - Powermock has an issue running on JDK 17 because of an outdated Mockito dependency
 - This PR refactors unit tests that rely on Powermock to Kotlin and `mockk`

 ### Checklist

- @sshropshire


build.gradle
```diff
@@ -66,20 +64,9 @@ buildscript {
...
+            "mockitoCore"                : "org.mockito:mockito-core:5.3.1",
             "jsonAssert"                 : "org.skyscreamer:jsonassert:1.5.1",
             "mockk"                      : "io.mockk:mockk:1.13.4",
-
-            "powermockJunit"             : "org.powermock:powermock-module-junit4:${versions.powermock}",
-            "powermockRule"              : "org.powermock:powermock-module-junit4-rule:${versions.powermock}",
-            "powermockMockito"           : "org.powermock:powermock-api-mockito2:${versions.powermock}",
-            "powermockClassloading"      : "org.powermock:powermock-classloading-xstream:${versions.powermock}",
-
-            // Legacy versions of the powermock libraries are required until 3DS and Visa Checkout tests can be updated to resolve robolectric/powermock test issues
```
jaxdesmarais, line 78: 🥳

build.gradle
```diff
@@ -66,20 +64,9 @@ buildscript {
             "junitTest"                  : "androidx.test.ext:junit:1.1.3",
             "robolectric"                : "org.robolectric:robolectric:4.7.3",
             "dexmakerMockito"            : "com.google.dexmaker:dexmaker-mockito:1.2",
-            "mockitoCore"                : "org.mockito:mockito-core:3.6.0",
+            "mockitoCore"                : "org.mockito:mockito-core:5.3.1",
```
jaxdesmarais, line 67: Do we need to call this version bump out in our changelog or does this only matter for our internal tests?

VisaCheckout/src/test/java/com/braintreepayments/api/VisaCheckoutClientUnitTest.kt
```diff
@@ -0,0 +1,172 @@
...
+    @Throws(Exception::class)
+    fun setup() {
+        configurationWithVisaCheckout = fromJson(Fixtures.CONFIGURATION_WITH_VISA_CHECKOUT)
+        visaPaymentSummary = mockk(relaxed = true)
+
+        every { visaPaymentSummary.callId } returns "stubbedCallId"
+        every { visaPaymentSummary.encKey } returns "stubbedEncKey"
+        every { visaPaymentSummary.encPaymentData } returns "stubbedEncPaymentData"
+
+//        every { VisaPaymentSummary::class.java}.withAnyArguments() returns visaPaymentSummary)
```
jaxdesmarais, line 38: Do we need this?

jaxdesmarais (approved):

build.gradle
sshropshire: For mockito it'll be fine. Since we mark it as a test dependency it won't ship when we release.

VisaCheckout/src/test/java/com/braintreepayments/api/VisaCheckoutClientUnitTest.kt
sshropshire: This doesn't appear to be needed. The original test [stubs the constructor with powermock](https://github.com/braintree/braintree_android/blob/e139d80dbe193eb568fb08dbd365d049ccc201e4/VisaCheckout/src/test/java/com/braintreepayments/api/VisaCheckoutClientUnitTest.java#L57), but I checked and we don't invoke the constructor internally, so the stub has no effect. I'm going to remove it.

PR #741, opened 2023-06-05
Remove Kount

sshropshire:
### Summary of changes

 - Removing Kount to improve Google Play Store compliance

 ### Checklist

 - [x] Added a changelog entry

- @sshropshire


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* BraintreeDataCollector
+  * Remove Kount dependency
```
jaxdesmarais, line 6:
```suggestion
  * Remove Kount dependency
  * Kount is no longer supported via Braintree, instead please use our [Fraud Protection Advanced product](https://developer.paypal.com/braintree/articles/guides/fraud-tools/premium/fraud-protection-advanced)
```

BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
```diff
@@ -354,9 +351,9 @@ open class Configuration internal constructor(configurationString: String?) {
         isBraintreeApiEnabled = braintreeApiConfiguration.isEnabled
         isFraudDataCollectionEnabled = cardConfiguration.isFraudDataCollectionEnabled
         isGraphQLEnabled = graphQLConfiguration.isEnabled
-        isKountEnabled = kountConfiguration.isEnabled
+        isKountEnabled = false
```
jaxdesmarais, line 354: Do we need `isKountEnabled` or `kountMerchantId` at all anymore?

jaxdesmarais: Cool - maybe we can add a `NEXT_MAJOR_VERSION` note here so we don't forget to remove it in v5

jaxdesmarais (approved):
👻

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* BraintreeDataCollector
```
jaxdesmarais, line 5: Do we know what kind of warning a merchant would get from the Goog? I wonder if it's worth calling it out specifically for future us if merchants come in asking about it or if we will just remember it's a Kount thing. 🤷

sshropshire: I'm not sure there's supposed to be a way to check by uploading the library in the Play console but I don't think it exists yet. That might be a good follow up question for them.

scannillo (approved):

PR #754, opened 2023-07-05
MXO-1160 - Nxo fix fallback from web and update nxo to GA

kelay858:

### Summary of changes

 - Updating nxo to GA
 - Adding a fix from the web fallback flow that pulls the user data from the approval data object from native checkout

 ### Checklist

 - [x] Added a changelog entry

-

### DIFF
--- a/PayPalNativeCheckout/build.gradle
+++ b/PayPalNativeCheckout/build.gradle
@@ -37,7 +37,7 @@ dependencies {
     implementation deps.appCompat
     implementation project(':PayPalDataCollector')
 
-    implementation('com.paypal.checkout:android-sdk:0.8.8') {
+    implementation('com.paypal.checkout:android-sdk:0.112.2-SNAPSHOT') {
         exclude group: 'com.paypal.android.sdk', module: 'data-collector'
     }
 
--- a/PayPalNativeCheckout/build.gradle
+++ b/PayPalNativeCheckout/build.gradle
@@ -37,7 +37,7 @@ dependencies {
     implementation deps.appCompat
     implementation project(':PayPalDataCollector')
 
-    implementation('com.paypal.checkout:android-sdk:0.112.2-SNAPSHOT') {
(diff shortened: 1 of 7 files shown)


PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutAccountNonce.java
```diff
@@ -198,6 +198,13 @@ public String getPayerId() {
         return payerId;
     }
 
+    public void setPayerInfo(String payerId, String firstName, String lastName, String email) {
```
sshropshire, line 201: We'll need docs for this since it's public.

kelay858: good callout, I added
```
    /**
     * Sets the payer info if on the nonce object. This is used when the payer information is
     * not returned from the json
     * 
     * @param payerId - The payer id associated with the user
     * @param firstName - The payer's first name
     * @param lastName - The payer's last name
     * @param email - The payer's email
     */
     ```

File: PayPalNativeCheckout/build.gradle [hunk 2]
```diff
@@ -37,7 +37,7 @@ dependencies {
     implementation deps.appCompat
     implementation project(':PayPalDataCollector')
 
-    implementation('com.paypal.checkout:android-sdk:0.8.8') {
+    implementation('com.paypal.checkout:android-sdk:1.0.0') {
```
sshropshire, line 40: q: is this version GA yet? I'm unable to find it on [mvnrepository.com](https://mvnrepository.com/artifact/com.paypal.checkout/android-sdk).

kelay858: nah it was just pushed late last night haha it should be live now though

sshropshire (approved):

PR #756, opened 2023-07-11
Add `getCardNetwork` to `GooglePayCardNonce`

jaxdesmarais:
### Summary of changes

 - Add `getCardNetwork` to `GooglePayCardNonce`
 - Update test to add `getCardNetwork`
 - Verified card network is returned as expected in demo app
     - **Note:** we do not have a way to explicity test Maestro cards since Google does not provide them in their test card suite and we do not not have access to a valid Maestro card to add to the Google Wallet

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/GooglePay/src/main/java/com/braintreepayments/api/GooglePayCardNonce.java
+++ b/GooglePay/src/main/java/com/braintreepayments/api/GooglePayCardNonce.java
@@ -228,7 +228,6 @@ public class GooglePayCardNonce extends PaymentMethodNonce {
         return cardNetwork;
     }
 
-
     @Override
     public void writeToParcel(Parcel dest, int flags) {
         super.writeToParcel(dest, flags);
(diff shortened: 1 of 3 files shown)


GooglePay/src/main/java/com/braintreepayments/api/GooglePayCardNonce.java
```diff
@@ -214,6 +220,15 @@ public BinData getBinData() {
...
     }
 
+    /**
+     * @return The card network. This card network value should not be displayed.
+     */
+    @NonNull
+    public String getCardNetwork() {
+        return cardNetwork;
+    }
+
```
scannillo, line 230:
```suggestion
```

sshropshire (approved):

GooglePay/src/main/java/com/braintreepayments/api/GooglePayCardNonce.java
```diff
@@ -214,6 +220,15 @@ public BinData getBinData() {
         return binData;
     }
 
+    /**
+     * @return The card network. This card network value should not be displayed.
```
scannillo, line 224: Take it or leave it comment - 
```suggestion
     * @return The card network. This card network value should not be displayed to the buyer.
```

scannillo (approved):

PR #765, opened 2023-08-07
NXO fix url response data sent inside native checkout client

kelay858:
Thank you for your contribution to Braintree. 

> Before submitting this PR, note that we cannot accept language translation PRs. We support the same languages that are supported by PayPal, and have a dedicated localization team to provide the translations. If there is an error in a specific translation, you may open an issue and we will escalate it to the localization team.

### Summary of changes

 - Changing the payment method response json to put the correct url to send back in
 - Setting the correct`UrlResponseData` on the account inside the native client

 ### Checklist

 - [x] Added a changelog entry

-

PR #767, opened 2023-08-10
Changing merchant id for native checkout nonce requests

kelay858:

### Summary of changes

 - The demo app had to make some changes to the merchant id for native checkout so that the requests would post using that specific merchant id
 - This should be merged after https://github.com/braintree/braintree_android/pull/765

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/CreateTransactionFragment.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/CreateTransactionFragment.java
@@ -1,6 +1,6 @@
 package com.braintreepayments.demo;
 
-import static com.braintreepayments.demo.PayPalNativeCheckoutRequestFactory.PAYPAL_NATIVE_CHECKOUT_CLIENT_ID;
+import static com.braintreepayments.demo.PayPalNativeCheckoutRequestFactory.PAYPAL_NATIVE_CHECKOUT_MERCHANT_ID;
 
 import android.os.Bundle;
 import android.text.TextUtils;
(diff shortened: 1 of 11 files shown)


Demo/src/main/java/com/braintreepayments/demo/CreateTransactionFragment.java
```diff
@@ -97,6 +100,9 @@ public void failure(RetrofitError error) {
             transactionRequest = new TransactionRequest(amount, nonceString, unionPayMerchantAccountId);
         } else {
             String merchantAccountId = Settings.getMerchantAccountId(activity);
+            if (merchantAccountId == null && nonce instanceof PayPalNativeCheckoutAccountNonce) {
+                merchantAccountId = PAYPAL_NATIVE_CHECKOUT_CLIENT_ID;
```
kelay858, line 104: since we create the ba token with this client id we need to execute the transaction with it as well

Demo/src/main/java/com/braintreepayments/demo/PayPalNativeCheckoutRequestFactory.java
```diff
@@ -8,6 +8,7 @@
 import com.braintreepayments.api.PostalAddress;
 
 public class PayPalNativeCheckoutRequestFactory {
+    public static final String PAYPAL_NATIVE_CHECKOUT_CLIENT_ID = "NativeXOTest";
```
ghbhatt, line 11: should we name this `PAYPAL_NATIVE_CHECKOUT_MERCHANT_ID`?

kelay858: yeah i will switch to merchant

PR #770, opened 2023-08-18
[3.x] Fix LPM Configuration Bug

jaxdesmarais:
### Summary of changes

 - Seems like something changed with client tokens/the gateway and `configuration.getPayPal().isEnabled()` now returns for merchants enabled for PayPal/Local Payment Methods. We note that `isEnabled` is deprecated and switching to use `isPayPalEnabled` returns the correct result for LPM enabled merchants.
 - The tokenization key we were using didn't support LPMs - pulled the one being used from iOS in the LPM ViewController to test things are now working as expected.

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/LocalPaymentsActivity.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/LocalPaymentsActivity.java
@@ -40,7 +40,7 @@ public class LocalPaymentsActivity extends BaseActivity implements PaymentMethod
         }
 
         try {
-            mBraintreeFragment = BraintreeFragment.newInstance(this, Settings.getLocalPaymentsTokenizationKey(this));
+            mBraintreeFragment = BraintreeFragment.newInstance(this, "sandbox_f252zhq7_hh4cpc39zq4rgjcg");
             mIdealButton.setEnabled(true);
         } catch (InvalidArgumentException e) {
             onError(e);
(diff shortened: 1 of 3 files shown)


Epreuve (approved):
🍰

sshropshire (approved):

Demo/src/main/java/com/braintreepayments/demo/LocalPaymentsActivity.java
```diff
@@ -40,7 +40,7 @@ protected void onAuthorizationFetched() {
         }
 
         try {
-            mBraintreeFragment = BraintreeFragment.newInstance(this, Settings.getLocalPaymentsTokenizationKey(this));
+            mBraintreeFragment = BraintreeFragment.newInstance(this, "sandbox_f252zhq7_hh4cpc39zq4rgjcg");
```
asos-edgeorge, line 43: 👀 Should this have been replaced?

jaxdesmarais: Yes, the tokenization key being used previously did not have local payment methods enabled so we could not test this flow at all in our demo app. The previous key is used across several internal apps, so updating it in our demo app allows us to test the LPM flow as expected. Additionally the new key is the same as what we use for LPM on iOS so allows for consistency across platforms.

PR #777, opened 2023-08-28
Update Cardinal to 2.2.7-4

sarahkoop:
Thank you for your contribution to Braintree. 

> Before submitting this PR, note that we cannot accept language translation PRs. We support the same languages that are supported by PayPal, and have a dedicated localization team to provide the translations. If there is an error in a specific translation, you may open an issue and we will escalate it to the localization team.

### Summary of changes

 - Update Cardinal SDK to 2.2.7-4

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

PR #779, opened 2023-08-29
Update Cardinal to 2.2.7-4 (Continued)

sshropshire:
### Summary of changes

 - Update Cardinal SDK to 2.2.7-4
 - Continuation of https://github.com/braintree/braintree_android/pull/777

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop 
- @sshropshire

### DIFF
--- a/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureActivity.java
+++ b/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureActivity.java
@@ -32,19 +32,15 @@ public class ThreeDSecureActivity extends AppCompatActivity implements CardinalV
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-        challengeObserver = new CardinalChallengeObserver(this, new CardinalValidateReceiver() {
-            @Override
-            public void onValidated(Context context, ValidateResponse validateResponse, String s) {
-                handleValidated(validateResponse, s);
-            }
-        });
+        challengeObserver = new CardinalChallengeObserver(
+                this, (context, validateResponse, s) -> handleValidated(validateResponse, s));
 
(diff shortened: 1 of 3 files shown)


ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureActivity.java
```diff
@@ -36,7 +38,15 @@ public void onValidated(Context context, ValidateResponse validateResponse, Stri
                 handleValidated(validateResponse, s);
             }
         });
-        onCreateInternal(cardinalClient);
+
+        new Handler(Looper.getMainLooper()).post(new Runnable() {
```
sarahkoop, line 42: I'm not sure I fully see the benefit of adding this. We get an error message without it (cca_continue error) vs this error (internal error). It doesn't seem that much more useful to us or the merchant to have that specific error and this adds a delay to every 3DS transaction. What do you think?

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureActivity.java
```diff
@@ -36,11 +38,29 @@ public void onValidated(Context context, ValidateResponse validateResponse, Stri
...
```
sshropshire, line 42: This works, but I'm wondering if there's a better way? We basically need to defer the call to `launchCardinalAuthChallenge()` so that `handleValidated()` gets a chance to run after a process kill.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureActivity.java
sshropshire: The delay won't be noticeable this is a way to allow `handleValidated()` a chance to be called.

sarahkoop (approved):

scannillo (approved):
Appreciate the helpful comments!

PR #780, opened 2023-08-30
Android 14 Support

sarahkoop:
### Summary of changes

 - Update compileSdkVersion and targetSdkVersion to API 34 (Android 14) 
 - TODO: Update browser switch version when released with [Android 14 support change](https://github.com/braintree/browser-switch-android/pull/69)

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

### DIFF
--- a/SharedUtils/src/main/java/com/braintreepayments/api/SynchronousHttpClient.java
+++ b/SharedUtils/src/main/java/com/braintreepayments/api/SynchronousHttpClient.java
@@ -73,8 +73,8 @@ class SynchronousHttpClient {
             httpRequest.dispose();
         }
 
-        int responseCode = connection.getResponseCode();
         try {
+            int responseCode = connection.getResponseCode();
             return parser.parse(responseCode, connection);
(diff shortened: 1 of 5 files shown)


SharedUtils/src/main/java/com/braintreepayments/api/SynchronousHttpClient.java
```diff
@@ -73,8 +73,8 @@ String request(HttpRequest httpRequest) throws Exception {
             httpRequest.dispose();
         }
 
-        int responseCode = connection.getResponseCode();
         try {
+            int responseCode = connection.getResponseCode();
```
sarahkoop, line 77: This change fixes an integration test where we are testing that an invalid endpoint returns an error. Currently no error is returned because no connection is ever made. I think this update is the intended behavior (returns error on no connection).

scannillo (approved):

sshropshire (approved):

PR #793, opened 2023-10-03
Add Detekt Linting

sarahkoop:
### Summary of changes

 - Add `detekt` for Kotlin linting 
 - Fix lint errors or suppress warnings for each issue (I attempted to resolve issues that didn't require actual logic/code changes and suppressed most of the "SwallowedException" errors that would require logic changes to keep this PR simple and to be resolved in the future)

 ### Checklist

### Authors

- @sarahkoop


BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
```diff
@@ -131,6 +137,7 @@ internal class AnalyticsClient @VisibleForTesting constructor(
         } catch (e: Exception) {
             ListenableWorker.Result.failure()
         }
+        }
```
jaxdesmarais, line 140: Is it just me or does this look weird with two `}` at the same level?

jaxdesmarais (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
sarahkoop: Oh yes the indentation was messed up here - fixed!

PR #795, opened 2023-10-04
Remove Preferred Payment Methods

sarahkoop:
### Summary of changes

 - Remove Preferred Payment Methods from BraintreeCore and demo app
 - _Open Question_: should this change go on v5 or main? Our docs say this feature is beta so could remove in v4 instead

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop


sshropshire (approved):

scannillo (approved):
Changes look good, but I would remove it in V4

PR #798, opened 2023-10-05
Add VenmoLauncher API

sarahkoop:
### Summary of changes

 - Add `VenmoLauncher` and associated API to decouple `VenmoClient` from activity result handling
 - Remove `VenmoClient` API for activity result handling

The `VenmoFragment` demo app integration and `v5_MIGRATION_GUIDE` show what this new integration pattern looks like. I opted to rename the existing `tokenizeVenmoAccount` method to `requestAuthChallenge`, then the merchant launches the auth challenge, and passes that result to a new `tokenizeVenmoAccount` method (previously internal method called `onVenmoResult`). 

I think keeping a `tokenizeVenmoAccount` method existing, and adding one step the merchant has to take before it makes most sense, but please weigh in if that feels complicated and we should just do new method names for both.

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop


sshropshire (approved):
🔥 ! I have a couple q's that are out of scope I'll add in. This is great!

Demo/src/main/java/com/braintreepayments/demo/VenmoFragment.java
```diff
@@ -35,14 +37,20 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
         return view;
     }
 
-    private void handleVenmoResult(VenmoAccountNonce venmoAccountNonce) {
+    private void handleVenmoResult(VenmoAccountNonce nonce, Exception error) {
+        if (nonce != null) {
+            handleVenmoAccountNonce(nonce);
+        } else {
+            handleError(error);
+        }
```
sshropshire, line 51: This isn't relevant to this PR, but worth mentioning. I remember from that one discussion a merchant dev mentioned nullability in callbacks being a hinderance. It may be worth considering a single result type for result callbacks so that `null` checks aren't required.

I recently thought about having an interface type in PPCP e.g. `VenmoResult` that would allow logic like:
```java
if (result instanceof VenmoSuccessResult) {
  // handle success
} else if (result instanceof VenmoErrorResult) {
  // handle error
}
``` 
There are other ways too, like using enums. Just a thought I wanted to share. We have to figure something out on PPCP too. It'll be a good topic to brainstorm on going forward.

Demo/src/main/java/com/braintreepayments/demo/VenmoFragment.java
```diff
@@ -35,14 +37,20 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
         venmoButton = view.findViewById(R.id.venmo_button);
         venmoButton.setOnClickListener(this::launchVenmo);
 
-        braintreeClient = getBraintreeClient();
-        venmoClient = new VenmoClient(this, braintreeClient);
-        venmoClient.setListener(this);
+        venmoLauncher = new VenmoLauncher(this, venmoAuthChallengeResult ->
+                venmoClient.tokenizeVenmoAccount(venmoAuthChallengeResult, this::handleVenmoResult));
```
sshropshire, line 41: More 🧠 🌦️ : I wonder if composing a reference to `BraintreeClient` within each Launcher would be a good design pattern once `ClientTokenProvider` is no longer required. It would make a method like `venmoLauncher.authenticateVenmoUser()` possible. It might be cool, but it might be worth holding off to keep it simple.

Demo/src/main/java/com/braintreepayments/demo/VenmoFragment.java
sarahkoop: Ah yes I remember that feedback - I think a single result type could be a good move. I'll go forward with these API changes with the dual results for now and create a ticket to reconsider the approach and align across all modules

Demo/src/main/java/com/braintreepayments/demo/VenmoFragment.java
sarahkoop: Agreed I have a separate ticket to do this across all the modules - I think it would make most merchant integration stories simpler and we can add a params object for customization if needed

jaxdesmarais (approved):
🔥

v5_MIGRATION_GUIDE.md
```diff
@@ -19,5 +20,46 @@ Now, you can tokenize with just the card details:
...
+```kotlin
 
+class MyActivity : FragmentActivity() {
+    
+    private lateinit var venmoLauncher: VenmoLauncher
+    private lateinit var braintreeClient: BraintreeClient
+    private lateinit var venmoClient: VenmoClient
+    
+    @override fun onCreate(savedInstanceState: Bundle?) {
+        venmoLauncher = VenmoLauncher(this) { authChallengeResult ->
```
scannillo, line 44: I think of "authentication challenge" as a barrier to protect against unauthorized users, and less so for general login. According to[ this definition](https://en.wikipedia.org/wiki/Challenge%E2%80%93response_authentication) though, a basic password authentication does qualify.

Just food for thought that this language might be confusing since for flows (3DS) the "challenge" is optional and not required.

scannillo: Maybe `continue` or `presentLogin` or `startLogin`?

v5_MIGRATION_GUIDE.md
```diff
@@ -19,5 +20,46 @@ Now, you can tokenize with just the card details:
...
+            }
+        }
+    }
+    
+    fun initializeClients() {
+        braintreClient = BraintreeClient(context, "TOKENIZATION_KEY_OR_CLIENT_TOKEN")
+        venmoClient = VenmoClient(braintreeClient)
+    }
+    
+    fun launchVenmoFlow() {
```
scannillo, line 57: Do we need this func? Looks like it isn't called anywhere in this sample snippet

v5_MIGRATION_GUIDE.md
sarahkoop: @sshropshire what are your thoughts on changing up this naming scheme? 

I'd be fine with something like `venmoClient.presentLogin` returning a `LoginResult` or keeping with this `authChallenge` naming. Thinking through the other modules if we want consistent naming:

PayPal and Venmo are "login" screens
Google Pay is just the Google Pay dialog so not sure either naming makes sense for that one (same with PayPalNativeCheckout)
3DS is a true auth challenge
Local Payment is likely a "login" screen
SEPA launches a mandate

Maybe it won't make sense to have naming consistency and we should just describe what each flow is doing? I do like the clarity across integrations of a single pattern though

v5_MIGRATION_GUIDE.md
sarahkoop: This is what would be invoked on the click of the pay with venmo button or however they are starting the payment flow. I didn't include the button/setup logic in this example - would that make it more clear or a code comment?

v5_MIGRATION_GUIDE.md
scannillo: Yeh, I'm with you on preferring a unified name across modules!

scannillo: I have a question which is _slightly tangential_, but I think could inform the discussion:

For 3DS, we'll still require merchants to call [`continuePerformVerification`](https://github.com/braintree/braintree_android/blob/d7e67f8bf1aa25a7560f4e2975543a0f33533778/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java#L265) in order to proceed with the challenge?

So would that look something like this (in pseudocode/short-hand)?

```
class MyActivity {
    fun onCreate() {
        threeDSecureLauncher { authChallengeResult ->
            threeDSecureClient.performVerification() {
                threeDSecureClient. continuePerformVerification()
            }
        }
    }

    func onClick() {
        threeDSecureClient.requestAuthChallenge()
    }
}

class MyClass: ThreeDSecureListener {
    fun onThreeDSecureSuccess() {
        // handle result/nonce
    }
}
```

Would the launcher & client steps be flipped?

Venmo/src/test/java/com/braintreepayments/api/VenmoLauncherUnitTest.java
```diff
@@ -0,0 +1,60 @@
...
+        String expectedKey = "com.braintreepayments.api.Venmo.RESULT";
+        ActivityResultRegistry activityResultRegistry = mock(ActivityResultRegistry.class);
+        FragmentActivity lifecycleOwner = new FragmentActivity();
+
+        VenmoLauncher sut = new VenmoLauncher(activityResultRegistry, lifecycleOwner, callback);
+
+        verify(activityResultRegistry).register(eq(expectedKey), same(lifecycleOwner),
+                Mockito.<ActivityResultContract<VenmoAuthChallenge, VenmoAuthChallengeResult>>any(),
+                Mockito.any());
+
```
scannillo, line 44:
```suggestion
```

Venmo/src/test/java/com/braintreepayments/api/VenmoLauncherUnitTest.java
```diff
@@ -0,0 +1,60 @@
...
+public class VenmoLauncherUnitTest {
+    @Mock
+    ActivityResultLauncher<VenmoAuthChallenge> activityResultLauncher;
+    private VenmoAuthChallengeResultCallback callback;
+
+    @Before
+    public void beforeEach()  {
+        MockitoAnnotations.openMocks(this);
+        callback = mock(VenmoAuthChallengeResultCallback.class);
+    }
```
scannillo, line 32:
```suggestion
    }
    
```

Venmo/src/test/java/com/braintreepayments/api/VenmoLauncherUnitTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+import org.robolectric.RobolectricTestRunner;
+
+@RunWith(RobolectricTestRunner.class)
+public class VenmoLauncherUnitTest {
+    @Mock
```
scannillo, line 24:
```suggestion

    @Mock
```

Venmo/src/main/java/com/braintreepayments/api/VenmoLauncher.java
```diff
@@ -0,0 +1,63 @@
...
+     * Used to launch the Venmo authentication flow to tokenize a Venmo account. This class must
+     * be instantiated in the OnCreate method of your Fragment.
+     *
+     * @param fragment an Android Fragment from which you will launch the Venmo app
+     * @param callback a {@link VenmoAuthChallengeResultCallback} to receive the result of the
+     *                 Venmo app switch authentication flow
+     */
+    public VenmoLauncher(@NonNull Fragment fragment,
+                         @NonNull VenmoAuthChallengeResultCallback callback) {
+        this(fragment.getActivity().getActivityResultRegistry(), fragment.getViewLifecycleOwner(),
```
scannillo, line 31: Question - is `fragment.getViewLifecycleOwner()` something that would be acceptable for a merchant to do? Just curious if we need to offer 2 VenmoLauncher inits, or if doing so prevents issues.

Venmo/src/main/java/com/braintreepayments/api/VenmoLauncher.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * Used to launch the Venmo authentication flow to tokenize a Venmo account. This class must
+     * be instantiated in the OnCreate method of your Activity.
+     *
+     * @param activity an Android Activity from which you will launch the Venmo app
+     * @param callback a {@link VenmoAuthChallengeResultCallback} to receive the result of the
+     *                 Venmo app switch authentication flow
+     */
```
scannillo, line 42: 👏 This is super clear

scannillo (approved):
No blockers, just a few questions!

v5_MIGRATION_GUIDE.md
sarahkoop: Updated method name 👍

v5_MIGRATION_GUIDE.md
```diff
@@ -19,5 +20,46 @@ Now, you can tokenize with just the card details:
...
+activity result handling. 
+
+`VenmoLauncher` has been added to handle the app switching portion of the Venmo flow for user 
+authentication via the Venmo app. This class uses the Android Activity Result API and therefore 
+must be instantiated in the `OnCreate` method of your Activity or Fragment.
+
+`BraintreeClient` and `VenmoClient` no longer require references to Fragment or Activity and do not 
+need to be instantiated in `OnCreate`.
+
+```kotlin
```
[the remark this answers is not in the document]
sarahkoop: Very cool - added that!

v5_MIGRATION_GUIDE.md
sarahkoop: In the 3DS case, the `performVerification` method becomes `requestAuthChallenge` (or we could leave the `performVerification` naming as is)

```
class MyActivity {
    fun onCreate() {
        threeDSecureLauncher = ThreeDSecureLauncher(this) { authChallengeResult ->
              threeDSecureClient.continuePerformVerification() { nonce, error -> 
                  // handle nonce/error
             }
        }
    }

    func onClick() {
        threeDSecureClient.requestAuthChallenge(this, request) { authChallenge, error -> 
          threeDSecureLauncher.launch(authChallenge)
        }
    }
}
```

sarahkoop: I've added a ticket to re-asses the naming of the `requestAuthChallenge` method and make sure it makes sense across modules so I'm going to leave it as is in this PR

Venmo/src/main/java/com/braintreepayments/api/VenmoLauncher.java
scannillo: That's really helpful, thanks Sarah!

v5_MIGRATION_GUIDE.md
sshropshire: I'm all for merchants having an intuitive interface. It's a good callout to try and improve naming whenever we can! 

I don't have a major preference, but with this new design pattern the `Launcher` will be presenting the login flow instead of the `Client`. We talked about potentially allowing the Launcher to have a reference to the Client, that would make methods like `launcher.startLogin()` possible.

The current naming is tightly coupled to the underlying Android frameworks we use. The way it is now helps us a lot more than it'll help merchants. We should iterate on naming before GA to maximize API clarity.

PR #804, opened 2023-10-12
Fix Publish and Release CI

sarahkoop:
### Summary of changes

 - Recent release failed on the close and release step (not possible to test without releasing)
 - Investigation shows that breakout of publish an close/release actions doesn't work for multiple modules with the new gradle publish plugin
 - Updated this process to match `paypal-android` publish step exactly
 - Tested the publish to sonatype and see all modules within `braintree_android` staging repository
- Will need to run another full release to confirm the close step works now

 ### Checklist

### Authors

- @sarahkoop


jaxdesmarais (approved):
👻

sshropshire (approved):

PR #805, opened 2023-10-12
Update to Java 11 and Cleanup Files

sarahkoop:
### Summary of changes

 - Update project to compile for Java 11
 
**Reformat Only (no functional code changes):**
 - Replace explicit callbacks with lambdas throughout codebase (java 11 feature)
 - Run Android studio automated code cleanup on files (mainly fixes too long lines)

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
-      - name: Setup Java 8
+      - name: Setup Java
         uses: ./.github/actions/setup
       # After decoding the secret key, place the file in ~ /. Gradle/ secring.gpg
       - name: Decode Signing Key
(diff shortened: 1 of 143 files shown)


sshropshire (approved):
🌊

.github/workflows/release.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
-      - name: Setup Java 8
+      - name: Setup Java
```
jaxdesmarais, line 17: Do we care to specify the version in the `Setup Java` step? The answer may be no - more just thinking of me trying to remember where the java version is specified for the SDK and and always forgetting. 🙈

jaxdesmarais (approved):

.github/workflows/release.yml
sarahkoop: This is the version of Java that's use to build the modules and run the tests - we should try and keep it in line with the version we compile for, but it was actually already bumped to Java 11 in CI 😆  (this is just removing the "8" from the name of the task since it's incorrect)

jaxdesmarais: LOL - here I was assuming it was right the whole time 🙃

PR #806, opened 2023-10-13
Adding auth email config

kelay858:

### Summary of changes

 - Adding in the auth config for nxo to be able to set the merchant passed email

 ### Checklist

 - [ ] Added a changelog entry

-

### DIFF
--- a/PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeRequest.java
+++ b/PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeRequest.java
@@ -51,7 +51,7 @@ public abstract class PayPalNativeRequest implements Parcelable {
     private String riskCorrelationId;
     private final ArrayList<PayPalNativeCheckoutLineItem> lineItems;
     private String returnUrl;
-
+    private String authEmail;
     /**
      * Constructs a request for PayPal Checkout and Vault flows.
      */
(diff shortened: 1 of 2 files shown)


PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeRequest.java
```diff
@@ -203,6 +203,21 @@ public void setReturnUrl(@NonNull String returnUrl) {
...
+    public String getUserAuthenticationEmail() {
+        return authEmail;
+    }
+
+    /**
+     * Optional: Sets the email of the user prior to calling authentication
+     *
+     * @param authEmail the users email address
+     */
+    public void setUserAuthenticationEmail(@NonNull String authEmail) {
```
sshropshire, line 216:
```suggestion
    public void setUserAuthenticationEmail(@Nullable String authEmail) {
```
I believe the annotations need to match up for Kotlin interop.

tdchow: Good catch!

sarahkoop (commented):
Code looks good but fix failing unit tests please

tdchow:
Thanks for the review!

It looks like there's a compile issue due to the `AuthConfig` param missing in the Checkout Config object. We'll need to update the Mobile Checkout SDK to 1.2.0 once it's available on maven central. So we can shelf this PR for now.

PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeRequest.java
```diff
@@ -203,6 +203,21 @@ public void setReturnUrl(@NonNull String returnUrl) {
         this.returnUrl = returnUrl;
     }
 
+    @Nullable
+    public String getUserAuthenticationEmail() {
+        return authEmail;
+    }
+
+    /**
+     * Optional: Sets the email of the user prior to launching authentication
```
scannillo, line 212: Can we use the same docstring across [iOS](https://github.com/braintree/braintree_ios/pull/1118) and Android?

sarahkoop:
@kelay858 I think this has been replaced by #829 - is it ok to close this one?

PR #807, opened 2023-10-13
PayPal Launcher API

sarahkoop:
### Summary of changes

 - Add PayPal launcher API (this is slightly different than the previous two launcher API PRs because this is a web based flow rather than an activity based flow)
 - **Note**: Method/callback/result object names have not been updated to keep this PR diff smaller while we determine the best naming convention to follow throughout modules

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop


PayPal/src/main/java/com/braintreepayments/api/PayPalLauncher.java
```diff
@@ -0,0 +1,67 @@
...
+     * Launches the PayPal flow by switching to a web browser for user authentication
+     *
+     * @param activity       an Android {@link FragmentActivity}
+     * @param payPalResponse the result of the PayPal web authentication flow received from invoking
+     *                       {@link PayPalClient#tokenizePayPalAccount(FragmentActivity,
+     *                       PayPalRequest, PayPalFlowStartedCallback)}
+     * @throws BrowserSwitchException if there is an error launching the PayPal web flow
+     */
+    public void launch(@NonNull FragmentActivity activity, @NonNull PayPalResponse payPalResponse)
+            throws BrowserSwitchException {
```
sarahkoop, line 41: Should this throw or how should this exception be handled? The Activity result based `launch` methods (Google Pay, Venmo) don't throw or callback any exception, so I opted to throw to keep the `launch` parameters cleaner. This does mean a try/catch block in the merchant code (see migration guide example). Other alternative would be to callback an error in this method, or give the `PayPalLauncher` a callback in the constructor to handle results (this would align with the Activity based launchers)

PayPal/src/main/java/com/braintreepayments/api/PayPalLauncher.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param context the context used to check for pending results
+     * @param intent  the intent to return to your application containing a deep link result from
+     *                the PayPal browser flow
+     * @return a {@link BrowserSwitchResult} to be passed to
+     * {@link PayPalClient#tokenizePayPalAccount(FragmentActivity, PayPalRequest,
+     * PayPalFlowStartedCallback)}
+     */
+    public BrowserSwitchResult deliverResult(@NonNull Context context, @NonNull Intent intent) {
+        BrowserSwitchResult result = browserSwitchClient.parseResult(context, PAYPAL, intent);
+        browserSwitchClient.clearActiveRequests(context);
```
sarahkoop, line 64: @sshropshire Can `clearActiveRequests` be called under the hood here or do we need to give the merchants an accessor to call this themselves after handling the result?

PayPal/src/main/java/com/braintreepayments/api/PayPalLauncher.java
sshropshire: Yeah we should design it to have the same API as the Activity launchers if possible. It'd be awesome if we're able to offer a familiar DX for onboarding each Activity and Browser-based payment flow.

PayPal/src/main/java/com/braintreepayments/api/PayPalLauncher.java
sshropshire: For this one I'm 50/50. We could call it for merchants, but run the risk of using too much abstraction. I'm in favor of having merchants call `launcher.clearActiveRequests()` (doesn't have to be the same name). A better solution may be to revisit the acceptance criteria for browser switching to make sure we have it right.

sarahkoop: Ok agreed for now I am going to keep it under the hood because I think it should always be called immediately after a successful parse, but created a ticket to revisit this pattern before GA

PayPal/src/main/java/com/braintreepayments/api/PayPalLauncher.java
sarahkoop: Ok updated this so that the launcher constructor takes a callback as a parameter (to align with Activity Result flows) and if there's an error in the `launch` method, that error will be delivered to the callback from the constructor.

sshropshire (approved):

PayPal/src/main/java/com/braintreepayments/api/PayPalBrowserSwitchResult.java
```diff
@@ -0,0 +1,23 @@
...
+        this.error = error;
+    }
+
+    BrowserSwitchResult getBrowserSwitchResult() {
+        return browserSwitchResult;
+    }
+
+    Exception getError() {
+        return error;
+    }
```
scannillo, line 22: Do these need to be public w/ docstrings?

v5_MIGRATION_GUIDE.md
```diff
@@ -150,3 +152,71 @@ class MyActivity : FragmentActivity() {
...
+`PayPalLauncher` has been added to handle the browser switching portion of the PayPal flow for
+payment authorization via PayPal login. This class is used to handle browser switch 
+results upon returning to your app from the web flow, therefore must be instantiated in or 
+before the `OnResume` method of your Activity or Fragment.
+
+`BraintreeClient` and `PayPalClient` no longer require references to Fragment or Activity and
+do not need to be instantiated in `OnCreate`.
+
+```diff
+class MyActivity : FragmentActivity() {
```
scannillo, line 169: Do we need to show calling `handleReturnToAppFromBrowser` in this code sample?

sarahkoop: Ah yes this was renamed from `deliverResult` - updated migration guide

PayPal/src/main/java/com/braintreepayments/api/PayPalBrowserSwitchResult.java
sarahkoop: No the idea with this result object is that the merchant doesn't have to do anything with it / be aware of any params and just passes it directly to the `PayPalClient#onBrowserSwitchResult` method, which will then handle any errors and call them back to the merchant. We could break it up and allow merchants to handle errors at each step, but this seems cleaner for integrations

CHANGELOG.md
```diff
@@ -36,6 +36,14 @@
     * Change `ThreeDSecureClient#performVerification` and `ThreeDSecureClient#initializeChallengeWithLookupResponse` parameters
     * Add `ThreeDSecureClient#onCardinalResult`
     * Remove `versionRequested` from `ThreeDSecureRequest`
+  * PayPal
+    * Remove `PayPalListener`
+    * Add `PayPalLauncher` and `PayPalResponse`
+    * Add parameter to `PayPalFlowStartedCallback`
+    * Remove overload constructors, `setListener`, `parseBrowserSwitchResult`
```
scannillo, line 43:
```suggestion
    * Remove overload constructors, `setListener`, `parseBrowserSwitchResult`,
```

scannillo (approved):

PR #811, opened 2023-10-25
v5 Naming Pattern Proposal

sarahkoop:
### Summary of changes

- This PR offers a proposal for different method/object naming patterns that we could implement for v5. View the markdown file in rich diff for easiest table comparison. 
- This PR should not be merged, but please comment ideas or feedback on the naming alternatives.
- My opinion is that we should follow the consistent naming pattern option, but I am not sure `paymentAuth` is the preferred naming pattern. 

 ### Checklist

### Authors

- @sarahkoop

### DIFF
--- a/METHOD_NAMING_PROPOSAL.md
+++ b/METHOD_NAMING_PROPOSAL.md
@@ -244,15 +244,15 @@ payPalClient.tokenizePayPalAccount(activity, request)
 <td>
 
 ```kotlin
-payPalLauncher = PayPalLauncher() { payPalBrowserSwitchResult ->
+payPalLauncher = PayPalLauncher() { payPalPaymentAuthResult ->
   // Parse the result from the PayPal browser flow and return a nonce
-  payPalClient.onBrowserSwitchResult(payPalBrowserSwitchResult) 
+  payPalClient.tokenize(payPalPaymentAuthResult) 
 }
 
 // Retrieve data required to launch the PayPal flow in a browser
-payPalClient.tokenizePayPalAccount(context, request) { payPalResponse, error ->
+payPalClient.createPaymentAuthRequest(context, request) { payPalPaymentAuthRequest, error ->
   // Launch the PayPal in a browser
-  payPalLauncher.launch(activity, payPalResponse) 
+  payPalLauncher.launch(activity, payPalPaymentAuthRequest) 
 }
 ```
 
@@ -282,15 +282,15 @@ localPaymentClient.startPayment(activity, request) { localPaymentResult, error -
 <td>
 
 ```kotlin
-localPaymentLauncher = LocalPaymentLauncher() { localPaymentBrowserSwitchResult ->
+localPaymentLauncher = LocalPaymentLauncher() { localPaymentAuthResult ->
   // Parse the result from the local payment browser flow and return a nonce
-  localPaymentClient.approveLocalPayment(localPaymentBrowserSwitchResult) 
+  localPaymentClient.tokenize(localPaymentAuthResult) 
 }
 
 // Retrieve data required to launch the local payment flow in a browser
-localPaymentClient.startPayment(context, request) { localPaymentResult, error ->
+localPaymentClient.createPaymentAuthRequest(context, request) { localPaymentAuthRequest, error ->
   // Launch the local payment in a browser
-  localPaymentLauncher.launch(activity, localPaymentResult) 
+  localPaymentLauncher.launch(activity, localPaymentAuthRequest) 
 }
 ```
 
@@ -320,16 +320,16 @@ threeDSecureClient.performVerification(activity, request) { threeDSecureResult, 
 <td>
 
 ```kotlin
-threeDSecureLauncher = ThreeDSecureLauncher() { authenticationResult ->
+threeDSecureLauncher = ThreeDSecureLauncher() { threeDSecurePaymentAuthResult ->
   // Parse the result from the 3DS auth flow and return a nonce
-  threeDSecureClient.continuePerformVerification(authenticationResult) 
+  threeDSecureClient.tokenize(threeDSecurePaymentAuthResult) 
 }
 
 // Retrieve data required to launch the 3DS flow
 // Instantiate Cardinal SDK
-threeDSecureClient.performVerification(context, request) { threeDSecureResult, error ->
+threeDSecureClient.createPaymentAuthRequest(context, request) { threeDSecurePaymentAuthRequest, error ->
   // Launch 3DS flow in Cardinal SDK 
-  threeDSecureLauncher.launch(threeDSecureResult) 
+  threeDSecureLauncher.launch(threeDSecurePaymentAuthRequest) 
 }
 ```
 
@@ -357,16 +357,16 @@ venmoClient.tokenizeVenmoAccount(activity, request)
 <td>
 
 ```kotlin
-venmoLauncher = VenmoLauncher(activity) { authChallengeResult ->
+venmoLauncher = VenmoLauncher(activity) { venmoPaymentAuthResult ->
   // Create nonce from payment context app switch result and vault if required
-  venmoClient.tokenizeVenmoAccount(authChallengeResult) 
+  venmoClient.tokenize(venmoPaymentAuthResult) 
 }
 
 
   // Create Venmo payment context
-venmoClient.requestAuthChallenge(context, request) { authChallenge, error ->
+venmoClient.createPaymentAuthRequest(context, request) { venmoPaymentAuthRequest, error ->
   // Launch Venmo app
-  venmoLauncher.launch(authChallenge) 
+  venmoLauncher.launch(venmoPaymentAuthRequest) 
 }
 ```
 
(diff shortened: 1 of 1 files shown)


METHOD_NAMING_PROPOSAL.md
```diff
@@ -0,0 +1,412 @@
...
+### Consistent Naming Pattern
+
+With this implementation, the method names do not align with v4, so there is greater merchant migration difficulty, but the naming pattern is consistent across payment methods and provides clarity to what is actually occurring.
+For this pattern, I've chosen to use the `paymentAuth` naming pattern.
+
+`client.createPaymentAuthRequest` returns `paymentAuthRequest`
+
+`launcher.launch(paymentAuthRequest)` returns `paymentAuthResult`
+
+`client.tokenize(paymentAuthResult)` returns `paymentMethodNonce`
```
jaxdesmarais, line 222: This may be a 🔥 take but do we need to use the work `auth` at all? Could we do:
```suggestion
`client.createPaymentRequest` returns `paymentRequest`

`launcher.launch(paymentRequest)` returns `paymentResult`

`client.tokenize(paymentResult)` returns `paymentMethodNonce`
```
Mainly thinking from the point of view that `auth` tends to signal an authorization attempt vs capturing a payment which could be confusing? Not sure what you/others think, so just throwing some ideas out there.

METHOD_NAMING_PROPOSAL.md
```diff
@@ -0,0 +1,412 @@
...
+// Return nonce to listener
+payPalClient.tokenizePayPalAccount(activity, request)
+```
+
+</td>
+
+<td>
+
+```kotlin
+payPalLauncher = PayPalLauncher() { payPalBrowserSwitchResult ->
```
hollabaq86, line 45: To confirm, this pattern would require a `launcher` object for each individual payment method, correct? A `launcher` object cannot and should not be shared across payment method flows?

METHOD_NAMING_PROPOSAL.md
```diff
@@ -0,0 +1,412 @@
+# v5 Method Naming Proposal
+
+## Overview
+
+In v5 of the Braintree Android SDK, we are introducing a new integration pattern that includes a
```
hollabaq86, line 5: Can we get more info on why the move to this `Launcher` class is being proposed? What are the potential benefits/setbacks to this design change?

METHOD_NAMING_PROPOSAL.md
```diff
@@ -0,0 +1,412 @@
...
+<td>
+
+```kotlin
+localPaymentLauncher = LocalPaymentLauncher() { localPaymentBrowserSwitchResult ->
+  // Parse the result from the local payment browser flow and return a nonce
+  localPaymentClient.approveLocalPayment(localPaymentBrowserSwitchResult) 
+}
+
+// Retrieve data required to launch the local payment flow in a browser
+localPaymentClient.startPayment(context, request) { localPaymentResult, error ->
```
hollabaq86, line 89: this pattern is a little confusing as an Android outsider, as localPaymentClient is calling itself within the startPayment function via localPaymentLauncher. What is the localPaymentLauncher doing in this workflow that localPaymentClient can do on its own?

hollabaq86: Is it the activity access and management that makes the difference?

METHOD_NAMING_PROPOSAL.md
```diff
@@ -0,0 +1,412 @@
...
+  threeDSecureClient.continuePerformVerification(activity, request, threeDSecureResult)
+}
+```
+
+</td>
+
+<td>
+
+```kotlin
+threeDSecureLauncher = ThreeDSecureLauncher() { authenticationResult ->
```
hollabaq86, line 121: it's been awhile but we may want to confirm if this is a result from the authentication result of 3DS or lookup result

hollabaq86 (commented):
left some questions as an absolute Android outsider, great work putting this together!

METHOD_NAMING_PROPOSAL.md
KunJeongPark: I think of Auth as permission, so it makes sense to me but I don't feel too strongly about it. Less change for typo is good also.

METHOD_NAMING_PROPOSAL.md
KunJeongPark: I was thinking about the launcher too, Holly. My understanding is that it will be separate for each payment method and abstracts away details of laughing webView or app switch. Except card module?

METHOD_NAMING_PROPOSAL.md
KunJeongPark: It would be helpful for me to see naming change proposals side by side (what is now, minimum change, consistent changes) but I think you all probably have them memorized.

sarahkoop: @jaxdesmarais I like plain `paymentRequest` too! 

@KunJeongPark You should be able to see the side by side comparison if you view the rich diff of the markdown file (I tried to make a table with two columns). The v4 column is what is now, and the v5 column is the proposed change)

METHOD_NAMING_PROPOSAL.md
sarahkoop: Yes it needs to be a separate launcher for each payment method because that class is responsible for actually launching the flow (i.e. opening PayPal browser or switching to Venmo app), so it is payment type specific.

METHOD_NAMING_PROPOSAL.md
sarahkoop: Yes see comment above - the reason for the Launcher is to separate out the portion of the flow that needs to be aware of Android Activity (launching apps/browsers) and the portion that doesn't (API calls/tokenizing), stemming from the Android API requirement to register activity result handlers in the OnCreate method of your Activity (before being able to fetch a client token). But open to suggestions for making this pattern more clear!

METHOD_NAMING_PROPOSAL.md
scannillo: Pretty sure this would be the lookupResult. And the merchant can decide here wether or not to show the challenge or not. If they decide to take on the liability themselves, they would skip the `threeDSecureClient.continuePerformVerification()` call and directly call `cardClient.tokenize()` if they choose

METHOD_NAMING_PROPOSAL.md
scannillo: 😈 advocate - `PaymentResult` might be easy to confuse with `PaymentMethodNonce`

METHOD_NAMING_PROPOSAL.md
```diff
@@ -0,0 +1,412 @@
...
+```
+
+</td>
+
+<td>
+
+```kotlin
+venmoLauncher = VenmoLauncher(activity) { authChallengeResult ->
+  // Create nonce from payment context app switch result and vault if required
+  venmoClient.tokenizeVenmoAccount(authChallengeResult)
```
scannillo, line 160: I like how this proposal keeps `tokenize()` as the last step in all of the flows!

METHOD_NAMING_PROPOSAL.md
```diff
@@ -0,0 +1,412 @@
...
+```
+
+</td>
+
+<td>
+
+```kotlin
+localPaymentLauncher = LocalPaymentLauncher() { localPaymentAuthResult ->
+  // Parse the result from the local payment browser flow and return a nonce
+  localPaymentClient.tokenize(localPaymentAuthResult)
```
scannillo, line 287: I like how this proposal keeps `tokenize()` as the last step in all of the flows!

scannillo:
> My opinion is that we should follow the consistent naming pattern option, but I am not sure `paymentAuth` is the preferred naming pattern.

I also vote for the consistent naming pattern option. The merchants will already have to restructure their integrations a bit, so using consistent names from the start will be more clear.

METHOD_NAMING_PROPOSAL.md
scannillo: I like the names you have proposed, Sarah! I can't think of anything better ATM

sshropshire: Another option would be to name the request/result pair to highlight the integration point of `Launcher` and `Client`.  `LauncherRequest` and `LauncherResult` might be helpful in that case. I do prefer the naming convention to be decoupled from the implementation though, that would be ideal.

Overall, the best name will make it clear that additional steps are required by the buyer, and the result of the auth challenge will be handled by the SDK before a `PaymentMethodNonce` is returned.

METHOD_NAMING_PROPOSAL.md
sshropshire: Yeah @hollabaq86 this has roots in Android's [Activity Result API](https://developer.android.com/training/basics/intents/result). This pattern allows us to further decouple feature clients from Android, which is a big win for merchants because they don't have unnecessary restrictions imposed on them.

We introduced [Client Token Provider](https://github.com/braintree/braintree_android/blob/main/v4.9.0%2B_MIGRATION_GUIDE.md#client-token-provider) in `v4` to provide a BraintreeFragment-like `v3` interface. There were benefits to providing a single entry point into each feature via Feature Clients before `startActivityForResult()` was deprecated. With `v4`, we decided to fully encapsulate the new Activity Result API within each feature client. This pattern works well for 90% of merchant integrations, but it can get complex when merchants integrate with the Braintree SDK in a headless manner i.e. without a UI.

All in all, complexity is conserved because we're adding a `Launcher` concept and (eventually) removing the `ClientTokenProvider` concept. I'd be down to deep dive into this some time! This pattern is the result of listening to our merchants and it will make onboarding `v5` a lot smoother.

sarahkoop:
Closing this with the decision to go with pattern:
`client.createPaymentAuthRequest` returns `paymentAuthRequest`

`launcher.launch(paymentAuthRequest)` returns `paymentAuthResult`

`client.tokenize(paymentAuthResult)` returns `paymentMethodNonce`

We can revisit before GA if the naming scheme doesn't make sense.

PR #810, opened 2023-10-24
Local Payment Launcher API

sarahkoop:
### Summary of changes

 - Add Local Payment launcher API (aligns with [PayPal Launcher API](https://github.com/braintree/braintree_android/pull/807))
- **Note:** Method/callback/result object names have not been updated to keep this PR diff smaller while we determine the best naming convention to follow throughout modules

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

PR #808, opened 2023-10-18
3DS Launcher API

sarahkoop:
### Summary of changes

- Add 3DS launcher API 
- Consolidate `performVerification` and `continuePerformVerification` with the addition of the `ThreeDSecureLauncher` the determination of whether to "continue" with the verification happens before launching now
- Remove some of the unsupported 3DS V1 logic to allow for simplified integration - the remaining V1 code cleanup will happen in a future PR
- **Note:** Method/callback/result object names have not been updated to keep this PR diff smaller while we determine the best naming convention to follow throughout modules (`Cardinal` public naming convention will not stay - just kept the name of the existing class made public for now)

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

PR #812, opened 2023-10-31
SEPA Launcher API [Source Code]

scannillo:
### Summary of changes

 - Replace existing SEPA integration offering with new Launcher API style (see #807, #808, & #810 for reference)

### Feedback Requested

- Prefixing everything with `SEPADirectDebit` is pretty wordy, but was our already-existing pattern. What do folks think? Can we move to just a `SEPA` prefix?
- There are some SEPA cases where a web-flow is not required (the mandate is already approved). In this case, I opted to populate an optional `SEPADirectDebitResponse.nonce` property. If present, the merchant is done, otherwise they have to continue on to call `SEPLauncher.launch(sepaResponse)`.
    - Would love folks thoughts on this.

### Note

- Tests expected to fail
    - To be addressed in follow-up PR to `sepa-launcher-api-refactor` branch. 
    - Once complete, SEPA feature branch can be merged into `v5`.

 ### Checklist

 - [X] Added a changelog entry

### Authors
@scannillo


SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitClient.java
```diff
@@ -113,37 +86,57 @@ public void tokenize(final FragmentActivity activity,
...
+        }
 
         int result = browserSwitchResult.getStatus();
         switch (result) {
             case BrowserSwitchStatus.CANCELED:
                 braintreeClient.sendAnalyticsEvent("sepa-direct-debit.browser-switch.canceled");
-                listener.onSEPADirectDebitFailure(
-                        new UserCanceledException("User canceled SEPA Debit."));
+                callback.onResult(null, new UserCanceledException("User canceled SEPA Debit."));
                 break;
```
scannillo, line 140: This code block never executes on a user cancel clicking the "X" in the CCT (same for [this block](https://github.com/braintree/braintree_android/blob/v5/PayPal/src/main/java/com/braintreepayments/api/PayPalClient.java#L237-L241) in the PayPalClient).

I talked with @sarahkoop about this a little, and since the Launcher pattern in V5 is using the "manual browser switching" pattern from V4 she mentioned that we don't have a way to distinguish cancel cases. Currently for the cancel case, the [Launcher's BroswerSwitchClient.parseResult() call](https://github.com/braintree/braintree_android/blob/e94dec16aa0608c080f67bfe718658dac6363619/PayPal/src/main/java/com/braintreepayments/api/PayPalLauncher.java#L75) returns a null intent, which no-ops the method and doesn't notify the merchant that the flow is complete. This is currently the same for both PP and SEPA v5 launcher flows.

Wanted to loop in @sshropshire too. It sounds like this is expected for v5? Sarah mentioned a potential customization to CCT as a workaround.

SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitClient.java
```diff
@@ -25,65 +27,34 @@ public class SEPADirectDebitClient {
...
      *
-     * @param activity               an Android FragmentActivity
-     * @param sepaDirectDebitRequest the {@link SEPADirectDebitRequest}.
+     * @param activity Android FragmentActivity
+     * @param sepaDirectDebitRequest {@link SEPADirectDebitRequest}
+     * @param callback {@link SEPADirectDebitFlowStartedCallback}
      */
-    public void tokenize(final FragmentActivity activity,
-                         final SEPADirectDebitRequest sepaDirectDebitRequest) {
+    public void tokenize(@NonNull final FragmentActivity activity,
```
sarahkoop, line 55: Can activity be removed from this method? Doesn't look like it's used anymore

SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitLauncher.java
```diff
@@ -0,0 +1,79 @@
...
+     * <p>
+     * This method will deliver a {@link SEPADirectDebitBrowserSwitchResult} to the
+     * {@link SEPADirectDebitLauncherCallback} used to instantiate this class. The
+     * {@link SEPADirectDebitBrowserSwitchResult} should be passed to
+     * {@link SEPADirectDebitClient#onBrowserSwitchResult(SEPADirectDebitBrowserSwitchResult, SEPADirectDebitBrowserSwitchResultCallback)}
+     *
+     * @param context the context used to check for pending results
+     * @param intent  the intent to return to your application containing a deep link result from
+     *                the SEPA mandate flow
+     */    public void handleReturnToAppFromBrowser(@NonNull Context context, @NonNull Intent intent) {
```
sarahkoop, line 72:
```suggestion
     */    
     public void handleReturnToAppFromBrowser(@NonNull Context context, @NonNull Intent intent) {
```

SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitLauncherCallback.java
```diff
@@ -0,0 +1,8 @@
+package com.braintreepayments.api;
+
+import androidx.annotation.NonNull;
+
+public interface SEPADirectDebitLauncherCallback {
```
sarahkoop, line 5: This will need docs (I think I missed them in PayPal too), but since the params/description will likely change based on the decision in #813 I think it's fine to leave blank for now

sarahkoop (approved):
⭐

SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitClient.java
scannillo: Done!

SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitLauncher.java
scannillo: Addressed in 16d1fcd0e3995d2d6804c4c4b525e85eda3e5973

SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitLauncherCallback.java
scannillo: Good catch! I gave it a shot and added a docstring in 16d1fcd0e3995d2d6804c4c4b525e85eda3e5973

SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitClient.java
sshropshire: Yeah we still haven't been able to find a good way to signal a cancelation. CCT is super restrictive.

PR #813, opened 2023-10-31
v5 Result Object Proposal

sarahkoop:
### Summary of changes

 - This document summarizes alternative result handling patterns we could follow in v5. Please provide input/discuss!

 ### Checklist

### Authors

- @sarahkoop

### DIFF
--- a/RESULT_PROPOSAL.md
+++ b/RESULT_PROPOSAL.md
@@ -157,14 +157,14 @@ public void onError(Exception error) {
 ```
 ## Single Result Object With Types
 
-This approach is the most Kotlin-first pattern since the return type handling and casting can be
-done in the least lines of code. It also resolves the nullability issues. However, the Java
-integration becomes somewhat complex with casting. This approach aligns with how other payment
-SDKs handle results.
+This approach is the most Kotlin-first pattern. It also resolves the nullability issues. However, 
+the Java integration becomes somewhat complex with casting. This approach aligns with how other 
+payment SDKs handle results.
 
 This approach relies on the Kotlin sealed class, so requires the `PaymentAuthResult` and
-`PaymentResult` objects to live in `BraintreeCore`. This reduces code duplication in our SDK, but
-requires casting by the merchants (ex: PaymentMethodNonce is returned instead of
+`PaymentResult` objects to live in `BraintreeCore`. It also requires the launchable result 
+objects to all extend a single class (ex: `LaunchableRequest`). This reduces code duplication in 
+our SDK, but requires casting by the merchants (ex: PaymentMethodNonce is returned instead of
 module-specific PayPalAccountNonce). Once the SDK is fully converted to Kotlin, these could be
 moved into payment module specific sealed classes (ex: `PayPalPaymentAuthResult`).
 
--- a/RESULT_PROPOSAL.md
+++ b/RESULT_PROPOSAL.md
@@ -1,0 +1,295 @@
+# Result Handling Android v5
+
+This document proposes alternative patterns for result handling in Braintree Android v5. In
+Braintree Android v4, there are two different patterns for delivering results: a listener with
+two methods (`onSuccess`/`onFailure`) or a callback with one method and two nullable parameters
+(`onResult(@Nullable nonce, @Nullable error)`). A merchant brought to our attention [some issues](https://github.com/braintree/braintree_android/issues/491)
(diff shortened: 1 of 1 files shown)


RESULT_PROPOSAL.md
```diff
@@ -0,0 +1,295 @@
...
+public void onCancel() {
+    // handle cancel
+}
+
+@Ovrride
+public void onError(Exception error) {
+    // handle error
+}
+```
+## Single Result Object With Types
```
jaxdesmarais, line 158: This _feels_ the best to me, but it also reads similar to Swift so I think that makes sense. It sounds like this is also what other folks have moved to so would be a good move to modernize our approach.

Is there a world in which we'd consider offering 2 similar approaches in v5 and in a future v6 go with 1? In iOS v6 we offer both [completions](https://github.com/braintree/braintree_ios/blob/b0db748229adf147f53573ea2a567ee6d74f43de/Sources/BraintreePayPal/BTPayPalClient.swift#L70-L75) and [async/await](https://github.com/braintree/braintree_ios/blob/b0db748229adf147f53573ea2a567ee6d74f43de/Sources/BraintreePayPal/BTPayPalClient.swift#L88-L98) with a plan to go to only async/await in v7. I am not sure if something like this is feasible on Android though or would over complicate things.

sarahkoop: It would be a bit of a docs and development burden, but I could see the case for including the single result callback pattern and the listener pattern (since listener was the suggested v4 approach) to ease integration burden and confirm which approach merchants prefer. But that might create a whole mess of docs (you'd need to invoke different methods if you set the listener vs not)

jaxdesmarais: Ah gotcha - that makes sense. We just document 1 for iOS but we can kinda just call the callbacks from the async/await so it's not a ton of extra code lift. Whereas in Android it seems like we'd be adding a ton of extra work for ourselves.

RESULT_PROPOSAL.md
```diff
@@ -0,0 +1,295 @@
...
+
+This approach is the most Kotlin-first pattern. It also resolves the nullability issues. However, 
+the Java integration becomes somewhat complex with casting. This approach aligns with how other 
+payment SDKs handle results.
+
+This approach relies on the Kotlin sealed class, so requires the `PaymentAuthResult` and
+`PaymentResult` objects to live in `BraintreeCore`. It also requires the launchable result 
+objects to all extend a single class (ex: `LaunchableRequest`). This reduces code duplication in 
+our SDK, but requires casting by the merchants (ex: PaymentMethodNonce is returned instead of
```
scannillo, line 167: > requires casting by the merchants

Will this casting still be required once fully converted to Kotlin?

RESULT_PROPOSAL.md
```diff
@@ -0,0 +1,295 @@
...
+    // handle error
+}
+```
+## Single Result Object With Types
+
+This approach is the most Kotlin-first pattern. It also resolves the nullability issues. However, 
+the Java integration becomes somewhat complex with casting. This approach aligns with how other 
+payment SDKs handle results.
+
+This approach relies on the Kotlin sealed class, so requires the `PaymentAuthResult` and
```
scannillo, line 164: > so requires the `PaymentAuthResult` and `PaymentResult` objects to live in `BraintreeCore`

Does this mean the payment-method specific return types will have to live in Core? Like `SEPADirectDebitNonce`?

RESULT_PROPOSAL.md
sarahkoop: Yes but Kotlin has [smart casting](https://kotlinlang.org/docs/typecasts.html#smart-casts) so for Kotlin merchants it's as simple as checking if the object `is PaymentResult.Success` and then it's automatically cast to that type. So the Kotlin code snippet below won't change but it's already relatively clean. 

Java merchants will always have the cumbersome type-checking/unwrapping.

RESULT_PROPOSAL.md
scannillo: @sarahkoop - After our mob PR review, I also vote for this option. We _can_ take a look at the past few inbounds in Slack & GH to see if they are mostly Kotlin (probably). 

Historically we've been more conservative, but I think we're at a point now we can be more aggressive and go for this. We could also introduce it module-by-module and put out beta releases to see if people use it and have feedback.

sarahkoop:
Closing this with decision to go with single result objects

PR #820, opened 2023-11-07
Create FPTI Analytics Event Enums

sarahkoop:
### Summary of changes

 - Prepare for FPTI analytics migration by creating analytics enums with events aligned with iOS for each module
 - These enums will be used in a future PR when FPTI feature is implemented

 ### Checklist

### Authors

- @sarahkoop


AmericanExpress/src/main/java/com/braintreepayments/api/AmericanExpressAnalytics.java
```diff
@@ -0,0 +1,21 @@
...
+    AmericanExpressAnalytics(String event) {
+        this.event = event;
+    }
+
+    String getEvent() {
+        return event;
+    }
+}
+
+
```
jaxdesmarais, line 21: Super nit: it looks like most of the files don't have spaces at the end, is that our preference?

Card/src/main/java/com/braintreepayments/api/CardAnalytics.java
```diff
@@ -0,0 +1,20 @@
+package com.braintreepayments.api;
+
+enum CardAnalytics {
+
+    // Tokenize Events
+    CARD_TOKENIZE_STARTED("card:tokenize:started"),
+    CARD_TOKENIZE_FAILED("card:tokenize:failed"),
+    CARD_TOKENIZE_SUCCEEDED("card:tokenize:succeeded"),
+    CARD_TOKENIZE_NETWORK_CONNECTION_LOST("card:tokenize:network-connection:failed");
```
jaxdesmarais, line 9: I think we actually got rid of this one on BT before GA! Unless there is a specific reason that we want it on Android.

jaxdesmarais (approved):

AmericanExpress/src/main/java/com/braintreepayments/api/AmericanExpressAnalytics.java
```diff
@@ -0,0 +1,21 @@
+package com.braintreepayments.api;
+
+enum AmericanExpressAnalytics {
+
+    // Get Rewards Balance Events
```
scannillo, line 5: The comments in the iOS files are a little inconsistent. [This doc](https://paypal.sharepoint.com/:x:/r/sites/BTSDK/_layouts/15/Doc.aspx?sourcedoc=%7BAAA7C353-5744-4F21-AC7D-DF85D8605817%7D&file=Northstar%20Analytics%20Events.xlsx&action=default&mobileredirect=true&cid=1b0afccd-68ab-448f-9760-60ed14df9811) is our source-of-truth for events. The colored ones (red, blue, green, orange) are the ones we're using to calculate conversion rates, whereas the black ones are superfluous / extra detail we're capturing in case we need to deep-dive on an issue.

scannillo: The intention with the Marks/comments on iOS was to make it clear which 3 or 4 events were the really important ones for "conversion" calculation. I think the [SEPA file](https://github.com/braintree/braintree_ios/blob/main/Sources/BraintreeSEPADirectDebit/BTSEPADirectAnalytics.swift) probably does it best

scannillo:
Referencing [our events doc](https://paypal.sharepoint.com/:x:/r/sites/BTSDK/_layouts/15/Doc.aspx?sourcedoc=%7BAAA7C353-5744-4F21-AC7D-DF85D8605817%7D&file=Northstar%20Analytics%20Events.xlsx&action=default&mobileredirect=true&cid=1b0afccd-68ab-448f-9760-60ed14df9811) again, I wonder if any of these jump out to you as not-trackable on Android? Mostly around the browser-presentation / challenge (CCT) succeeded, failed, and canceled. I know we have limitations on Android to our insights into the CCT.

AmericanExpress/src/test/java/com/braintreepayments/api/AmericanExpressAnalyticsUnitTest.java
```diff
@@ -0,0 +1,18 @@
+package com.braintreepayments.api;
```
scannillo, line 1: Curious what the plan is for adding new files in Java vs Kotlin? Are we waiting to add anymore Kotlin until Post v5?

sarahkoop: We don't have an official plan yet, but these can be in Kotlin since it's cleaner that way - converted them and the tests.

AmericanExpress/src/main/java/com/braintreepayments/api/AmericanExpressAnalytics.java
sarahkoop: My bad - was looking at an iOS PR instead of that doc - updated now. I removed some of the internal comments where all the events are conversion events an clarified others.

Card/src/main/java/com/braintreepayments/api/CardAnalytics.java
sarahkoop: Fixed!

AmericanExpress/src/main/java/com/braintreepayments/api/AmericanExpressAnalytics.java
sarahkoop: No there should be newlines - fixed 👍

sarahkoop:
> Referencing [our events doc](https://paypal.sharepoint.com/:x:/r/sites/BTSDK/_layouts/15/Doc.aspx?sourcedoc=%7BAAA7C353-5744-4F21-AC7D-DF85D8605817%7D&file=Northstar%20Analytics%20Events.xlsx&action=default&mobileredirect=true&cid=1b0afccd-68ab-448f-9760-60ed14df9811) again, I wonder if any of these jump out to you as not-trackable on Android? Mostly around the browser-presentation / challenge (CCT) succeeded, failed, and canceled. I know we have limitations on Android to our insights into the CCT.

I _think_ everything in there should be trackable on Android - but if we want to change any during implementation that's fine too. For cancellation, we can't track the actual cancel from CCT currently, but there is still the "cancel" button on the actual PayPal page that can be tracked (returned as cancel via parsed URL).

AmericanExpress/src/main/java/com/braintreepayments/api/AmericanExpressClient.java
```diff
@@ -36,6 +36,7 @@ public void getRewardsBalance(@NonNull String nonce, @NonNull String currencyIso
                 .appendQueryParameter("currencyIsoCode", currencyIsoCode)
                 .build()
                 .toString();
+        String analytis = AmericanExpressAnalytics.REWARDS_BALANCE_SUCCEEDED.event;
```
scannillo, line 39: Accidental commit?

```suggestion
```

sarahkoop: Ah yes whoops was testing to confirm the Kotlin conversion worked correctly!

scannillo:
> I think everything in there should be trackable on Android - but if we want to change any during implementation that's fine too. For cancellation, we can't track the actual cancel from CCT currently, but there is still the "cancel" button on the actual PayPal page that can be tracked (returned as cancel via parsed URL).

Gotcha! If whatever we _do_ send differs in meaning from iOS, we should update the description/info in our excel doc. Also if events don't apply to one platform vs the other we should make a note of it (for example the alert-cancel ones). Maybe we can add a column with a dropdown like 'iOS/Android/Both' next to the events? I don't mind adding that if we think it would help.

PayPal/src/main/java/com/braintreepayments/api/PayPalAnalytics.kt
```diff
@@ -0,0 +1,14 @@
...
+
+    // Conversion Events
+    TOKENIZATION_STARTED("paypal:tokenize:started"),
+    TOKENIZATION_FAILED("paypal:tokenize:failed"),
+    TOKENIZATION_SUCCEEDED("paypal:tokenize:succeeded"),
+    BROWSER_LOGIN_CANCELED("paypal:tokenize:browser-login:canceled"),
+
+    // Browser Presentation Events
+    BROWSER_SWITCH_SUCCEEDED("paypal:tokenize:browser-presentation:succeeded"),
+    BROWSER_SWITCH_FAILED("paypal:tokenize:browser-presentation:failed")
```
scannillo, line 13: [On iOS (https://github.com/braintree/braintree_ios/blob/74f056ac0600c48bdb2c4908075fcf5580a71a86/Sources/BraintreePayPal/BTPayPalClient.swift#L293-L298) we have a synchronous `bool` return value from ASWeb into if the ASWeb view presented or didn't.

Do we have this insight on Android? A way to know if the CCT displayed or didn't?

sarahkoop: [`BrowserSwitchClient#start`](https://github.com/braintree/browser-switch-android/blob/main/browser-switch/src/main/java/com/braintreepayments/api/BrowserSwitchClient.java#L47) (which launches the CCT) throws a `BrowserSwitchException` in some cases, mainly for improper deep linking configuration, but also if the host activity is finishing while trying to launch CCT. In those cases if we receive that `BrowserSwitchException`, we would send that `BROWSER_SWITCH_FAILED` analytic event on Android. So it's not a 1 to 1 match with the iOS meaning of that event, but I think it's still useful for Android

sarahkoop:
> > I think everything in there should be trackable on Android - but if we want to change any during implementation that's fine too. For cancellation, we can't track the actual cancel from CCT currently, but there is still the "cancel" button on the actual PayPal page that can be tracked (returned as cancel via parsed URL).
> 
> Gotcha! If whatever we _do_ send differs in meaning from iOS, we should update the description/info in our excel doc. Also if events don't apply to one platform vs the other we should make a note of it (for example the alert-cancel ones). Maybe we can add a column with a dropdown like 'iOS/Android/Both' next to the events? I don't mind adding that if we think it would help.

I'll add a comment to each of the analytics implementation tickets to update the source of truth doc with the event description for Android and to note if any don't apply. I think it will be easier to confirm once the events are actually migrated/implemented. Agreed a column for iOS/Android/Both would be good

PayPal/src/main/java/com/braintreepayments/api/PayPalAnalytics.kt
scannillo: Awesome, this is great - thanks for the explanation

GooglePay/src/main/java/com/braintreepayments/api/GooglePayAnalytics.kt
```diff
@@ -0,0 +1,12 @@
+package com.braintreepayments.api
+
+internal enum class GooglePayAnalytics(@JvmField val event: String) {
+
+    PAYMENT_REQUEST_STARTED("google-pay:payment-request:started"),
+    PAYMENT_REQUEST_FAILED("google-pay:payment-request:failed"),
+    PAYMENT_REQUEST_SUCCEEDED("google-pay:payment-request:succeeded"),
+
+    TOKENIZE_STARTED("google-pay:tokenize:started"),
```
scannillo, line 9: Nit - should we add the `// Conversion events` comment here too?

PR #824, opened 2023-11-09
Rename Google Pay APIs

sarahkoop:
### Summary of changes

- Rename Google Pay APIs to `createPaymentAuthRequest` and `tokenize` and rename associated result objects and callbacks
- Cleanup GooglePayClient - use lambdas

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop


sshropshire (approved):

GooglePay/src/main/java/com/braintreepayments/api/GooglePayClient.java
```diff
@@ -200,63 +189,55 @@ public void requestPayment(@NonNull final GooglePayRequest request,
 
         //noinspection ConstantConditions
         if (request == null) {
-            callback.onGooglePayIntentData(null, new BraintreeException(
```
sshropshire, line 203: Random 🤔 : There may be a benefit to having `on<PAYMENT_METHOD>Result()` style callbacks. It would help in the extreme case that a merchant decides to implement multiple payment method listener interfaces on the same class. If they're all named the same it could get confusing. It's more verbose, but for single response objects they'll be shortened to lambdas on Java 8 and up.

sarahkoop: Good point - I'll plan to refactor the naming of these methods in the PRs that move to single result object callbacks to make sure that the naming all makes sense.

jaxdesmarais (approved):

PR #827, opened 2023-11-13
Rename 3DS APIs

sarahkoop:
### Summary of changes

 - Rename 3DS APIs to follow `createPaymentAuthRequest` and `tokenize` pattern
 - Remove internal `continuePerformVerification` method and rename `startVerificationFlow` internal method to `sendAnalyticsAndCallbackResult` (this method can also likely be refactored and removed when analytics are updated).
 - 3DS is slightly different than the other modules because it uses a single `ThreeDSecureResult` object for both steps of the flow, there is also a single callback for those results that is used internally and externally. For now, I left the `ThreeDSecureResult` object naming the same and kept the callback. I've added TODOs to refactor that object and those callbacks in the PR to migrate to a single result object pattern.

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,10 +35,12 @@
     * Add `GooglePayClient#tokenize` 
   * ThreeDSecure
     * Remove `ThreeDSecureListener`
-    * Add `ThreeDSecureLauncher`, `CardinalResult`, and `CardinalResultCallback`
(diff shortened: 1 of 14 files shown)


jaxdesmarais (approved):
🚀

sshropshire (approved):

PR #828, opened 2023-11-13
Venmo Result Objects Refactor

sarahkoop:
### Summary of changes

 - Refactor Venmo results into 3 result object classes and updated callbacks to return single non-null result object 

`VenmoClient#isReadyToPay` -> `VenmoPaymentReadinessResult`
`VenmoClient#createPaymentAuthRequest` -> `VenmoPaymentAuthRequest`
`VenmoClient#tokenize` -> `VenmoResult`

The naming schemes with the second two result objects will be used across modules so please weigh in if we want to make changes to those!

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop


Demo/src/main/java/com/braintreepayments/demo/VenmoFragment.java
```diff
@@ -86,12 +91,12 @@ public void launchVenmo(View v) {
                 lineItems.add(new VenmoLineItem(VenmoLineItem.KIND_DEBIT, "Two Items", 2, "10"));
                 venmoRequest.setLineItems(lineItems);
 
-                venmoClient.createPaymentAuthRequest(requireActivity(), venmoRequest, (venmoAuthChallenge, authError) -> {
-                    if (authError != null) {
-                        handleError(authError);
-                        return;
+                venmoClient.createPaymentAuthRequest(requireActivity(), venmoRequest, (paymentAuthRequest) -> {
+                    if (paymentAuthRequest instanceof VenmoPaymentAuthRequest.Failure) {
```
sarahkoop, line 95: Do we still think `VenmoPaymentAuthRequest` is the best naming or is it too wordy for the Java integration?

Venmo/src/main/java/com/braintreepayments/api/VenmoInternalCallback.java
```diff
@@ -0,0 +1,16 @@
+package com.braintreepayments.api;
+
+import androidx.annotation.Nullable;
+
+/**
+ * Callback for receiving result of
+ * {@link VenmoClient#tokenize(VenmoPaymentAuthResult, VenmoTokenizeCallback)}.
+ */
+interface VenmoInternalCallback {
```
sarahkoop, line 9: Internal and external methods shared the same callback for delivering (nonce, exception) results - I opted to only refactor the public facing callbacks for now, so kept this one and renamed it to internal. Future code cleanup will be to migrate all internal code to use the single result object pattern too.

Venmo/src/main/java/com/braintreepayments/api/VenmoIsReadyToPayCallback.java
```diff
@@ -11,8 +11,7 @@
 public interface VenmoIsReadyToPayCallback {
 
     /**
-     * @param isReadyToPay true if Venmo is ready; false otherwise.
-     * @param error an exception that occurred while checking if Venmo is ready
+     * @param venmoReadinessResult true if Venmo is ready; false otherwise.
      */
-    void onResult(boolean isReadyToPay, @Nullable Exception error);
+    void onVenmoReadinessResult(@NonNull VenmoReadinessResult venmoReadinessResult);
```
sarahkoop, line 16: This PR changes all public callbacks to follow the result method naming pattern of `on<ObjectBeingReturned>Result`. It can provide some code clarity if merchants are creating multiple callbacks (rather than all having the same `onResult` method), but it also feels pretty verbose. What do you think?

Venmo/src/main/java/com/braintreepayments/api/VenmoPaymentAuthRequest.kt
```diff
@@ -0,0 +1,17 @@
...
+
+    /**
+     * The request was successfully created and is ready to be launched by [VenmoLauncher]
+     */
+    class ReadyToLaunch(val requestParams: VenmoPaymentAuthRequestParams) : VenmoPaymentAuthRequest()
+
+    /**
+     * There was an [error] creating the request
+     */
+    class Failure(val error: Exception) : VenmoPaymentAuthRequest()
```
sarahkoop, line 16: I'm not sure if `Failure` feels right for returning a `Request` object - it might give the impression that the request itself failed, but in this case it's that the creating of the request failed. Is this naming pattern clear or is there a better way?

Venmo/src/main/java/com/braintreepayments/api/VenmoReadinessResult.kt
```diff
@@ -0,0 +1,25 @@
...
+    /**
+     * User is not ready to pay with Venmo. Call [VenmoClient.showVenmoInGooglePlayStore] to
+     * allow the user to install the Venmo app.
+     */
+    object NotReadyToPay : VenmoReadinessResult()
+
+    /**
+     * An [error] occurred when determining if the user is ready to pay with Venmo.
+     */
+    class Failure(val error: Exception) : VenmoReadinessResult()
```
sarahkoop, line 22: Do we prefer three result options here (explicitly ready, explicitly not ready, or error), or would it be better to combine not ready and failure? The failure is only if there's an error fetching the config so we can't determine readiness. In that case it would be ReadyToPay() or NotReadyToPay(error?) with an optional error. 

Merchants likely will handle not ready and failure in the same way. But, they do have the option if a user is not ready to pay, to call `showVenmoInCGooglePayStore`. They might not want to call that on the failure case though.

Venmo/src/main/java/com/braintreepayments/api/VenmoResult.kt
```diff
@@ -0,0 +1,22 @@
+package com.braintreepayments.api
+
+/**
+ * Result of tokenizing a Venmo account
+ */
+sealed class VenmoResult {
```
sarahkoop, line 6: I dropped the `Payment` from this result object naming - open to opinions if it should be `VenmoPaymentResult` instead

Demo/src/main/java/com/braintreepayments/demo/VenmoFragment.java
jaxdesmarais: I think it's fine unless you had ideas on another name, though we probably would want to update everything not just Venmo if so.

Venmo/src/main/java/com/braintreepayments/api/VenmoReadinessResult.kt
jaxdesmarais: 🤔 do we think merchants would call `showVenmoInGooglePayStore` with the result or when instantiating the `GooglePayClient` if they have a failure here? If it's with the `GooglePayClient` I would expect we could handle them both in 1 (ready and failure) result as they would want to parse the result in either case. That would give them time to set `showVenmoInGooglePayStore` in either case if desired.

But this is also assuming `showVenmoInGooglePayStore` is somewhere in the Google module and not here so I may be misunderstanding.

Venmo/src/main/java/com/braintreepayments/api/VenmoIsReadyToPayCallback.java
jaxdesmarais: Yeah it feels a bit verbose but imo not _too_ bad. We do have `SEPADirectDebit` all over so not sure anything can top that with verbosity 🙃

sshropshire: In most cases I assume it'll be reduced to a lambda `(venmoReadinessResult) -> {}`. At the same time, it does offer flexibility to merchants who aren't using lambdas.

Venmo/src/main/java/com/braintreepayments/api/VenmoPaymentAuthRequest.kt
```diff
@@ -0,0 +1,17 @@
+package com.braintreepayments.api
+
+/**
+ * A request used to launch the Venmo app for continuation of the Venmo payment flow.
+ */
+sealed class VenmoPaymentAuthRequest {
```
sshropshire, line 6: nit: There's [some interesting interop](https://stackoverflow.com/a/53771710) cases between Kotlin and Java with sealed classes.

Venmo/src/main/java/com/braintreepayments/api/VenmoReadinessResult.kt
jaxdesmarais: Oh! Lol that makes way more sense. Thanks for the explanation. 🙏

Venmo/src/test/java/com/braintreepayments/api/VenmoClientUnitTest.java
```diff
@@ -116,11 +116,13 @@ public void createPaymentAuthRequest_whenCreatePaymentContextFails_collectAddres
...
 
-        verify(venmoPaymentAuthRequestCallback).onPaymentAuthRequest(isNull(), captor.capture());
+        verify(venmoPaymentAuthRequestCallback).onVenmoPaymentAuthRequest(captor.capture());
         verify(braintreeClient).sendAnalyticsEvent("pay-with-venmo.app-switch.failed");
+        VenmoPaymentAuthRequest paymentAuthRequest = captor.getValue();
+        assertTrue(paymentAuthRequest instanceof VenmoPaymentAuthRequest.Failure);
         assertEquals(
                 "Cannot collect customer data when ECD is disabled. Enable this feature in the Control Panel to collect this data.",
-                captor.getValue().getMessage());
+                ((VenmoPaymentAuthRequest.Failure) paymentAuthRequest).getError().getMessage());
```
sshropshire, line 125: Seeing now that the class interop works well in the Java unit test. I'd only heard about the interop before, never seen it in action. 👍

sshropshire (approved):

jaxdesmarais (approved):

PR #830, opened 2023-11-16
Instantiate BraintreeClient Within Clients

sarahkoop:
### Summary of changes

 - Remove `BraintreeClient` public constructors and instead instantiate `BraintreeClient` under the hood within each payment method client
 - Add `ClientParams` object to be configurable with parameters required to instantiate clients (and BraintreeClient)
 - Update client public constructors to take `ClientParams`

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop


sarahkoop:
Open question for this PR: Is this cleaner/preferred over having merchants instantiate `BraintreeClient`? BraintreeClient feels a little bit like exposing unnecessary internals to the merchant (and that was feedback we received), but constructing a `ClientParams` object is essentially the same in code as instantiating a `BraintreeClient`, so this could be unnecessary migration lift for merchants.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -13,6 +13,7 @@ import com.braintreepayments.api.IntegrationType.Integration
  * Core Braintree class that handles network requests.
  */
 @Suppress("LargeClass", "LongParameterList", "TooManyFunctions")
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 open class BraintreeClient @VisibleForTesting internal constructor(
```
scannillo, line 17: Does this change impact DropIn at all?

v5_MIGRATION_GUIDE.md
```diff
@@ -335,9 +345,9 @@ class MyActivity : FragmentActivity() {
 
 
     fun initializeClients() {
-        braintreClient = BraintreeClient(context, "TOKENIZATION_KEY_OR_CLIENT_TOKEN")
+-       braintreClient = BraintreeClient(context, "TOKENIZATION_KEY_OR_CLIENT_TOKEN")
 -       localPaymentClient = LocalPaymentClient(this, braintreeClient)
-+       localPaymentClient = LocalPaymentClient(braintreeClient)
++       localPaymentClient = LocalPaymentClient(ClientParams(context, "TOKENIZATION_KEY_OR_CLIENT_TOKEN")
 -       localPaymentClient.setListener(this)
```
scannillo, line 351: I like this! It's a change we can make on iOS too (in a major version), dropping the merchant step to instantiate a `BTAPIClient`. It always felt redundant to me to have merchants instantiate that class when there was no public functionality on it they would ever use.

BraintreeCore/src/main/java/com/braintreepayments/api/ClientParams.kt
```diff
@@ -0,0 +1,10 @@
+package com.braintreepayments.api
+
+import android.content.Context
+
+data class ClientParams(val context: Context,
```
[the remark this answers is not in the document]
sshropshire: Yeah `ClientParams` is more tech-debt than API. It's internal in `v4`, it can be removed in `v5`.

v5_MIGRATION_GUIDE.md
jaxdesmarais: It'd be great to coordinate this major version alongside the `package` access modifier in Swift 5.9+ ([proposal here](https://github.com/apple/swift-evolution/blob/main/proposals/0386-package-access-modifier.md)) as a while. But totally agree having merchant just access the client like this would be dope:
```swift
// iOS v6
let apiClient = BTAPIClient(authorization: "AUTHORIZATION")
let cardClient = BTCardClient(apiClient: apiClient)

// iOS v7
let cardClient = BTCardClient(authorization:"AUTHORIZATION")
```
Added this to our v7 doc!

BraintreeCore/src/main/java/com/braintreepayments/api/ClientParams.kt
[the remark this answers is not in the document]
sshropshire: On second thought, the [parameter object pattern](https://www.baeldung.com/java-best-practices-many-parameters-method#1-parameter-object) here can be useful. We created too many convenience initializers in `v4` to support multiple integration patterns. With `v5` we do get to wipe the slate clean and have a single constructor, but theoretically a param object like `ClientParams` can help to keep the constructor count low as we add new features.

v5_MIGRATION_GUIDE.md
scannillo: OMG if `package` is approved that will benefit us in so many ways! 🤞

BraintreeCore/src/main/java/com/braintreepayments/api/ClientParams.kt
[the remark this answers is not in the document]
sarahkoop: Good idea I think that's a good compromise here -  I'll go with that

sarahkoop:
Updated this PR to remove the `ClientParams` object and just construct each payment method client with a `context` and `authorization` string directly (which will hopefully align with an iOS v7 approach).

For the overload parameters (currently just browser switching params like `returnUrlScheme`), the plan will be to add those parameters to the payment method request objects themselves (ex: `PayPalRequest.setReturnUrlScheme()`). This is a simpler integration path for merchants as those request objects are already configurable, and it removes the need to make payment method specific parameters available to all clients. There is another ticket to handle moving those browser switch BraintreeClient params to the payment method specific request objects.

Demo/src/main/java/com/braintreepayments/demo/DemoActivity.java
```diff
@@ -46,16 +50,11 @@ protected void onCreate(Bundle savedInstanceState) {
 
     public BraintreeClient getBraintreeClient() {
```
sshropshire, line 51: We should be able to remove the `getBrainreeClient()` eventually. If it can be removed in this PR that would be a nice win. Generally the less `DemoActivity` does the better.

sarahkoop: Good call - removed

v5_MIGRATION_GUIDE.md
```diff
@@ -81,9 +92,9 @@ class MyActivity : FragmentActivity() {
     }
     
     fun initializeClients() {
-        braintreClient = BraintreeClient(context, "TOKENIZATION_KEY_OR_CLIENT_TOKEN")
+-       braintreClient = BraintreeClient(context, "TOKENIZATION_KEY_OR_CLIENT_TOKEN")
 -       venmoClient = VenmoClient(this, braintreeClient)
-+       venmoClient = VenmoClient(braintreeClient)
++       venmoClient = VenmoClient(context, "TOKENIZATION_KEY_OR_CLIENT_TOKEN")
```
scannillo, line 97: 👏 So clean!

SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitClient.java
```diff
@@ -26,11 +26,17 @@ public class SEPADirectDebitClient {
...
+     * @param context an Android Context
+     * @param authorization a Tokenization Key or Client Token used to authenticate
      */
-    public SEPADirectDebitClient(@NonNull BraintreeClient braintreeClient) {
+    public SEPADirectDebitClient(@NonNull Context context, @NonNull String authorization) {
+       this(new BraintreeClient(context, authorization));
+    }
+
+    @VisibleForTesting
+    SEPADirectDebitClient(@NonNull BraintreeClient braintreeClient) {
```
scannillo, line 39: Are we planning to remove these in a follow-up? Or are we keeping for ease of unit testing?

Demo/src/main/java/com/braintreepayments/demo/GooglePayFragment.java
```diff
@@ -57,27 +53,13 @@ public View onCreateView(@NonNull LayoutInflater inflater,
     public void onResume() {
         super.onResume();
 
-        braintreeClient.getConfiguration((configuration, error) -> {
```
scannillo, line 60: Curious - why were we doing config fetches in the demo app?

scannillo (approved):
A few small questions, but LGTM!

Demo/src/main/java/com/braintreepayments/demo/GooglePayFragment.java
sshropshire: Yeah we probably could've deleted this a while back. This has [roots in v3](https://github.com/braintree/braintree_android/blob/3.x/Braintree/src/main/java/com/braintreepayments/api/models/GooglePaymentConfiguration.java#L75). I remember removing it from the Core SDK since it would require a runtime dependency on the Google Pay SDK, but we probably kept it in when we went to GA because we needed to get a release out quickly.

SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitClient.java
sarahkoop: For now they are nice to keep for unit testing since we pass in a `MockBraintreeClient` with responses to various API calls

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
sarahkoop: It shouldn't - drop-in uses the `BraintreeClient` constructor with `BraintreeOptions` param, and has the same "com.braintreepayments.api" group ID as the core libraries, so they all can share these "package-private"/library group methods.

sshropshire (approved):

Demo/src/main/java/com/braintreepayments/demo/BaseFragment.java
```diff
@@ -47,11 +48,15 @@ protected void showDialog(String message) {
...
-    protected BraintreeClient getBraintreeClient() {
+
+    protected void fetchAuthorization(BraintreeAuthorizationCallback callback) {
         DemoActivity demoActivity = getDemoActivity();
         if (demoActivity != null) {
-            return demoActivity.getBraintreeClient();
+            demoActivity.fetchAuthorization(callback);
         }
-        return null;
+    }
```
sshropshire, line 57: 2cents: I always felt like phasing out `BaseFragment` would be the best call. We should be able to fetch client tokens within each feature fragment. Do you feel like it makes sense to remove eventually?

sarahkoop: Agreed - will look into that in another demo refactor PR

PR #831, opened 2023-11-16
Remove ClientTokenProvider

sarahkoop:
### Summary of changes

 - Remove `ClientTokenProvider` and asynchronous loading of authorization in favor of having merchants instantiate clients with authorization directly (see #830).
 - Remove overload `BraintreeClient` constructor and add `authorization` parameter
 - Remove internal `AuthorizationLoader` class
 - The effect of this change for merchants is that if they pass an invalid authentication string to their client constructors, it will be returned as a config error (error fetching configuration, rather than error fetching authorization).
 - Making `BraintreeClient` internal and removing public constructors also has the side effect of making `getConfiguration` private. As far as I can tell, getConfiguration is not publicly available in iOS, so I think this is an ok change.

*Note*: Update base to `v5` after #830 is merged.

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop


GooglePay/src/main/java/com/braintreepayments/api/GooglePayClient.java
```diff
@@ -154,21 +154,16 @@ public void isReadyToPay(@NonNull final FragmentActivity activity,
...
-                        return;
-                    }
-                    callback.onResult(
-                            getTokenizationParameters(configuration, authorization),
-                            getAllowedCardNetworks(configuration));
-                });
-            } else {
+        braintreeClient.getConfiguration((configuration, e) -> {
+            if (configuration == null) {
                 callback.onResult(null, null);
```
sshropshire, line 159: Unrelated, but this is like the only callback with two success parameters and no error parameter. We should be able to fix that for v5 since we can make breaking changes.

sshropshire (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -69,111 +74,26 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      */
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
     constructor(clientParams: ClientParams) :
```
scannillo, line 76: > The effect of this change for merchants is that if they pass an invalid authentication string to their client constructors, it will be returned as a config error (error fetching configuration, rather than error fetching authorization).

On iOS, the `BTAPIClient` (closest to android's `BraintreeClient`) has a [failable init](https://github.com/braintree/braintree_ios/blob/d422ab673680fc8cbf2c6a333f3dc2ebb79fb426/Sources/BraintreeCore/BTAPIClient.swift#L65) which gets triggered if some of the SDKs manual checks on the clientToken after base64 decoding fail.

We _could_ move up that logic in the Android sequence of events to more closely mirror iOS and return an error sooner if the auth is invalid (I think it lives [here](https://github.com/braintree/braintree_android/blob/ed09dabd22f8f594088ce5659a800e7c4309abde/BraintreeCore/src/main/java/com/braintreepayments/api/Authorization.kt#L48-L54)). Total take it or leave it comment - curious your thoughts

sarahkoop: The concept of a failable init doesn't really exist in Kotlin or Java - there are some hacky ways to achieve that behavior but I don't think it would be an expected integration pattern for merchants. We could always throw an error in the `init` method, but that means merchants have to handle the error which I think is less ideal than calling back an error further in the flow. To make it more explicit, we could add a check to the four BraintreeClient methods  that fail for auth (http calls and getConfiguration) if auth is `InvalidAuthorization` and callback an authorization error at the top of those methods so that the errors are more clear. Requires some code duplication but I think that's probably the best merchant experience - what do you think?

sarahkoop: @scannillo I decided to go ahead and callback an auth error in the four BraintreeClient methods that would error on config fetch for invalid auth. In v4 we had a helpful auth error message directing merchants to docs so I think that's a good message to keep.

jaxdesmarais (approved):

PR #834, opened 2023-11-29
Send analytics to FPTI

scannillo:
### Summary

Send existing analytics to FPTI instead of Arachne

### Changes

 - Remove extraction of analytics URL from config
     - Remove `AnalyticsConfiguration.kt`
     - Remove `Configuration.isAnalyticsEnabled`  & `Configuration.analyticsUrl` properties
     - Use hard-coded FPTI analytics URL
- Update metadata/analytics payload params
     - Align with iOS on tags sent (see [FPTIBatchData.swift](https://github.com/braintree/braintree_ios/blob/main/Sources/BraintreeCore/Analytics/FPTIBatchData.swift))
     - Remove no-longer needed tags (userOrientation, dropInVersion, networkType, devicePersistentUUID, isPayPalInstalled, isVenmoInstalled)
- Update POST formatting to match FPTI requirement

### Next

Next, an audit of all current events needs to be done to mirror iOS. This will reduce our number of events sent & increase consistency of when events are sent across platforms.

### Verification

I verified using Looker Event Explorer that our events (and metadata/tags) are successfully being sent to FPTI:

Screenshot 2023-11-29 at 10 36 47 AM

 ### Checklist

 - [X] Added a changelog entry

### Authors
@scannillo


BraintreeCore/src/test/java/com/braintreepayments/api/BraintreeClientUnitTest.kt
```diff
@@ -639,17 +620,19 @@ class BraintreeClientUnitTest {
...
         sut.reportCrash()
 
+        // TODO: - How to make this test wait for the getConfiguration() callback to load first
         verify {
             analyticsClient.reportCrash(
                 applicationContext,
+                any(),
                 "session-id",
                 IntegrationType.CUSTOM,
                 authorization
```
scannillo, line 638: 👋 I need some help from our Android pros on how to properly test `BraintreeClient.reportCrash()` (see source code [here](https://github.com/braintree/braintree_android/blob/aad4455d68263a33f4d8c761672e04c03a0c489c/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt#L431-L441)).

Is there a way with MockK to easily mock the callback from `getConfiguration()` [here](https://github.com/braintree/braintree_android/blob/aad4455d68263a33f4d8c761672e04c03a0c489c/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt#L185)? I couldn't get anything working

scannillo: (Note this test is currently failing)

sarahkoop: It's not pretty, but I got the test working (committed the code). It'll get cleaner when we remove the auth call when #830 is merged.

BraintreeCore/src/main/java/com/braintreepayments/api/DeviceInspector.kt
```diff
@@ -137,18 +117,5 @@ class DeviceInspector @VisibleForTesting internal constructor(
                     "$VENMO_APP_PACKAGE.$VENMO_APP_SWITCH_ACTIVITY"
                 )
             )
-
-        internal fun getDropInVersion(): String? {
```
sarahkoop, line 141: Will we have a way to track the drop in version once these changes go into drop-in? With this it looks like even in a drop-in integration the SDK would report the underlying core SDK version in analytics

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -429,7 +429,15 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      */
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
     fun reportCrash() = authorizationLoader.authorizationFromCache?.let { authorization ->
-        analyticsClient.reportCrash(applicationContext, sessionId, integrationType, authorization)
+        getConfiguration { configuration, _ ->
+            analyticsClient.reportCrash(
+                applicationContext,
+                configuration ?: null,
```
sarahkoop, line 435:
```suggestion
                configuration,
```
If config is null it'll just pass as null 
```suggestion
                configuration,
```

BraintreeCore/src/test/java/com/braintreepayments/api/BraintreeClientUnitTest.kt
scannillo: I see your changes, thank you!

It's seems a little weird to me to do that config fetch when trying to report an app crash (in case resources are limited) but that's the only way we'll get merchantID and env, which seem pretty critical to have in the log. We _could_ not do that config fetch and pass `null` for the config (which again would result in no env or merchantID param in the FPTI payload).

We can leave as-is, but just wanted to call it out in case you felt strongly

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## unreleased
 
+* Move from Braintree to PayPal analytics service (FPTI)
```
sarahkoop, line 5: Should this have a public CHANGELOG line? Seems internal maybe - except is it now required for all merchants to send analytics where it previously was optional?

BraintreeCore/src/main/java/com/braintreepayments/api/DeviceInspector.kt
scannillo: Great catch! On iOS, we aren't yet using iOS v6 (with FPTI) in Drop In - so the standard hasn't yet been set for this.

I can add this back, and then update our [tags doc](https://paypal.sharepoint.com/:x:/r/sites/BTSDK/_layouts/15/Doc.aspx?sourcedoc=%7B6F8E06EE-ECD0-47BD-9380-567AE4D3F2EF%7D&file=FPTI%20Tags%20-%20Mobile%20SDKs.xlsx&action=default&mobileredirect=true) to add `drop_in_sdk_version`. We _could_ re-use `c_sdk_ver` for the DropIn version, as well if you think that is more clear than having 2 separate?

scannillo: Eh - conflating the 2 values into one tag could get confusing

CHANGELOG.md
scannillo: Mostly added it for our own tracking of when the switch was made for easier tracking (also to align w/ iOS)

BraintreeCore/src/main/java/com/braintreepayments/api/DeviceInspector.kt
scannillo: Re-added & updated our tags doc!

BraintreeCore/src/test/java/com/braintreepayments/api/BraintreeClientUnitTest.kt
sarahkoop: Yeah the analytics without merchantID don't seem very valuable so probably fine to leave as is

sarahkoop (approved):

scannillo:
For those curious - we get `networkType` for free when hitting FPTI! The tag is `connctn_type`

PR #837, opened 2023-12-04
Make PostalAddress Data Class

sarahkoop:
### Summary of changes

 - Convert `PostalAddress` to a data class and add Kotlin parcelize plugin

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
@@ -4,9 +4,6 @@ import androidx.annotation.RestrictTo
 import org.json.JSONException
 import org.json.JSONObject
 
-// NEXT MAJOR VERSION: remove 'open' modifiers, Java classes/methods are open by default
-// Ref: https://kotlin-quick-reference.com/102c-R-open-final-classes.html
-
 /**
  * Contains the remote configuration for the Braintree Android SDK.
  *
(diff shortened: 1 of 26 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/PostalAddress.kt
```diff
@@ -1,78 +1,36 @@
...
-    open var locality: String? = null
-    open var region: String? = null
-    open var postalCode: String? = null
-    open var sortingCode: String? = null
-    open var countryCodeAlpha2: String? = null
-
-    constructor()
+@Parcelize
+data class PostalAddress(
+    var recipientName: String?,
```
saperi22, line 12: Do you have a preference on making all data class members to be `val`s?
The data class object would then be immutable and copy() constructor could be used to make changes.

saperi22 (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/PostalAddress.kt
sarahkoop: Hmm in general if it can be a `val` it probably should be... but in this case these parameters should actually be mutable I think. It also lets merchants use `apply` when constructing which seems to be a common pattern. But I could go either way on this one. @sshropshire do you have a preference?

sshropshire: Yeah it's totally for convenience to allow merchants to use `apply`. When we were 100% Java, we used to offer a builder pattern to give a similar Developer Experience to merchants. I personally like the flexibility of it, but I also like immutable types so I'm 50/50.

sshropshire: Last thought: Immutability may be a heavy concept for a dev that's just learning.

sarahkoop: Ok yeah for the `apply` convenience I'm going to leave these as `var`

PR #838, opened 2023-12-04
Configuration Major Version Updates

sarahkoop:
### Summary of changes

 - Resolve NEXT_MAJOR_VERSION notes in `Configuration`

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
@@ -4,9 +4,6 @@ import androidx.annotation.RestrictTo
 import org.json.JSONException
 import org.json.JSONObject
 
-// NEXT MAJOR VERSION: remove 'open' modifiers, Java classes/methods are open by default
-// Ref: https://kotlin-quick-reference.com/102c-R-open-final-classes.html
-
 /**
  * Contains the remote configuration for the Braintree Android SDK.
  *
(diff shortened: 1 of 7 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt
```diff
@@ -4,9 +4,6 @@ import androidx.annotation.RestrictTo
 import org.json.JSONException
 import org.json.JSONObject
 
-// NEXT MAJOR VERSION: remove 'open' modifiers, Java classes/methods are open by default
```
saperi22, line 7: Please correct me if I'm understanding this wrong.
As this is a Kotlin class, removing `open` from `Configuration` will remove the possibility of it being extended by the implementer. Is that the intended change?

saperi22: The comment `Java classes/methods are open by default` on a kotlin class confused me a bit.

sarahkoop: Yes I believe this class was converted from Java to Kotlin in a minor version, and therefore there was the possibility that a merchant was extending this class (which they shouldn't be, but was technically possible when it was a Java class), this class was made `open` to prevent a breaking change. Now that we are making breaking changes we can remove `open` since it shouldn't be extended anyways.

saperi22 (approved):

sshropshire (approved):

PR #839, opened 2023-12-04
Resolve 3DS Next Major Version Notes

sarahkoop:
### Summary of changes

 - Remove implementation of `CardinalValidateReceiver`
 - Make `ThreeDSecureLookup` empty constructor package-private

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@
     * Remove `ThreeDSecureV1UICustomization`
     * Remove `versionRequesed` from `ThreeDSecureRequest` as version 1 is no longer supported
     * Remove `ThreeDSecureV2BaseCustomization`
+    * Remove `CardinalValidateReceiver` from `ThreeDSecureActivity`
+    * Make empty `ThreeDSecureLookup` constructor package-private
   * PayPal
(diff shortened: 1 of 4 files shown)


saperi22 (approved):

sarahkoop:
> Hey @sarahkoop, great job summarizing the changes! Not sure if everyone would benefit from this, but I'd like to understand the `why` behind a change. IMO what changed can be inferred from the PR itself.

Good callout - I can definitely start including the "why" in these PR descriptions! 

For this one, the `CardinalValidateReceiver` is being instantiated when used in the `CardinalChallengeObserver` (this was a change with a previous update in the Cardinal SDK) but we couldn't remove the old way of using the `CardinalValidateReceiver` since it would technically be a breaking change on a public class. Now that we are making breaking changes in this major version, we can go ahead and remove the old implementation.

Similarly, the `ThreeDSecureLookup` empty constructor should only be used internally, but we had accidentally left it public, and it is technically a breaking change to make private, so we are doing that now.

sshropshire (approved):

PR #840, opened 2023-12-04
Analytics Major Version Updates

sarahkoop:
### Summary of changes

 - Remove unecessary `AnalyticsBaseWorker` base class in favor of extending `Worker` where required
 - Remove `open` modifiers on analytics classes that should not be extended, and add library scope restriction
 
Note: Room automatically generated the schema (I believe due to dependency updates) - it might be good to have in case we change things in the future, but I can also remove.

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop


sshropshire (approved):

sarahkoop:
Set the base to `gradle_updates` so I could confirm everything is working with the latest work/room versions, but will be changed to `v5` once that PR is merged.

jaxdesmarais (approved):

PR #842, opened 2023-12-05
Google pay single result

sarahkoop:
### Summary of changes

 - Add single result object patter to all public Google Pay APIs to align with result handling pattern across payment modules

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop


GooglePay/src/main/java/com/braintreepayments/api/GooglePayReadinessResult.kt
```diff
@@ -0,0 +1,19 @@
+package com.braintreepayments.api
+
+/**
+ * Result of [GooglePayClient.isReadyToPay]
+ */
+sealed class GooglePayReadinessResult {
```
sarahkoop, line 6: Still open to other ideas on the naming of this result object

GooglePay/src/main/java/com/braintreepayments/api/GooglePayReadinessResult.kt
```diff
@@ -0,0 +1,19 @@
...
+     * The Google Pay API is supported and set up on this device. Show the Google Pay button for
+     * Google Pay.
+     */
+    object ReadyToPay : GooglePayReadinessResult()
+
+    /**
+     * The Google Pay API is supported or not set up on this device, or there was an [error]
+     * determining readiness.
+     */
+    class NotReadyToPay(val error: Exception?) : GooglePayReadinessResult()
```
sarahkoop, line 18: Decided to combine not ready with any failures fetching readiness - but can split them out into two types if that's preferred.

GooglePay/src/main/java/com/braintreepayments/api/GooglePayReadinessResult.kt
jaxdesmarais: Looking through the Google docs they seem to use "readiness" as well so I think that's fine. 🤷

GooglePay/src/main/java/com/braintreepayments/api/GooglePayClient.java
```diff
@@ -97,25 +97,24 @@ public void isReadyToPay(@NonNull final FragmentActivity activity,
         try {
             Class.forName(PaymentsClient.class.getName());
         } catch (ClassNotFoundException | NoClassDefFoundError e) {
-            callback.onResult(false, null);
+            callback.onGooglePayReadinessResult(new GooglePayReadinessResult.NotReadyToPay(null));
```
sshropshire, line 100:
```suggestion
            callback.onGooglePayReadinessResult(new GooglePayReadinessResult.NotReadyToPay(e));
```
We could forward the cause here, or wrap it in a `BraintreeExcpetion`.

sarahkoop: Good catch

GooglePay/src/main/java/com/braintreepayments/api/GooglePayReadinessResult.kt
sshropshire, line 18:
```suggestion
    class NotReadyToPay(val cause: Exception?) : GooglePayReadinessResult()
```
nit: There is an [Exception constructor](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/lang/Exception.html#Exception(java.lang.Throwable)) `public Exception(Throwable cause)`. We could borrow the name `cause` to signal to merchants that the exception was produced by an underlying cause.

sshropshire: ^ The `cause` property also allows us to chain stack traces. It'll be super useful for GitHub inbounds if we leverage that constructor internally when creating `BraintreeException` instances.

sarahkoop: Updated - do you think we should change to cause for all Failure results or just this one?

sshropshire: We should keep it as `Exception` yeah,  `Exception extends Throwable` so we shouldn't have to change any types. There may not always be a `cause` too, and that's fine `cause == null` in that case.

If we do have an exception that signifies the root cause, then a merchant can inspect `braintreeException?.cause` to give us more info. Also, the `toString()` implementation for `Exception` prints out the combined stack trace of the exception chain.

`new BraintreeException("Error with Config", ioExceptionCausedByConfigFetching)` as an example would allow us to chain the exception returned to the merchant by the SDK to the original root cause exception, like in this example the user's WiFi could  be disabled.

sarahkoop: Ok actually I think the `cause` wording here is confusing - we aren't returning the reason why the user isn't ready to pay (we are just returning NotReadyToPay in those cases). We are only returning an `error` if there was an error determining if the user was ready to pay (in which case it's not really the cause of why they aren't ready to pay, it's that we can't determine if they are ready to pay or not). I'm going to put it back to `error` but I still have a ticket to investigate using cause/throwables

jaxdesmarais (approved):

GooglePay/src/main/java/com/braintreepayments/api/GooglePayReadinessResult.kt
hollabaq86: Yeah I think that's great- we align with Google's naming conventions so it's easier for devs to understand the integration.

Demo/src/main/java/com/braintreepayments/demo/GooglePayFragment.java
```diff
@@ -99,8 +105,14 @@ public void launchGooglePay(View v) {
                 .addAllowedCountryCodes(Settings.getGooglePayAllowedCountriesForShipping(activity))
                 .build());
 
-        googlePayClient.createPaymentAuthRequest(googlePayRequest,
-                (paymentAuthRequest, error) -> googlePayLauncher.launch(paymentAuthRequest));
+        googlePayClient.createPaymentAuthRequest(googlePayRequest, (paymentAuthRequest) -> {
+            if (paymentAuthRequest instanceof GooglePayPaymentAuthRequest.ReadyToLaunch) {
+                googlePayLauncher.launch(
+                        ((GooglePayPaymentAuthRequest.ReadyToLaunch) paymentAuthRequest).getRequestParams());
+            } else if (paymentAuthRequest instanceof GooglePayPaymentAuthRequest.Failure) {
```
hollabaq86, line 112: are there any other possible values for `GooglePayPaymentAuthRequest`? Curious if we need a fallback for scenarios we don't catch.

Demo/src/main/java/com/braintreepayments/demo/GooglePayFragment.java
```diff
@@ -38,11 +42,13 @@ public View onCreateView(@NonNull LayoutInflater inflater,
...
-                            if (error != null) {
-                                handleError(error);
-                            } else {
-                                handleGooglePayActivityResult(paymentMethodNonce);
+                        (googlePayResult) -> {
+                            if (googlePayResult instanceof GooglePayResult.Failure) {
+                                handleError(((GooglePayResult.Failure) googlePayResult).getError());
+                            } else if (googlePayResult instanceof GooglePayResult.Success){
+                                handleGooglePayActivityResult(((GooglePayResult.Success) googlePayResult).getNonce());
+                            } else if (googlePayResult instanceof GooglePayResult.Cancel) {
```
hollabaq86, line 50: should we add an `else` block for other values, or are these the only options from google?

GooglePay/src/main/java/com/braintreepayments/api/GooglePayClient.java
```diff
@@ -170,9 +169,9 @@ public void getTokenizationParameters(
     }
 
     /**
-     * Start the Google Pay payment flow. This will return a {@link GooglePayPaymentAuthRequest} that will
+     * Start the Google Pay payment flow. This will return a {@link GooglePayPaymentAuthRequestParams} that will
```
hollabaq86, line 172:
```suggestion
     * Start the Google Pay payment flow. This will return {@link GooglePayPaymentAuthRequestParams} that are
```

GooglePay/src/main/java/com/braintreepayments/api/GooglePayClient.java
```diff
@@ -170,9 +169,9 @@ public void getTokenizationParameters(
...
      * be used to present Google Pay payment sheet in
```
hollabaq86, line 173:
```suggestion
     * used to present Google Pay payment sheet in
```

GooglePay/src/main/java/com/braintreepayments/api/GooglePayClient.java
```diff
@@ -182,39 +181,38 @@ public void createPaymentAuthRequest(@NonNull final GooglePayRequest request,
         braintreeClient.sendAnalyticsEvent("google-payment.selected");
 
         if (!validateManifest()) {
-            callback.onResult(null, new BraintreeException(
-                    "GooglePayActivity was " + "not found in the " + "Android " +
-                            "manifest, or did not have a theme of R.style.bt_transparent_activity"));
+            callback.onGooglePayPaymentAuthRequest(new GooglePayPaymentAuthRequest.Failure(new BraintreeException(
+                    "GooglePayActivity was not found in the Android " +
+                            "manifest, or did not have a theme of R.style.bt_transparent_activity")));
```
hollabaq86, line 186: this is _oddly_ specific heh. No action, just commenting

GooglePay/src/main/java/com/braintreepayments/api/GooglePayClient.java
```diff
@@ -182,39 +181,38 @@ public void createPaymentAuthRequest(@NonNull final GooglePayRequest request,
...
                 return;
             }
 
             if (!configuration.isGooglePayEnabled()) {
-                callback.onResult(null, new BraintreeException(
-                        "Google Pay " +
-                                "is not enabled for your Braintree account," +
-                                " or Google Play Services are not configured correctly."));
+                callback.onGooglePayPaymentAuthRequest(new GooglePayPaymentAuthRequest.Failure(new BraintreeException(
+                        "Google Pay is not enabled for your Braintree account," +
```
hollabaq86, line 214: is there an Android specific reason why this isn't one big string?

GooglePay/src/main/java/com/braintreepayments/api/GooglePayInternalClient.java
```diff
@@ -18,14 +15,16 @@ void isReadyToPay(FragmentActivity activity, Configuration configuration, IsRead
...
-        paymentsClient.isReadyToPay(isReadyToPayRequest).addOnCompleteListener(new OnCompleteListener<Boolean>() {
-            @Override
-            public void onComplete(@NonNull Task<Boolean> task) {
-                try {
-                    callback.onResult(task.getResult(ApiException.class), null);
-                } catch (ApiException e) {
-                    callback.onResult(false, e);
+        paymentsClient.isReadyToPay(isReadyToPayRequest).addOnCompleteListener(task -> {
+            try {
+                Boolean isReady = task.getResult(ApiException.class);
```
hollabaq86, line 20: Maybe this is some android syntax nuance that I'm missing, but this currently reads like:

"isReady = is there an API Exception?"
if yes, provide readiness result (which can be a failure?) **<-- can I get a confirmation of this expectation?**
otherwise not ready

GooglePay/src/main/java/com/braintreepayments/api/GooglePayReadinessResult.kt
hollabaq86: Let's also document this design decision to include `cause` where we think it's _applicable_, and some loose guidelines around what "applicable" means somewhere internally.

I can see one of us in the future taking a look at the code and thinking "now why did we do this here but that there?", or even an external dev wondering why we did it 😄

hollabaq86 (commented):
Great work on some meaningful cleanup. Got a few Qs and things I'd like to confirm before we merge

GooglePay/src/main/java/com/braintreepayments/api/GooglePayReadinessResult.kt
sarahkoop: I put this back to `error`, which aligns with all other `Failure` results because I think it is more descriptive of what we are actually returning - so no additional design decision doc needed 👍

Demo/src/main/java/com/braintreepayments/demo/GooglePayFragment.java
sarahkoop: These are the only options - the benefit of defining `GooglePayPaymentAuthRequest` as a sealed class is that it is exhaustive, and when we migrate this file to Kotlin, we can use the `when` switch statement without a default/fallback scenario.

Demo/src/main/java/com/braintreepayments/demo/GooglePayFragment.java
sarahkoop: We define the `GooglePayResult` object as a sealed class so these are the only options.

GooglePay/src/main/java/com/braintreepayments/api/GooglePayClient.java
sarahkoop: Yeah... it is getting pretty verbose open to alternative callback naming! The previous pattern was always naming it `onResult` which is a little confusing when the result we are returning is actually a request object

GooglePay/src/main/java/com/braintreepayments/api/GooglePayClient.java
sarahkoop: No that was just weird editor line break - fixed

GooglePay/src/main/java/com/braintreepayments/api/GooglePayInternalClient.java
sarahkoop: This is saying isReady = the result of the task to fetch readiness status (which will be true or false if successfully fetched), but if there is an error executing this task, then throw an APIException.

GooglePay/src/main/java/com/braintreepayments/api/GooglePayReadinessResult.kt
sshropshire: Yeah `cause` is mostly useful for acquiring stack traces from merchants when possible.

Demo/src/main/java/com/braintreepayments/demo/GooglePayFragment.java
sshropshire: > Curious if we need a fallback for scenarios we don't catch.

This is one thing that keeps me up at night lol. Adding a new case to an existing enum can totally break a merchant app. The tradeoff is reduced compiler enforcement in exchange for not breaking a merchant integration.

I do think it's worth calling out in our docs to encourage merchants to catch "UNKNOWN" results so that their integrations are forward compatible.

GooglePay/src/main/java/com/braintreepayments/api/GooglePayInternalClient.java
sshropshire: randomtwocents: The Google Pay API is legit weird IMO. Merchant wins when we wrap! 🌯

sshropshire:
> Let's also document this design decision to include cause where we think it's applicable, and some loose guidelines around what "applicable" means somewhere internally.

It's mostly objective, the standard `Exception` type in Java takes a `cause` property in its constructor, and `BraintreeException extends Exception`. If the exception we're reporting to the merchant has a root cause e.g. CardinalException, BrowserSwitchException, then we can still provide our generic `BraintreeException` message to the merchant like `3D Secure verification failed.`, and then `error.cause == CardinalException`.

Also `BraintreeException.printStackTrace()` will include the OG cardinal exception as well, and we can take that to them to help diagnose the cause of an inbound.

v5_MIGRATION_GUIDE.md
```diff
@@ -143,10 +143,21 @@ class MyActivity : FragmentActivity() {
...
+        
 +       googlePayLauncher = GooglePayLauncher(this) { paymentAuthResult ->
-+            googlePayClient.tokenize(paymentAuthResult) { paymentMethodNonce, error ->
-+                error?.let { /* handle error */ }
-+                paymentMethodNonce?.let { /* handle nonce */ }
++            googlePayClient.tokenize(paymentAuthResult) { googlePayResult ->
++               when (googlePayResult) {
++                   is googlePayResult.Failure -> { /* handle error */ }
++                   is googlePayResult.Cancel -> { /* handle cancel */ }
++                   is googlePayResult.Success -> { /* handle nonce */ }
```
sshropshire, line 159:
```suggestion
+                   is GooglePayResult.Failure -> { /* handle error */ }
+                   is GooglePayResult.Cancel -> { /* handle cancel */ }
+                   is GooglePayResult.Success -> { /* handle nonce */ }
```

v5_MIGRATION_GUIDE.md
```diff
@@ -160,9 +171,12 @@ class MyActivity : FragmentActivity() {
     
     fun onGooglePayButtonClick() {
 -       googlePayClient.requestPayment(activity, request)
-+       googlePayClient.createPaymentAuthRequest(this, request) { paymentAuthRequest, error ->
-+            error?.let { /* handle error */ }
-+            paymentAuthRequest?.let { googlePayLauncher.launch(it) }
++       googlePayClient.createPaymentAuthRequest(this, request) { paymentAuthRequest ->
++           when (paymentAuthRequest) {
++            is paymentAuthRequest.Failure -> { /* handle error */ }
++            is paymentAuthRequest.ReadyToLaunch -> {
```
sshropshire, line 177:
```suggestion
+            is GooglePayPaymentAuthRequest.Failure -> { /* handle error */ }
+            is GooglePayPaymentAuthRequest.ReadyToLaunch -> { 
```
Also would `GooglePayAuthRequest` also work?

sshropshire (approved):

PR #844, opened 2023-12-06
Add Payment Insights Client public API shell

MaxHastingsPP:
### Ticket

https://paypal.atlassian.net/browse/DTBTSDK-3179

### Summary of changes

 - Created `BraintreeShopperInsightsClient` with a hard coded response in `getRecommendedPaymentMethods` implementation, until API is available ( different ticket )
 - Created `ShopperInsightCallback`, `ShopperInsightRequest` and `ShopperInsightResponse` for merchant integration.
 - Created `ShopperInsightsClientUnitTest` to test `getRecommendedPaymentMethods`
 - Added `kotlin-test` test dependency to take advantage of `assertIs`

 ### Checklist

 - [ ] Added a changelog entry

### Authors

- @MaxHastingsPP @saperi22


BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightResult.kt
```diff
@@ -0,0 +1,8 @@
+package com.braintreepayments.api
+
+sealed class ShopperInsightResult {
```
sarahkoop, line 3: I like this pattern a lot, and it is what we are moving towards in `v5`, but it doesn't align with the current result pattern in `v4` (where we use callbacks with dual (result, error) params across all methods/modules). Should we align with `v4` here or go forward with `v5` pattern since this is an alpha feature?

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeShopperInsightsClient.kt
```diff
@@ -0,0 +1,41 @@
+package com.braintreepayments.api
+
+import androidx.annotation.VisibleForTesting
+
+/**
+ * Use `BTShopperInsightsClient` to optimize your checkout experience
```
sarahkoop, line 6:
```suggestion
 * Use [ShopperInsightsClient] to optimize your checkout experience
```

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeShopperInsightsClient.kt
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Retrieves recommended payment methods based on the provided shopper insights request.
+     *
+     * This function takes a `ShopperInsightRequest` object as input and returns a `ShopperInsightResult`.
+     * The result includes flags indicating whether specific payment methods,
+     * such as PayPal or Venmo, are potentially a high priority for the shopper.
+     *
+     * @param request The `ShopperInsightRequest` containing information about the shopper.
+     * @return A `ShopperInsightResult` object indicating the recommended payment methods.
+     */
```
sarahkoop, line 23:
```suggestion
     * Retrieves recommended payment methods based on the provided shopper insights request.
     *
     * @param request The [ShopperInsightRequest] containing information about the shopper.
     * @return A [ShopperInsightResult] object indicating the recommended payment methods.
     */
```

I don't think the extra explanation is necessary since we link to the classes where those details are documented, and we have param documentation. But open to whatever others think. Also I think [] syntax links the classes in kdoc.

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightCallback.kt
```diff
@@ -0,0 +1,5 @@
+package com.braintreepayments.api
+
+interface ShopperInsightCallback {
```
sarahkoop, line 3: Missing docs

BraintreeCore/build.gradle
```diff
@@ -82,6 +82,7 @@ dependencies {
     testImplementation project(':TestUtils')
     testImplementation project(':UnionPay')
     testImplementation project(':Venmo')
+    testImplementation("org.jetbrains.kotlin:kotlin-test:1.8.0")
```
sarahkoop, line 85: Can we put this in the top-level `build.gradle` deps section (open to other ways to centralize dependencies, but that's currently the pattern we follow)? 

Also curious if the existing `kotlin-test-junit` library could be used or if this is preferred?

BraintreeCore/src/test/java/com/braintreepayments/api/ShopperInsightsClientUnitTest.kt
```diff
@@ -0,0 +1,44 @@
+package com.braintreepayments.api
+
+import org.junit.Assert.assertNotNull
+import org.junit.Before
+import org.junit.Test
+import kotlin.test.assertIs
+
+/**
+ * Unit tests for BraintreeShopperInsightsClient.
```
sarahkoop, line 9: Fine to leave but for future work we don't typically add docs for tests (or internal methods unless extra explanation is required).

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightResult.kt
MaxHastingsPP: I would say if this feature is being released with v5, then keep it with v5. But if not, v4 would make sense to be consistent. I will find this out.

sarahkoop:
This work should be merged into a feature branch until the entire feature is complete and ready to be released - can we make a new `payment-insights-complete-feature` or something like that branch and put this PR against that? (`main` should be in a releasable state at all times in case we need to do other bug fixes etc for merchants)

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightRequest.kt
```diff
@@ -0,0 +1,14 @@
+package com.braintreepayments.api
+
+/**
+ * Data class representing a request for shopper insights.
+ *
+ * @property email The email address of the shopper.
+ * @property phoneCountryCode The international country code for the shopper's phone number (e.g., "+1" for the United States).
+ * @property phoneNationalNumber The national segment of the shopper's phone number (excluding the country code).
+ */
+data class ShopperInsightRequest(
```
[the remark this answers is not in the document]
MaxHastingsPP: Makes sense, didn't know that!

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightResult.kt
scannillo: We're targeting to release it in Feb, which will likely be within v4. I think we _could_ have it mirror the V5 patterns though like Sarah mentioned since it will be a beta feature

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -0,0 +1,37 @@
+package com.braintreepayments.api
+
+import androidx.annotation.VisibleForTesting
+
+/**
+ * Use [ShopperInsightsClient] to optimize your checkout experience
+ * by prioritizing the customer’s preferred payment methods in your UI.
+ * By customizing each customer’s checkout experience,
+ * you can improve conversion, increase sales/repeat buys and boost user retention/loyalty.
+ * - Note: This feature is in beta. It's public API may change in future releases.
```
scannillo, line 10: This ` - Note: ` syntax is specific to Swift/Xcode to get the below formatting. Not sure if there's an equivalent in Android


Screenshot 2023-12-06 at 11 33 58 AM

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightRequest.kt
```diff
@@ -0,0 +1,16 @@
+package com.braintreepayments.api
+
+/**
+ * Data class representing a request for shopper insights.
+ *
+ * @property email The email address of the shopper.
+ * @property phoneCountryCode The international country code for the shopper's phone number
+ * (e.g., "+1" for the United States).
+ * @property phoneNationalNumber The national segment of the shopper's phone number
+ * (excluding the country code).
```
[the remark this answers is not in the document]
MaxHastingsPP: Okay yeah to play it safe I removed the +

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightResponse.kt
```diff
@@ -0,0 +1,18 @@
...
+ * This class holds information about the recommended payment methods for a shopper
+ * The recommendations include flags for whether payment methods like PayPal or Venmo
+ * should be displayed with high priority in the user interface.
+ *
+ * @property isPayPalRecommended If true, indicates that the PayPal payment option
+ * should be given high priority in the checkout UI.
+ * @property isVenmoRecommended If true, indicates that the Venmo payment option
+ * should be given high priority in the checkout UI.
+ */
+data class ShopperInsightResponse(
```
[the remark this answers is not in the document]
sarahkoop: `ShopperInsightsResult` as the top level object returned would align with Android `v5`. We don't really use `Response` anywhere in the SDK currently - maybe `ShopperInsightsInfo` or `ShopperInsightsDetails`?

BraintreeCore/src/test/java/com/braintreepayments/api/ShopperInsightsClientUnitTest.kt
```diff
@@ -0,0 +1,44 @@
...
+            phoneNationalNumber = "fake-national-phone"
+        )
+        sut.getRecommendedPaymentMethods(request, object : ShopperInsightCallback {
+            override fun onResult(result: ShopperInsightResult) {
+                assertNotNull(result)
+                val successResult = assertIs<ShopperInsightResult.Success>(result)
+                assertNotNull(successResult.response.isPayPalRecommended)
+                assertNotNull(successResult.response.isVenmoRecommended)
+            }
+        })
```
sshropshire, line 42:
```suggestion
        sut.getRecommendedPaymentMethods(request) { result ->
                assertNotNull(result)
                val successResult = assertIs<ShopperInsightResult.Success>(result)
                assertNotNull(successResult.response.isPayPalRecommended)
                assertNotNull(successResult.response.isVenmoRecommended)
        }
```
^ will be possible after making `ShopperInsightCallback` a `fun interface`.

sshropshire (approved):
Very clean! 🌊

scannillo:
> This work should be merged into a feature branch until the entire feature is complete and ready to be released - can we make a new payment-insights-complete-feature or something like that branch and put this PR against that? (main should be in a releasable state at all times in case we need to do other bug fixes etc for merchants)

I made `payment-insights-feature` for iOS as well as Android to house this. Nitpick, but we could update the base of this branch to keep them aligned.

BraintreeCore/build.gradle
scannillo: +1, ideally we can leverage one of the dependencies we already have imported

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightRequest.kt
```diff
@@ -0,0 +1,31 @@
...
+    var email: String
+)
+
+sealed class ShopperInsightRequest {
+    data class Email(var email: BuyerEmail) : ShopperInsightRequest()
+    data class Phone(
+        var phone: BuyerPhone
+    ) : ShopperInsightRequest()
+    data class EmailAndPhone(
+        var email: BuyerEmail,
```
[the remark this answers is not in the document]
sshropshire: +1 a string would be sufficient here.

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightRequest.kt
```diff
@@ -0,0 +1,31 @@
...
+ * @property phoneCountryCode The international country code for the shopper's phone number
+ * (e.g., "1" for the United States).
+ * @property phoneNationalNumber The national segment of the shopper's phone number
+ * (excluding the country code).
+ */
+
+data class BuyerPhone(
+    var phoneCountryCode: String,
+    var phoneNationalNumber: String
+)
```
sshropshire, line 16:
```suggestion
data class BuyerPhone(
    var countryCode: String,
    var nationalNumber: String
)
```

BraintreeCore/build.gradle
saperi22: It looks like `kotlin-test-junit` builds on top of `kotlin-test`.
I was able to use the existing dependency and it still worked.

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightRequest.kt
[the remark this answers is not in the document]
saperi22: I kind of like the `BuyerEmail` format. Agree that it is only a string, but we could provide a companion object to take the email string and do validation at that point. The validation logic would be scoped to the class and not be littered everywhere we decide to accept email.

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightResult.kt
saperi22: Since this is a generally accepted pattern and easy to understand for the merchant developers, should we keep it as it is (`v5` pattern)?

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightRequest.kt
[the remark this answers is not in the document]
scannillo: We won't be doing any client-side validation on the format of the email, FWIW

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightResult.kt
saperi22: Will be using `v5` pattern as we are moving ShopperInsight to its own module.

jaxdesmarais: Is the plan to also move iOS to a new module @scannillo? I don't think we want to move this to a new module unless we are doing the same on iOS. This will make it easier for merchants to implement features across both platforms.

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightInfo.kt
```diff
@@ -0,0 +1,18 @@
...
+ * This class holds information about the recommended payment methods for a shopper
+ * The recommendations include flags for whether payment methods like PayPal or Venmo
+ * should be displayed with high priority in the user interface.
+ *
+ * @property isPayPalRecommended If true, indicates that the PayPal payment option
+ * should be given high priority in the checkout UI.
+ * @property isVenmoRecommended If true, indicates that the Venmo payment option
+ * should be given high priority in the checkout UI.
+ */
+data class ShopperInsightInfo(
```
scannillo, line 15: Needs beta note in docstring

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightsRequest.kt
```diff
@@ -0,0 +1,57 @@
...
+    fun toJson(): JSONObject = JSONObject().apply {
+            put("countryCode", countryCode)
+            put("nationalNumber", nationalNumber)
+        }
+}
+
+/**
+ * Data class representing a request for shopper insights.
+ */
+sealed class ShopperInsightsRequest {
```
scannillo, line 26: Similar comment re: adding beta note to each docstring

scannillo (approved):

BraintreeCore/src/test/java/com/braintreepayments/api/ShopperInsightsClientUnitTest.kt
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Tests if the getRecommendedPaymentMethods method returns paypal and venmo recommendations
+     * when providing a shopping insight request.
+     */
+    @Test
+    fun testGetRecommendedPaymentMethods_returnsDefaultRecommendations() {
+        val request = ShopperInsightsRequest.Email("fake-email")
+        sut.getRecommendedPaymentMethods(request
+        ) { result ->
```
sshropshire, line 44:
```suggestion
        sut.getRecommendedPaymentMethods(request) { result ->
```

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightsRequest.kt
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    private fun toJson(email: String? = null, phone: BuyerPhone? = null): String {
+        return JSONObject().apply {
+            put(KEY_CUSTOMER, JSONObject().apply {
+                putOpt(KEY_EMAIL, email)
+                putOpt(KEY_PHONE, phone?.toJson())
+            })
+        }.toString()
+    }
```
sshropshire, line 58: I could see these `toJson()` methods moving to `PaymentReadyAPI`. One thing we've found in the past on Braintree is that a JSON representation of a value object can be ambiguous if there is more than one potential transport context e.g. REST vs. GraphQL. The ideal scenario would be to have these value objects remain pure `data class` instances and we would then move the business logic for constructing the HTTP body to the API layer.

BraintreeCore/src/test/java/com/braintreepayments/api/ShopperInsightsClientUnitTest.kt
```diff
@@ -0,0 +1,138 @@
...
+            verify {
+                paymentApi.processRequest(capture(slot))
+            }
+        }
+
+        val customer = JSONObject(slot.captured).getJSONObject(KEY_CUSTOMER)
+        val email = customer.getString(KEY_EMAIL)
+        val phone = customer.getJSONObject(KEY_PHONE)
+        val countryCode = phone.getString(KEY_COUNTRY_CODE)
+        val nationalNumber = phone.getString(KEY_NATIONAL_NUMBER)
```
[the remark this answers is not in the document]
saperi22: This test started out that way.
The json was initially a single line json and `detekt` didn't like it exceeded 120 characters. I had to format the json into multi-line string, that resulted in the test failing because the indentation wasn't correct. After that, I moved to using this.

Testing string equivalency for json is too strict IMO. As long as the key value pairs are correct, the json will result in the same output, so strings are a bit limited.
Now, if we had a deterministic way of converting object to json (eg: gson, moshi etc.) we could then rely on that for converting our object to strings.

Let me know if you still don't agree, and I'll try to find a better way to figure out the object to string conversion.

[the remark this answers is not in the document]
sshropshire: Using gson / moshi would be nice, but we've always had a rule to only use support library (i.e. JetPack) dependencies in our SDK.

In our demo apps we typically use gson / moshi style libraries when communicated with our sample merchant servers, but we take a different approach in the SDK. Historically we've always done this, and it's benefited the SDK and merchants by reducing potential conflicts.

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightsRequest.kt
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Representation of a user phone number.
+ * @property countryCode The international country code for the shopper's phone number
+ * (e.g., "1" for the United States).
+ * @property nationalNumber The national segment of the shopper's phone number
+ * (excluding the country code).
+ */
+
+data class BuyerPhone(
```
sarahkoop, line 13: Can we move each data class to it's own file? I know they are small but we don't typically combine classes in files

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightsRequest.kt
```diff
@@ -0,0 +1,65 @@
...
+ */
+sealed class ShopperInsightsRequest {
+    data class Email(var email: String) : ShopperInsightsRequest()
+    data class Phone(
+        var phone: BuyerPhone
+    ) : ShopperInsightsRequest()
+    data class EmailAndPhone(
+        var email: String,
+        var phone: BuyerPhone
+    ) : ShopperInsightsRequest()
```
sarahkoop, line 41: I'm not sure I understand the purpose of these data classes? Can we just have a ShopperInsightsRequest class with `email: String?`, `phone: BuyerPhone?` properties?

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightsRequest.kt
```diff
@@ -0,0 +1,65 @@
...
+data class BuyerPhone(
+    var countryCode: String,
+    var nationalNumber: String
+) {
+    companion object {
+        internal const val KEY_COUNTRY_CODE = "countryCode"
+        internal const val KEY_NATIONAL_NUMBER = "nationalNumber"
+    }
+
+    fun toJson(): JSONObject = JSONObject().apply {
```
tdchow, line 22: Would this function only be used inside the SDK? Could we make it internal?

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightsRequest.kt
tdchow: +1 

It could be a little confusing to merchants on which class is needed based on what info they want to pass in. I think the nullable fields route may work. We'd just need to do a check to see if both fields are null and handle that case.

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightsRequest.kt
```diff
@@ -0,0 +1,65 @@
...
+    data class Email(var email: String) : ShopperInsightsRequest()
+    data class Phone(
+        var phone: BuyerPhone
+    ) : ShopperInsightsRequest()
+    data class EmailAndPhone(
+        var email: String,
+        var phone: BuyerPhone
+    ) : ShopperInsightsRequest()
+
+    fun toJson(): String {
```
tdchow, line 43: Same here - could we make this function internal?

BraintreeCore/src/test/java/com/braintreepayments/api/ShopperInsightsClientUnitTest.kt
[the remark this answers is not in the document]
saperi22: Moved the conversion to API class.

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightsRequest.kt
sarahkoop: Primarily from a merchant integration standpoint, it's not a pattern we use anywhere else in the SDK for request objects - we typically have configurable request objects with multiple required or optional parameters that can be set via request properties. This pattern also seems pretty limiting for any future configurable parameters (ex: if we add a user name property, then do we need to add every possible combination of email/phone/name as data classes?)

saperi22: > We'd just need to do a check to see if both fields are null and handle that case.
Handling that case would be throw an exception to the merchant who has to look out for it and add more code to handle it.

IMHO this approach is cleaner.

saperi22: > This pattern also seems pretty limiting for any future configurable parameters

Agreed.

tdchow: I agree going the sealed class route is cleaner with not having to handle the null checks, but still feel the interface is not as easy to work with than having the optional email and phone fields.

A merchant may think that multiple `ShopperInsightsRequest` can be passed in if they see `Email` and `Phone` fields. 

I think we could use the existing result callback to notify the merchant of the error due to both of the fields being null. They shouldn't need to add additional logic to handle this specific case.

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightsRequest.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Data class representing a request for shopper insights.
+ *
+ * Note: **This feature is in beta. It's public API may change in future releases.**
+ */
+data class ShopperInsightsRequest(
+    val email: String?,
+    val phone: BuyerPhone?
+) {
+    init {
+        require(email != null || phone != null) {
```
saperi22, line 13: @tdchow @sarahkoop placing the validation inline.

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightsRequest.kt
```diff
@@ -0,0 +1,17 @@
+package com.braintreepayments.api
+
+/**
+ * Data class representing a request for shopper insights.
+ *
+ * Note: **This feature is in beta. It's public API may change in future releases.**
+ */
+data class ShopperInsightsRequest(
+    val email: String?,
```
saperi22, line 9: @sarahkoop we set these as `var`s, right?

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightsRequest.kt
saperi22: changed

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightsRequest.kt
sarahkoop: Yes 👍

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightsRequest.kt
sarahkoop: I think we want to call back an error later in the flow if one of these is null (in whatever method they pass this request object into) - the merchant will already have a callback setup to handle errors for that method, so it's no extra implementation lift on their end. If we throw an error here, they have to have a separate check

BraintreeCore/src/main/java/com/braintreepayments/api/ShopperInsightsRequest.kt
```diff
@@ -0,0 +1,11 @@
+package com.braintreepayments.api
+
+/**
+ * Data class representing a request for shopper insights.
+ *
+ * Note: **This feature is in beta. It's public API may change in future releases.**
+ */
+data class ShopperInsightsRequest(
+    var email: String?,
```
sarahkoop, line 9: Can we add docs for these properties? 

Ex:
email The buyer's email address. If not set, then [phone] is required.

sarahkoop (approved):
One last suggestion for doc strings then LGTM

tdchow (approved):
Looks good! Thanks for making the udpates!

PR #846, opened 2023-12-07
Update Venmo with FPTI Analytics Events

sarahkoop:
### Summary of changes

 - Update analytics events throughout Venmo flow with new FPTI analytics events

 ### Checklist

### Authors

- @sarahkoop

### DIFF
--- a/Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
+++ b/Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
@@ -93,7 +93,8 @@ public class VenmoClient {
                 return;
             }
             if (!deviceInspector.isVenmoAppSwitchAvailable(activity)) {
-                callbackPaymentAuthAppSwitchFailure(callback,
+                braintreeClient.sendAnalyticsEvent(VenmoAnalytics.APP_SWITCH_FAILED);
+                callbackPaymentAuthFailure(callback,
                         new VenmoPaymentAuthRequest.Failure(new AppSwitchNotAvailableException("Venmo is not installed")));
                 return;
             }
@@ -252,11 +253,6 @@ public class VenmoClient {
                 callback.onVenmoReadinessResult(new VenmoReadinessResult.Failure(configError));
             }
         });
-    }
-
-    private void callbackPaymentAuthAppSwitchFailure(VenmoPaymentAuthRequestCallback callback, VenmoPaymentAuthRequest request) {
-        braintreeClient.sendAnalyticsEvent(VenmoAnalytics.APP_SWITCH_FAILED);
-        callback.onVenmoPaymentAuthRequest(request);
     }
 
     private void callbackPaymentAuthFailure(VenmoPaymentAuthRequestCallback callback, VenmoPaymentAuthRequest request) {
--- a/Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
+++ b/Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
@@ -83,23 +83,18 @@ public class VenmoClient {
         braintreeClient.sendAnalyticsEvent(VenmoAnalytics.TOKENIZE_STARTED);
         braintreeClient.getConfiguration((configuration, error) -> {
             if (configuration == null && error != null) {
-                callback.onVenmoPaymentAuthRequest(new VenmoPaymentAuthRequest.Failure(error));
-                braintreeClient.sendAnalyticsEvent(VenmoAnalytics.TOKENIZE_FAILED);
+                callbackPaymentAuthFailure(callback, new VenmoPaymentAuthRequest.Failure(error));
                 return;
             }
 
-            String exceptionMessage = null;
             if (!configuration.isVenmoEnabled()) {
-                exceptionMessage = "Venmo is not enabled";
-                braintreeClient.sendAnalyticsEvent(VenmoAnalytics.TOKENIZE_FAILED);
-            } else if (!deviceInspector.isVenmoAppSwitchAvailable(activity)) {
-                exceptionMessage = "Venmo is not installed";
-                braintreeClient.sendAnalyticsEvent(VenmoAnalytics.APP_SWITCH_FAILED);
+                callbackPaymentAuthFailure(callback,
+                        new VenmoPaymentAuthRequest.Failure(new AppSwitchNotAvailableException("Venmo is not enabled")));
+                return;
             }
-
-            if (exceptionMessage != null) {
-                callback.onVenmoPaymentAuthRequest(
-                        new VenmoPaymentAuthRequest.Failure(new AppSwitchNotAvailableException(exceptionMessage)));
+            if (!deviceInspector.isVenmoAppSwitchAvailable(activity)) {
+                callbackPaymentAuthAppSwitchFailure(callback,
+                        new VenmoPaymentAuthRequest.Failure(new AppSwitchNotAvailableException("Venmo is not installed")));
                 return;
             }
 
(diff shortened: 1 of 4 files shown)


Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -80,25 +80,26 @@ public void showVenmoInGooglePlayStore(@NonNull FragmentActivity activity) {
     public void createPaymentAuthRequest(@NonNull final FragmentActivity activity,
                                          @NonNull final VenmoRequest request,
                                          @NonNull VenmoPaymentAuthRequestCallback callback) {
-        braintreeClient.sendAnalyticsEvent("pay-with-venmo.selected");
+        braintreeClient.sendAnalyticsEvent(VenmoAnalytics.TOKENIZE_STARTED.event);
         braintreeClient.getConfiguration((configuration, error) -> {
             if (configuration == null && error != null) {
                 callback.onVenmoPaymentAuthRequest(new VenmoPaymentAuthRequest.Failure(error));
-                braintreeClient.sendAnalyticsEvent("pay-with-venmo.app-switch.failed");
+                braintreeClient.sendAnalyticsEvent(VenmoAnalytics.TOKENIZE_FAILED.event);
```
sarahkoop, line 87: We used "app-switch" failed more previously, but to be more accurate, I updated this to only be "app switch failed" for errors returned from actual app switching, or if Venmo is not installed.

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -110,7 +111,7 @@ public void createPaymentAuthRequest(@NonNull final FragmentActivity activity,
                 callback.onVenmoPaymentAuthRequest(new VenmoPaymentAuthRequest.Failure(new BraintreeException(
                         "Cannot collect customer data when ECD is disabled. Enable this feature " +
                                 "in the Control Panel to collect this data.")));
-                braintreeClient.sendAnalyticsEvent("pay-with-venmo.app-switch.failed");
+                braintreeClient.sendAnalyticsEvent(VenmoAnalytics.TOKENIZE_FAILED.event);
```
sshropshire, line 114:
```suggestion
                braintreeClient.sendAnalyticsEvent(VenmoAnalytics.TOKENIZE_FAILED);
```
nit: Would it be better to convert `VenmoAnalytics.TOKENIZE_FAILED` to a `String` and remove the `.event` property access?

sshropshire (approved):

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
sshropshire: Yeah I'd go companion `const val` to reduce some of the boilerplate. If we were performing a `switch / when` on it i'd keep it an enum, but for `braintreeClient.sendAnalyticsEvent()` a plain string value is equally sufficient.

sarahkoop: Cool - updated 👍

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -110,7 +111,7 @@ public void createPaymentAuthRequest(@NonNull final FragmentActivity activity,
                 callback.onVenmoPaymentAuthRequest(new VenmoPaymentAuthRequest.Failure(new BraintreeException(
                         "Cannot collect customer data when ECD is disabled. Enable this feature " +
                                 "in the Control Panel to collect this data.")));
-                braintreeClient.sendAnalyticsEvent("pay-with-venmo.app-switch.failed");
+                braintreeClient.sendAnalyticsEvent(VenmoAnalytics.TOKENIZE_FAILED);
```
scannillo, line 114: On BT iOS (and PPCP Android & iOS) we moved to this "notify pattern" (see [BT iOS](https://github.com/braintree/braintree_ios/blob/441f6d903ec8fd027459d124e61abb0b1b99f230/Sources/BraintreeVenmo/BTVenmoClient.swift#L399-L421) & [PPCP Android ](https://github.com/paypal/paypal-android/blob/fe8d82a0174e4bf49aaa2af5b8acd91bf428d459/CardPayments/src/main/java/com/paypal/android/cardpayments/CardClient.kt#L183-L196)examples) for firing analytics. The goal is that each analytics string constant should only be referenced in 1 place in source code. 

The purpose was to reduce errors in our firing of the events, since we have previous SDK version with bugs for forgetting to send analytics in certain cases. This way we know that every time we return a callback result/cancel/error to the merchant, we're firing the right event (and only once).

What do you think of this?

sarahkoop: I think that's a good idea - except for Android since we have two different callbacks for the two different parts of the flow, we'd need 5 analytics helper methods (one success, and then a cancel and failure for each callback) - do you still think that would be helpful?

sarahkoop: On second thought, to make sure we always fire an analytics event with each completion callback, I think it makes sense even if it requires 5 methods

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -270,4 +253,29 @@ public void isReadyToPay(final Context context, final VenmoIsReadyToPayCallback
...
+
+    private void callbackPaymentAuthAppSwitchFailure(VenmoPaymentAuthRequestCallback callback, VenmoPaymentAuthRequest request) {
+        braintreeClient.sendAnalyticsEvent(VenmoAnalytics.APP_SWITCH_FAILED);
+        callback.onVenmoPaymentAuthRequest(request);
+    }
+
+    private void callbackPaymentAuthFailure(VenmoPaymentAuthRequestCallback callback, VenmoPaymentAuthRequest request) {
+        braintreeClient.sendAnalyticsEvent(VenmoAnalytics.TOKENIZE_FAILED);
+        callback.onVenmoPaymentAuthRequest(request);
+    }
```
scannillo, line 265: If the app switch fails, I think we want that to count as a holistic "tokenize_failed", right? For that case on iOS we did [this](https://github.com/braintree/braintree_ios/blob/441f6d903ec8fd027459d124e61abb0b1b99f230/Sources/BraintreeVenmo/BTVenmoClient.swift#L346-L347), and send both a "conversion" event (tokenize_failed) + the "extra details" event (app_switch_failed).

This way we can use the "conversion" event of tokenize_failed to capture all failures, to use to measure against the total # of "start" events.

Though we went back and forth a few times on this with iOS, so definitely down to talk through this and come up with a pattern that makes sense across both.

sarahkoop: Ok agreed I think that makes sense - updated the app switch failure to send both the conversion analytic event and the app switch failure one. I agree with that approach for all the events 👍

Venmo/src/test/java/com/braintreepayments/api/VenmoClientUnitTest.java
```diff
@@ -77,19 +76,6 @@ public void showVenmoInGooglePlayStore_opensVenmoAppStoreURL() {
                 "https://play.google.com/store/apps/details?id=com.venmo");
     }
 
-    @Test
-    public void showVenmoInGooglePlayStore_sendsAnalyticsEvent() {
```
MaxHastingsPP, line 81: Why are we removing this unit test ?

MaxHastingsPP: I see. Should we consider removing this line of code in VenmoClient Line 60?

`braintreeClient.sendAnalyticsEvent("android.pay-with-venmo.app-store.invoked");`

sarahkoop: Ah yes good catch! That was meant to be removed.

scannillo: Wild - this same event slipped through the cracks on iOS too ([see code snippet](https://github.com/braintree/braintree_ios/blob/1528d60c7b6ed12b3433952e855c4f80c7aa4c22/Sources/BraintreeVenmo/BTVenmoClient.swift#L232)). So we actually are _still_ sending it on iOS v6.

Do we think this is a useful event? Should we add it to our events doc?

scannillo: cc: @jaxdesmarais

sarahkoop: Discussed and confirmed we will leave this event out for now 👍

PR #848, opened 2023-12-11
VenmoListener NPE Fix

### LINKED ISSUE #832: Crash NPE on onVenmoFailure listener is null
### Braintree SDK Version

4.33.0

### Environment

Production

### Android Version & Device

_No response_

### Braintree dependencies

com.braintreepayments.api:paypal:4.33.0,
com.braintreepayments.api:venmo:4.33.0,
com.braintreepayments.api:data-collector:4.33.0,

### Describe the bug

`NullPointerException - Attempt to invoke interface method 'void com.braintreepayments.api.i2.a` caused by `VenmoClient.java:305`
The listeners are null at that point, oddly it only happens for failure, it is very rare occasions, something is happening with the activity that we are losing the listeners, we also have set the listeners for paypal the same way, but we have no issue there. When looking into your codebase, we have found that paypal checks for the listener to not be null

PaypalClient.java:155
```
if (error != null && listener != null) {
                    listener.onPayPalFailure(error);
}
```

while for Venmo there are no checks for the listener not being null,
VenmoClient.java:305

```
else if (venmoResult.getError() != null) {
            if (venmoResult.getError() instanceof UserCanceledException) {
                braintreeClient.sendAnalyticsEvent("pay-with-venmo.app-switch.canceled");
            }
            listener.onVenmoFailure(venmoResult.getError());
        }
```

### To reproduce

We couldn't reproduce, it has been happening in production.

### Expected behavior

The fix for this is very simple, only call the listeners for the VenmoClient if they are not null.

### Screenshots

_No response_

sarahkoop:
### Summary of changes

 - Fix NPE in `VenmoClient` when listener is null by checking if listener exists before returning success or failure (does nothing on result without listener, but does not crash) 
 - Fixes #832 

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

### DIFF
--- a/Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
+++ b/Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
@@ -34,8 +34,6 @@ public class VenmoClient {
     static final String EXTRA_USERNAME = "com.braintreepayments.api.EXTRA_USER_NAME";
     static final String EXTRA_RESOURCE_ID = "com.braintreepayments.api.EXTRA_RESOURCE_ID";
 
-    static final String LOGGING_TAG = "Braintree";
-
     private final BraintreeClient braintreeClient;
     private final VenmoApi venmoApi;
     private final VenmoSharedPrefsWriter sharedPrefsWriter;
@@ -313,7 +311,7 @@ public class VenmoClient {
         if (listener != null) {
             listener.onVenmoSuccess(venmoAccountNonce);
         } else {
-            Log.w(LOGGING_TAG, "Unable to deliver result to null listener");
+            Log.w(LoggingUtils.TAG, LoggingUtils.LISTENER_WARNING);
         }
     }
 
@@ -321,7 +319,7 @@ public class VenmoClient {
         if (listener != null) {
             listener.onVenmoFailure(error);
         } else {
-            Log.w(LOGGING_TAG, "Unable to deliver result to null listener");
+            Log.w(LoggingUtils.TAG, LoggingUtils.LISTENER_WARNING);
         }
     }
 
--- a/Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
+++ b/Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
@@ -131,7 +131,7 @@ public class VenmoClient {
         tokenizeVenmoAccount(activity, request, new VenmoTokenizeAccountCallback() {
             @Override
             public void onResult(@Nullable Exception error) {
-                if (error != null ) {
+                if (error != null) {
                     deliverVenmoFailure(error);
                 }
             }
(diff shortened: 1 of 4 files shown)


Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -302,7 +302,19 @@ public void onResult(@Nullable VenmoAccountNonce venmoAccountNonce, @Nullable Ex
             if (venmoResult.getError() instanceof UserCanceledException) {
                 braintreeClient.sendAnalyticsEvent("pay-with-venmo.app-switch.canceled");
             }
-            listener.onVenmoFailure(venmoResult.getError());
+            deliverVenmoFailure(venmoResult.getError());
+        }
+    }
+
+    private void deliverVenmoSuccess(VenmoAccountNonce venmoAccountNonce) {
+        if (listener != null) {
```
jaxdesmarais, line 310: Do we want to handle if the listener is null in either of these cases? Not sure of the best way to do that or even if that'd be helpful to a merchant.

jaxdesmarais (approved):

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
sarahkoop: I'm not really sure there is any way to handle when the listener is null aside from just not crashing - without the listener we can't deliver a result, and updating the method to throw an error would be a breaking change.

jaxdesmarais: 🤔 is there any way for us to output something to the console? Mainly thinking for merchants who do something weird to have anything to point them potentially towards what they may have done wrong.

sarahkoop: Added warning logs here if the listeners are null

sshropshire (approved):

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -33,6 +34,8 @@ public class VenmoClient {
     static final String EXTRA_USERNAME = "com.braintreepayments.api.EXTRA_USER_NAME";
     static final String EXTRA_RESOURCE_ID = "com.braintreepayments.api.EXTRA_RESOURCE_ID";
 
+    static final String LOGGING_TAG = "Braintree";
```
sshropshire, line 37:
```suggestion
    static final String LOGGING_TAG = "Braintree SDK";
```
nit: take or leave. Also I could see us moving this constant to a `Core` file one day.

sarahkoop: Update and moved to a LoggingUtils class 👍

PR #850, opened 2023-12-11
Add Card FPTI Analytics Events

sarahkoop:
### Summary of changes

 - Update card tokenize analytics with new FPTI events

 ### Checklist

### Authors

- @sarahkoop

### DIFF
--- a/Card/src/main/java/com/braintreepayments/api/CardClient.java
+++ b/Card/src/main/java/com/braintreepayments/api/CardClient.java
@@ -62,9 +62,10 @@ public class CardClient {
      * @param callback {@link CardTokenizeCallback}
      */
     public void tokenize(@NonNull final Card card, @NonNull final CardTokenizeCallback callback) {
+        braintreeClient.sendAnalyticsEvent(CardAnalytics.CARD_TOKENIZE_STARTED);
         braintreeClient.getConfiguration((configuration, error) -> {
             if (error != null) {
-                callback.onCardResult(new CardResult.Failure(error));
+                callbackFailure(callback, new CardResult.Failure(error));
                 return;
             }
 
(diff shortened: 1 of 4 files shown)


sshropshire (approved):

Card/src/main/java/com/braintreepayments/api/CardClient.java
```diff
@@ -95,17 +96,22 @@ private void handleTokenizeResponse(JSONObject tokenizationResponse, Exception e
...
-
+                callbackSuccess(callback, new CardResult.Success(cardNonce));
             } catch (JSONException e) {
-                callback.onCardResult(new CardResult.Failure(e));
-                braintreeClient.sendAnalyticsEvent("card.nonce-failed");
+                callbackFailure(callback, new CardResult.Failure(e));
             }
         } else if (exception != null) {
-            callback.onCardResult(new CardResult.Failure(exception));
-            braintreeClient.sendAnalyticsEvent("card.nonce-failed");
```
tdchow, line 108: Were these existing events already being sent to FPTI? Just want to make sure that this change doesn't break any existing reports.

sarahkoop: As far as I know, there aren't any existing reports that are based on the old events, and the new dashboards will be based on these new planned events - but @scannillo correct me if I'm wrong.

tdchow (approved):

Card/src/main/java/com/braintreepayments/api/CardClient.java
scannillo: That's a great question. So on this V5 branch, we already [merged in a change](https://github.com/braintree/braintree_android/pull/834) that sends our analytics directly to FPTI versus sending to the old BT analytics service (Arachne).

The old events weren't being used in any dashboard, but these new events now mirror the ones on BT iOS V6, and will be able to be used in our new conversion dashboard that we share b/w BT & PPCP mobile.

With Android V5, we're doing this analytics cleanup to keep things consistent across platform & product

Card/src/test/java/com/braintreepayments/api/CardClientUnitTest.java
```diff
@@ -45,6 +43,18 @@ public void beforeEach() throws JSONException {
         graphQLDisabledConfig = Configuration.fromJson(Fixtures.CONFIGURATION_WITHOUT_ACCESS_TOKEN);
     }
 
+
```
scannillo, line 46:
```suggestion
```

scannillo (approved):

PR #853, opened 2023-12-12
Create ShopperInsights module and move files there.

saperi22:
### Summary of changes

https://paypal.atlassian.net/browse/DTBTSDK-3265

 - Creating a separate module for Shopper Insights and moving files into it.

 ### Checklist

 - [ ] Added a changelog entry

- @saperi22


ShopperInsights/build.gradle
```diff
@@ -0,0 +1,69 @@
...
+        targetCompatibility versions.javaTargetCompatibility
+    }
+    kotlinOptions {
+        jvmTarget = '1.8'
+    }
+}
+
+dependencies {
+
+    implementation 'androidx.core:core-ktx:1.9.0'
```
saperi22, line 45: cleanup the dependencies

ShopperInsights/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
sarahkoop, line 1: we have one `gitignore` file for the entire project - I don't think we need the module-specific one

ShopperInsights/README.md
```diff
@@ -0,0 +1,11 @@
+# Shopper Insights Module
```
sarahkoop, line 1: We haven't previously had module-specific READMEs - I'm open to it if we think it's useful and are going to do the same on iOS. What do other think about this?

ShopperInsights/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
sarahkoop, line 1: This can be removed if we don't need any proguard rules

ShopperInsights/src/androidTest/java/com/braintreepayments/api/shopperinsights/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import org.junit.Assert.*
+
+/**
+ * Instrumented test, which will execute on an Android device.
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+@RunWith(AndroidJUnit4::class)
+class ExampleInstrumentedTest {
```
sarahkoop, line 17: Let's rename this to the actual class or remove if we aren't going to have instrumentation tests here (sometimes they live in `Demo` module integration tests only

ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/BuyerPhone.kt
```diff
@@ -1,4 +1,4 @@
-package com.braintreepayments.api
+package com.braintreepayments.api.shopperinsights
```
sarahkoop, line 1: Package on all of these should be `com.braintreepayments.api` so that we can leverage package-private classes/methods across the entire SDK (everything lives in `com.braintreepayments.api`)

sarahkoop: To clarify - the directory setup should be 
`braintree_android/ShopperInsights/src/main/java/com/braintreepayments/api/BuyerPhone.kt`

sarahkoop:
This new module will also need to be added to the github actions `release.yml` and `release_snapshot.yml` files, and the detekt config in the top-level `build.gradle` file

ShopperInsights/build.gradle
```diff
@@ -0,0 +1,61 @@
+plugins {
+    id 'com.android.library'
+    id 'org.jetbrains.kotlin.android'
```
sarahkoop, line 3: Let's add the dokka plugin here too so we can have kotlin docs for these new classes

ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/BuyerPhone.kt
saperi22: > Package on all of these should be `com.braintreepayments.api` so that we can leverage package-private classes/methods across the entire SDK (everything lives in `com.braintreepayments.api`)

Good to know.
Package-private works for java code, but how does that look for Kotlin code?
`internal` at a module level, so I'm not sure how the code should look.

ShopperInsights/README.md
saperi22: I don't have a strong preference of where to put it, but I'd like to see a short description of each of the modules and what capabilities they provide to the merchant/user.

jaxdesmarais: I think if we want to move to module specific READMEs this should be done in a separate PR where it's added for all modules vs just one. Though I do feel like a README within a module would be generally difficult to find for merchants. We also typically rely on our public docs to communicate capabilities.

ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/BuyerPhone.kt
sarahkoop: It might not actually apply to any classes in this module currently, since they look like they can all be public or internal currently, but for Kotlin classes that we want to use across modules, we use the `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)` annotation (ex: BraintreeClient).

ShopperInsights/README.md
saperi22: > We also typically rely on our public docs to communicate capabilities.

Do you mind showing me an example?
While there's the overall capabilities of the SDK, there's probably a distinction we want to make where we let the merchants know what exactly they get by including a particular module.

saperi22: I can totally see a root level MODULES.md or something similar where we can briefly describe what the module does and also what one needs to do to create a new module.

sarahkoop (approved):
+1 on removing the README from this PR, but otherwise LGTM

ShopperInsights/README.md
jaxdesmarais: Here is an example of the PayPal public docs: https://developer.paypal.com/braintree/docs/guides/paypal/overview/android/v4. The overview handles the product overview, then we have the client side + feature specific docs for the client side SDKs as well (our team is responsible for writing these). Typically most features also have a server side integration guide (our team doesn't typically write these) as well as a testing/going live section.

I think we tried having a form of docs live in both the repo and publicly on PPCP but it often resulted in one of the two being out of date so never truly having a "source of truth" for our merchants.

tdchow: One option might be to have links on the root readme file to the different module's readme files. This would help with the visibility issue.

If the content on each of the module's readme file is small, it might make more sense in the root readme. This way a developer doesn't need to navigate to see what each module does.

saperi22: > One option might be to have links on the root readme file to the different module's readme files. This would help with the visibility issue.

We already do that. I probably should have called it out to make it more clear.

saperi22:
> +1 on removing the README from this PR, but otherwise LGTM

removed it

hollabaq86 (approved):
🚢

ShopperInsights/README.md
hollabaq86: We need to leverage the public dev docs where possible and avoid additional doc maintenance on our. We're too small and don't have the bandwidth.

If it's common or accepted practice to the Android dev community that there's a README for each individual module, we can include that file with a basic template that points folks to the public dev docs.

Otherwise, let's not adopt this for the SDK.

PR #860, opened 2023-12-15
PayPal with BrowserSwitchLauncher

sarahkoop:

### Summary of changes

 - 

 ### Checklist

 - [ ] Added a changelog entry

-


sarahkoop:
See #883 for new approach

PR #857, opened 2023-12-14
Single Result Objects for Public Methods

sarahkoop:
### Summary of changes

 - Update remaining public methods (American Express, Data Collector, Visa Checkout, Google Pay) to return single result objects to callbaks

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

PR #862, opened 2023-12-15
Google Pay Public Methods Single Result

sarahkoop:
### Summary of changes

 - Break out Google Pay changes from #857 
 - Update additional Google Pay public method with single result object

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop

### DIFF
--- a/GooglePay/src/main/java/com/braintreepayments/api/GooglePayTokenizationParameters.kt
+++ b/GooglePay/src/main/java/com/braintreepayments/api/GooglePayTokenizationParameters.kt
@@ -1,0 +1,20 @@
+package com.braintreepayments.api
+
+import com.google.android.gms.wallet.PaymentMethodTokenizationParameters
+
+/**
+ * A request used to launch the Venmo app for continuation of the Venmo payment flow.
+ */
+sealed class GooglePayTokenizationParameters {
+
+    /**
+     * The request was successfully created and is ready to be launched by [VenmoLauncher]
+     */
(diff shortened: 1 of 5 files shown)


GooglePay/src/main/java/com/braintreepayments/api/GooglePayTokenizationParameters.kt
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * A request used to launch the Venmo app for continuation of the Venmo payment flow.
+ */
+sealed class GooglePayTokenizationParameters {
+
+    /**
+     * The request was successfully created and is ready to be launched by [VenmoLauncher]
+     */
+    class Success(val parameters: PaymentMethodTokenizationParameters, val allowedCardNetworks:
```
tdchow, line 13: This class might read a little easier with the vals on separate lines:

```kotlin
class Success(
    val parameters: PaymentMethodTokenizationParameters,
    val allowedCardNetworks: Collection<Integer>
) : GooglePayTokenizationParameters()
```

tdchow (approved):

sshropshire (approved):

GooglePay/src/main/java/com/braintreepayments/api/GooglePayTokenizationParameters.kt
sarahkoop: Fixed 👍

PR #866, opened 2023-12-19
Send FPTI Analytics Events in American Express

sarahkoop:
### Summary of changes

 - Update `AmericanExpressClient` to send FPTI analytics events
 - Update `AmericanExpressAnalytics` to be a simple object rather than enum

 ### Checklist

### Authors

- @sarahkoop

### DIFF
--- a/AmericanExpress/src/main/java/com/braintreepayments/api/AmericanExpressClient.java
+++ b/AmericanExpress/src/main/java/com/braintreepayments/api/AmericanExpressClient.java
@@ -55,16 +55,23 @@ public class AmericanExpressClient {
                 try {
                     AmericanExpressRewardsBalance rewardsBalance =
                             AmericanExpressRewardsBalance.fromJson(responseBody);
-                    callback.onAmericanExpressResult(new AmericanExpressResult.Success(rewardsBalance));
-                    braintreeClient.sendAnalyticsEvent(AmericanExpressAnalytics.REWARDS_BALANCE_SUCCEEDED);
+                    callbackSuccess(new AmericanExpressResult.Success(rewardsBalance), callback);
                 } catch (JSONException e) {
-                    braintreeClient.sendAnalyticsEvent(AmericanExpressAnalytics.REWARDS_BALANCE_FAILED);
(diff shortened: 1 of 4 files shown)


AmericanExpress/src/main/java/com/braintreepayments/api/AmericanExpressClient.java
```diff
@@ -49,21 +49,21 @@ public void getRewardsBalance(@NonNull String nonce, @NonNull String currencyIso
...
+                    braintreeClient.sendAnalyticsEvent(AmericanExpressAnalytics.REWARDS_BALANCE_SUCCEEDED);
                 } catch (JSONException e) {
-                    braintreeClient.sendAnalyticsEvent("amex.rewards-balance.parse.failed");
+                    braintreeClient.sendAnalyticsEvent(AmericanExpressAnalytics.REWARDS_BALANCE_FAILED);
                     callback.onAmericanExpressResult(new AmericanExpressResult.Failure(e));
                 }
             } else if (httpError != null) {
                 callback.onAmericanExpressResult(new AmericanExpressResult.Failure(httpError));
-                braintreeClient.sendAnalyticsEvent("amex.rewards-balance.error");
+                braintreeClient.sendAnalyticsEvent(AmericanExpressAnalytics.REWARDS_BALANCE_FAILED);
```
sarahkoop, line 66: I know in other modules we are refactoring to have a single `success` and `error` method for callbacks and analytics to confirm we always send analytics with the callback. That seemed a little overkill in this module since there is only one method, but I can add those in if we think it's still beneficial

scannillo (approved):

AmericanExpress/src/main/java/com/braintreepayments/api/AmericanExpressClient.java
scannillo: We did it on iOS, just for consistency - but whatever you think works on Android 👍

sarahkoop: Updated for consistency too 👍

tdchow (approved):

PR #869, opened 2023-12-20
Shopper Insights - Merchant analytics

saperi22:
### Summary of changes
https://paypal.atlassian.net/browse/DTBTSDK-3296

 - Adding methods for merchants to send analytic events for shopper insights feature.

 ### Checklist

 - [ ] Added a changelog entry

- @saperi22

### DIFF
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsAnalytics.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsAnalytics.kt
@@ -1,8 +1,8 @@
 package com.braintreepayments.api
 
 sealed class ShopperInsightsAnalytics(val event: String) {
-    object PayPalPresented: ShopperInsightsAnalytics("shopper-insights:paypal-presented")
-    object PayPalSelected: ShopperInsightsAnalytics("shopper-insights:paypal-selected")
-    object VenmoPresented: ShopperInsightsAnalytics("shopper-insights:venmo-presented")
-    object VenmoSelected: ShopperInsightsAnalytics("shopper-insights:venmo-selected")
+    object PayPalPresented : ShopperInsightsAnalytics("shopper-insights:paypal-presented")
+    object PayPalSelected : ShopperInsightsAnalytics("shopper-insights:paypal-selected")
+    object VenmoPresented : ShopperInsightsAnalytics("shopper-insights:venmo-presented")
+    object VenmoSelected : ShopperInsightsAnalytics("shopper-insights:venmo-selected")
 }
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsAnalytics.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsAnalytics.kt
@@ -1,0 +1,8 @@
+package com.braintreepayments.api
+
+sealed class ShopperInsightsAnalytics(val event: String) {
+    object PayPalPresented: ShopperInsightsAnalytics("shopper-insights:paypal-presented")
+    object PayPalSelected: ShopperInsightsAnalytics("shopper-insights:paypal-selected")
+    object VenmoPresented: ShopperInsightsAnalytics("shopper-insights:venmo-presented")
+    object VenmoSelected: ShopperInsightsAnalytics("shopper-insights:venmo-selected")
+}
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsAnalytics.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsAnalytics.kt
@@ -1,8 +1,8 @@
 package com.braintreepayments.api
 
-sealed class ShopperInsightsAnalytics(val event: String) {
-    object PayPalPresented : ShopperInsightsAnalytics("shopper-insights:paypal-presented")
-    object PayPalSelected : ShopperInsightsAnalytics("shopper-insights:paypal-selected")
-    object VenmoPresented : ShopperInsightsAnalytics("shopper-insights:venmo-presented")
-    object VenmoSelected : ShopperInsightsAnalytics("shopper-insights:venmo-selected")
+internal object ShopperInsightsAnalytics {
+    const val PAYPAL_PRESENTED = "shopper-insights:paypal-presented"
+    const val PAYPAL_SELECTED = "shopper-insights:paypal-selected"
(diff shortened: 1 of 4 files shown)


ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsAnalytics.kt
```diff
@@ -0,0 +1,8 @@
+package com.braintreepayments.api
+
+sealed class ShopperInsightsAnalytics(val event: String) {
```
tdchow, line 3: I think we're moving away from a sealed class and having analytics strings as string constants within an `object`.

Recent PR with this change: https://github.com/braintree/braintree_android/pull/867/files#diff-fa9e940f487fd9297512740b520d3e1936bb71c7ca35f1ae96a20c3f0316501e

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -47,4 +48,36 @@ class ShopperInsightsClient @VisibleForTesting internal constructor(
             )
         )
     }
+
+    /**
+     * Call this method when the PayPal button has been successfully displayed to the buyer.
+     * This method sends analytics to help improve the Shopper Insights feature experience.
+     */
+    fun sendPayPalPresentedEvent() {
```
tdchow, line 56: Do we have a way to ensure that merchants invoke these analytics functions? Not sure if there's a way when the merchant owns the PayPal/Venmo button 🤔

saperi22: https://github.com/braintree/braintree_ios/pull/1162
I pretty much emulated the iOS PR.

> Do we have a way to ensure that merchants invoke these analytics functions?

I had the same question, but pretty sure we are at their mercy. I don't see how we can do client side analytics when they own the UI.

jaxdesmarais: ➕ 1 yeah we are adding these in hopes that merchants actually send the events but can't really force merchants to do anything with their own UI like Sai said.

tdchow (approved):

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsAnalytics.kt
```diff
@@ -0,0 +1,8 @@
+package com.braintreepayments.api
+
+internal object ShopperInsightsAnalytics {
+    const val PAYPAL_PRESENTED = "shopper-insights:paypal-presented"
+    const val PAYPAL_SELECTED = "shopper-insights:paypal-selected"
+    const val VENMO_PRESENTED = "shopper-insights:venmo-presented"
+    const val VENMO_SELECTED = "shopper-insights:venmo-selected"
```
jaxdesmarais, line 7: Not sure if it'll be a follow up PR but I thought I had heard we don't need to care about Venmo anymore for this feature. Totally fine to 🏈 on it if we are still waiting to get clarity, but just didn't want to add events if we already know Venmo will be removed.

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -1,6 +1,10 @@
 package com.braintreepayments.api
 
 import androidx.annotation.VisibleForTesting
+import com.braintreepayments.api.ShopperInsightsAnalytics.PAYPAL_PRESENTED
```
jaxdesmarais, line 4: This may be my lack of Android knowledge but can we just do `import com.braintreepayments.api.ShopperInsightsAnalytics` or do we need to import each event individually?

sarahkoop: In other modules we have been fully qualifying the events when used (ex: on line 61 `braintreeClient.sendAnalyticsEvent(ShopperInsightsAnalytics.PAYPAL_PRESENTED)`) that being said, either way is fine with me

tdchow: Yep, we could use a wildcard import here `import com.braintreepayments.api.ShopperInsightsAnalytics.*`, but Google is opinionated about import statements. https://developer.android.com/kotlin/style-guide#import_statements

I think there are pros and cons to both ways, but Android Studio is pretty good at formatting imports and auto imports.

jaxdesmarais: Nice, thanks for sending over that doc!

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsAnalytics.kt
saperi22: I not 100%, but we will likely have Venmo.
I can DM you the exact thing I heard in the meeting yesterday.

jaxdesmarais: Nah no worries - I trust you! You're more keyed into the going ons with this feature than I am. 😄

jaxdesmarais (approved):

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
saperi22: Some of the cons of using wildcard imports might no longer apply to kotlin, as it gives us a way to use import alias as shown below, but the tools and practices are slower to catch up to the capabilities of the newer languages, so we just go with whatever the tools suggest :D

But yeah, no reason not to use wild card imports in this current scenario.

```
import org.example.Message // Message is accessible
import org.test.Message as TestMessage // TestMessage stands for 'org.test.Message'
```
See: https://kotlinlang.org/docs/packages.html#imports

saperi22: If you are interested, here's how things could go wrong using wildcard imports.
https://www.javadude.com/posts/20040522-import-on-demand-is-evil/

sshropshire: There's a way to turn off autogen of `import *` in Android Studio in preferences, I forget exactly where.

PR #871, opened 2023-12-26
Adding imageUrl, upcType, upcCode to PayPalLineItem

vankads:

### Summary of changes

 - Added the ability for merchants to pass imageUrl, upcCode, upcType in PayPalLineItem.

 ### Checklist

 - [x] Added a changelog entry

- @vankads

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,7 @@
 # unreleased
 
 * BraintreePayPal
-  * Add following properties to PayPalLineItem
-    * imageUrl
-    * upcCode
-    * upcType
+  * Add imageUrl, upcCode, and upcType to PayPalLineItem
 
 ## 4.40.1 (2023-12-13)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Braintree Android SDK Release Notes
 
-# unreleased
+## unreleased
 
 * PayPal
   * Add imageUrl, upcCode, and upcType to PayPalLineItem
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 # unreleased
 
-* BraintreePayPal
+* PayPal
   * Add imageUrl, upcCode, and upcType to PayPalLineItem
 
 ## 4.40.1 (2023-12-13)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Braintree Android SDK Release Notes
 
+# unreleased
+
+* BraintreePayPal
+  * Add following properties to PayPalLineItem
+    * imageUrl
+    * upcCode
+    * upcType
+
 ## 4.40.1 (2023-12-13)
 
 * BraintreeCore
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalLineItem.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalLineItem.java
@@ -29,6 +29,22 @@ public class PayPalLineItem implements Parcelable {
     public static final String KIND_CREDIT = "credit";
     public static final String KIND_DEBIT = "debit";
 
+    /**
+     * The upc type of PayPal line item.
+     */
+    @Retention(RetentionPolicy.SOURCE)
+    @StringDef({PayPalLineItem.UPC_TYPE_A, PayPalLineItem.UPC_TYPE_B, PayPalLineItem.UPC_TYPE_C, PayPalLineItem.UPC_TYPE_D, PayPalLineItem.UPC_TYPE_E, PayPalLineItem.UPC_TYPE_2, PayPalLineItem.UPC_TYPE_5})
+    @interface PayPalLineItemUpcType {
+    }
+
+    public static final String UPC_TYPE_A = "UPC-A";
+    public static final String UPC_TYPE_B = "UPC-B";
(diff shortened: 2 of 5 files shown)


PayPal/src/main/java/com/braintreepayments/api/PayPalLineItem.java
```diff
@@ -178,6 +227,21 @@ public String getUrl() {
         return url;
     }
 
+    @Nullable
```
sshropshire, line 230:
```suggestion
    @NonNull
```
The getter and setter annotations need to match. Not sure why, but I remember running into this issue a while back.

PayPal/src/main/java/com/braintreepayments/api/PayPalLineItem.java
```diff
@@ -137,6 +159,33 @@ public void setUrl(@NonNull String url) {
...
+    public void setUpcCode(@NonNull String upcCode) {
+        this.upcCode = upcCode;
+    }
+
+    /**
+     * UPC type of the item.
+     *
+     * @param upcType The UPC type.
+     */
+    public void setUpcType(@NonNull String upcType) {
```
sshropshire, line 185:
```suggestion
    public void setUpcType(@Nullable String upcType) {
```
I'd actually go for `@Nullable` for these properties to align the getter/setter annotations. `upcType: String?` makes sense since it's initially `null`.

PayPal/src/main/java/com/braintreepayments/api/PayPalLineItem.java
vankads: These properties are optional fields. So, we can leave it as @Nullable. Let me know if you think otherwise.

ragavi98 (commented):
Looks good to me! Would approve if I could

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Braintree Android SDK Release Notes
 
+# unreleased
+
+* BraintreePayPal
+  * Add following properties to PayPalLineItem
+    * imageUrl
+    * upcCode
+    * upcType
```
sarahkoop, line 9:
```suggestion
* PayPal
  * Add `imageUrl`, `upcCode`, and `upcType` to `PayPalLineItem`
```
Nitpick but keeps the changelog a little more concise

PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutLineItem.java
```diff
@@ -137,6 +159,33 @@ public void setUrl(@NonNull String url) {
...
+    public void setUpcCode(@NonNull String upcCode) {
+        this.upcCode = upcCode;
+    }
+
+    /**
+     * UPC type of the item.
+     *
+     * @param upcType The UPC type.
+     */
+    public void setUpcType(@NonNull String upcType) {
```
sarahkoop, line 185:
```suggestion
    public void setUpcType(@NonNull @PayPalLineItemUpcType String upcType) {
```

This type should be added here and in the getter right (and in doc string too probably)? Or is that interface used somewhere else?

PayPal/src/main/java/com/braintreepayments/api/PayPalLineItem.java
vankads: Currently all the existing properties have @NonNull for setters and @Nullable for getters, as these are not required fields and merchant may or may not use setter methods. We didn't see any issues so far.

sarahkoop: I think that must have been an oversight when those properties were added. When these annotations don't match up, the Kotlin property setters don't seem to work as expected (ex: `lineItem.upcType = "some type"` won't work in Kotlin, they would have to set `lineItem.setUpcType("some type")`), which isn't the ideal Kotlin integration experience. Although I'm not certain if newer versions of Kotlin have resolved this issue.

sshropshire: Oh cool yeah if it does work in newer versions of Kotlin i'm 👍 . We should check though. The way it's written now makes total sense to me, but I have seen Kotlin get weird when the annotations aren't aligned.

sarahkoop: I just confirmed with the v5 branch that uses the latest Kotlin that this issue still exists 😞 - I created a ticket to update the annotations in v5 where they don't currently align, but we should add these new properties with matching annotations.

vankads: Updated setter method annotation from `@NonNull` to `@nullable` to keep it consistent with getter method annotations

PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutLineItem.java
vankads: Added interface type to setter and getter annotations

sarahkoop (approved):

sshropshire (approved):
🌊

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Braintree Android SDK Release Notes
 
+# unreleased
+
+* BraintreePayPal
```
sarahkoop, line 5:
```suggestion
* PayPal
```

PR #854, opened 2023-12-13
Add Shopping Insights Demo App Integration

MaxHastingsPP:
### Summary of changes

 - Created Shopping Insights Fragment
 - Created Shopping Insights View Model
 - Created Shopping Insights UI
 - Added Kotlin language support to demo app module
 - Added navigation to shopping insight fragment from main fragment

### Images

Screenshot 2023-12-15 at 10 36 43 AM

 ### Checklist

### Authors

- @MaxHastingsPP

PR #872, opened 2023-12-26
Adding PayPal and Venmo apps installed check

saperi22:
### Summary of changes
DTBTSDK-3182

 - Checking if PayPal or Venmo apps are installed and returning a result based on the installed apps.

 ### Checklist

 - [ ] Added a changelog entry

-


ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -16,18 +17,25 @@ import com.braintreepayments.api.ShopperInsightsAnalytics.VENMO_SELECTED
...
+     * @param context Android context
      * @param request The [ShopperInsightsRequest] containing information about the shopper.
      * @return A [ShopperInsightsResult] object indicating the recommended payment methods.
      */
     fun getRecommendedPaymentMethods(
+        context: Context,
         request: ShopperInsightsRequest,
         callback: ShopperInsightsCallback
     ) {
+        if(checkForInstalledApps(context, callback)) {
```
sshropshire, line 35: qq: Does the API team suggest we skip making a call if we can gather insights on our own?

saperi22: I'm going based off of the ticket.
It looks like we are to skip the check if both venmo and paypal apps are installed. I'm not sure if we should be trying to get any other insights.

sshropshire: Oh I see now my bad, I hadn't seen the ticket. The `Boolean` did jump out to me though, I wonder if we can leverage Kotlin if-statement expressions here. Something like:

```kotlin
val result = if (isPayPalInstalled && isVenmoInstalled) {
  val shopperInsights = ShopperInsightsInfo(isPayPalRecommended = true, isVenmoRecommended = true)
  ShopperInsightsResult.Success(shopperInsights)
} else {
  // validate request logic and fetch shopper insights from API
}
callback.onResult(result)
```
could work? It would allow us to avoid the `guard`-like return statement (and having to forward the callback reference).

saperi22: I could change the logic to use that. Was trying to reduce the nesting and the logic in a single function. :)

saperi22: Hey @sshropshire, while trying to refactor, I felt that the guard style actually makes sense (at least for me) as we are short circuiting the logic in a couple distinct ways; one is for the app installed checks and the other is for when the request params are all null. Not sure if I want to force them both in an if else block.

Let me know if you still strongly prefer not having guard style block and I'll change the logic again.

sshropshire: That works. I usually say go for what feels right and maximizes readability for someone who's looking at the code for the first time.

scannillo:
Have the [Demo app changes](https://github.com/braintree/braintree_android/pull/854) been merged into this branch yet?

These changes look good to me, but we'll need to update the callsite of `getRecommendedPaymentMethods()` to take in the context

tdchow (approved):

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -16,18 +17,43 @@ import com.braintreepayments.api.ShopperInsightsAnalytics.VENMO_SELECTED
...
     fun getRecommendedPaymentMethods(
+        context: Context,
         request: ShopperInsightsRequest,
         callback: ShopperInsightsCallback
     ) {
+        val applicationContext = context.applicationContext
+        val isVenmoAppInstalled = deviceInspector.isVenmoInstalled(applicationContext)
+        val isPayPalAppInstalled = deviceInspector.isPayPalInstalled(applicationContext)
+
+        if (isVenmoAppInstalled && isPayPalAppInstalled) {
```
sarahkoop, line 45: I'm confused why this is an && statement - can you provide more context on why we only want to return both PayPal and Venmo recommended together when both apps are installed, and not one or the other?

ShopperInsights/src/test/java/com/braintreepayments/api/ShopperInsightsClientUnitTest.kt
```diff
@@ -35,18 +43,32 @@ class ShopperInsightsClientUnitTest {
...
+        sut.getRecommendedPaymentMethods(context, request) { result ->
             assertNotNull(result)
             val successResult = assertIs<ShopperInsightsResult.Success>(result)
             assertNotNull(successResult.response.isPayPalRecommended)
             assertNotNull(successResult.response.isVenmoRecommended)
         }
     }
 
+    @Test
+    fun testGetRecommendedPaymentMethods_hasBothAppsInstalled_returnsSuccessResult() {
```
sarahkoop, line 55: Can we add a test for the negative case here too?

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
saperi22: I'm not sure. That's what the ticket said.
I'll have to ask @scannillo to shed some context there.

ShopperInsights/src/test/java/com/braintreepayments/api/ShopperInsightsClientUnitTest.kt
saperi22: I've updated the other tests to test the negative cases.

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
sarahkoop: Ah ok I see baed on the ticket description this looks good 👍

ShopperInsights/src/test/java/com/braintreepayments/api/ShopperInsightsClientUnitTest.kt
```diff
@@ -17,36 +20,75 @@ import org.junit.Test
...
-    fun testGetRecommendedPaymentMethods_returnsDefaultRecommendations() {
+    fun testGetRecommendedPaymentMethods_noInstalledApps_returnsDefaultRecommendations() {
+        every { deviceInspector.isVenmoInstalled(applicationContext) } returns false
+        every { deviceInspector.isPayPalInstalled(applicationContext) } returns false
+
+        val request = ShopperInsightsRequest("fake-email", null)
+        sut.getRecommendedPaymentMethods(context, request) { result ->
+            assertNotNull(result)
+            val successResult = assertIs<ShopperInsightsResult.Success>(result)
+            assertNotNull(successResult.response.isPayPalRecommended)
```
sarahkoop, line 52: Let's assert false here for now, since that is what the API is hard coded to do - and then we won't forget to update these tests based on the actual results in the future.

sarahkoop (approved):

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
scannillo: Basically, we'll only want to skip the PaymentReady API call if _both_ Venmo & PayPal apps are installed. The logic should be more clear once that API call is added.

[For now on iOS](https://github.com/braintree/braintree_ios/pull/1167/files), I wrote it like this:

```
        if isVenmoAppInstalled() && isPayPalAppInstalled() {
            return BTShopperInsightsResult(isPayPalRecommended: true, isVenmoRecommended: true)
        } else {
            // TODO: - Make API call to PaymentReadyAPI. DTBTSDK-3176
            return BTShopperInsightsResult()
        }
```

There will prob be even further logic after the API call to prefer the `true` value from the app installed check even if the API call returns `false`.

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -16,18 +17,43 @@ import com.braintreepayments.api.ShopperInsightsAnalytics.VENMO_SELECTED
...
+                    ShopperInsightsInfo(
+                        isPayPalRecommended = true,
+                        isVenmoRecommended = true
+                    )
+                )
+            )
+            return
+        }
+
         if (request.email == null && request.phone == null) {
```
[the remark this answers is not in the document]
saperi22: Good question.
My line of thought was that we had sufficient information to give the merchants a result they need to display the buttons, so we'd return the result.

The focus was on the final result rather than the API consistency.
I don't know if we need to prefer one way or the other. Do we need to consult with product for a better understanding of how this needs to work?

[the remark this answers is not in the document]
scannillo: Nah - I think we can make this call ourselves. Whatever we think delivers the best merchant dev experience.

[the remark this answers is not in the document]
saperi22: what does iOS do in this case?

MaxHastingsPP (approved):

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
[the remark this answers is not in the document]
saperi22: Changed logic to do the validation first.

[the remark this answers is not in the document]
scannillo: On iOS, we require the email or phone in the init of the request obj, so we don't have to do the validation in the same was as Android

PR #880, opened 2024-01-08
Shopping Insights Default Values

MaxHastingsPP:
### Ticket

https://paypal.atlassian.net/browse/DTBTSDK-3344

Shopping Insight api postman collection can be found in jira ticket for stage testing.

### Summary of changes

- Adding the new request and response objects for shopping insights api. Adding default values.

- Decoupling the shopping insights create body json into a separate class from the api call.

### Authors
- @MaxHastingsPP


ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightApiResult.kt
```diff
@@ -0,0 +1,34 @@
+/**
+ * Represents the result from the Shopper Insight API.
+ *
+ * @property eligible_methods Contains the payment methods available to the shopper.
+ */
+data class ShopperInsightApiResult (
+    val eligible_methods: PaymentMethods
```
saperi22, line 7: do we want to use @SerializedName and keep the property name camel case?

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -66,18 +69,34 @@ class ShopperInsightsClient @VisibleForTesting internal constructor(
...
+                countryCode = countryCode
+            )
+        )
         // Hardcoded result
         callback.onResult(
             ShopperInsightsResult.Success(
                 ShopperInsightsInfo(
-                    isPayPalRecommended = false,
-                    isVenmoRecommended = false
+                    isPayPalRecommended = isPaymentRecommended(result.eligible_methods.paypal),
```
saperi22, line 89: please update the tests for this logic

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightApiResult.kt
MaxHastingsPP: Good catch, fixed!

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
MaxHastingsPP: I will fix the unit tests at the end once we are all happy with the rest of the changes.

ShopperInsights/build.gradle
```diff
@@ -37,7 +37,7 @@ dependencies {
     implementation deps.coreKtx
 
     api project(':BraintreeCore')
-
+    implementation 'com.google.code.gson:gson:2.8.9'
```
saperi22, line 40: I believe we don't want to add a 3rd party dependency. Let's check if there's an equivalent to @SerializedName that comes out of the box.

saperi22: @sshropshire wondering how we've handled property names without importing gson. Do we just use snake case where required if that's what the API expects?

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
MaxHastingsPP: fixed tests. I just made it one unit test for json to string conversion.

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsApiCall.kt
```diff
@@ -0,0 +1,32 @@
+package com.braintreepayments.api
+
+import PaymentMethodDetails
+import PaymentMethods
+import ShopperInsightApiResult
+
+internal class ShoppingInsightsApiCall(
```
scannillo, line 7: Take it or leave it comment - since the endpoint for this API is called `v2/payments/find-eligible-methods` maybe renaming to FindEligiblePaymentsAPI will be easier, since it's more 1:1 with the underlying API

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsApiRequest.kt
```diff
@@ -0,0 +1,27 @@
...
+ */
+data class ShopperInsightsApiRequest(
+    var request: ShopperInsightsRequest,
+    var merchantId: String,
+    var currencyCode: String,
+    var countryCode: String,
+    var accountDetails: Boolean,
+    var vaultTokens: Boolean,
+    var constraintType: String,
+    var paymentSources: List<String>
```
MaxHastingsPP, line 26: does List make sense for paymentSources? It is later converted to a JSONArray in `ShoppingInsightsCreateBody`

ShopperInsights/build.gradle
sarahkoop: Yes we just use string values for keys (see [PayPaLineItem example](https://github.com/braintree/braintree_android/blob/main/PayPal/src/main/java/com/braintreepayments/api/PayPalLineItem.java)). 

+1 to not including third party dependencies here

MaxHastingsPP: removed Gson, adding custom fromJson method

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsApiCall.kt
MaxHastingsPP: My only concern is that we break the naming consistency we have for everything else called Shopping Insights

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightApiResult.kt
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Represents the result from the Shopper Insight API.
+ *
+ * @property eligibleMethods Contains the payment methods available to the shopper.
+ */
+data class ShopperInsightApiResult(
+    val eligibleMethods: PaymentMethods
+) {
+    companion object {
+        fun fromJson(jsonString: String): ShopperInsightApiResult {
```
sarahkoop, line 12: Let's restrict this to library scope - merchants shouldn't need to call this

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightApiResult.kt
```diff
@@ -0,0 +1,63 @@
...
+    }
+}
+
+/**
+ * Contains details about the available payment methods.
+ *
+ * @property paypal Details about the PayPal payment method.
+ * @property venmo Details about the Venmo payment method.
+ */
+data class PaymentMethods(
```
sarahkoop, line 26: Nit - let's move each class to it's own file, that's typically the pattern we follow in the SDK currently

sarahkoop: Also let's prefix all of these with `ShopperInsights`

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -66,18 +69,44 @@ class ShopperInsightsClient @VisibleForTesting internal constructor(
...
+        // TODO: get correct merchant ID from SDK
+        val merchantId = "MXSJ4F5BADVNS"
+
+        // Default values
+        val countryCode = "US"
+        val currencyCode = "USD"
+        val constraintType = "INCLUDE"
+        val paymentSources = listOf("PAYPAL", "VENMO")
+        val includeVaultTokens = true
+        val includeAccountDetails = true
```
scannillo, line 81: Do we have to send these bools?

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsApiCall.kt
```diff
@@ -0,0 +1,32 @@
...
+    private val shoppingInsightsCreateBody: ShoppingInsightsCreateBody
+) {
+    fun execute(request: ShopperInsightsApiRequest): ShopperInsightApiResult {
```
sarahkoop, line 10:
```suggestion
    fun fetchShopperInsights(request: ShopperInsightsApiRequest): ShopperInsightApiResult {
```
Or any other descriptive name that makes sense to you all

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsApiCall.kt
sarahkoop: I'm good with Sammy's suggestion, if leaving it as is, let's drop the `Call` from the name since we don't use that elsewhere. It looks like we aren't entirely consistent throughout the SDK but we use either `PaymentFeatureApi` or `PaymentFeatureInternalClient` for the classes that handle making API calls

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsCreateBody.kt
```diff
@@ -0,0 +1,66 @@
+package com.braintreepayments.api
+
+import org.json.JSONArray
+import org.json.JSONObject
+
+internal class ShoppingInsightsCreateBody {
```
sarahkoop, line 6: I think this functionality should all live in the `ShopperInsightsRequest` class

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightApiResult.kt
MaxHastingsPP: I updated to make ShopperInsightApiResult as internal. Merchants will only have `ShopperInsightsInfo` returned to them when they make a call to ShoppingInsightsClient.

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
MaxHastingsPP: `includeAccountDetails` yes.
`includeVaultTokens` from testing it did not make a difference but just added it just in case.

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsApiCall.kt
MaxHastingsPP: would `fetch` work? Class name already has `ShoppingInsights`

sarahkoop: Assuming there might be more API calls added to this API client in the future, I think it would be helpful for the method to indicate what API call is actually being made - or do we think this will be the only API call for this feature?

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightApiResult.kt
MaxHastingsPP: fixed

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsApiCall.kt
sarahkoop: I don't feel that strongly about it though 🤷

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsApiCall.kt
MaxHastingsPP: I am cool with dropping the Call name part. fixed

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
scannillo: Gotcha - I think we should remove the ones that don't make a difference, start with as little as possible and then add if needed

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -66,18 +68,44 @@ class ShopperInsightsClient @VisibleForTesting internal constructor(
             return
         }
 
-        paymentReadyAPI.processRequest(request)
+        // TODO: get correct merchant ID from SDK
+        val merchantId = "MXSJ4F5BADVNS"
+
+        // Default values
+        val countryCode = "US"
```
scannillo, line 75: Similar Q for the countryCode - I thought we could omit this one based on one of our Slack threads with Harish (and that only currencyCode was required)

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsApiCall.kt
MaxHastingsPP: If we need to add more API calls in the future I would assume we would be creating new classes for them.

sarahkoop: We typically have one API "client" per payment feature that handles all internal API calls

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
MaxHastingsPP: When testing using postman I needed country code to get Venmo recommendation back.

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsApiCall.kt
MaxHastingsPP: What would a good example to look at in our codebase to get a better understanding?

sarahkoop: [ThreeDSecureAPI ](https://github.com/braintree/braintree_android/blob/main/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureAPI.java) provides an example

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsApi.kt
```diff
@@ -0,0 +1,30 @@
+package com.braintreepayments.api
+
+import ShopperInsightApiResult
+
+internal class ShoppingInsightsApi(
```
saperi22, line 5: rename to `ShopperInsightsApi`?

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
scannillo: Thanks for confirming!

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsApiCall.kt
MaxHastingsPP: @sarahkoop Thanks! I am concerned about API classes getting bloated if we follow that pattern, especially the unit test files.

ShopperInsights/src/test/java/com/braintreepayments/api/ShopperInsightApiResultTest.kt
```diff
@@ -0,0 +1,47 @@
...
+        val result = ShopperInsightApiResult.fromJson(jsonString)
+
+        // Assertions for PayPal
+        val paypal = result.eligibleMethods.paypal
+        assertTrue(paypal.canBeVaulted == true)
+        assertFalse(paypal.eligibleInPaypalNetwork == true)
+        assertTrue(paypal.recommended == true)
+        assertEquals(1, paypal.recommendedPriority)
+
+        // Assertions for Venmo
```
tdchow, line 40: What are your thoughts on breaking the venmo assertions into a separate unit test?

A benefit I see in having smaller, single-assertion unit tests is that when a test fails it's much easier to debug.

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsApiCall.kt
MaxHastingsPP: @tdchow What are your thoughts ?

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
MaxHastingsPP: removed `includeVaultTokens` not needed.

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsApi.kt
MaxHastingsPP: agreed. fixed

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsApiRequest.kt
```diff
@@ -0,0 +1,27 @@
...
+ * @property accountDetails Include account details
+ * @property vaultTokens Include vault tokens
+ * @property constraintType The constraint type
+ * @property paymentSources Payment sources, ie. PAYPAL VENMO
+ *
+ * [merchantId] [currencyCode] [countryCode] are needed for Venmo recommended results
+ *
+ */
+internal data class ShopperInsightsApiRequest(
+    var request: ShopperInsightsRequest,
```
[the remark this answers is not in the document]
MaxHastingsPP: yeah you are right. fixed.

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -16,12 +16,14 @@ import com.braintreepayments.api.ShopperInsightsAnalytics.VENMO_SELECTED
  * Note: **This feature is in beta. It's public API may change in future releases.**
  */
 class ShopperInsightsClient @VisibleForTesting internal constructor(
-    private val paymentReadyAPI: PaymentReadyApi,
+    private val shoppingInsightsCreateBody: ShoppingInsightsApi,
```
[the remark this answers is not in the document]
MaxHastingsPP: good catch. fixed.

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsPaymentMethodDetails.kt
```diff
@@ -0,0 +1,29 @@
...
+ * @property canBeVaulted Indicates if the payment method can be saved for future transactions.
+ * @property eligibleInPaypalNetwork Indicates if the payment method is eligible in the PayPal network.
+ * @property recommended Indicates if this payment method is recommended for the shopper.
+ * @property recommendedPriority The priority ranking of this payment method if recommended; null if not applicable.
+ */
+internal data class ShopperInsightsPaymentMethodDetails(
+    val canBeVaulted: Boolean?,
+    val eligibleInPaypalNetwork: Boolean?,
+    val recommended: Boolean?,
+    val recommendedPriority: Int? = null
```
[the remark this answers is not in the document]
MaxHastingsPP: actually I am going to just remove default being null and make them non optionals since the fromJson() method creates the data object with default values anyways. ie. false and 0

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsPaymentMethodDetails.kt
```diff
@@ -0,0 +1,29 @@
...
+ * Details of a specific payment method.
+ *
+ * @property canBeVaulted Indicates if the payment method can be saved for future transactions.
+ * @property eligibleInPaypalNetwork Indicates if the payment method is eligible in the PayPal network.
+ * @property recommended Indicates if this payment method is recommended for the shopper.
+ * @property recommendedPriority The priority ranking of this payment method if recommended; null if not applicable.
+ */
+internal data class ShopperInsightsPaymentMethodDetails(
+    val canBeVaulted: Boolean?,
+    val eligibleInPaypalNetwork: Boolean?,
```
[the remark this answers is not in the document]
MaxHastingsPP: fixed

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsPaymentMethods.kt
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Contains details about the available payment methods.
+ *
+ * @property paypal Details about the PayPal payment method.
+ * @property venmo Details about the Venmo payment method.
+ */
+internal data class ShopperInsightsPaymentMethods(
+    val paypal: ShopperInsightsPaymentMethodDetails,
+    val venmo: ShopperInsightsPaymentMethodDetails
```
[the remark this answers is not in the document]
MaxHastingsPP: Yes!! good catch thank you!

ShopperInsights/src/test/java/com/braintreepayments/api/ShoppingInsightsCreateBodyTest.kt
```diff
@@ -0,0 +1,69 @@
+package com.braintreepayments.api
+
+import org.junit.Test
+import org.skyscreamer.jsonassert.JSONAssert
+
+class ShoppingInsightsCreateBodyTest {
```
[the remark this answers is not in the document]
MaxHastingsPP: agree. fixed

ShopperInsights/src/test/java/com/braintreepayments/api/ShopperInsightApiResultTest.kt
MaxHastingsPP: I agree. Fixed.

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsApiCall.kt
tdchow: i think having APIs separated into their own class was a pattern that was introduced at the very start of the mobile checkout SDK, I'm not entirely sure what the reasoning was.

most projects I've seen have an API class that contains a logical grouping of multiple API calls. although they'd use a networking library like retrofit. so each function/api call wouldn't have any customized logic. (retrofit demo example: https://github.com/square/retrofit/blob/a4b81ed4ffc40a33c35bb466c4be707749b8c858/samples/src/main/java/com/example/retrofit/SimpleService.java#L39C27-L39C27)

in our case, I'm leaning towards the rename and remaining flexible on the possibility of adding an additional API call to this class.

MaxHastingsPP: What about a wrapper class that groups all the api method calls into one class for a particular module, similar to the `ThreeDSecureAPI` example. The only difference being that you move each api implementation into their own class file. It would help keep each class and its unit tests smaller in size and easier to read, even if a client needed to make many api calls everything would still continue to be manageable. Thoughts ?

sarahkoop: That sounds good to me 👍

tdchow: I agree, sounds like a good solution!

MaxHastingsPP: Cool, fixed

ShopperInsights/src/main/java/com/braintreepayments/api/ShoppingInsightsCreateBody.kt
MaxHastingsPP: fixed!

MaxHastingsPP:
Work stream for this feature is now paused until further notice. :(

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsApi.kt
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * [ShopperInsightsApi] is a wrapper class for all api related classes that
+ * [ShopperInsightsClient] depends on.
+ */
+internal class ShopperInsightsApi(
+    private val paymentsApi: FindEligiblePaymentsApi
+) {
+
+    fun findEligiblePayments(request: FindEligiblePaymentsApiRequest): FindEligiblePaymentsApiResult {
```
tdchow, line 15: I'm not sure if we're utilizing coroutines for async work. Will we need to make this a suspend function or add a callback for this API call?

ShopperInsights/src/main/java/com/braintreepayments/api/findeligiblepayments/FindEligiblePaymentMethodDetails.kt
```diff
@@ -0,0 +1,29 @@
+package com.braintreepayments.api.findeligiblepayments
```
sarahkoop, line 1: Can we move this out of the the `findeligiblepayments` directory? We use packge-private scope throughout the SDK, which relies on everything living in `com.braintreepayments.api`

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsApi.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.braintreepayments.api.findeligiblepayments.FindEligiblePaymentsApiResult
+import com.braintreepayments.api.findeligiblepayments.FindEligiblePaymentsApi
+import com.braintreepayments.api.findeligiblepayments.FindEligiblePaymentsApiRequest
+
+/**
+ * [ShopperInsightsApi] is a wrapper class for all api related classes that
+ * [ShopperInsightsClient] depends on.
+ */
+internal class ShopperInsightsApi(
+    private val paymentsApi: FindEligiblePaymentsApi
```
tdchow, line 12:
```suggestion
    private val findEligiblePaymentsApi: FindEligiblePaymentsApi
```

Would we want to match the variable name to the class name here?

ShopperInsights/src/main/java/com/braintreepayments/api/findeligiblepayments/FindEligiblePaymentMethodDetails.kt
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Details of a specific payment method.
+ *
+ * @property canBeVaulted Indicates if the payment method can be saved for future transactions.
+ * @property eligibleInPayPalNetwork Indicates if the payment method is eligible in the PayPal network.
+ * @property recommended Indicates if this payment method is recommended for the shopper.
+ * @property recommendedPriority The priority ranking of this payment method if recommended.
+ */
+internal data class FindEligiblePaymentMethodDetails(
```
sarahkoop, line 13: More nitpick but can we drop the `Find` on all these classes? It doesn't align with other naming patterns (we don't typically start with a verb, most data/result objects are just nouns)

tdchow (approved):
Looks good! just had a few non blocking comments.

ShopperInsights/src/main/java/com/braintreepayments/api/findeligiblepayments/FindEligiblePaymentMethodDetails.kt
MaxHastingsPP: fixed

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsApi.kt
MaxHastingsPP: Yeah we still have not decided yet. We will change this later once we work on API implementation.

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsApi.kt
MaxHastingsPP: fixed!

sarahkoop (approved):

saperi22 (approved):

PR #882, opened 2024-01-09
Add analytics events

saperi22:
### Summary of changes

 - Add analytics events for Shopper Insights

 ### Checklist

 - [ ] Added a changelog entry

- @saperi22

### DIFF
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
@@ -2,6 +2,9 @@ package com.braintreepayments.api
 
 import android.content.Context
 import androidx.annotation.VisibleForTesting
+import com.braintreepayments.api.ShopperInsightsAnalytics.GET_RECOMMENDED_PAYMENTS_FAILED
+import com.braintreepayments.api.ShopperInsightsAnalytics.GET_RECOMMENDED_PAYMENTS_STARTED
+import com.braintreepayments.api.ShopperInsightsAnalytics.GET_RECOMMENDED_PAYMENTS_SUCCEEDED
 import com.braintreepayments.api.ShopperInsightsAnalytics.PAYPAL_PRESENTED
 import com.braintreepayments.api.ShopperInsightsAnalytics.PAYPAL_SELECTED
 import com.braintreepayments.api.ShopperInsightsAnalytics.VENMO_PRESENTED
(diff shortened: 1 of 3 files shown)


ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -47,6 +51,7 @@ class ShopperInsightsClient @VisibleForTesting internal constructor(
                     )
                 )
             )
+            braintreeClient.sendAnalyticsEvent(GET_RECOMMENDED_PAYMENTS_FAILED)
```
MaxHastingsPP, line 54: Would it be possible to put why it failed? Merchant passing bad data

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -76,6 +82,7 @@ class ShopperInsightsClient @VisibleForTesting internal constructor(
                 )
             )
         )
+        braintreeClient.sendAnalyticsEvent(GET_RECOMMENDED_PAYMENTS_SUCCEEDED)
```
sarahkoop, line 85: For other modules in v5, to make sure that we always send an analytics event with each callback, we've been creating helper methods like `callbackRecommendedPaymentsSuccess()` which would include the calling back and sending analytics (for success and failure).

sarahkoop (approved):
Take it or leave it comment on callback methods otherwise LGTM 👍

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
saperi22: Yep, I remember the pattern from one of your PRs. I did initially write it that way, but then I took a second look and realized I was abstracting out 2 calls into a method. I concluded that the abstraction wasn't worth it at this time.

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
saperi22: Max, I don't see us sending any data in any of the other places. We'll have to look into the API to figure out if it accepts a key, value data object. I'll keep this PR as is and figure that out as part of another ticket if the need arises.

MaxHastingsPP: Makes sense.

MaxHastingsPP (approved):

PR #883, opened 2024-01-10
PayPalLauncher with v3 BrowserSwitchClient

sarahkoop:
### Summary of changes

 - Update PayPalLauncher to use browser switch pattern defined in https://github.com/braintree/browser-switch-android/pull/81 (CI to fail until that PR is merged)
 - Update demo app to handle persisting pending PayPal request for process kills

 ### Checklist

 - [x] Added a changelog entry

### Authors

- @sarahkoop


Demo/src/main/java/com/braintreepayments/demo/PendingRequestStore.java
```diff
@@ -0,0 +1,52 @@
+package com.braintreepayments.demo;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+
+import com.braintreepayments.api.PayPalPendingRequest;
+
+public class PendingRequestStore {
```
sarahkoop, line 8: This just an example of how a merchant could store these requests - could use Room or whatever other storage mechanism they want

PayPal/src/main/java/com/braintreepayments/api/PayPalPendingRequest.kt
```diff
@@ -0,0 +1,44 @@
...
+ * [PayPalLauncher.handleReturnToAppFromBrowser]
+ */
+sealed class PayPalPendingRequest {
+
+    /**
+     * A pending request was successfully started. This [PayPalPendingRequest.Started] should be
+     * stored and passed to [PayPalLauncher.handleReturnToAppFromBrowser]
+     */
+    class Started(val request: PayPalBrowserSwitchRequest) : PayPalPendingRequest() {
+        fun toJsonString(): String {
```
sarahkoop, line 18: Added these methods to make it easier for merchants to store the pending request as a String rather than having to deal with parcelizing/serializing/storing as an object. Open to other ideas for how to make this as merchant-friendly as possible.

v5_MIGRATION_GUIDE.md
```diff
@@ -359,10 +362,24 @@ class MyActivity : FragmentActivity() {
 +       payPalClient.createPaymentAuthRequest(this, request) { paymentAuthRequest ->
 +           when(paymentAuthRequest) {
 +               is PayPalPaymentAuthRequest.ReadyToLaunch -> {
-+                   payPalLauncher.launch(this@MyActivity, paymentAuthRequet)
++                   val pendingRequest = payPalLauncher.launch(this@MyActivity, paymentAuthRequet)
++                   when(pendingRequest) {
++                       is (PayPalPendingRequest.Success) { /* store pending request */ }
++                       is (PayPalPendingRequest.Failure) { /* handle error */ }
++                   }
```
scannillo, line 369: Does it matter in what order they call `launch()` or store the pending request?

v5_MIGRATION_GUIDE.md
```diff
@@ -324,26 +324,29 @@ class MyActivity : FragmentActivity() {
...
     
     // ALL OTHER ACTIVITY LAUNCH MODES 
     override fun onResume() {
-+       payPalLauncher.handleReturnToAppFromBrowser(requireContext(), requireActivity().intent)
++       handleReturnToAppFromBrowser(requireActivity().intent)
+    }
+    
+    fun handleReturnToAppFromBrowser(intent: Intent) {
+       // fetch stored PayPalPendingRequest.Success 
+       fetchPendingRequest()?.let {
```
scannillo, line 342: I think these lines also need ` +` prefix, since a merchant will have to add this code

PayPal/src/main/java/com/braintreepayments/api/PayPalLauncher.java
```diff
@@ -1,93 +1,95 @@
...
+     * the PayPal payment flow. Returns null if the user closed the browser to cancel the payment
+     * flow, or returned to the app without completing the PayPal authentication flow.
      */
-    public void handleReturnToAppFromBrowser(@NonNull Context context, @NonNull Intent intent) {
-        BrowserSwitchResult result = browserSwitchClient.parseResult(context, PAYPAL, intent);
+    @Nullable
+    public PayPalPaymentAuthResult handleReturnToAppFromBrowser(
+            @NonNull PayPalPendingRequest.Started pendingRequest, @NonNull Intent intent) {
+        BrowserSwitchResult result = browserSwitchClient.parseResult(pendingRequest.getRequest().getBrowserSwitchPendingRequest(), intent);
+        PayPalPaymentAuthResult paymentAuthResult = null;
```
scannillo, line 85: Take it or leave it comment - curious here why return `null` versus throw an error for the user-did-not-complete-flow case?

sarahkoop: This is to avoid being overly opinionated in the SDK - the browser switch method under the hood essentially just checks to see if there is a matching result for the intent and pending request passed in, and if there is, returns a success. If there isn't, the merchant can infer that the user didn't complete the flow based on their own setup and the request they passed in, but the SDK can't be certain that it means the user didn't complete the flow, because the merchant could have passed any request in at any point

v5_MIGRATION_GUIDE.md
sarahkoop: Ah yes good catch - fixed

PayPal/src/main/java/com/braintreepayments/api/PayPalPendingRequest.kt
sarahkoop: See comments in [this PR](https://github.com/braintree/browser-switch-android/pull/83/files) for alternative options for this

v5_MIGRATION_GUIDE.md
scannillo: 🤦 Ah, yes - that makes sense.

Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
```diff
@@ -62,7 +56,27 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
     @Override
     public void onResume() {
         super.onResume();
-        payPalLauncher.handleReturnToAppFromBrowser(requireContext(), requireActivity().getIntent());
+        PayPalPendingRequest.Started pendingRequest = getPendingRequest();
```
sshropshire, line 59:
```suggestion
        PayPalPendingRequest pendingRequest = getPendingRequest();
```
nit: Would it be the base class here since it could technically be `.Started` or `.Failure`?

Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
```diff
@@ -62,7 +56,27 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
+        if (pendingRequest != null) {
+            PayPalPaymentAuthResult paymentAuthResult = payPalLauncher.handleReturnToAppFromBrowser(pendingRequest, requireActivity().getIntent());
```
sshropshire, line 61:
```suggestion
            payPalClient.continueTokenize(pendingRequest, (result) -> { /* handle result logic */ };
```
nit: We could have a "continue" / "finish" method to pass control back to the SDK to complete the flow.

Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
```diff
@@ -62,7 +56,27 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
+            if (paymentAuthResult != null) {
+                completePayPalFlow(paymentAuthResult);
+            } else {
+                handleError(new Exception("User did not complete payment flow"));
+            }
+        }
+    }
+
+    private void storePendingRequest(PayPalPendingRequest.Started request) {
+        PendingRequestStore.putPayPalPendingRequest(requireContext(), request);
```
sshropshire, line 72:
```suggestion
        PendingRequestStore.getInstance(requireContext()).putPayPalPendingRequest(request);
```
nit: Singleton accessor.

Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
sarahkoop: Our docs tell merchants to only store a `PayPalRequest.Started`, so that's what the persistent store returns here. If the pending request was `PayPalPendingRequest.Failure` there is no need to store it, because the user can't complete that flow anyways, so they should handle an error immediately after calling `start`. Does that seem unclear for merchants?

Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
sarahkoop: Good call - fixed

PayPal/src/main/java/com/braintreepayments/api/PayPalBrowserSwitchRequest.java
```diff
@@ -0,0 +1,17 @@
+package com.braintreepayments.api;
+
+/**
+ * Request parameters associated with a {@link PayPalPendingRequest.Started}
+ */
+public class PayPalBrowserSwitchRequest {
```
sarahkoop, line 6: Is this class helpful/necessary? It just wraps a `BrowserSwitchPendingRequest.Started`, which is already a public class for direct browser switch integrations. We could just have `PayPalPendingRequest.Started` have a `BrowserSwitchPendingRequest.Started` as a param instead. Thoughts?

PayPal/src/main/java/com/braintreepayments/api/PayPalLauncher.java
scannillo: Got it, thank you!

Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
sshropshire: Oh got it yeah that makes sense. I'm leaning toward a `PayPalLaunchResult.Success` type with a `PayPalPendingRequest` associated value but re-using the `PayPalRequest.Started` type also makes sense.

I'm also thinking from a PPCP perspective, the requirements are bit different on our side.

Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
sshropshire: Yeah it's take or leave. This is another case where PPCP may be different. We probably won't have a Launcher since we only have browser-based flows at the moment.

PayPal/src/main/java/com/braintreepayments/api/PayPalBrowserSwitchRequest.java
sshropshire: Only thing it does is wrap browser switch. Keeping browser switch encapsulated by the SDK may be its only responsibility at this point.

scannillo (approved):

Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
sarahkoop: Since they can't pass it back into the `parseResult` method (which only accepts a `PayPalPendingRequest.Started`) I think while setting up the code it would become clear not to store a `Failure`, but the benefit of returning a `Failure` rather than `null` is that they can get the Exception `cause` from the `Failure` for more info, where they can't with a null response.

Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
sarahkoop: I've added this to the ticket to discuss the `PendingRequest` object naming - I think we can determine a good method naming pattern that makes sense with whatever we decide to call that request object as well.

PayPal/src/main/java/com/braintreepayments/api/PayPalBrowserSwitchRequest.java
sarahkoop: Ok will keep it in 👍

v5_MIGRATION_GUIDE.md
```diff
@@ -359,10 +362,24 @@ class MyActivity : FragmentActivity() {
 +       payPalClient.createPaymentAuthRequest(this, request) { paymentAuthRequest ->
```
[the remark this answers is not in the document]
sarahkoop: Oh good call I bet we can - I'll make a ticket to update that

PayPal/src/test/java/com/braintreepayments/api/PayPalPendingRequestUnitTest.kt
```diff
@@ -0,0 +1,64 @@
...
+
+import android.net.Uri
+import org.json.JSONException
+import org.json.JSONObject
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
```
tdchow, line 11: Is Robolectric needed for this unit test? Do we use Robolectric for all of our unit tests?

PayPal/src/test/java/com/braintreepayments/api/PayPalPendingRequestUnitTest.kt
```diff
@@ -0,0 +1,64 @@
...
+        false
+    )
+
+    private val browserSwitchPendingRequest =
+        BrowserSwitchPendingRequest.Started(browserSwitchRequest)
+
+    private val payPalBrowserSwitchRequest = PayPalBrowserSwitchRequest(browserSwitchPendingRequest)
+
+    @Test
+    fun startedConstructor_fromString_createsBrowserSwitchRequest() {
```
tdchow, line 28: could we use backticks around the function name to allow for spaces? it would make the file and reporting more readable.

```suggestion
    fun `started constructor fromString creates BrowserSwitchRequest`() {
```

tdchow (approved):
Just had a few non blocking questions about a unit test file.

PayPal/src/test/java/com/braintreepayments/api/PayPalPendingRequestUnitTest.kt
sarahkoop: We use it in most but not all unit tests for mocking Android dependencies. In these specific tests we get an error that `parse` in `Uri` is not mocked without Robolectric. We use the `AndroidJUnitRunner` for some tests where Robolectric is not required. Definitely open to improvements in our test suite, Robolectric has just been the default for a while now.

PayPal/src/test/java/com/braintreepayments/api/PayPalPendingRequestUnitTest.kt
sarahkoop: Ah yes good idea - most of our tests are still in Java, so we don't have a naming convention around this. But agreed that backticks and readable method names is helpful and should be used going forward.

PR #888, opened 2024-01-16
Add PayPal Messaging Dependency

jaxdesmarais:
This work is going into the `paypal-messaging-feature` branch

### Summary of changes

 - Now that a version of the SDK has been released we can add it as a dependency and pick back up on wrapping this feature

### Checklist

### Authors

@jaxdesmarais

### DIFF
--- a/PayPalMessaging/build.gradle
+++ b/PayPalMessaging/build.gradle
@@ -8,7 +8,7 @@ android {
 
     defaultConfig {
         // required to be out of sync with other modules due to messaging SDK min version
-        minSdkVersion 23
+        minSdkVersion rootProject.minSdkVersionPayPalMessaging
         targetSdkVersion rootProject.targetSdkVersion
 
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
--- a/PayPalMessaging/build.gradle
+++ b/PayPalMessaging/build.gradle
@@ -40,6 +40,7 @@ dependencies {
     implementation deps.coreKtx
     implementation deps.kotlinStdLib
     implementation deps.appCompat
+    implementation('com.paypal.messages:paypal-messages:1.0.0-alpha.0-SNAPSHOT')
 
     testImplementation deps.junitTest
 }
--- a/PayPalMessaging/build.gradle
+++ b/PayPalMessaging/build.gradle
@@ -7,7 +7,8 @@ android {
     compileSdkVersion rootProject.compileSdkVersion
 
     defaultConfig {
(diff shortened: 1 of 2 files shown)


agedd (approved):
🚀

PayPalMessaging/build.gradle
```diff
@@ -40,10 +41,19 @@ dependencies {
     implementation deps.coreKtx
     implementation deps.kotlinStdLib
     implementation deps.appCompat
+    implementation('com.paypal.messages:paypal-messages:1.0.0-alpha.00-SNAPSHOT')
 
     testImplementation deps.junitTest
 }
 
+allprojects {
```
sarahkoop, line 49: Can we add a TODO to remove this before feature release?

jaxdesmarais: ✅ : https://github.com/braintree/braintree_android/pull/888/commits/a66d782dd41b374f932d83b898b76b49add187c5

PayPalMessaging/build.gradle
```diff
@@ -7,7 +7,8 @@ android {
     compileSdkVersion rootProject.compileSdkVersion
 
     defaultConfig {
-        minSdkVersion rootProject.minSdkVersion
+        // required to be out of sync with other modules due to messaging SDK min version
+        minSdkVersion 23
```
MaxHastingsPP, line 11: I wonder if instead of hardcoding version 23 here we reference the rootProject such as

`rootProject.midSdkVersionNativeCheckout`

thoughts?

jaxdesmarais: I think there could be a future where the min SDK version for PayPal Checkout and the min version for PayPal Messaging dependencies could be different. I'm not necessarily opposed though if we feel that it'd be helpful.

Not sure if this will become a non issue in v5 @sarahkoop?

sarahkoop: It will become a non-issue in v5 for API 23, but not a bad idea to establish a convention around what to do if one module deviates from the others on API versions. Agreed not to link it with native checkout, but maybe it's own dependency in the top-level gradle file that is `rootProject.minSdkVersionPayPalMessaging`. Then it's more obvious when we do make changes that there are modules that don't align and make sure to update those too.

jaxdesmarais: Done: https://github.com/braintree/braintree_android/pull/888/commits/127f32add8131b64fe181921bf65d092afdc29a4

sarahkoop (approved):

MaxHastingsPP (approved):

PR #894, opened 2024-01-22
Add network call for shopper insights

MaxHastingsPP:
### Summary of changes

 - Added network call for eligible payments api
 - Updated demo app to show response

### Screenshots

Screenshot 2024-01-23 at 11 57 30 AM

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### Authors
@MaxHastingsPP

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
@@ -124,7 +124,7 @@ internal class BraintreeHttpClient(
         if (authorization is TokenizationKey) {
             request.addHeader(CLIENT_KEY_HEADER, authorization.bearer)
         }
-        request.addHeader("Authorization", "Bearer " + authorization?.bearer)
+        authorization?.bearer?.let { token -> request.addHeader("Authorization", "Bearer $token")
         httpClient.sendRequest(request, callback)
     }
 
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
@@ -83,6 +83,7 @@ internal class BraintreeHttpClient(
      * @param authorization
      * @param callback [HttpResponseCallback]
      */
+    @Suppress("CyclomaticComplexMethod")
     fun post(
         path: String,
         data: String,
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
@@ -124,6 +124,9 @@ internal class BraintreeHttpClient(
         if (authorization is TokenizationKey) {
             request.addHeader(CLIENT_KEY_HEADER, authorization.bearer)
         }
+        if (authorization is ClientToken) {
+            request.addHeader("Authorization", "Bearer " + authorization.authorizationFingerprint)
+        }
         httpClient.sendRequest(request, callback)
     }
 
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
@@ -124,9 +124,7 @@ internal class BraintreeHttpClient(
         if (authorization is TokenizationKey) {
             request.addHeader(CLIENT_KEY_HEADER, authorization.bearer)
         }
-        if (authorization is ClientToken) {
-            request.addHeader("Authorization", "Bearer " + authorization.authorizationFingerprint)
-        }
+        request.addHeader("Authorization", "Bearer " + authorization?.bearer)
         httpClient.sendRequest(request, callback)
     }
 
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
@@ -125,7 +125,7 @@ internal class BraintreeHttpClient(
         if (authorization is TokenizationKey) {
             request.addHeader(CLIENT_KEY_HEADER, authorization.bearer)
         }
-        authorization?.bearer?.let { token -> request.addHeader("Authorization", "Bearer $token")
+        authorization?.bearer?.let { token -> request.addHeader("Authorization", "Bearer $token") }
         httpClient.sendRequest(request, callback)
     }
 
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
@@ -15,9 +15,7 @@ internal class EligiblePaymentsApi(
             jsonBody,
             object : HttpResponseCallback {
                 override fun onResult(responseBody: String?, httpError: Exception?) {
-                    if (httpError != null) {
-                        callback.onResult(null, httpError)
-                    } else if (responseBody != null) {
+                    if (responseBody != null) {
                         try {
                             callback.onResult(
                                 EligiblePaymentsApiResult.fromJson(responseBody),
@@ -27,7 +25,7 @@ internal class EligiblePaymentsApi(
                             callback.onResult(null, e)
                         }
                     } else {
-                        callback.onResult(null, Exception("null body"))
+                        callback.onResult(null, httpError)
                     }
                 }
             }
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
@@ -17,23 +17,10 @@ internal class EligiblePaymentsApi(
                 override fun onResult(responseBody: String?, httpError: Exception?) {
                     if (responseBody != null) {
                         try {
-                            val responseJson = JSONObject(responseBody)
-                            callback.onResult(EligiblePaymentsApiResult(
-                                eligibleMethods = EligiblePaymentMethods(
-                                    paypal = EligiblePaymentMethodDetails(
-                                        canBeVaulted = true,
-                                        eligibleInPayPalNetwork = true,
-                                        recommended = true,
-                                        recommendedPriority = 1
-                                    ),
-                                    venmo = EligiblePaymentMethodDetails(
-                                        canBeVaulted = true,
-                                        eligibleInPayPalNetwork = true,
-                                        recommended = true,
-                                        recommendedPriority = 1
-                                    )
-                                )
-                            ), null)
+                            callback.onResult(
+                                EligiblePaymentsApiResult.fromJson(responseBody),
+                                null
+                            )
                         } catch (e: JSONException) {
                             callback.onResult(null, e)
                         }
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
@@ -8,27 +8,33 @@ internal class EligiblePaymentsApi(
 ) {
     fun execute(request: EligiblePaymentsApiRequest, callback: EligiblePaymentsCallback) {
         val jsonBody = request.toJson()
-        // TODO: Move url to PaypalHttpClient class when it is created
-        val url = "https://api.sandbox.paypal.com/v2/payments/find-eligible-methods"
-        braintreeClient.sendPOST(
-            url,
-            jsonBody,
-            object : HttpResponseCallback {
-                override fun onResult(responseBody: String?, httpError: Exception?) {
-                    if (responseBody != null) {
-                        try {
-                            callback.onResult(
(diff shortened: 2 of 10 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
```diff
@@ -124,6 +124,7 @@ internal class BraintreeHttpClient(
         if (authorization is TokenizationKey) {
             request.addHeader(CLIENT_KEY_HEADER, authorization.bearer)
         }
+        request.addHeader("Authorization", "Bearer " + authorization?.bearer)
```
scannillo, line 127: We don't want to send this with other API requests that hit the Braintree GW. On iOS we did a check like [this](https://github.com/braintree/braintree_ios/blob/235c6994f3c4b5373db25f0f1cd46770e6a38fdd/Sources/BraintreeCore/BTHTTP.swift#L275-L278)

ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
```diff
@@ -1,28 +1,34 @@
...
-                    recommendedPriority = 1
-                )
-            )
+internal class EligiblePaymentsApi(
+    private val braintreeClient: BraintreeClient
+) {
+    fun execute(request: EligiblePaymentsApiRequest, callback: EligiblePaymentsCallback) {
+        val jsonBody = request.toJson()
+        // TODO: Move url to PaypalHttpClient class when it is created
+        val url = "https://api.sandbox.paypal.com/v2/payments/find-eligible-methods"
```
scannillo, line 12: This URL should be able to switch b/w `api.sandbox.paypal.com` & `api.paypal.com` based on the env of the ClientToken provided to initialize the `BraintreeClient`. We get the env from Config lookup call [here](https://github.com/braintree/braintree_android/blob/13140a52d349533d1285d7c0a0257852c3f1cd57/BraintreeCore/src/main/java/com/braintreepayments/api/ConfigurationLoader.kt#L16) (see [Configuration.environment ](https://github.com/braintree/braintree_android/blob/13140a52d349533d1285d7c0a0257852c3f1cd57/BraintreeCore/src/main/java/com/braintreepayments/api/Configuration.kt#L80)property)

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
MaxHastingsPP: Yeah I agree, I just did not see an easy way to do it that wouldn't be messy. I can look more into it.

sshropshire:
```suggestion
        authorization?.bearer?.let { token -> request.addHeader("Authorization", "Bearer $token")
```
^ is a good option. It guards against `Authorization: Bearer null` from being sent.

ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
```diff
@@ -1,28 +1,34 @@
...
+        braintreeClient.sendPOST(
+            url,
+            jsonBody,
+            object : HttpResponseCallback {
+                override fun onResult(responseBody: String?, httpError: Exception?) {
```
tdchow, line 17: Do we want to do something with the `httpError`? We might want to pipe it through the callback if it's not null.

ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
```diff
@@ -1,28 +1,34 @@
...
+                        try {
+                            callback.onResult(
+                                EligiblePaymentsApiResult.fromJson(responseBody),
+                                null
+                            )
+                        } catch (e: JSONException) {
+                            callback.onResult(null, e)
+                        }
+                    } else {
+                        callback.onResult(null, Exception("null body"))
```
tdchow, line 28: Do we return a generic `Exception` in other API calls when the reponseBody is null? It might be good to create a specific exception type for this.

MaxHastingsPP: Good point, let me see what we do elsewhere, maybe it already exists.

ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
MaxHastingsPP: Fixed. bbe66c375bb173fcdf315fdfe5064f93ccba8fd3

ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
MaxHastingsPP: Changed to match other api classes. 73740f7a38912977469a9f0e964ebf2072a4db13

They all return httpError if responseBody is null.

ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
MaxHastingsPP: Ah, I did not see that. Thank you. Fixed 8c921affcfcfcf64f2fbce2c9c0787cbe6358595

ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
```diff
@@ -1,28 +1,40 @@
...
+    private val braintreeClient: BraintreeClient
+) {
+    fun execute(request: EligiblePaymentsApiRequest, callback: EligiblePaymentsCallback) {
+        val jsonBody = request.toJson()
+        braintreeClient.getConfiguration { configuration, configError ->
+            // TODO: Move url to PaypalHttpClient class when it is created
+            val baseUrl = when (configuration?.environment) {
+                "production" -> "https://api.paypal.com"
+                else -> "https://api.sandbox.paypal.com"
+            }
```
tdchow, line 16: Does this baseUrl logic exist in the other API calls? I wonder if we can extract this to a shared function/class.

tdchow (commented):
Changes look good! Just had a question about extracting the baseUrl logic. Can we also add unit test coverage for these changes?

ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
MaxHastingsPP: One exists for Braintree but not PayPal. https://github.com/braintree/braintree_android/blob/8c921affcfcfcf64f2fbce2c9c0787cbe6358595/BraintreeCore/src/main/java/com/braintreepayments/api/TokenizationKey.kt#L35

I think the plan is to create a `PaypalHttpClient` and have the baseUrl logic live in there.

MaxHastingsPP:
> Changes look good! Just had a question about extracting the baseUrl logic. Can we also add unit test coverage for these changes?

Yeah I can look into adding tests.

MaxHastingsPP:
> Changes look good! Just had a question about extracting the baseUrl logic. Can we also add unit test coverage for these changes?

I added some EligiblePaymentApi unit tests. 3dddac2c37674031a18783739ab628819c58a2f0

tdchow:
> > Changes look good! Just had a question about extracting the baseUrl logic. Can we also add unit test coverage for these changes?
> 
> I added some EligiblePaymentApi unit tests. [3dddac2](https://github.com/braintree/braintree_android/commit/3dddac2c37674031a18783739ab628819c58a2f0)

Thanks for adding the tests! Could we also add coverage for the changes in `ShopperInsightsClient` and `BraintreeHttpClient`?

ShopperInsights/src/test/java/com/braintreepayments/api/EligiblePaymentsApiUnitTest.kt
```diff
@@ -0,0 +1,123 @@
...
+    @Before
+    fun setup() {
+        callback = mockk(relaxed = true)
+        braintreeClient = mockk(relaxed = true)
+        setupBraintreeClientToReturnConfiguration()
+        sut = EligiblePaymentsApi(braintreeClient)
+    }
+
+    @Test
+    fun test_sendPOST_Error() {
```
tdchow, line 26: Nit - in Kotlin we can use backticks to include spaces in the test function names.


```suggestion
    fun `when sendPost is called and an error occurs, callback onResult is invoked with the error`() {
```

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
```diff
@@ -124,6 +125,7 @@ internal class BraintreeHttpClient(
         if (authorization is TokenizationKey) {
             request.addHeader(CLIENT_KEY_HEADER, authorization.bearer)
         }
+        authorization?.bearer?.let { token -> request.addHeader("Authorization", "Bearer $token") }
```
scannillo, line 128: Can we have a test to cover this? To make sure it is only added in the cases we expect?

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -91,16 +85,25 @@ class ShopperInsightsClient @VisibleForTesting internal constructor(
...
-                )
-            )
+            ),
+            callback = { result, error ->
+                if (error == null) {
+                    callback.onResult(
+                        ShopperInsightsResult.Success(
+                            ShopperInsightsInfo(
+                                isPayPalRecommended = isPaymentRecommended(result?.eligibleMethods?.paypal),
+                                isVenmoRecommended = isPaymentRecommended(result?.eligibleMethods?.venmo)
```
scannillo, line 95: If the `result` comes back `null`/missing, would we instead want to surface an error to the merchant, versus giving them back a `false` for `is<Payment>Recommended`?

tdchow: I'm adding this logic now - do we expect to always get Venmo and PayPal in the response? Are there any cases where we'd only get one or the other?

```kotlin
            result?.eligibleMethods?.paypal == null ||
                    result.eligibleMethods.venmo == null -> {
                callback.onResult(
                    ShopperInsightsResult.Failure(
                        NullPointerException("Missing data in API response")
                    )
                )
            }
```

This is what I currently have, if either of methods is missing I'm invoking the failure.

agedd: i think based on the [doc](https://paypal.atlassian.net/wiki/spaces/MI/pages/738195099/Internal+Shopper+Insights+-+BT+Web+SDK+Integration) here we can either get both, either paypal or venmo, or neither. Thinking of throwing a custom error like `BTShopperInsightsError.emptyBodyReturned` in the case none are eligible. wdyt?

tdchow: thanks for sharing that doc!

if we're able to get one method and not the other, we might want to only have a check to see if both Venmo and PayPal are missing from the response and return the failure. 

If we get one and not the other, we could treat it as a successful onResult, but default the one that's missing to not recommended.

in the case that both are returned in the API but both are not eligible, it might make sense to return a success to the merchant but have both `isPayPalRecommended` and `isVenmoRecommended` as false. 

lots of edge cases here to handle 😅

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
tdchow: I added a few tests covering this logic.

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
tdchow: i just pushed a commit updating this logic to && instead of ||:

```kotlin
            result?.eligibleMethods?.paypal == null &&
                    result.eligibleMethods.venmo == null -> {
                callback.onResult(
                    ShopperInsightsResult.Failure(
                        NullPointerException("Missing data in API response")
                    )
                )
            }
```

what do you all think? does this logic make sense?

saperi22: Nit: I wouldn't use `NullPointerException`, but a more specific exception for the use-case. Otherwise, I'm good with the logic.

saperi22 (approved):

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
agedd: the logic looks good to me! i tried to match the same on iOS and am throwing a custom error -`BTShopperInsightsError.emptyBodyReturned` - in this case.

agedd (approved):
lgtm!

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
tdchow: I'll update this to a specific exception matching `emptyBodyReturned` on iOS

PR #897, opened 2024-01-24
Update LocalPaymentLauncher for Browser Switch v3

sarahkoop:
### Summary of changes

 - Update LocalPaymentLauncher to use browser switch pattern defined in https://github.com/braintree/browser-switch-android/pull/81
 - Update SEPADirectDebit and PayPal demo fragment pending request storage

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors
@sarahkoop 

- [ ] ASAP (let's discuss outside this PR)
- [ ] Soon, here's our target release date: <DD/MM/YYYY>
- [x] Whenever, no rush


BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeDeepLinkActivity.kt
```diff
@@ -7,11 +7,10 @@ import androidx.appcompat.app.AppCompatActivity
  * This Activity is used by the SDK Activity to handle deep links on behalf of the merchant
  * It should not be used directly.
  */
+// NEXT_MAJOR_VERSION : remove this class
 class BraintreeDeepLinkActivity : AppCompatActivity() {
-    private val browserSwitchClient = BrowserSwitchClient()
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        browserSwitchClient.captureResult(this)
```
sarahkoop, line 14: This is unrelated, but the latest browser switch snapshot removes this method so it no longer compiles

v5_MIGRATION_GUIDE.md
```diff
@@ -409,35 +409,36 @@ do not need to be instantiated in `OnCreate`.
...
     // ALL OTHER ACTIVITY LAUNCH MODES 
     override fun onResume() {
-+       localPaymentLauncher.handleReturnToAppFromBrowser(requireContext(), requireActivity().
-+           intent)
++       handleReturnToAppFromBrowser(requireActivity().intent)
+    }
+    
+    fun handleReturnToAppFromBrowser(intent: Intent) {
+       // fetch stored LocalPaymentPendingRequest.Success 
++       fetchPendingRequest()?.let {
```
sshropshire, line 433:
```suggestion
+       fetchPendingRequestFromPersistentStorage()?.let {
```
🧶 : Could also use `getPersistentRequest()`. Maybe I'm being overly cautious, but the take-or-leave for me is to avoid sounding like a network request is being executed.

sshropshire (approved):

Demo/src/main/java/com/braintreepayments/demo/LocalPaymentFragment.java
```diff
@@ -35,18 +37,23 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
     }
 
     @Override
     public void onResume() {
         super.onResume();
-        localPaymentLauncher.handleReturnToAppFromBrowser(requireContext(),
-                requireActivity().getIntent());
+        LocalPaymentPendingRequest.Started pendingRequest = getPendingRequest();
+        if (getPendingRequest() != null) {
+            LocalPaymentAuthResult paymentAuthResult = localPaymentLauncher.handleReturnToAppFromBrowser(pendingRequest, requireActivity().getIntent());
```
tdchow, line 48: nit - is there a coding style / line limit that the BT SDK adheres to?

Demo/src/main/java/com/braintreepayments/demo/PendingRequestStore.java
```diff
@@ -36,10 +40,52 @@ public static PayPalPendingRequest.Started getPayPalPendingRequest(Context conte
...
+            try {
+                return new LocalPaymentPendingRequest.Started(requestString);
+            } catch (JSONException e) {
+                return null;
+            }
+        }
+        return null;
+    }
+
+    public  void clearLocalPaymentPendingRequest(Context context) {
```
tdchow, line 64:
```suggestion
    public void clearLocalPaymentPendingRequest(Context context) {
```

Demo/src/main/java/com/braintreepayments/demo/PendingRequestStore.java
```diff
@@ -36,10 +40,52 @@ public static PayPalPendingRequest.Started getPayPalPendingRequest(Context conte
...
+            try {
+                return new SEPADirectDebitPendingRequest.Started(requestString);
+            } catch (JSONException e) {
+                return null;
+            }
+        }
+        return null;
+    }
+
+    public  void clearSEPADirectDebitPendingRequest(Context context) {
```
tdchow, line 85:
```suggestion
    public void clearSEPADirectDebitPendingRequest(Context context) {
```

LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentBrowserSwitchRequest.kt
```diff
@@ -0,0 +1,22 @@
+package com.braintreepayments.api
+
+import androidx.annotation.RestrictTo
+
+/**
+ * Request parameters associated with a [LocalPaymentPendingRequest.Started]
+ */
+class LocalPaymentBrowserSwitchRequest internal constructor(
+    browserSwitchPendingRequest: BrowserSwitchPendingRequest.Started
```
tdchow, line 9: Could we simplify this class by defining the variable in the constructor?

```suggestion
    private val browserSwitchPendingRequest: BrowserSwitchPendingRequest.Started
```

This would remove the need to define `browserSwitchPendingRequest` within the class and the init

LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentBrowserSwitchRequest.kt
```diff
@@ -0,0 +1,22 @@
...
+) {
+
+    private var browserSwitchPendingRequest: BrowserSwitchPendingRequest.Started
+
+    init {
+        this.browserSwitchPendingRequest = browserSwitchPendingRequest
+    }
+
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
tdchow, line 18: Is this annotation needed in Kotlin?

Could we make `LocalPaymentBrowserSwitchRequest` internal which would obfuscate the entire class from the merchant?

tdchow (approved):
Looks good! just had a few non blocking comments.

LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentBrowserSwitchRequest.kt
sarahkoop: This is a good call I think it actually warrants a separate PR to align across all the modules that use browser switch. I think we can remove the need for this wrapper class, and just have a `BrowserSwitchPendingRequest.Started` as a `private val` in the `LocalPaymentPendingRequest.Started`. The whole point of this `LocalPaymentBrowserSwitchRequest` was to encapsulate browser switch classes, but if we make that class a private val, it's still encapsulated from the merchant perspective.

Demo/src/main/java/com/braintreepayments/demo/LocalPaymentFragment.java
sarahkoop: Rules are enforced for Kotlin files (max line length 120) but not for Java files currently 🤦‍♀️ - I just rely on the Android studio reformat file, but sometimes that adds unrelated formatting changes that distract from PRs. It's a work in progress. Updated this line though

v5_MIGRATION_GUIDE.md
sarahkoop: Good call - updated

LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentBrowserSwitchRequest.kt
sarahkoop: See comment above - will fix this across modules in a separate PR 👍

Demo/src/main/java/com/braintreepayments/demo/LocalPaymentFragment.java
tdchow: got it! it might not be a huge issue moving forward with more classes getting converted to kotlin.

PR #899, opened 2024-01-30
Remove PayPalBrowserSwitchRequest and Kotlin Conversion

sarahkoop:
### Summary of changes

 - Remove make `PayPalPendingRequest.Started` constructor variable internal and remove unnecessary wrapper class `PayPalBrowserSwitchRequest`
 - Convert `PayPalLauncher` and unit tests to Kotlin - the Kotlin to Java interop with internal constructor variables doesn't seem to work well (looks like `pendingRequest.request$PayPal_debug()` when used in a Java class within same module), so converted to Kotlin for cleanliness.
 - If changes look good, will make the same change in SEPADirectDebit and LocalPayment modules

### Checklist

### Authors
@sarahkoop 

- [ ] ASAP (let's discuss outside this PR)
- [ ] Soon, here's our target release date: <DD/MM/YYYY>
- [x] Whenever, no rush


PayPal/src/main/java/com/braintreepayments/api/PayPalLauncher.kt
```diff
@@ -0,0 +1,94 @@
...
+import androidx.activity.ComponentActivity
+
+/**
+ * Responsible for launching PayPal user authentication in a web browser
+ */
+class PayPalLauncher internal constructor(private val browserSwitchClient: BrowserSwitchClient) {
+    /**
+     * Used to launch the PayPal flow in a web browser and deliver results to your Activity
+     */
+    constructor() : this(BrowserSwitchClient())
```
tdchow, line 13: Could we instead make this a default value in the main constructor? 

```kotlin
class PayPalLauncher internal constructor(
    private val browserSwitchClient: BrowserSwitchClient = BrowserSwitchClient()
)
```

edit: actually is this to only expose the empty constructor to the merchant?

PayPal/src/main/java/com/braintreepayments/api/PayPalLauncher.kt
```diff
@@ -0,0 +1,94 @@
...
+     * invoking [PayPalLauncher.launch]
+     * @param intent         the intent to return to your application containing a deep link result
+     * from the PayPal browser flow
+     * @return a [PayPalPaymentAuthResult] that should be passed to
+     * [PayPalClient.tokenize] to complete
+     * the PayPal payment flow. Returns null if the user closed the browser to cancel the payment
+     * flow, or returned to the app without completing the PayPal authentication flow.
+     */
+    fun handleReturnToAppFromBrowser(
+        pendingRequest: PayPalPendingRequest.Started, intent: Intent
```
tdchow, line 68: Nit - not sure if we have a specific Kotlin style guide or if detekt will enforce these rule sets. What are your thoughts on having each parameter on a separate line when the primary constructor needs to expand more than a single line?

https://kotlinlang.org/docs/coding-conventions.html#class-headers

PayPal/src/test/java/com/braintreepayments/api/PayPalLauncherUnitTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    private val paymentAuthRequestParams: PayPalPaymentAuthRequestParams = mockk(relaxed = true)
+    private val intent: Intent = mockk(relaxed = true)
+    private val options: BrowserSwitchOptions = mockk(relaxed = true)
+    private val browserSwitchRequest = BrowserSwitchRequest(
+        1,
+        Uri.parse("http://"),
+        JSONObject().put("test_key", "test_value"),
+        "return-url-scheme",
+        false
+    )
```
tdchow, line 33: What are your thoughts on adding named parameters to this constructor? It might add some clarity to some of the arguments.

PayPal/src/test/java/com/braintreepayments/api/PayPalLauncherUnitTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    fun setup() {
+        every { paymentAuthRequestParams.browserSwitchOptions } returns options
+    }
+
+    @Test
+    fun launch_startsBrowserSwitch_returnsPendingRequest() {
+        val startedPendingRequest = BrowserSwitchPendingRequest.Started(browserSwitchRequest)
+        every { browserSwitchClient.start(activity, options) } returns startedPendingRequest
+
+        val sut = PayPalLauncher(browserSwitchClient)
```
tdchow, line 45: Would we be able to move `sut` to be a lateinit var on the test class, then in `setup` save a new instance of PayPalLauncher? It could minimally clean up each of the test functions.

tdchow (approved):
Changes look good! Just had a few non-blocking comments.

PayPal/src/main/java/com/braintreepayments/api/PayPalLauncher.kt
sarahkoop: Yes we don't want to expose `BrowserSwitchClient` to merchants, but we want them to have an empty constructor.

PayPal/src/test/java/com/braintreepayments/api/PayPalLauncherUnitTest.kt
tdchow: Ah, i didn't realize `BrowserSwitchRequest` is a Java file.

PayPal/src/test/java/com/braintreepayments/api/PayPalLauncherUnitTest.kt
sarahkoop: Done 👍

PayPal/src/main/java/com/braintreepayments/api/PayPalLauncher.kt
sarahkoop: Manually fixed this one but to enforce it across the project I think we need to turn on ArgumentListWrapping in the detekt config (and will add a bunch of unrelated files to this PR to fix). Maybe a separate PR would be a good time to evaluate the entire Kotlin style guide that we want to follow and update any rules. We don't currently have a strong Kotlin convention in this project so am open to any ruleset you all prefer.

tdchow: Sounds good. Yep definitely in a separate PR. 

We could add a baseline file to prevent all existing code from breaking the new ruleset. It would only enforce new code or modified code.

As for coding styles, I think Detekt's default ruleset follows the kotlin and Android style guides:
* https://developer.android.com/kotlin/style-guide
* https://kotlinlang.org/docs/coding-conventions.html

but we can also modify them if we need

PR #904, opened 2024-02-05
Implement PayPal FPTI Analytics

sarahkoop:
### Summary of changes

 - Update PayPal flow with FPTI analytics events

### Checklist

 - ~[ ] Added a changelog entry~
 - [x] Relevant test coverage

### Authors
@sarahkoop

PR #908, opened 2024-02-05
Update Local Payments to Send FPTI Events

tdchow:
### Summary of changes

 - Updates to the Local Payments module to send analytics events to FPTI
 - Bundled all failure events under the same `PAYMENT_FAILED` event - additional details around the error will be sent as a param in a separate PR

### Checklist

 ~- [ ] Added a changelog entry|~

[ ] Whenever, no rush


LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentClient.java
```diff
@@ -116,20 +117,27 @@ void buildBrowserSwitchOptions(@NonNull
...
                     .put("merchant-account-id",
                             localPaymentAuthRequestParams.getRequest().getMerchantAccountId())
                     .put("payment-type", localPaymentAuthRequestParams.getRequest().getPaymentType()));
         } catch (JSONException e) {
-            callback.onLocalPaymentAuthRequest(new LocalPaymentAuthRequest.Failure(new BraintreeException("Error parsing local payment request")));
+            authRequestFailure(
+                new BraintreeException("Error parsing local payment request"),
+                callback
+            );
+            // Should we return here? or move the lines following the try/catch into the try block?
```
sarahkoop, line 130: Let's return here

LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentAnalytics.kt
```diff
@@ -1,17 +1,14 @@
...
-    PAYMENT_FAILED("local-payment:start-payment:failed"),
-    PAYMENT_CANCELED("local-payment:start-payment:browser-login:canceled"),
+    const val PAYMENT_STARTED = "local-payment:start-payment:started"
+    const val PAYMENT_SUCCEEDED = "local-payment:start-payment:succeeded"
+    const val PAYMENT_FAILED = "local-payment:start-payment:failed"
+    const val PAYMENT_CANCELED = "local-payment:start-payment:browser-login:canceled"
 
     // Browser Presentation Events
-    BROWSER_SWITCH_SUCCEEDED("local-payment:start-payment:browser-presentation:succeeded"),
-    BROWSER_SWITCH_FAILED("local-payment:start-payment:browser-presentation:failed"),
```
sarahkoop, line 13: I'm fine removing these as they don't add much value in this flow, but we should maybe align the approach with this PR (#904) where I was just sending the browser failed event with the tokenize failed event. What do you think makes more sense?

tdchow: I feel that it might be redundant to send multiple events back to back for a certain scenario. I think this is where adding detail params would be able to capture the same data while minimizing the amount of events we're sending.

Should I align the PR to match the one you shared or create a story to ensure we're sending the browser switch detail for this error?

sarahkoop: Ok agreed let's keep this PR as is then and can update the PayPal analytics when the detail is added to the error in a future PR.

saperi22 (approved):

PR #909, opened 2024-02-06
Remove App Check From Shopper Insights and Analytics Updates

tdchow:
### Summary of changes

 - Removed the PayPal and Venmo app installed checks from the Shopper Insights module - the API call is now always executed.
 - Rearrange a few analytics calls to match iOS

### Checklist

[ ] Whenever, no rush

### DIFF
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
@@ -90,18 +90,10 @@ class ShopperInsightsClient @VisibleForTesting internal constructor(
             }
 
             else -> {
-                braintreeClient.sendAnalyticsEvent(GET_RECOMMENDED_PAYMENTS_SUCCEEDED)
-                callback.onResult(
-                    ShopperInsightsResult.Success(
-                        ShopperInsightsInfo(
-                            isPayPalRecommended = isPaymentRecommended(
-                                result.eligibleMethods.paypal
-                            ),
-                            isVenmoRecommended = isPaymentRecommended(
-                                result.eligibleMethods.venmo
-                            )
-                        )
-                    )
+                callbackSuccess(
+                    callback = callback,
+                    isPayPalRecommended = isPaymentRecommended(result.eligibleMethods.paypal),
+                    isVenmoRecommended = isPaymentRecommended(result.eligibleMethods.venmo)
                 )
             }
         }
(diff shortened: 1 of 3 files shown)


sarahkoop (approved):

ShopperInsights/src/test/java/com/braintreepayments/api/ShopperInsightsClientUnitTest.kt
```diff
@@ -24,91 +20,30 @@ import org.junit.Test
...
-        }
         verifyStartedAnalyticsEvent()
-        verifySuccessAnalyticsEvent()
     }
 
     @Test
-    fun testGetRecommendedPaymentMethods_hasBothAppsInstalled_returnsSuccessResult() {
-        every { deviceInspector.isVenmoInstalled(applicationContext) } returns true
-        every { deviceInspector.isPayPalInstalled(applicationContext) } returns true
+    fun `when getRecommendedPaymentMethods is called, succeeded event is sent`() {
```
scannillo, line 42: Looks like the test name here and on line 247 got mixed up?

tdchow: Shoot, good catch! let me update this.

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -103,16 +80,17 @@ class ShopperInsightsClient @VisibleForTesting internal constructor(
...
+
+            result?.eligibleMethods?.paypal == null && result?.eligibleMethods?.venmo == null -> {
+                callbackFailure(
+                    callback = callback,
+                    error = BraintreeException("Required fields missing from API response body")
                 )
             }
+
             else -> {
+                braintreeClient.sendAnalyticsEvent(GET_RECOMMENDED_PAYMENTS_SUCCEEDED)
```
scannillo, line 93: Nit - I know we talked about this is another PR, but maybe we should align with the other payment method features also move the result return & analytics firing for the success case into a helper method similar to how we're doing it for the error case

scannillo (approved):

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
tdchow: I had left out the success function since this is the only spot where we send the success event and invoke the success on the callback, but it looks like the other modules still have the success function. I'll update this to align!

PR #911, opened 2024-02-07
Venmo Universal/App Links

jaxdesmarais:
### Summary of changes

 -   Add `setFallbackToWeb()` to `VenmoRequest`
     - If set to `true` customers will fallback to a web based Venmo flow if the Venmo app is not installed
     - This method uses App Links instead of Deep Links
- Add `VenmoClient#parseBrowserSwitchResult(Context, Intent)` method
- Add `VenmoClient#clearActiveBrowserSwitchRequests(Context)` method
- Add `VenmoClient#onBrowserSwitchResult(BrowserSwitchResult, VenmoOnActivityResultCallback)` method
- This feature has been added to the Demo app, select "Venmo Fallback to Web" in the Settings menu to test this flow
- Add unit tests

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors

@jaxdesmarais @sarahkoop @jwarmkessel @tdchow


Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -428,6 +445,196 @@ private Intent getLaunchIntent(Configuration configuration, String profileId, St
...
+
+                        String resourceIdFromAppSwitch = "";
+                        String paymentMethodNonceFromAppSwitch = "";
+                        String usernameFromAppSwitch = "";
+                        if (resourceIdFromBrowserSwitch == null && paymentMethodNonceFromBrowserSwitch == null && usernameFromBrowserSwitch == null) {
+                            String cleanedAppSwitchUri = String.valueOf(deepLinkUri).replaceFirst("&","?");
+                            resourceIdFromAppSwitch = Uri.parse(String.valueOf(cleanedAppSwitchUri)).getQueryParameter("resource_id");
+                            paymentMethodNonceFromAppSwitch = Uri.parse(String.valueOf(cleanedAppSwitchUri)).getQueryParameter("payment_method_nonce");
+                            usernameFromAppSwitch = Uri.parse(String.valueOf(cleanedAppSwitchUri)).getQueryParameter("username");
+                        }
```
jaxdesmarais, line 478: This logic doesn't _feel_ great but is due to getting 2 slightly different deeplink URLs back from the App vs Web based flow. If there is a cleaner way to do this I am down for sure!

App based `deepLinkUri`:
```
com.braintreepayments.demo.braintree://x-callback-url/vzero/auth/venmo/success&username=a-username&payment_method_nonce=fake-venmo-account-nonce&resource_id=a-resource-id
```

Web based `deepLinkUri`:
```
com.braintreepayments.demo.braintree://x-callback-url/vzero/auth/venmo/success?resource_id=a-resource-id&username=a-username&payment_method_nonce=fake-venmo-account-nonce
```

Specifically we cannot get the query parameters from the App based flow since the deep link uri we get back starts with an `&` instead of a `?`. Right now we are doing a replace on the first `&` if present and parsing the rest of the uri as we do on the web.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* Venmo
+  * Add `setFallbackToWeb()` to `VenmoRequest`
+    * If set to `true` customers will fallback to a web based Venmo flow if the Venmo app is not installed
+    * This method uses App Links and is more secure than the Deep Link based flow
```
jaxdesmarais, line 8:
```suggestion
    * This method uses App Links instead of Deep Links
```

Demo/src/main/java/com/braintreepayments/demo/VenmoFragment.java
```diff
@@ -43,11 +48,13 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
     }
 
     private void handleVenmoResult(VenmoAccountNonce venmoAccountNonce) {
-        super.onPaymentMethodNonceCreated(venmoAccountNonce);
+        if (venmoAccountNonce != null) {
```
sarahkoop, line 51: Is this just fixing lint? If this is only called in the onSuccess listener method it shouldn't ever be null I think

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -163,10 +169,12 @@ public void onResult(@Nullable final Configuration configuration, @Nullable Exce
                 }
 
                 String exceptionMessage = null;
-                if (!configuration.isVenmoEnabled()) {
-                    exceptionMessage = "Venmo is not enabled";
-                } else if (!deviceInspector.isVenmoAppSwitchAvailable(activity)) {
-                    exceptionMessage = "Venmo is not installed";
+                if (!request.getFallbackToWeb()) {
+                    if (!configuration.isVenmoEnabled()) {
```
sarahkoop, line 173: The `enabled` check for the merchant account should happen even if `fallBackToWeb` is true right?

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -428,6 +445,196 @@ private Intent getLaunchIntent(Configuration configuration, String profileId, St
...
+                } else if (error != null) {
+                    listener.onVenmoFailure(error);
+                }
+            }
+        });
+
+        this.pendingBrowserSwitchResult = null;
+    }
+
+    private void startUniversalLinkFlow(FragmentActivity activity, VenmoIntentData input, Context context) throws JSONException, BrowserSwitchException {
```
sarahkoop, line 603: And `FragmentActivity` is an Android `Context` so I think we should be able to remove the extra `context` param and access all the required info from the `activity` param

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
sarahkoop: Is that `&` instead of `?` intentional or something the Venmo team could fix? Otherwise this logic seems like the best option - but maybe for clarity we could break out all this URL parsing into helper methods in this class like `parseAppBasedReturnUrl` and `parseWebBasedReturnUrl`? Or something like that

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -428,6 +445,196 @@ private Intent getLaunchIntent(Configuration configuration, String profileId, St
...
+            case BrowserSwitchStatus.SUCCESS:
+                Uri deepLinkUri = browserSwitchResult.getDeepLinkUrl();
+                if (deepLinkUri != null) {
+                    if (deepLinkUri.getPath().contains("success")) {
+                        braintreeClient.sendAnalyticsEvent("pay-with-venmo.browser-switch.success");
+                        String resourceIdFromBrowserSwitch = Uri.parse(String.valueOf(deepLinkUri)).getQueryParameter("resource_id");
+                        String paymentMethodNonceFromBrowserSwitch = Uri.parse(String.valueOf(deepLinkUri)).getQueryParameter("payment_method_nonce");
+                        String usernameFromBrowserSwitch = Uri.parse(String.valueOf(deepLinkUri)).getQueryParameter("username");
+
+                        String resourceIdFromAppSwitch = "";
```
sarahkoop, line 470:
```suggestion
                        String resourceIdFromAppSwitch;
```

Do we need these to be defined as empty strings? It seems like if not set they should be null

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -428,6 +445,196 @@ private Intent getLaunchIntent(Configuration configuration, String profileId, St
...
+                                @Override
+                                public void onResult(@Nullable VenmoAccountNonce venmoAccountNonce, @Nullable Exception error) {
+                                    if (venmoAccountNonce != null) {
+                                        callback.onResult(venmoAccountNonce, null);
+                                    } else {
+                                        callback.onResult(null, error);
+                                    }
+                                }
+                            });
+                        } else if ((paymentMethodNonceFromBrowserSwitch != null && usernameFromBrowserSwitch != null) || (paymentMethodNonceFromAppSwitch != null && usernameFromAppSwitch != null)) {
```
sarahkoop, line 498: Not sure if this is overkill but it might be helpful for future understanding to pull these if statements into helper methods too that say what these conditions mean - I'm reading this as if `returnToAppSuccessWithPaymentContext()` else if `returnToAppSuccessWithoutPaymentContext()` is that right?

Demo/src/main/java/com/braintreepayments/demo/VenmoFragment.java
jaxdesmarais: Ah, good catch, removed: https://github.com/braintree/braintree_android/pull/911/commits/1bd1e1d979c7d9673797c23ffe77aece150bdb8b

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: You're right, update: https://github.com/braintree/braintree_android/pull/911/commits/86ce5513e20bae65950237b13d8f42fce49024d1

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -428,6 +447,196 @@ private Intent getLaunchIntent(Configuration configuration, String profileId, St
...
+     * @param context The context used to clear pending browser switch requests
+     */
+    public void clearActiveBrowserSwitchRequests(@NonNull Context context) {
+        braintreeClient.clearActiveBrowserSwitchRequests(context);
+    }
+
+    void onBrowserSwitchResult(@NonNull BrowserSwitchResult browserSwitchResult) {
+        this.pendingBrowserSwitchResult = browserSwitchResult;
+        if (listener != null) {
+            // NEXT_MAJOR_VERSION: determine if browser switch logic can be further decoupled
```
sarahkoop, line 566: We can remove these NEXT_MAJOR_VERSION notes and maybe just have one that says "remove all manual browser switching methods" since we know what we will be doing in the next major version

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: The more you know. Lol. ✅ : https://github.com/braintree/braintree_android/pull/911/commits/52fa91869690ce39e690d252965f559242cd9419

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: It seems like the compiler doesn't like that syntax, but does like `String resourceIdFromAppSwitch = null;` for defaulting to null?

jaxdesmarais: Updated here: https://github.com/braintree/braintree_android/pull/911/commits/a25792e1bf8098124108d1d0c90455b1e94b71f8

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: Yeah, we could do that, I think this logic is mostly similar to the logic in `onVenmoResult` ([here](https://github.com/braintree/braintree_android/blob/fdd6bdc23e5a9c076ff800939d79096c9f64a0dc/Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java#L247-L296)). Agree it's a bit hard to reason about so if we think it'd be more clear to extract into a helper method I am down.

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: That's a good question if it's intentional or not. I can ask! Since we need all 3 things what kind of object would the helper methods return? Then we would need to parse the return types or duplicate the logic below. What do you think?

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -428,6 +447,195 @@ private Intent getLaunchIntent(Configuration configuration, String profileId, St
...
+
+    BrowserSwitchResult deliverBrowserSwitchResultFromNewTask(FragmentActivity activity) {
+        return braintreeClient.deliverBrowserSwitchResultFromNewTask(activity);
+    }
+
+    private void deliverBrowserSwitchResultToListener(final BrowserSwitchResult browserSwitchResult) {
+        onBrowserSwitchResult(browserSwitchResult, new VenmoOnActivityResultCallback() {
+            @Override
+            public void onResult(@Nullable VenmoAccountNonce venmoAccountNonce, @Nullable Exception error) {
+                if (venmoAccountNonce != null) {
```
sarahkoop, line 592: Let's check that listener is not null before either of these outcomes - if null we just can't return anything, but better that than to crash

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -428,6 +447,195 @@ private Intent getLaunchIntent(Configuration configuration, String profileId, St
...
+                } else if (error != null) {
+                    listener.onVenmoFailure(error);
+                }
+            }
+        });
+
+        this.pendingBrowserSwitchResult = null;
+    }
+
+    private void startUniversalLinkFlow(FragmentActivity activity, VenmoIntentData input) throws JSONException, BrowserSwitchException {
```
sarahkoop, line 603:
```suggestion
    private void startAppLinkFlow(FragmentActivity activity, VenmoIntentData input) throws JSONException, BrowserSwitchException {
```
Not sure if we want to align with iOS in naming but I prefer using the Android terminology here

Venmo/src/test/java/com/braintreepayments/api/VenmoActivityResultContractUnitTest.java
```diff
@@ -61,6 +62,17 @@ public void createIntent_returnsIntentWithExtras() throws JSONException {
         JSONAssert.assertEquals(expectedBraintreeData, new JSONObject(intent.getStringExtra(EXTRA_BRAINTREE_DATA)), JSONCompareMode.STRICT);
     }
 
+    @Test
+    public void createIntentForFallbackToWebWithVenmoInstalled_returnsIntentWithExtras() throws JSONException {
```
sarahkoop, line 66: The title of this test doesn't really match the assertion - should it be null or should we assert on the extras?

Venmo/src/test/java/com/braintreepayments/api/VenmoClientUnitTest.java
```diff
@@ -1496,4 +1523,82 @@ public void onVenmoResult_withPaymentContext_withFailedVaultCall_forwardsErrorTo
...
+
+        VenmoApi venmoApi = new MockVenmoApiBuilder()
+                .createNonceFromPaymentContextSuccess(nonce)
+                .build();
+
+        VenmoClient sut = new VenmoClient(activity, lifecycle, braintreeClient, venmoApi, sharedPrefsWriter, deviceInspector);
+        sut.pendingBrowserSwitchResult = browserSwitchResult;
+        sut.setListener(listener);
+
+        verify(listener).onVenmoSuccess(any(VenmoAccountNonce.class));
```
sarahkoop, line 1552:
```suggestion
        verify(listener).onVenmoSuccess(same(nonce));
```
Nitpick but just a stronger assertion

Venmo/src/test/java/com/braintreepayments/api/VenmoClientUnitTest.java
```diff
@@ -433,6 +437,29 @@ public void tokenizeVenmoAccount_whenVenmoNotInstalled_forwardsExceptionToListen
         verify(braintreeClient).sendAnalyticsEvent("pay-with-venmo.app-switch.failed");
     }
 
+    @Test
+    public void tokenizeVenmoAccount_whenFallbackToWebTrueAndAppInstalled_appSwitchIsSuccessful() {
```
sarahkoop, line 441: Feels like we are missing the other cases like `whenFallbackToWebTrueAndAppNotInstalled` - are those tested somewhere else?

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
sarahkoop: Also maybe we should make this method package private and @VisibleForTesting and test it directly? It seems like there is a lot of parsing logic for building the URL that could use a unit test

sarahkoop:
Left a few nitpick comments around naming and testing but otherwise 🔥

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: Oooo, yeah I had not thought of that, updated here: https://github.com/braintree/braintree_android/pull/911/commits/45b2a20f192b435f4ecc947c641d8e2ff6468ed4

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: Done: https://github.com/braintree/braintree_android/pull/911/commits/45b2a20f192b435f4ecc947c641d8e2ff6468ed4

Venmo/src/test/java/com/braintreepayments/api/VenmoActivityResultContractUnitTest.java
jaxdesmarais: I think this can be removed entirely, we had originally built the URL and appended it to the intent but don't have that logic in the VenmoAcitivityResultContract at all anymore. Removing this and will also update the missing test cases.

Venmo/src/test/java/com/braintreepayments/api/VenmoClientUnitTest.java
jaxdesmarais: Added some additional tests here: https://github.com/braintree/braintree_android/pull/911/commits/0a8e84284f11fd1308c528fdd1c5f2c0cb0c75be

Venmo/src/test/java/com/braintreepayments/api/VenmoClientUnitTest.java
jaxdesmarais: 👍 : https://github.com/braintree/braintree_android/pull/911/commits/84aa6b3a0599a1776f36fb0ecbf6ee14090af030

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
sarahkoop: Ah you're right there isn't a single return object, maybe is fine as is then 👍

sarahkoop (approved):

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: I ended up extracting the parsing logic into it's own methods (one for each value - nonce, username, resource id) here: https://github.com/braintree/braintree_android/pull/911/commits/b01835d10649518ab30edba7b588059970649444. I think it cleans this up a bit and hopefully makes it a bit more clear vs all living in this method. 🙏

tdchow (approved):

jaxdesmarais:
Waiting to merge this, there are some issues with the Venmo app return currently on both the standard flow and this new flow. Waiting for those to get resolved to do a final round of testing before merging.

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -163,10 +169,14 @@ public void onResult(@Nullable final Configuration configuration, @Nullable Exce
                 }
 
                 String exceptionMessage = null;
+                if (!request.getFallbackToWeb()) {
+                     if (!deviceInspector.isVenmoAppSwitchAvailable(activity)) {
```
scannillo, line 173: I thought we were going to keep the logic that checks if the app is installed before doing the switch, even if the switch is via app links?

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -428,6 +447,219 @@ private Intent getLaunchIntent(Configuration configuration, String profileId, St
...
+                .appendQueryParameter("customerClient", "MOBILE_APP")
+                .build();
+
+        BrowserSwitchOptions browserSwitchOptions = new BrowserSwitchOptions()
+                .requestCode(BraintreeRequestCodes.VENMO)
+                .url(venmoBaseURL)
+                .returnUrlScheme(braintreeClient.getReturnUrlScheme());
+
+        braintreeClient.startBrowserSwitch(activity, browserSwitchOptions);
+        braintreeClient.sendAnalyticsEvent("pay-with-venmo.browser-switch.started");
```
scannillo, line 660: I think we'll want a way to differentiate the old vs new switch methods in our analytics. Maybe that's in an `additionalDetails` tag? (It could also cover the PP checkout vs vault distinction we also wanted to cover at some point too)

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -428,6 +447,219 @@ private Intent getLaunchIntent(Configuration configuration, String profileId, St
         return venmoIntent;
     }
 
+    /**
+     * Use this method with the manual browser switch integration pattern.
+     *
+     * @param browserSwitchResult a {@link BrowserSwitchResult} with a {@link BrowserSwitchStatus}
+     * @param callback            {@link VenmoOnActivityResultCallback}
+     */
+    public void onBrowserSwitchResult(@NonNull BrowserSwitchResult browserSwitchResult, @NonNull final VenmoOnActivityResultCallback callback) {
```
scannillo, line 456: These new methods are only required for the app links flow (in the case of a web fallback) right? If so, can we call that out in the docstrings that if you opt-in to app links, you must implement these methods?

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: Are you thinking for analytics purposes like sending a new event here? We wouldn't want to set the exception message if we are falling back to web. On like 182 we check the value of this message and callback with the exception if it exists. For the web fallback case, we should expect the exception to be null below.

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: These methods would be for both the web fallback and app switch flows using app links, but _only_ for merchants using the manual browser switch pattern. The app links flow with app installed also uses this method (we get a different result back from the app links flow for both app install and web fallback).

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: I went with the "browser-switch" name because that's what all of the other modules use, but they only support 1 browser switch method vs app links that this module supports for both flows. But we could update this to "app-links" instead. This method is used for both the app and web fallback flow via app links.

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
scannillo: No not for analytics purposes. I thought the purpose of moving to universal links was to replace the URL scheme switch and hope for less drop-off with "unknown" errors that happen, not to remove the check for if the app is installed entirely.

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
scannillo: Ah - I'm remembering that this version is still on Arachne. So adding a new tag name in the POST body means an Arachne deploy ...

So adding a new event name to distinguish the 2 switch types is probably our best bet

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
scannillo: Gotcha! So maybe in our `VenmoRequest.useWebFallback` docstring we can call out that these must be implemented

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: Yeah, still on arachne, but it is a good point since we will need to port this to v5. The intent based flow (old flow) uses: `"pay-with-venmo.app-switch.<ACTION>"` as the event name so that naming is totally seperate. 

Do you think we should go with `"pay-with-venmo.browser-switch.<ACTION>"` to remain consistent across v4 modules or update this PR to use `"pay-with-venmo.app-link.<ACTION>"` to prepare for what we may want to call this in v5?

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: Yeah, we can call that out there for merchants using the manual browser switch pattern. Updated: https://github.com/braintree/braintree_android/pull/911/commits/33063cd16c6cfd7ac732913a0be99e7e6df7f51f

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: We shouldn't need the app installed check for this flow since we rely on the app links URL + OS to make this determination. We don't fallback to the old Intent flow if the app is not installed. In both cases for app links we hit `https://venmo.com/go/checkout` and the OS does the checks. I am not sure why we'd still want to do the check additionally.

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
scannillo: Great question! Hm - I lean toward moving it closer to the v5 naming convention. But really as long as we document it, either should be usable.

Maybe we can add a new tabbed section to our events Excel doc to call this out (since the existing tab is for V5+)

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
scannillo: Gotcha - I get that the check is not necessary to perform the switch. I just thought we were planning to keep it until the next major version. 

Keeping it would let us more accurately compare the performance of intents vs app link switches at the OS level.

scannillo: ^ This doesn't have to be a blocking comment, but this might come up w/ the Venmo team again

jaxdesmarais: 🤔 I may be completely misunderstanding what you are saying, so if I am happy to talk irl, but it doesn't look like we send anything in the check for `isVenmoAppSwitchAvailable` ([it calls this method under the hood](https://github.com/braintree/braintree_android/blob/fdd6bdc23e5a9c076ff800939d79096c9f64a0dc/BraintreeCore/src/main/java/com/braintreepayments/api/DeviceInspector.kt#L58-L64)) that would allow us to compare performance. Again, maybe I am completely misunderstanding what you're saying. 😅

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: Done: https://github.com/braintree/braintree_android/pull/911/commits/85114837ccefb1e73efef3ffeb02c709cb2e3d7a

Also added a new tab for Android v4 on the excel. Though these are the only events there for now, but at least we will have somewhere to point to for the event names if needed! Not sure if we plan to add iOS v5 + the rest of the Android v4 events at some point. Could be a good "tech debt" task at some point so we can point folks with questions on those versions somewhere.

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
scannillo: Totally down to chat IRL! Also don't let this convo block the PR.

What I mean is: I imagine we'll want to be comparing "conversion" rate for the intent flow versus the app link flow, so comparing # start to # end events. That comparison is more apples-to-apples if the intentional web fallback is removed from the equation, if both situations check for the app being installed before attempting to switch.

Let's say the app link flow (which currently doesn't check for app install) shows us awful conversion rates. We won't know if that's because the web fallback conversion stinks, _or_ the if the general app <--> app switch behavior on the app link rails is problematic and inconsistent.

scannillo: Apologies for the confusion! More of a general ❓

scannillo: To close this thread - we're going forward with the logic, as-is! For intentionally falling back to the web flow in the cases of no app installed.

Confirmed with Jax IRL and also with the Venmo product team.

scannillo: @jaxdesmarais Should I leave this thread unresolved until we can lang on an analytics piece? Also happy to 🏈 punt on the analytics piece here for a future PR too.

IRL we discussed that knowing what portion of users don't have the app installed will allow us to determine what % of users are being sent to the web fallback, or not.

jaxdesmarais: Do with think it'd be helpful to add analytics in the case of web fallback + app not installed? We could also add this in a future PR if we want more granular data on the flow. 🤔

jaxdesmarais: LOL, we asked the same question. I think I lean towards adding it in this PR as it could give us some insight between the two flows (intent vs app links) as well as the performance of app not installed + web fallback vs app installed + web fallback. The question is really do we want to send it for _all_ flows or just not installed + web fallback? What do you think?

scannillo (approved):
👏 Awesome test coverage

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -428,6 +447,219 @@ private Intent getLaunchIntent(Configuration configuration, String profileId, St
...
+                                        callback.onResult(venmoAccountNonce, null);
+                                    } else {
+                                        callback.onResult(null, error);
+                                    }
+                                }
+                            });
+                        } else if (paymentMethodNonce != null && username != null) {
+                            braintreeClient.sendAnalyticsEvent("pay-with-venmo.app-links.success");
+                            VenmoAccountNonce venmoAccountNonce = new VenmoAccountNonce(paymentMethodNonce, username, false);
+                            callback.onResult(venmoAccountNonce, null);
```
sarahkoop, line 487: Should there be vaulting logic here?

jaxdesmarais: Ah! You're absolutely right there should be. Made this update to match `onActivityResult`: https://github.com/braintree/braintree_android/pull/911/commits/4b538dc18c54c52203a24521fe0fcb1cc7ccb7a3

jaxdesmarais:
We have completed our test cases for this flow based on current versions of the Venmo app. The Venmo team will be completing backwards compatibility testing off of this branch. Once their tests are complete as well we should be able to remove the "do not merge" tag.

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -225,7 +235,16 @@ private void startVenmoActivityForResult(
         sharedPrefsWriter.persistVenmoVaultOption(activity, shouldVault);
         if (observer != null) {
             VenmoIntentData intentData = new VenmoIntentData(configuration, venmoProfileId, paymentContextId, braintreeClient.getSessionId(), braintreeClient.getIntegrationType());
-            observer.launch(intentData);
+            if (request.getFallbackToWeb()) {
+                try {
+                    startAppLinkFlow(activity, intentData);
+                } catch (JSONException | BrowserSwitchException exception) {
+                    braintreeClient.sendAnalyticsEvent("pay-with-venmo.app-links.failure");
```
scannillo, line 242: Just want to bump [this comment ](https://github.com/braintree/braintree_ios/pull/1193#issuecomment-1961935900)that we'll need to send paymentContextID for Android too.

jaxdesmarais: Yep, I added it to the tags confluence page as well as the android ticket for that workstream

PR #912, opened 2024-02-07
Add PayPal-Client-Metadata-Id to Shopper Insights API Call

tdchow:
### Summary of changes

 - Add the `PayPal-Client-Metadata-Id` header to the Shopper Insights API call
 - Pass `addtionalHeaders` all the way down to `BraintreeHttpClient` to be added to the request
 - Added named parameters to certain call sites to utilize the empty map default value while keeping the callback param the last in the list

### Checklist

[ ] Whenever, no rush

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
@@ -35,7 +35,6 @@ class ApiClient(private val braintreeClient: BraintreeClient) {
             sendPOST(
                 url = url,
                 data = paymentMethod.buildJSON().toString(),
-                additionalHeaders = emptyMap(),
             ) { responseBody, httpError ->
                 parseResponseToJSON(responseBody)?.let { json ->
                     sendAnalyticsEvent("card.rest.tokenization.success")
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
@@ -32,17 +32,19 @@ class ApiClient(private val braintreeClient: BraintreeClient) {
             paymentMethod.setSessionId(braintreeClient.sessionId)
 
             sendAnalyticsEvent("card.rest.tokenization.started")
-            sendPOST(url, paymentMethod.buildJSON().toString(), object : HttpResponseCallback {
-                override fun onResult(responseBody: String?, httpError: Exception?) {
-                    parseResponseToJSON(responseBody)?.let { json ->
-                        sendAnalyticsEvent("card.rest.tokenization.success")
-                        callback.onResult(json, null)
-                    } ?: httpError?.let { error ->
-                        sendAnalyticsEvent("card.rest.tokenization.failure")
-                        callback.onResult(null, error)
-                    }
+            sendPOST(
+                url = url,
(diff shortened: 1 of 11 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
```diff
@@ -32,17 +32,19 @@ class ApiClient(private val braintreeClient: BraintreeClient) {
...
-                        callback.onResult(json, null)
-                    } ?: httpError?.let { error ->
-                        sendAnalyticsEvent("card.rest.tokenization.failure")
-                        callback.onResult(null, error)
-                    }
+            sendPOST(
+                url = url,
+                data = paymentMethod.buildJSON().toString(),
+                additionalHeaders = emptyMap(),
+            ) { responseBody, httpError ->
```
tdchow, line 39: Converted the anonymous inner class to a lambda

BraintreeCore/src/main/java/com/braintreepayments/api/ApiClient.kt
```diff
@@ -32,17 +32,19 @@ class ApiClient(private val braintreeClient: BraintreeClient) {
...
-                        sendAnalyticsEvent("card.rest.tokenization.success")
-                        callback.onResult(json, null)
-                    } ?: httpError?.let { error ->
-                        sendAnalyticsEvent("card.rest.tokenization.failure")
-                        callback.onResult(null, error)
-                    }
+            sendPOST(
+                url = url,
+                data = paymentMethod.buildJSON().toString(),
+                additionalHeaders = emptyMap(),
```
sarahkoop, line 38: Do we need to set this to `emptyMap()` here since it's the default?

SharedUtils/src/main/java/com/braintreepayments/api/HttpResponseCallback.kt
```diff
@@ -7,7 +7,7 @@ import androidx.annotation.RestrictTo
  * @suppress
  */
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-interface HttpResponseCallback {
+fun interface HttpResponseCallback {
```
sarahkoop, line 10: Nice 😎 - tech debt item but we should really do this with all of the SDK interfaces

ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
```diff
@@ -8,33 +8,36 @@ internal class EligiblePaymentsApi(
 ) {
     fun execute(request: EligiblePaymentsApiRequest, callback: EligiblePaymentsCallback) {
         val jsonBody = request.toJson()
-        braintreeClient.getConfiguration { configuration, configError ->
+        braintreeClient.getConfiguration { configuration, _ ->
```
sarahkoop, line 11: Out of scope of this PR maybe but I think we are missing an outer layer of error handling - we should be checking that configuration is not null first otherwise calling back `configError` right?

tdchow: i noticed this as well. i'll create a follow up ticket for adding the missing error handling.

sarahkoop (approved):

jaxdesmarais (approved):
🚀

PR #913, opened 2024-02-08
[Shopper Insights] Demo App Parity

scannillo:
See similar [iOS PR](https://github.com/braintree/braintree_ios/pull/1185). The goal is to align our Demo app's functionality for QA testers to easily use.

### Summary of changes

 - Add Venmo & PayPal Vault buttons to ShopperInsights demo for use after `getRecommendedPaymentMethods()` returns a result
    - Add Venmo & PP listeners
    - Add navigation logic to nonce creation & transaction.sale() screen
    - Disable Venmo & PP buttons until `getRecommendedPaymentMethods()` returns `isRecommended` booleans
 - Add default email & phone test values to trigger mocked API response

### Notes

- This demo requires that Settings select "Client Token" as the auth type
    - We should consider returning an error to the merchant if they attempt to use a tokenization key in follow up PRs
- There's a mixed use of "shopper" and "shopping" terminology throughout the demo app. Let's stay consistent and use "shopper" only in a follow-up PR

### Compare iOS & Android Demos

| iOS      | Android |
| ----------- | ----------- |
|  ScreenRecording2024-02-07at5 58 07PM-ezgif com-video-to-gif-converter | ScreenRecording2024-02-07at5 56 38PM-ezgif com-video-to-gif-converter       |

### Checklist

 - ~Added a changelog entry~
 - ~Relevant test coverage~

### Authors
@scannillo

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragment.kt
+++ b/Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragment.kt
@@ -21,6 +21,8 @@ class ShopperInsightsFragment : BaseFragment() {
 
     private lateinit var responseTextView: TextView
     private lateinit var actionButton: Button
+    private lateinit var paypalVaultButton: Button
+    private lateinit var venmoButton: Button
     private lateinit var emailInput: TextInputLayout
     private lateinit var countryCodeInput: TextInputLayout
     private lateinit var nationalNumberInput: TextInputLayout
@@ -53,6 +55,10 @@ class ShopperInsightsFragment : BaseFragment() {
         nationalNumberInput = view.findViewById(R.id.nationalNumberInput)
         emailNullSwitch = view.findViewById(R.id.emailNullSwitch)
         phoneNullSwitch = view.findViewById(R.id.phoneNullSwitch)
+
+        emailInput.editText?.setText("PR1_merchantname@personal.example.com")
+        nationalNumberInput.editText?.setText("408-232-1001")
+        countryCodeInput.editText?.setText("1")
     }
 
     private fun setupActionButton() {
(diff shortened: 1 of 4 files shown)


scannillo:
For whatever reason the PaymentReady API was returning `false` for Venmo in the screenshots I recorded. I did test that the Venmo flow does properly return back to the Demo app.

Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragment.kt
```diff
@@ -36,48 +50,123 @@ class ShopperInsightsFragment : BaseFragment() {
...
+            ShopperInsightsRequest(email, null)
+        }
+
+        shopperInsightsClient.getRecommendedPaymentMethods(
+            request
+        ) { result ->
+            when (result) {
+                is ShopperInsightsResult.Success -> {
+                    payPalVaultButton.isEnabled = result.response.isPayPalRecommended
+                    venmoButton.isEnabled = result.response.isVenmoRecommended
```
tdchow, line 113: Would we want to show/hide the button based on if the payment method is recommended instead of setting it to enabled/disabled? I was thinking that merchants would probably show/hide the button for their integration.

tdchow (approved):
Changes look good! Just had a question on how we want to handle when a payment method is recommended in the demo app.

Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragment.kt
scannillo: Definitely - a merchant should not be disabling the Venmo or PayPal buttons 😆. Maybe we can 🏈 punt that for another PR and align with how iOS shows the buttons?

tdchow: Sounds good!

agedd (approved):
🚀

PR #915, opened 2024-02-09
Add PayPal Messaging Feature to Demo App

jaxdesmarais:
This work is going into the `paypal-messaging-feature` branch

### Summary of changes

 - Add `PayPalMessagingFragment` to Demo app
     - Add button to demo app to select this feature
     - Implement `PayPalMessagingView`
- Update `PayPalMessagingView` method and subclass `FrameLayout` so the class is a view type
    - Move context to constructor vs in start method as it is needed for `FrameLayout` subclass
    - Update CHANGELOG with these changes

### Checklist

 - [x] Added a changelog entry

### Authors

@jaxdesmarais

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/PayPalMessagingFragment.kt
+++ b/Demo/src/main/java/com/braintreepayments/demo/PayPalMessagingFragment.kt
@@ -1,0 +1,58 @@
+package com.braintreepayments.demo
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import com.braintreepayments.api.PayPalMessagingListener
+import com.braintreepayments.api.PayPalMessagingLogoType
+import com.braintreepayments.api.PayPalMessagingOfferType
+import com.braintreepayments.api.PayPalMessagingRequest
+import com.braintreepayments.api.PayPalMessagingTextAlignment
+import com.braintreepayments.api.PayPalMessagingView
+
+class PayPalMessagingFragment: BaseFragment(), PayPalMessagingListener {
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View? {
+        return inflater.inflate(R.layout.fragment_paypal_messaging, container, false)
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        val payPalMessagingRequest = PayPalMessagingRequest(
+            2.0,
+            null,
+            PayPalMessagingOfferType.PAY_LATER_LONG_TERM,
+            "US",
+            PayPalMessagingLogoType.PRIMARY,
+            PayPalMessagingTextAlignment.CENTER,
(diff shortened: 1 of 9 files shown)


Demo/src/main/java/com/braintreepayments/demo/PayPalMessagingFragment.kt
```diff
@@ -0,0 +1,68 @@
...
+            amount = 2.0,
+            placement = null,
+            offerType = PayPalMessagingOfferType.PAY_LATER_LONG_TERM,
+            buyerCountry = "US",
+            logoType = PayPalMessagingLogoType.PRIMARY,
+            textAlignment = PayPalMessagingTextAlignment.CENTER,
+            color = null
+        )
+
+        var payPalMessagingView = PayPalMessagingView(braintreeClient, requireActivity())
```
tdchow, line 37: Can we make this `var` a `val` here? It doesn't seem like we're resetting its value.

Demo/src/main/java/com/braintreepayments/demo/PayPalMessagingFragment.kt
```diff
@@ -0,0 +1,68 @@
...
+        print("Loading PayPalMessagingView")
+    }
+
+    override fun onPayPalMessagingSuccess() {
+        print("PayPalMessagingView displayed to user")
+    }
+
+    override fun onPayPalMessagingFailure(error: Exception) {
+        print("PayPalMessagingView returned the error:" + error.message)
+    }
```
tdchow, line 67: We might want to use `println()` here instead of `print()`.

`println()` will add a new line character to the end of the message: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/println.html

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingView.kt
```diff
@@ -16,16 +17,18 @@ import com.paypal.messages.config.message.PayPalMessageViewStateCallbacks
  * Note: **This module is in beta. It's public API may change or be removed in future releases.**
  * @property braintreeClient a {@link BraintreeClient}
```
tdchow, line 18: Would we want to update the kdoc with the `context` param?

Demo/src/main/java/com/braintreepayments/demo/PayPalMessagingFragment.kt
jaxdesmarais: Done: https://github.com/braintree/braintree_android/pull/915/commits/35b1b76d31195efd27639d320fbaa58265a52f7c

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingView.kt
jaxdesmarais: Good catch, updated: https://github.com/braintree/braintree_android/pull/915/commits/35b1b76d31195efd27639d320fbaa58265a52f7c

tdchow (approved):
Looks great! Thanks for making the updates!

sarahkoop (approved):

PR #920, opened 2024-02-13
SEPA Direct Debit FPTI Analytics

sarahkoop:
### Summary of changes

 - Implement SEPA FPTI analytics (this includes non-conversion events that will likely be moved to events details in future PR)
- Fix incorrect failure that should be cancel result

### Checklist

 - [x] Relevant test coverage

### Authors
@sarahkoop

### DIFF
--- a/SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitClient.java
+++ b/SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitClient.java
@@ -62,6 +62,7 @@ public class SEPADirectDebitClient {
                 (result, createMandateError) -> {
                     if (result != null) {
                         if (URLUtil.isValidUrl(result.getApprovalUrl())) {
+                            braintreeClient.sendAnalyticsEvent(SEPADirectDebitAnalytics.CREATE_MANDATE_CHALLENGE_REQUIRED);
                             try {
                                 SEPADirectDebitPaymentAuthRequestParams params =
                                         new SEPADirectDebitPaymentAuthRequestParams(buildBrowserSwitchOptions(result));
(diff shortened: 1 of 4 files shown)


SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitClient.java
```diff
@@ -56,53 +56,43 @@ public SEPADirectDebitClient(@NonNull Context context, @NonNull String authoriza
...
                             try {
                                 SEPADirectDebitPaymentAuthRequestParams params =
                                         new SEPADirectDebitPaymentAuthRequestParams(buildBrowserSwitchOptions(result));
+                                braintreeClient.sendAnalyticsEvent(SEPADirectDebitAnalytics.CREATE_MANDATE_SUCCEEDED);
                                 callback.onResult(new SEPADirectDebitPaymentAuthRequest.ReadyToLaunch(params));
                             } catch (JSONException exception) {
-                                braintreeClient.sendAnalyticsEvent(
-                                        "sepa-direct-debit.browser-switch.failure");
-                                callback.onResult(new SEPADirectDebitPaymentAuthRequest.Failure(exception));
+                                braintreeClient.sendAnalyticsEvent(SEPADirectDebitAnalytics.CREATE_MANDATE_FAILED);
```
tdchow, line 72: Do we want to add create mandate failed as a param instead of a separate event?

tdchow (approved):
Looks good! just had a question around sending multiple events for the same code path.

SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitClient.java
sarahkoop: Yes I just put these in to make it easier to know what extra info should be added to each of these events - they can be removed when the extra params are added to analytics (separate tickets)

sarahkoop: But also can remove the entirely if we think it's simpler that way

tdchow: That makes sense, no I think keeping the second event in there will help keep track of the data we need to pass.

jaxdesmarais (approved):

PR #921, opened 2024-02-14
Venmo Final Amount

jaxdesmarais:
### Summary of changes

 -   Add `setIsFinalAmount()` to `VenmoRequest`
     - This removes the text "Subject to change" as seen below

#### Visuals

| Final Amount = false (default) | Final Amount = true |
|-----|-----|
|Screenshot_20240214-084546|Screenshot_20240214-084609|

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors

@jaxdesmarais

### DIFF
--- a/Venmo/src/main/java/com/braintreepayments/api/VenmoApi.java
+++ b/Venmo/src/main/java/com/braintreepayments/api/VenmoApi.java
@@ -27,6 +27,7 @@ class VenmoApi {
             input.put("merchantProfileId", venmoProfileId);
             input.put("customerClient", "MOBILE_APP");
             input.put("intent", "CONTINUE");
+            input.put("isFinalAmount", request.getFinalAmountAsString());
             JSONObject paysheetDetails = new JSONObject();
             paysheetDetails.put("collectCustomerShippingAddress", request.getCollectCustomerShippingAddressAsString());
             paysheetDetails.put("collectCustomerBillingAddress", request.getCollectCustomerBillingAddressAsString());
--- a/Venmo/src/main/java/com/braintreepayments/api/VenmoApi.java
+++ b/Venmo/src/main/java/com/braintreepayments/api/VenmoApi.java
@@ -27,7 +27,7 @@ class VenmoApi {
             input.put("merchantProfileId", venmoProfileId);
             input.put("customerClient", "MOBILE_APP");
             input.put("intent", "CONTINUE");
-            input.put("isFinalAmount", request.getFinalAmountAsString());
+            input.put("isFinalAmount", request.getIsFinalAmountAsString());
             JSONObject paysheetDetails = new JSONObject();
             paysheetDetails.put("collectCustomerShippingAddress", request.getCollectCustomerShippingAddressAsString());
             paysheetDetails.put("collectCustomerBillingAddress", request.getCollectCustomerBillingAddressAsString());
--- a/Venmo/src/main/java/com/braintreepayments/api/VenmoRequest.java
+++ b/Venmo/src/main/java/com/braintreepayments/api/VenmoRequest.java
@@ -25,6 +25,7 @@ public class VenmoRequest implements Parcelable {
     private String taxAmount;
     private String shippingAmount;
     private ArrayList<VenmoLineItem> lineItems;
+    private boolean finalAmount;
 
     private final @VenmoPaymentMethodUsage int paymentMethodUsage;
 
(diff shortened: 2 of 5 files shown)


Venmo/src/main/java/com/braintreepayments/api/VenmoRequest.java
```diff
@@ -248,6 +249,22 @@ public ArrayList<VenmoLineItem> getLineItems() {
         return lineItems;
     }
 
+    /**
+     * @param finalAmount Optional - Indicates whether the purchase amount is the final amount.
+     *                    Defaults to false.
+     */
+    public void setFinalAmount(boolean finalAmount) {
```
sarahkoop, line 256: Nitpick - I feel like `setIsFinalAmount` would make it more clear that we are setting a boolean here, not a final amount dollar value - what do you think?

Also, I let's add a public getter for this too for consistency and Kotlin property interop

jaxdesmarais: That totally makes sense, I went back and fourth on it. Updated here: https://github.com/braintree/braintree_android/pull/921/commits/cbdecdc6a88094b13ed7543c474ec628ef68068f

Venmo/src/main/java/com/braintreepayments/api/VenmoApi.java
```diff
@@ -27,6 +27,7 @@ void createPaymentContext(@NonNull final VenmoRequest request, String venmoProfi
             input.put("merchantProfileId", venmoProfileId);
             input.put("customerClient", "MOBILE_APP");
             input.put("intent", "CONTINUE");
+            input.put("isFinalAmount", request.getIsFinalAmountAsString());
```
sarahkoop, line 30: Does this need to be a string? I think you an just put a boolean into the JSONObject

jaxdesmarais: The underlying API needs a string value, same as these two above that are bools on our end but strings for the API:
```java
paysheetDetails.put("collectCustomerShippingAddress", request.getCollectCustomerShippingAddressAsString());
paysheetDetails.put("collectCustomerBillingAddress", request.getCollectCustomerBillingAddressAsString());
```

Venmo/src/test/java/com/braintreepayments/api/VenmoApiUnitTest.java
```diff
@@ -168,6 +168,64 @@ public void createPaymentContext_whenGraphQLPostError_forwardsErrorToCallback()
         verify(callback).onResult(null, error);
     }
 
+    @Test
+    public void createPaymentContext_withTotalAmountAndSetsFinalAmountToTrue() throws JSONException {
```
sarahkoop, line 172: Does it work ok if total amount is not set and final amount is set to true? No error just doesn't show total amount at all?

tdchow (approved):

agedd (approved):
lgtm! 🚀

Venmo/src/test/java/com/braintreepayments/api/VenmoApiUnitTest.java
sarahkoop: The only use case I'd be concerned about is if total amount not set, final amount true (unsure what the behavior would be for that), but as long as it is still ok then, Probably every combination doesn't need a separate unit test

sarahkoop (approved):

Venmo/src/test/java/com/braintreepayments/api/VenmoApiUnitTest.java
jaxdesmarais: Yep, total amount is optional for both `VenmoPaymentMethodUsage.SINGLE_USE` and `VenmoPaymentMethodUsage.MULTI_USE`. I just tested both cases with only setting `setIsFinalAmount(true)` without `setTotalAmount` and there are no issues.

Venmo/src/main/java/com/braintreepayments/api/VenmoRequest.java
```diff
@@ -248,6 +249,29 @@ public ArrayList<VenmoLineItem> getLineItems() {
...
+        return isFinalAmount;
+    }
+
+    /**
+     * @return Whether or not the purchase amount is the final amount as a string value.
+     */
+    String getIsFinalAmountAsString() {
+        return String.valueOf(this.isFinalAmount);
+    }
+
```
jaxdesmarais, line 273:
```suggestion
```

PR #924, opened 2024-02-16
Venmo Universal Links v5

sarahkoop:
### Summary of changes

 - Implement Venmo universal links on v5
 - Removes fallbackToWeb parameter as this is the only available flow in v5
 - Update integration pattern to follow browser based launcher pattern instead of app (using app link is browser switch flow)

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors
@sarahkoop


Demo/src/main/java/com/braintreepayments/demo/VenmoFragment.java
```diff
@@ -71,32 +87,55 @@ public void launchVenmo(View v) {
 
         boolean shouldVault =
                 Settings.vaultVenmo(activity) && !TextUtils.isEmpty(Settings.getCustomerId(activity));
+        boolean fallbackToWeb = Settings.venmoFallbackToWeb(activity);
```
jaxdesmarais, line 90: nit: I don't think we need this at all anymore. Same with the settings changes since we now default to this flow.

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -230,6 +292,7 @@ public boolean isVenmoAppSwitchAvailable(@NonNull Context context) {
      * @param context  Android Context
      * @param callback {@link VenmoIsReadyToPayCallback}
      */
+    // TODO: How should this method behave with web fallback enabled?
```
sshropshire, line 295: I wonder if this method is still relevant. I'm not sure what the original intent was historically, but it seems like it's a good check to make before redirecting a user to Venmo's page in Google Play?

sshropshire (approved):

Demo/src/main/java/com/braintreepayments/demo/VenmoFragment.java
```diff
@@ -74,29 +90,51 @@ public void launchVenmo(View v) {
...
+                } else if (pendingRequest instanceof VenmoPendingRequest.Failure) {
+                    handleError(((VenmoPendingRequest.Failure) pendingRequest).getError());
                 }
-            });
-        } else {
-            showDialog("Please install the Venmo app first or confirm that Venmo is enabled for this merchant account");
-        }
+            }
+        });
+
```
jaxdesmarais, line 123:
```suggestion
```

Demo/src/main/res/xml/settings.xml
```diff
@@ -179,6 +179,12 @@
             android:summary="@string/venmo_fallback_to_web_summary"
             android:defaultValue="false" />
 
+        <CheckBoxPreference
+            android:key="venmo_fallback_to_web"
+            android:title="@string/venmo_fallback_to_web"
+            android:summary="@string/venmo_fallback_to_web_summary"
+            android:defaultValue="false" />
+
```
jaxdesmarais, line 187: Do we need this anymore since it's the default behavior?

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
```diff
@@ -132,9 +130,38 @@ private void createPaymentAuthRequest(
...
+                .buildUpon()
+                .appendQueryParameter("x-success", braintreeClient.getReturnUrlScheme() + "://x-callback-url/vzero/auth/venmo/success")
+                .appendQueryParameter("x-error", braintreeClient.getReturnUrlScheme() + "://x-callback-url/vzero/auth/venmo/error")
+                .appendQueryParameter("x-cancel", braintreeClient.getReturnUrlScheme() + "://x-callback-url/vzero/auth/venmo/cancel")
+                .appendQueryParameter("x-source", applicationName)
+                .appendQueryParameter("braintree_merchant_id", venmoProfileId)
+                .appendQueryParameter("braintree_access_token", configuration.getVenmoAccessToken())
+                .appendQueryParameter("braintree_environment", configuration.getVenmoEnvironment())
+                .appendQueryParameter("resource_id", paymentContextId)
+                .appendQueryParameter("braintree_sdk_data", braintreeData.toString())
```
jaxdesmarais, line 154: Note that we will want to make sure the metadata changes in both places from PR #945 get make in this PR as well.

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
jaxdesmarais: 🤔 I would think we can remove this now and just do the config check for `isVenmoEnabled` internally. With the fallback to web flow, we no longer care (nor should merchants) if `isVenmoAppSwitchAvailable` aside from if we want to send analytics in the future if the app is not installed. I would think we can remove this method entirely, but curious to hear what y'all think.

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
sarahkoop: Updated 👍

Demo/src/main/res/xml/settings.xml
sarahkoop: Ah no - removed

Venmo/src/main/java/com/braintreepayments/api/VenmoClient.java
sarahkoop: Yeah I think this can be removed, since it doesn't really provide any value anymore (shopper insights will likely replace this anyways)

jaxdesmarais (approved):
General question, will the changes to add `clientSdkMetadata` to `VenmoApi` get merged in next time we merge `main` into `v5`? Just want to make sure we don't miss that. This looks great! 👏

sarahkoop:
> General question, will the changes to add `clientSdkMetadata` to `VenmoApi` get merged in next time we merge `main` into `v5`? Just want to make sure we don't miss that. This looks great! 👏

Those changes actually already exist in `v5` from merging `main` into it 👍

PR #926, opened 2024-02-20
Cardinal Cleanup Method

jaxdesmarais:
### Summary of changes

 - Call Cardinal cleanup method to resolve memory leak from `Cardinal.getInstance`
     - Verified via Leak Canary that the `getInstance` memory leak is gone
     - Note: there is still a memory leak on `Cardinal.cca_continue` that will be resolve in a future Cardinal release

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors

@jaxdesmarais

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * Venmo
   * Add `setIsFinalAmount()` to `VenmoRequest`
+* ThreeDSecure
+  * Call cleanup method to resolve `Cardinal.getInstance` memory leak (fixes #898)
 
 ## 4.41.0 (2024-01-18)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * Venmo
   * Add `setIsFinalAmount()` to `VenmoRequest`
 * ThreeDSecure
-  * Call cleanup method to resolve `Cardinal.getInstance` memory leak (fixes #898)
+  * Call cleanup method to resolve `Cardinal.getInstance` memory leak
 
 ## 4.41.0 (2024-01-18)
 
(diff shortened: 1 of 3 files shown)


sshropshire (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 * Venmo
   * Add `setIsFinalAmount()` to `VenmoRequest`
+* ThreeDSecure
+  * Call cleanup method to resolve `Cardinal.getInstance` memory leak (fixes #898)
```
sarahkoop, line 8: Does this fix 898? I think that is the memory leak that is still open with Cardinal, but I could be wrong

jaxdesmarais: Ah you may be right, this only fixes _half_ of the memory leak, so maybe not worth calling out until we get a new version from Cardinal that fully fixes it. What do you think?

ThreeDSecure/src/main/java/com/braintreepayments/api/CardinalClient.java
```diff
@@ -58,6 +58,8 @@ void continueLookup(ThreeDSecureResult threeDSecureResult, CardinalChallengeObse
         } catch (RuntimeException e) {
             throw new BraintreeException("Cardinal SDK cca_continue Error.", e);
         }
+
+        Cardinal.getInstance().cleanup();
```
sarahkoop, line 62: I'm wondering if this should be invoked in the `onValidated` callback (line 37) instead? But if we tested and confirmed it resolves the leak when called here that's good with me too

CHANGELOG.md
jaxdesmarais, line 8:
```suggestion
  * Call cleanup method to resolve `Cardinal.getInstance` memory leak
```

ThreeDSecure/src/main/java/com/braintreepayments/api/CardinalClient.java
jaxdesmarais: When I was testing I noticed there was still a memory leak with it in `onValidated` since we call `Cardinal.getInstance().init` afterwards. I think actually having it down here may resolve both memory leaks but maybe I am doing something wrong with Leak Canary.

sarahkoop (approved):

PR #927, opened 2024-02-21
Cleanup PayPal Messaging Implementation

jaxdesmarais:
### Summary of changes

 - Move to `setListener()` pattern used in other modules
 - Remove unneeded else block 

### Checklist

### Authors

@jaxdesmarais

### DIFF
--- a/PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingView.kt
+++ b/PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingView.kt
@@ -22,7 +22,15 @@ class PayPalMessagingView(
     private val braintreeClient: BraintreeClient,
     context: Context
 ) : FrameLayout(context) {
-    var payPalMessagingListener: PayPalMessagingListener? = null
+    private var listener: PayPalMessagingListener? = null
+
+    /**
+     * Add a {@link PayPalMessagingListener} to your client to receive results or errors from the PayPal Messaging flow.
+     *
+     * @param listener a {@link PayPalMessagingListener}
+     */fun setListener(listener: PayPalMessagingListener) {
+        this.listener = listener
+    }
 
     /**
      * Creates a view to be displayed to promote offers such as Pay Later and PayPal Credit to customers.
@@ -88,7 +96,7 @@ class PayPalMessagingView(
 
         val viewStateCallbacks = PayPalMessageViewStateCallbacks(
             onLoading = {
-                payPalMessagingListener?.onPayPalMessagingLoading()
+                listener?.onPayPalMessagingLoading()
             },
             onError = { error ->
                 notifyFailure(error)
@@ -100,10 +108,10 @@ class PayPalMessagingView(
 
         val eventsCallbacks = PayPalMessageEventsCallbacks(
             onClick = {
-                payPalMessagingListener?.onPayPalMessagingClick()
+                listener?.onPayPalMessagingClick()
             },
             onApply = {
-                payPalMessagingListener?.onPayPalMessagingApply()
+                listener?.onPayPalMessagingApply()
             }
         )
 
@@ -119,11 +127,11 @@ class PayPalMessagingView(
 
     private fun notifySuccess() {
         braintreeClient.sendAnalyticsEvent(PayPalMessagingAnalytics.SUCCEEDED)
-        payPalMessagingListener?.onPayPalMessagingSuccess()
+        listener?.onPayPalMessagingSuccess()
     }
 
     private fun notifyFailure(error: Exception) {
         braintreeClient.sendAnalyticsEvent(PayPalMessagingAnalytics.FAILED)
-        payPalMessagingListener?.onPayPalMessagingFailure(error)
+        listener?.onPayPalMessagingFailure(error)
     }
 }
--- a/PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingView.kt
+++ b/PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingView.kt
@@ -51,7 +51,8 @@ class PayPalMessagingView(
                     )
                     notifyFailure(error = clientIdError)
                 } else {
-                    val payPalMessageView = constructPayPalMessageView(context, clientId, configuration, request)
+                    val payPalMessageView =
+                        constructPayPalMessageView(context, clientId, configuration, request)
                     payPalMessageView.layoutParams = ViewGroup.LayoutParams(
                         ViewGroup.LayoutParams.MATCH_PARENT,
                         ViewGroup.LayoutParams.MATCH_PARENT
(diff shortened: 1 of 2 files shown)


PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingView.kt
```diff
@@ -22,7 +22,15 @@ class PayPalMessagingView(
...
     context: Context
 ) : FrameLayout(context) {
-    var payPalMessagingListener: PayPalMessagingListener? = null
+    private var listener: PayPalMessagingListener? = null
+
+    /**
+     * Add a {@link PayPalMessagingListener} to your client to receive results or errors from the PayPal Messaging flow.
+     *
+     * @param listener a {@link PayPalMessagingListener}
+     */fun setListener(listener: PayPalMessagingListener) {
```
jaxdesmarais, line 31:
```suggestion
     */
     fun setListener(listener: PayPalMessagingListener) {
```

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingView.kt
```diff
@@ -22,7 +22,16 @@ class PayPalMessagingView(
     private val braintreeClient: BraintreeClient,
     context: Context
 ) : FrameLayout(context) {
-    var payPalMessagingListener: PayPalMessagingListener? = null
+    private var listener: PayPalMessagingListener? = null
```
sshropshire, line 25: Interesting I never realized that we make the listener value private. It's probably always been like this since the pre-Kotlin days.

We should consider moving to property-based access in `v5` to make the API more Kotlin friendly.

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingView.kt
```diff
@@ -22,7 +22,16 @@ class PayPalMessagingView(
...
+
+    /**
+     * Add a {@link PayPalMessagingListener} to your client to receive results or errors from the PayPal Messaging flow.
+     *
+     * @param listener a {@link PayPalMessagingListener}
+     */
+     fun setListener(listener: PayPalMessagingListener) {
+        this.listener = listener
+    }
```
sshropshire, line 34:
```suggestion
     fun setListener(listener: PayPalMessagingListener) {
        this.listener = listener
     }
```

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingView.kt
jaxdesmarais: Yeah I am down if we want to make that change across the board in v5. For now, just wanted to keep how this module sets the listener consistent with the other modules.

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingView.kt
```diff
@@ -43,19 +51,15 @@ class PayPalMessagingView(
...
                         ViewGroup.LayoutParams.MATCH_PARENT
                     )
 
                     addView(payPalMessageView)
                 }
-            } else {
-                val unknownError = BraintreeException(
-                    "Fetching Braintree configuration resulted in no error or configuration returned."
-                )
-                notifyFailure(error = unknownError)
```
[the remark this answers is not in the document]
sshropshire: Yeah this is a good catch. Technically it won't, but this code is good for completeness since it's theoretically possible for both values to be null (maybe through a bug of some sort).

It's definitely a code smell. Our callback pattern also has Pre-Kotlin roots. Ideally we'll move to a `sealed class Result` type that behaves similar to Swift's `Result` type that will allow us to guarantee at compile time that a `(null, null)` callback scenario never happens.

sshropshire (approved):
🌊

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingView.kt
sshropshire: Def yeah that's the right call. I only brought it up to make note.

jaxdesmarais: Oh for sure, it's also the more Swift-y approach which is almost certainly why I wrote it that way initially before looking at the other modules. 🙈

sarahkoop: Listeners are for the most part removed in v5, so I think with this module since we are kind of bridging v4 and v5 we can take whatever approach we want (assuming we keep the listener in v5 for this module only)

jaxdesmarais: Oooo interesting, would this be the more "kotlin" approach and `setListener` is the more java approach? I guess also is the a reason we wouldn't want both to exist or does the kotlin way get bridged somehow to a more java friendly interface? These may be super basic android questions. Lol.

sarahkoop (approved):

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingView.kt
sshropshire: As far as Java v. Kotlin here, the __set__ prefix e.g. `setProperty()` is more of a traditional naming convention in Java. Kotlin actually has `get` and `set` baked [into the language](https://www.baeldung.com/kotlin/getters-setters#getters-and-setters-for-properties).

It's an important callout here too because the Java style getters and setters are a big source of boilerplate that bloats code size. The compiler will synthesize Java-style getters and setters for any `val` or `var` Kotlin properties, so anyone consuming our libs in Java won't have to know the underlying implementation is Kotlin-based.

We still need to codify our Kotlin and Java style guides, but I lean on Swift API Naming Guidelines and the JetBrains [Kotlin Style Guide](https://kotlinlang.org/docs/coding-conventions.html#idiomatic-use-of-language-features). Right now it's good to bend the rules until we figure out what feels right for us.

jaxdesmarais: Oh nice! Yeah if under the hood we get what we need, I am down to revert the `setListener` change and keep the dot syntax. Is that what you are thinking @sshropshire?

sshropshire: TLDR from me, I think it's fine how you have it, but a good discussion to continue for `v5`.

jaxdesmarais: Cool cool 😅 - I'll keep it how it is in this PR for now and like you said as we solidify our style for v5 we can make updates.

PR #937, opened 2024-03-05
Convert Callback Interfaces to Kotlin

tdchow:
### Summary of changes

 - Convert all callbacks to Kotlin function interfaces
 - Rename `onResult` to more specific function names

### Checklist

[ ] Whenever, no rush


saperi22 (approved):

GooglePay/src/main/java/com/braintreepayments/api/GooglePayLauncher.java
```diff
@@ -49,7 +49,7 @@ public GooglePayLauncher(@NonNull ComponentActivity activity,
     GooglePayLauncher(ActivityResultRegistry registry, LifecycleOwner lifecycleOwner,
                       GooglePayLauncherCallback callback) {
         activityLauncher = registry.register(GOOGLE_PAY_RESULT, lifecycleOwner,
-                new GooglePayActivityResultContract(), callback::onResult);
+                new GooglePayActivityResultContract(), callback::onGooglePayLauncherResult);
```
jaxdesmarais, line 52: It looks like this one is public - do we need to call out this change in the CHANGELOG?

SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitClient.java
```diff
@@ -67,7 +67,7 @@ public void createPaymentAuthRequest(@NonNull final SEPADirectDebitRequest sepaD
                                 SEPADirectDebitPaymentAuthRequestParams params =
                                         new SEPADirectDebitPaymentAuthRequestParams(buildBrowserSwitchOptions(result));
                                 braintreeClient.sendAnalyticsEvent(SEPADirectDebitAnalytics.CREATE_MANDATE_SUCCEEDED);
-                                callback.onResult(new SEPADirectDebitPaymentAuthRequest.ReadyToLaunch(params));
+                                callback.onSEPADirectDebitPaymentAuthResult(new SEPADirectDebitPaymentAuthRequest.ReadyToLaunch(params));
```
jaxdesmarais, line 70: It looks like this one also may be public and worth calling out somewhere

GooglePay/src/main/java/com/braintreepayments/api/GooglePayLauncher.java
tdchow: Good catch! Just pushed an update to the changelog.

PR #942, opened 2024-03-12
Update American Express Package Name

tdchow:
### Summary of changes

 - Rename American Express package name from `com.braintreepayments.api` to `com.braintreepayments.americanexpress`
 - Add `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)` annotation to the Json class

### Checklist

 - [x] Added a changelog entry

[ ] Whenever, no rush

### DIFF
--- a/AmericanExpress/src/main/java/com/braintreepayments/americanexpress/AmericanExpressAnalytics.kt
+++ b/AmericanExpress/src/main/java/com/braintreepayments/americanexpress/AmericanExpressAnalytics.kt
@@ -1,4 +1,4 @@
-package com.braintreepayments.api
+package com.braintreepayments.americanexpress
 
 internal object AmericanExpressAnalytics {
 
--- a/SharedUtils/src/main/java/com/braintreepayments/api/Json.java
+++ b/SharedUtils/src/main/java/com/braintreepayments/api/Json.java
@@ -1,7 +1,10 @@
 package com.braintreepayments.api;
 
+import androidx.annotation.RestrictTo;
+
 import org.json.JSONObject;
 
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 public class Json {
 
     /**
@@ -15,6 +18,7 @@ public class Json {
      * @param fallback
      * @return {@link String}
      */
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
     public static String optString(JSONObject json, String name, String fallback) {
         if (json == null || json.isNull(name)) {
             return fallback;
@@ -23,7 +27,8 @@ public class Json {
         }
     }
 
-    static Boolean optBoolean(JSONObject json, String name, Boolean fallback) {
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    public static Boolean optBoolean(JSONObject json, String name, Boolean fallback) {
         if (json == null || json.isNull(name)) {
             return fallback;
         } else {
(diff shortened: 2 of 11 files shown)


sshropshire (approved):

AmericanExpress/src/main/java/com/braintreepayments/americanexpress/AmericanExpressAnalytics.kt
```diff
@@ -1,4 +1,4 @@
-package com.braintreepayments.api
+package com.braintreepayments.americanexpress
```
sshropshire, line 1: random: I wonder if we want to keep `com.braintreepayments.api` prefix since it's the group ID for the artifacts in Maven Central?

tdchow: Yeah, that's a good call. Let me rename to `com.braintreepayments.api.americanexpress`.

I was originally thinking that having `api` would be redundant, but I think we should stay aligned with the gradle group ID.

jaxdesmarais: ➕ 1 to keeping the group ID

SharedUtils/src/main/java/com/braintreepayments/api/Json.java
```diff
@@ -1,8 +1,11 @@
 package com.braintreepayments.api;
 
+import androidx.annotation.RestrictTo;
+
 import org.json.JSONObject;
 
-class Json {
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
jaxdesmarais, line 7: It seems like some of these restriction are 💥 some of the Unit Tests

AmericanExpress/src/main/java/com/braintreepayments/americanexpress/AmericanExpressAnalytics.kt
tdchow: i updated the package name to `com.braintreepayments.api.americanexpress`!

SharedUtils/src/main/java/com/braintreepayments/api/Json.java
tdchow: resolved the broken tests 👍

AmericanExpress/src/main/java/com/braintreepayments/api/americanexpress/AmericanExpressAnalytics.kt
```diff
@@ -1,4 +1,4 @@
-package com.braintreepayments.api
+package com.braintreepayments.api.americanexpress
```
jaxdesmarais, line 1: This is going to be a _super basic_ question, but what is the advantage to moving to appending `americanexpress`? 😅

jaxdesmarais: I also assume we will make this change in all of the modules and not just amex or no? Neither of these are blockers just for my own knowledge.

jaxdesmarais (approved):

AmericanExpress/src/main/java/com/braintreepayments/api/americanexpress/AmericanExpressAnalytics.kt
jaxdesmarais: Ah nice! That is super helpful. 🙏

tdchow: > I also assume we will make this change in all of the modules and not just amex or no?

Yep! We'll be updating the package name for all modules. I spit it up to make the PRs more digestible.

PR #944, opened 2024-03-14
Add optional param, payerEmail, to PayPalVaultRequest 

sarahkoop:
### Summary of changes

 - Separating out adding `payerEmail` to `PayPalVaultRequest` from the rest of the shopper insights feature branch 
 - This work was previously merged into feature branch in #918 

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors
@tdchow

### DIFF
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
@@ -38,10 +38,11 @@ public class PayPalVaultRequest extends PayPalRequest implements Parcelable {
      *
      * @param userAuthenticationEmail - email address of the payer
      */
-    public void setUserAuthenticationEmail(String userAuthenticationEmail) {
+    public void setUserAuthenticationEmail(@Nullable String userAuthenticationEmail) {
         this.userAuthenticationEmail = userAuthenticationEmail;
     }
 
+    @Nullable
     public String getUserAuthenticationEmail() {
         return this.userAuthenticationEmail;
     }
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
@@ -4,6 +4,8 @@ import android.os.Parcel;
 import android.os.Parcelable;
 import android.text.TextUtils;
 
+import androidx.annotation.Nullable;
+
 import org.json.JSONException;
 import org.json.JSONObject;
 
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
@@ -63,9 +63,7 @@ public class PayPalVaultRequest extends PayPalRequest implements Parcelable {
             parameters.put(DESCRIPTION_KEY, billingAgreementDescription);
         }
 
-        if (userAuthenticationEmail != null) {
-            parameters.put(PAYER_EMAIL_KEY, userAuthenticationEmail);
-        }
+        parameters.putOpt(PAYER_EMAIL_KEY, userAuthenticationEmail);
 
         JSONObject experienceProfile = new JSONObject();
         experienceProfile.put(NO_SHIPPING_KEY, !isShippingAddressRequired());
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
@@ -14,6 +14,8 @@ public class PayPalVaultRequest extends PayPalRequest implements Parcelable {
 
     private boolean shouldOfferCredit;
 
+    private String userAuthenticationEmail;
+
     public PayPalVaultRequest() {
     }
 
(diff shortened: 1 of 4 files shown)


PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
```diff
@@ -47,6 +63,10 @@ String createRequestBody(Configuration configuration, Authorization authorizatio
             parameters.put(DESCRIPTION_KEY, billingAgreementDescription);
         }
 
+        if (userAuthenticationEmail != null) {
```
jaxdesmarais, line 66: General non-blocking question, does the API throw any sort of error if we send an empty string in here or is java smart enough to equate an empty string to null?

jaxdesmarais (approved):

PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
```diff
@@ -30,6 +32,20 @@ public boolean getShouldOfferCredit() {
         return shouldOfferCredit;
     }
 
+    /**
+     * Optional: User email to initiate a quicker authentication flow in cases where the user has a
+     * PayPal Account with the same email.
+     *
+     * @param userAuthenticationEmail - email address of the payer
+     */
+    public void setUserAuthenticationEmail(String userAuthenticationEmail) {
```
sshropshire, line 41:
```suggestion
    public void setUserAuthenticationEmail(@Nullable String userAuthenticationEmail) {
```

PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
```diff
@@ -30,6 +32,20 @@ public boolean getShouldOfferCredit() {
...
+        this.userAuthenticationEmail = userAuthenticationEmail;
+    }
+
+    public String getUserAuthenticationEmail() {
```
sshropshire, line 45:
```suggestion
    @Nullable
    public String getUserAuthenticationEmail() {
```

sshropshire (approved):

PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
sarahkoop: I brought this up on the initial PR and @tdchow found that [it wasn't necessary](https://github.com/braintree/braintree_android/pull/918#discussion_r1486903583) - thoughts?

sshropshire: Oh ok maybe it's changed we should check the type signature of this property on the Kotlin side to be sure. Without annotations it used to default to `String!!` rather than `String?`.

PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
tdchow: I'd have to double-check, but my guess is that an empty string would be set on this JSON property. Do you know if we're checking for an empty string on iOS?

PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
```diff
@@ -47,6 +63,10 @@ String createRequestBody(Configuration configuration, Authorization authorizatio
             parameters.put(DESCRIPTION_KEY, billingAgreementDescription);
         }
 
+        if (userAuthenticationEmail != null) {
+            parameters.put(PAYER_EMAIL_KEY, userAuthenticationEmail);
+        }
```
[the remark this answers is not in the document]
tdchow: awesome! i wasn't aware of `putOpt()`.

PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
jaxdesmarais: 🤔 Just looked through the PR on iOS and we are only using `if let` which does the same thing you have (just checks that it's not nil). I guess maybe the question is _should_ we be checking that it's not just an empty string?

PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
tdchow: I just did some testing. Looks like there's a runtime exception thrown when trying to set `userAuthenticationEmail` to null. I thought I'd be good to ensure a merchant passes in a non null value here, and I thought we'd get a compile time check here. 

I'm good to add the nullable annotations to prevent this.

PayPal/src/main/java/com/braintreepayments/api/PayPalVaultRequest.java
tdchow: i don't feel strongly one way or the other. it wouldn't hurt to add the empty check, also the API probably ignores an empty email address as well.

maybe we leave it as is and pipe the empty string to the API? less clientside logic to maintain 🤔

jaxdesmarais: That makes sense, as long as the API doesn't throw an error if we pass in an empty string I am a-ok with keeping it as is

tdchow: Just tested the API sending in an empty string. No error is thrown!

jaxdesmarais: Perfect! I'll resolve this.

tdchow:
Live and Sandbox testing has passed!

PR #954, opened 2024-03-26
Add has user location consent to Data Collector and PayPal Data Collector

tdchow:
### Summary of changes

 - Update `DataCollector` and `PayPalDataCollector` to take in a new request object. This is an additive change, so all methods are overloaded to make this a non-breaking change.
 - Add `DataCollectorRequest` and `PayPalDataCollectorRequest` that contains a `hasUserLocationConsent` field
 
### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

[ ] Whenever, no rush


BraintreeDataCollector/src/main/java/com/braintreepayments/api/DataCollector.java
```diff
@@ -36,33 +36,35 @@ public DataCollector(@NonNull BraintreeClient braintreeClient) {
      * @param context  Android Context
      * @param callback {@link DataCollectorCallback}
      */
-    public void collectDeviceData(@NonNull Context context, @NonNull DataCollectorCallback callback) {
-        collectDeviceData(context, null, callback);
+    public void collectDeviceData(
```
sarahkoop, line 39: Can be future PR but we should also deprecate this if we want merchants to pass in the request every time

BraintreeDataCollector/src/main/java/com/braintreepayments/api/DataCollectorRequest.kt
```diff
@@ -0,0 +1,5 @@
+package com.braintreepayments.api
+
+data class DataCollectorRequest(
```
sarahkoop, line 3: Also can be TODO but doc strings for this

.github/workflows/release_snapshot.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - 'main'
+  workflow_dispatch:
```
tdchow, line 7: Looks like this needs to be merge into main for the run workflow button to show up in github actions. Should we open a separate PR for this?

BraintreeDataCollector/src/test/java/com/braintreepayments/api/DataCollectorUnitTest.java
```diff
@@ -76,4 +77,22 @@ public void collectDeviceData_getsDeviceDataJSONWithCorrelationIdFromPayPal() th
         JSONObject json = new JSONObject(deviceData);
         assertEquals("sample_correlation_id", json.getString("correlation_id"));
     }
+
+    @Test
+    public void collectDeviceData_without_DataCollectorRequest_sets_hasUserLocationConsent_to_false() {
```
sarahkoop, line 82: I think we should have a test for with DataCollectorRequest passes the value from the request (to test the true case)

BraintreeDataCollector/src/main/java/com/braintreepayments/api/DataCollector.java
tdchow: 👍 let me add the deprecation annotation

BraintreeDataCollector/src/main/java/com/braintreepayments/api/DataCollectorRequest.kt
scannillo: +1 to define what user "consent" means

BraintreeDataCollector/src/main/java/com/braintreepayments/api/DataCollector.java
```diff
@@ -76,16 +78,42 @@ public void onResult(@Nullable Configuration configuration, @Nullable Exception
...
+     * <p>
+     * Use the return value on your server, e.g. with `Transaction.sale`.
+     *
+     * @param context    Android Context
+     * @param merchantId Optional - Custom Kount merchant id. Leave blank to use the default.
+     * @param callback   {@link DataCollectorCallback}
+     * @deprecated Kount is officially deprecated, use {@link PayPalDataCollector#collectDeviceData(Context, String, PayPalDataCollectorCallback)} instead.
+     */
+    @Deprecated
+    public void collectDeviceData(
```
tdchow, line 95: I didn't add an overloaded method for this one since it's deprecated.

BraintreeDataCollector/src/main/java/com/braintreepayments/api/DataCollectorRequest.kt
tdchow: 👍 i have a todo item for double checking javadocs and kdocs

.github/workflows/release_snapshot.yml
sarahkoop: Yeah 👍

tdchow: I just updated the workflow to trigger off of the new feature branch. As a longer term solution, we could still open up the PR adding this to main.

PayPalDataCollector/src/main/java/com/braintreepayments/api/PayPalDataCollectorInternalRequest.java
```diff
@@ -7,17 +7,25 @@
 /**
  * Used to configuration the PayPalDataCollector request
  */
-class PayPalDataCollectorRequest {
+class PayPalDataCollectorInternalRequest {
```
scannillo, line 10: Instead of renaming this and making a new `PayPalDataCollectorRequest` class. Could we instead just only publicly expose the `hasUserLocationConsent` bool on here?

BraintreeDataCollector/src/test/java/com/braintreepayments/api/DataCollectorUnitTest.java
tdchow: Added the missing test case

PayPalDataCollector/src/main/java/com/braintreepayments/api/PayPalDataCollectorInternalRequest.java
tdchow: I kinda like the clearer separation between internal and merchant-facing classes. Once we expose one field, we'll have a hard time ever refactoring this class if we need. I'm also not sure how the constructor would look once we convert the request class to Kotlin 🤔 

But definitely open to this if folks feel like it's the better route.

sarahkoop (approved):
Approving for feature branch snapshot release - will re-review for PR to main

PayPalDataCollector/src/main/java/com/braintreepayments/api/PayPalDataCollectorInternalRequest.java
sshropshire: I'm +1 for the sake of simplicity. Internal properties on `PayPalDataCollectorRequest` will result in less code since it's a Kotlin data class, which is a nice win for maintainability.

PayPalDataCollector/src/main/java/com/braintreepayments/api/PayPalDataCollectorRequest.kt
```diff
@@ -0,0 +1,5 @@
+package com.braintreepayments.api
+
+data class PayPalDataCollectorRequest(
+    val hasUserLocationConsent: Boolean
```
sshropshire, line 4: 🧶 : thought about `isLocationDataCollectionConsentGrantedByUser: Boolean`. It's an obnoxiously long variable name, but leading with is leans in to some default Java / Kotlin interop behavior with [getters and setters](https://kotlinlang.org/docs/java-interop.html#getters-and-setters). Total take-or-leave, but wanted to point it out for good measure.

sshropshire (approved):

.github/workflows/release_snapshot.yml
tdchow: I needed to add `workflow_dispatch:` back in to run the workflow.

PayPalDataCollector/src/main/java/com/braintreepayments/api/PayPalDataCollectorRequest.kt
tdchow: That's a good call. I think in this case, there wouldn't be a setter since `hasUserLocationConsent` is defined as a `val` and a merchant wouldn't realistically need to get the value from the request object since they'd be passing it into the SDK. However, I could see merchants using the getter in a unit test.

I'm in favor of sticking with this naming since it aligns with the MXO and Magnes SDKs.

sshropshire: Got it yeah that makes sense. The part without the setter is key I hadn't realized it at first 👍 .

PayPalDataCollector/src/main/java/com/braintreepayments/api/PayPalDataCollectorInternalRequest.java
tdchow: I'm going to merge the PR in its current state but will open a separate PR into the feature branch if we find a solution that works 👍

sshropshire: Ok cool. Yeah basically the top-level constructor is made internal, and the convenience constructor is made public. We do this a lot [on PPCP](https://github.com/paypal/paypal-android/blob/cf95ac7982b529f3caf25afd901e5687dfd8e810/PayPalWebPayments/src/main/java/com/paypal/android/paypalwebpayments/PayPalWebCheckoutClient.kt#L14).

tdchow: Thanks for sharing this example! I think the addition of another constructor might get confusing for internal use, as that public constructor would need to set a default value for the `internalRequiredParam`. This use case might be slightly different as it looks like we have an internal constructor for dependency injection purposes.

PR #956, opened 2024-03-28
Add has user location consent param in PayPal, PayPal Native, and Local Payment modules

tdchow:
### Summary of changes

 - Add `hasUserLocationConsent` param to PayPal, PayPal Native, and Local Payment modules
 - Update missing changelog for Data Collector and PayPal Data Collector

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

[ ] Whenever, no rush

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@
 * PayPalNativeCheckout
   * Add required property `hasUserLocationConsent` to `PayPalNativeCheckoutRequest`, `PayPalNativeVaultRequest` and `PayPalNativeCheckoutRequest`
   * Deprecate existing constructors that do not pass in `hasUserLocationConsent`
+* BraintreeDataCollector
+  * Add `DataCollectorRequest` to pass in `hasUserLocationConsent`
+  * Update `DataCollector.collectDeviceData()` to take in `DataCollectorRequest`
+  * Deprecate existing `DataCollector.collectDeviceData()`
+* PayPalDataCollector
+  * Add `PayPalDataCollectorRequest` to pass in `hasUserLocationConsent`
+  * Update `PayPalDataCollector.collectDeviceData()` to take in `PayPalDataCollectorRequest`
+  * Deprecate existing `PayPalDataCollector.collectDeviceData()`
 
 ## 4.43.0 (2024-03-19)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* Local Payment
+  * Add required property `hasUserLocationConsent` to `LocalPaymentRequest`
+  * Deprecate existing constructor that does not pass in `hasUserLocationConsent`
+* PayPal
+  * Add required property `hasUserLocationConsent` to `PayPalCheckoutRequest`, `PayPalVaultRequest` and `PayPalRequest`
+  * Deprecate existing constructors that do not pass in `hasUserLocationConsent`
+* PayPalNativeCheckout
+  * Add required property `hasUserLocationConsent` to `PayPalNativeCheckoutRequest`, `PayPalNativeVaultRequest` and `PayPalNativeCheckoutRequest`
+  * Deprecate existing constructors that do not pass in `hasUserLocationConsent`
+
 ## 4.43.0 (2024-03-19)
 
 * Move from Braintree to PayPal analytics service (FPTI)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,23 +2,34 @@
 
 ## unreleased
 
+
 * Local Payment
-  * Add required property `hasUserLocationConsent` to `LocalPaymentRequest`
-  * Deprecate existing constructor that does not pass in `hasUserLocationConsent`
+  * Fixes Google Play Store Rejection
+    * See [developer documentation](link) for required updates for user location data compliance
+    * Add `hasUserLocationConsent` property to `LocalPaymentRequest`
+    * Deprecate existing constructor that does not pass in `hasUserLocationConsent`
 * PayPal
-  * Add required property `hasUserLocationConsent` to `PayPalCheckoutRequest`, `PayPalVaultRequest` and `PayPalRequest`
-  * Deprecate existing constructors that do not pass in `hasUserLocationConsent`
+  * Fixes Google Play Store Rejection
+    * See [developer documentation](link) for required updates for user location data compliance
+    * Add `hasUserLocationConsent` property to `PayPalCheckoutRequest`, `PayPalVaultRequest` and `PayPalRequest`
+    * Deprecate existing constructors that do not pass in `hasUserLocationConsent`
 * PayPalNativeCheckout
-  * Add required property `hasUserLocationConsent` to `PayPalNativeCheckoutRequest`, `PayPalNativeVaultRequest` and `PayPalNativeCheckoutRequest`
-  * Deprecate existing constructors that do not pass in `hasUserLocationConsent`
+  * Fixes Google Play Store Rejection
+    * See [developer documentation](link) for required updates for user location data compliance
+    * Add `hasUserLocationConsent` property to `PayPalNativeCheckoutRequest`, `PayPalNativeVaultRequest` and `PayPalNativeCheckoutRequest`
+    * Deprecate existing constructors that do not pass in `hasUserLocationConsent`
 * BraintreeDataCollector
-  * Add `DataCollectorRequest` to pass in `hasUserLocationConsent`
-  * Update `DataCollector.collectDeviceData()` to take in `DataCollectorRequest`
-  * Deprecate existing `DataCollector.collectDeviceData()`
+  * Fixes Google Play Store Rejection
+    * See [developer documentation](link) for required updates for user location data compliance
+    * Add `DataCollectorRequest` to pass in `hasUserLocationConsent`
+    * Update `DataCollector.collectDeviceData()` to take in `DataCollectorRequest`
+    * Deprecate existing `DataCollector.collectDeviceData()`
 * PayPalDataCollector
-  * Add `PayPalDataCollectorRequest` to pass in `hasUserLocationConsent`
-  * Update `PayPalDataCollector.collectDeviceData()` to take in `PayPalDataCollectorRequest`
-  * Deprecate existing `PayPalDataCollector.collectDeviceData()`
+  * Fixes Google Play Store Rejection
+    * See [developer documentation](link) for required updates for user location data compliance
+    * Add `PayPalDataCollectorRequest` to pass in `hasUserLocationConsent`
+    * Update `PayPalDataCollector.collectDeviceData()` to take in `PayPalDataCollectorRequest`
+    * Deprecate existing `PayPalDataCollector.collectDeviceData()`
 
 ## 4.43.0 (2024-03-19)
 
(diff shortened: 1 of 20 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,25 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* Local Payment
```
jaxdesmarais, line 5: I know historically we've added something along the lines of "fixes Google Play store blah blah violation". May be helpful here for merchants/internal folks to see what version fixes this.

LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentRequest.java
```diff
@@ -45,6 +45,22 @@ public class LocalPaymentRequest {
...
+
+    /**
+     * Deprecated. Use {@link LocalPaymentRequest#LocalPaymentRequest(boolean)} instead.
+     **/
+    @Deprecated
+    public LocalPaymentRequest() {
+        hasUserLocationConsent = false;
+    }
+
+    /**
```
jaxdesmarais, line 58: Should these exist in this PR?

LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentRequest.java
```diff
@@ -206,6 +222,10 @@ public String getSurname() {
         return surname;
     }
 
+    public boolean hasUserLocationConsent() {
```
jaxdesmarais, line 225: I think we will want to docstring this as well

PayPal/src/main/java/com/braintreepayments/api/PayPalCheckoutRequest.java
```diff
@@ -57,8 +59,26 @@ public class PayPalCheckoutRequest extends PayPalRequest implements Parcelable {
...
+    /**
+     * @param amount The transaction amount in currency units (as * determined by setCurrencyCode).
+     *               For example, "1.20" corresponds to one dollar and twenty cents. Amount must be a non-negative
+     *               number, may optionally contain exactly 2 decimal places separated by '.' and is
+     *               limited to 7 digits before the decimal point.
+     *               <p>
+     *               This amount may differ slightly from the transaction amount. The exact decline rules
+     *               for mismatches between this client-side amount and the final amount in the Transaction
+     *               are determined by the gateway.
+     * TODO: add javadoc for hasUserLocationConsent param
```
jaxdesmarais, line 78: Same here on missing docs

jaxdesmarais (commented):
I added a couple comments but not sure if the updated public docs should exist in this PR or if that will be another follow up?

CHANGELOG.md
tdchow: Good call, let me add a bullet for that. Would it make sense to make these bullets a sub-point under the "fixes Google Play Store ..." bullet?

LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentRequest.java
tdchow: We don't have the approved copy yet. I feel that having them merged into the feature branch will save time once we have the copy. But happy to remove them if folks feel differently!

LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentRequest.java
tdchow: I saw that the above getter functions do not have any docstrings. I feel that merchants would not realistically use these getters, since they'd be the ones to instantiate these request objects. We could probably mark them as internal, but maybe a larger discussion outside the scope of this PR.

Would we still want to add a docstring here?

CHANGELOG.md
sarahkoop: Yeah I would say we should mirror whatever wording we are going to put in the dev docs in the changelog as well something like 
```
* Fixes Google Play Store Rejection
  * See [developer documentation](link) for required updates for user location data compliance
```

LocalPayment/src/test/java/com/braintreepayments/api/LocalPaymentClientUnitTest.java
```diff
@@ -764,6 +764,42 @@ public void clearActiveBrowserSwitchRequests_forwardsInvocationToBraintreeClient
...
+
+        LocalPaymentRequest request = getIdealLocalPaymentRequest();
+        LocalPaymentClient sut = new LocalPaymentClient(activity, lifecycle, braintreeClient, payPalDataCollector, localPaymentApi);
+
+        sut.startPayment(request, localPaymentStartCallback);
+
+        sut.setListener(listener);
+        sut.onBrowserSwitchResult(activity, browserSwitchResult);
+
+        verify(payPalDataCollector).getClientMetadataId(any(), eq(payPalEnabledConfig), eq(true));
```
sshropshire, line 800:
```suggestion
        verify(payPalDataCollector).getClientMetadataId(any(), same(payPalEnabledConfig), eq(true));
```
nit: seems more semantically correct to verify that we're forwarding a reference to the `getClientMetadataId()` invocation.

sshropshire (approved):

LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentRequest.java
jaxdesmarais: 🤔 interesting, it looks like we are pretty inconsistent across the SDK, for example [3DS has them docstringed](https://github.com/braintree/braintree_android/blob/6f2324098431753ac74778edaa28d2f7cf4d1b1c/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java#L389) but other requests don't. I'm curious what other folks think or if this is tech debt that we should consider addressing in v5 (either docstring them all or making them internal).

CHANGELOG.md
jaxdesmarais: What Sarah said makes sense to me!

LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentRequest.java
jaxdesmarais: Ah gotcha, yeah if we are waiting on copy to address the docstrings in a future PR before merging the feature into main that's 👍 to me.

CHANGELOG.md
```diff
@@ -1,5 +1,25 @@
...
+  * Add required property `hasUserLocationConsent` to `LocalPaymentRequest`
```
sarahkoop, line 6: It's not a required property in this major version though right? That would be a breaking change

tdchow: Ah yeah, it's not a required property since the old constructor still exists. I'll update the wording here.

LocalPayment/src/test/java/com/braintreepayments/api/LocalPaymentClientUnitTest.java
tdchow: Good call! i've updated to use `same()` instead.

CHANGELOG.md
tdchow: Sounds good. I've updated the wording in the PR. Will have a follow up PR once we get the approved copy.

LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentRequest.java
tdchow: I think tackling this in v5 makes sense but also open to what other folks think

sarahkoop: +1 to adding this to the v5 backlog

sarahkoop: I _think_ the getters need to exist for Kotlin interop (making the property accessible), but not 100% sure on that, and if we convert these request objects to Kotlin in v5 might be resolved on it's own

tdchow: Yep, I thought about having the getter named `isHasUserLocationConsent()` which would make it property accessible in Kotlin, but the only use of this property is in Java files. I think you're right in that when we do the conversion to Kotlin, the issue will be resolved.

saperi22 (approved):

PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
```diff
@@ -70,13 +70,17 @@ public abstract class PayPalRequest implements Parcelable {
...
     private String riskCorrelationId;
     private final ArrayList<PayPalLineItem> lineItems;
+    private final boolean hasUserLocationConsent;
 
     /**
      * Constructs a request for PayPal Checkout and Vault flows.
+     *
+     * TODO: add javadoc for hasUserLocationConsent param
      */
-    public PayPalRequest() {
```
sarahkoop, line 77: This should be deprecated not removed I think

sarahkoop: Likely should be non-public in v5 since I don't think merchants ever use it but still _technically_ a breaking change

PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutVaultRequest.java
```diff
@@ -14,7 +14,19 @@ public class PayPalNativeCheckoutVaultRequest extends PayPalNativeRequest implem
 
     private boolean shouldOfferCredit;
 
+    /**
+     * Deprecated. Use {@link PayPalNativeCheckoutVaultRequest#PayPalNativeCheckoutVaultRequest(boolean)} instead.
+     */@Deprecated
+
```
sarahkoop, line 20:
```suggestion
```

sarahkoop, line 20:
```suggestion
    /**
     * Deprecated. Use {@link PayPalNativeCheckoutVaultRequest#PayPalNativeCheckoutVaultRequest(boolean)} instead.
     */
     @Deprecated
```

LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentRequest.java
sarahkoop: Oh but don't we want merchants to be able to set this property if they are integrating in Kotlin?

PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
tdchow: That was my thought in removing the empty constructor. I'll add the deprecated empty constructor back.

Do we always handle breaking changes for flows that are potentially misuse of the SDK?

LocalPayment/src/main/java/com/braintreepayments/api/LocalPaymentRequest.java
tdchow: Since it's a required field, I was thinking it'd be defined as a `val` in the constructor when we convert to a Kotlin data class. So it wouldn't have a separate setter function generated.

I may have prematurely made this assumption 😅 Is this the direction we'd want to go in?

PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
sarahkoop: Historically I think we have still treated misuse as breaking changes, and added `NEXT_MAJOR_VERSION` notes to resolve, but open to the rest of the teams thoughts on that

sarahkoop (approved):
Docstrings to be added when confirmed across teams 👍

PR #968, opened 2024-04-08
Update Shared Utils Package Name

tdchow:
### Summary of changes

 - Update Shared Utils package name to `com.braintreepayments.api.sharedutils`

### Checklist

 - [x] Added a changelog entry


SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/HttpNoResponse.java
```diff
@@ -0,0 +1,11 @@
+package com.braintreepayments.api.sharedutils;
+
+import androidx.annotation.RestrictTo;
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public class HttpNoResponse implements HttpResponseCallback {
+
+    @Override
+    public void onResult(String responseBody, Exception httpError) {
```
richherrera, line 9: Should we add a _TODO_ for future implementations, or describe if something is missing?

richherrera (approved):

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/HttpNoResponse.java
tdchow: Shoot, this shouldn't be showing up as a new class. This was an existing class that was moved to the new package.

The only change was making the class public and adding the `@RestrictTo()` annotation.

PR #970, opened 2024-04-10
Update braintree core package name

saperi22:
### Summary of changes

 - Rename core package to `com.braintreepayments.api.core`

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
```diff
@@ -73,8 +74,13 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      */
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
     constructor(context: Context, authorization: String) :
-            this(BraintreeOptions(context = context, authorization = Authorization.fromString
-                (authorization)))
+            this(
+                BraintreeOptions(
+                    context = context,
+                    authorization = Authorization.fromString
+                        (authorization)
```
sshropshire, line 81:
```suggestion
                    authorization = Authorization.fromString(authorization)
```
Just a qq: I've been seeing formatting like this a lot lately. I was wondering is there an automated tool that causes these lines to break like this?

sshropshire (approved):
🌊

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
saperi22: IDK. I think I converted the file from Java to Kotlin and that resulted in some weird formatting.
I've tried to reformat code, and looks like this is an accepted format for the tool.

I'd say it is probably the combination of the two.
Wrong formatting by code converter and not fixing by existing formatter.

sshropshire: Got it yeah it's probably something in our toolchain. I think its cool since it's consistent, but it may be a good thing for us to revisit as a team in the future to improve maintainability / readability.

jaxdesmarais (approved):

richherrera (approved):
🥇

PR #976, opened 2024-04-15
Send `link_type` FPTI tag

richherrera:
### Summary of changes

 -  Send `link_type` in `event_params` to PayPal's analytics service (FPTI) - this will help us determine if users are going through the `universal link` or `deeplink` flow
 - Updated UnitTests

### Checklist

 - [X] Added a changelog entry
 - [X] Relevant test coverage

### Authors

- @richherrera 

Screenshot 2024-04-18 at 2 06 48 p m


BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
```diff
@@ -188,6 +189,7 @@ internal class AnalyticsClient @VisibleForTesting constructor(
             val singleEventJSON = JSONObject()
                 .put(EVENT_NAME_KEY, analyticsEvent.name)
                 .putOpt(PAYPAL_CONTEXT_ID_KEY, analyticsEvent.payPalContextId)
+                .putOpt(PAYPAL_CONTEXT_ID_KEY, analyticsEvent.linkType)
```
jaxdesmarais, line 192:
```suggestion
                .putOpt(LINK_TYPE_KEY, analyticsEvent.linkType)
```

jaxdesmarais (commented):
Couple of questions:

1. Have we had a chance to verify that we are seeing this as expected in FPTI?
2. Are we able to add a test case for the "universal" link type?

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
richherrera: Good catch!!

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.kt
```diff
@@ -5,13 +5,15 @@ import androidx.room.AutoMigration
...
+import androidx.sqlite.db.SupportSQLiteDatabase
 
 // Ref: https://developer.android.com/training/data-storage/room/migrating-db-versions
 @Database(
-        version = 2,
+        version = 3,
         entities = [AnalyticsEvent::class],
         autoMigrations = [
-            AutoMigration(from = 1, to = 2)
+            AutoMigration(from = 2, to = 3)
```
saperi22, line 16: AFAIK you _add_ to the auto migration, never replace. Are you sure you want to replace?
What if we have a user on version 1 and they skip several updates to update to this version. How would their db migrate?

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.kt
```diff
@@ -23,15 +25,24 @@ internal abstract class AnalyticsDatabase : RoomDatabase() {
...
+                        database.execSQL("ALTER TABLE analytics_event ADD COLUMN link_type TEXT")
+                    }
+                }
+
                 val instance = Room.databaseBuilder(
                     context.applicationContext,
                     AnalyticsDatabase::class.java,
                     "analytics_database"
-                ).build()
+                ).addMigrations(MIGRATION_2_3).build()
```
saperi22, line 45: We didn't have a migration before this, but here too, I'd think we'd want to have a 1-2, 2-3 migration. Am I missing something? 🤔

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.kt
richherrera: Thanks for the heads up! I've added a database migration to ensure functionality for merchants who might upgrade directly from version 1 to version 3. This approach is based on the Kotlin [documentation](https://developer.android.com/training/data-storage/room/migrating-db-versions) for database management. Since we are modifying the `AnalyticsEvent` entity, it's necessary to increase the database version number; otherwise, the SDK will throw an exception like this:

_java.lang.IllegalStateException: Room cannot verify the data integrity. Looks like you've changed schema but forgot to update the version number. You can simply fix this by increasing the version number._ 

With the version number incremented, we need to perform a database migration from v2 to v3 to avoid an exception like:

_java.lang.IllegalStateException: A migration from 2 to 3 was required but not found. Please provide the necessary Migration path via RoomDatabase.Builder.addMigration(Migration ...) or allow for destructive migrations via one of the RoomDatabase.Builder.fallbackToDestructiveMigration* methods._

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.kt
richherrera: In this specific case (migrating the database from version 2 to version 3), a manual migration was necessary, as outlined in the [documentation](https://developer.android.com/training/data-storage/room/migrating-db-versions#manual). That’s why we don't have any records of past migrations. The exception we resolved with this approach is as follows:

java.lang.IllegalStateException: Migration didn't properly handle: analytics_event(com.braintreepayments.api.AnalyticsEvent).
 Expected:
TableInfo{name='analytics_event', columns={name=Column{name='name', type='TEXT', affinity='2', notNull=true, primaryKeyPosition=0, defaultValue='null'}, link_type=Column{name='link_type', type='TEXT', affinity='2', notNull=false, primaryKeyPosition=0, defaultValue='null'}, _id=Column{name='_id', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=1, defaultValue='null'}, paypal_context_id=Column{name='paypal_context_id', type='TEXT', affinity='2', notNull=false, primaryKeyPosition=0, defaultValue='null'}, timestamp=Column{name='timestamp', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=0, defaultValue='null'}}, foreignKeys=[], indices=[]}
 Found:
TableInfo{name='analytics_event', columns={name=Column{name='name', type='TEXT', affinity='2', notNull=true, primaryKeyPosition=0, defaultValue='null'}, _id=Column{name='_id', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=1, defaultValue='null'}, paypal_context_id=Column{name='paypal_context_id', type='TEXT', affinity='2', notNull=false, primaryKeyPosition=0, defaultValue='null'}, timestamp=Column{name='timestamp', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=0, defaultValue='null'}}, foreignKeys=[], indices=[]}


It appears that the migration does not recognize that a new column was added to the table. In the manual migration, we are indicating that the column `link_type` was added to the table `analytics_event`.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.kt
sarahkoop: +1 to Sai's concern - I think what we want to do is is create a new auto migration from 2 to 3 (i.e. copy the AutoMigration from 1 to 2 and add another below it for the changes from 2 to 3)

saperi22 (changes requested):
Revert changes made to schema 2.json

richherrera:
> Revert changes made to schema 2.json

Reverted it, thanks for your help!!!

saperi22 (approved):

jaxdesmarais (approved):

PR #980, opened 2024-04-26
Add appLinkReturnUri to BraintreeClient

tdchow:
### Summary of changes

 - Add `appLinkReturnUri` as a parameter passed to the `BraintreeClient` constructor
 - Remove unused `BraintreeClient` constructors that were marked as internal but only used in test classes
 - Collapse public `BraintreeClient` constructors by using default values and `@JvmOverloads` annotation

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
@@ -41,7 +41,12 @@ open class BraintreeClient @VisibleForTesting internal constructor(
     private val manifestValidator: ManifestValidator,
     private val returnUrlScheme: String,
     private val braintreeDeepLinkReturnUrlScheme: String,
-    private val appLinkReturnUri: Uri?,
+
+    /**
+     * @suppress
+     */
+    @get:RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    val appLinkReturnUri: Uri?,
 ) {
 
     private val crashReporter: CrashReporter
@@ -84,7 +89,7 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      * @param authorization   The tokenization key or client token to use. If an invalid
      * authorization is provided, a [BraintreeException] will be returned via callback.
      * @param returnUrlScheme A custom return url to use for browser and app switching
-     * @param appLinkReturnUri TODO
+     * @param appLinkReturnUri A [Uri] containing the Android App Link to use for app switching
      */
     @JvmOverloads
     constructor (
@@ -115,7 +120,7 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      * @param clientTokenProvider An implementation of [ClientTokenProvider] that [BraintreeClient]
      * will use to fetch a client token on demand.
      * @param returnUrlScheme     A custom return url to use for browser and app switching
-     * @param appLinkReturnUri TODO
+     * @param appLinkReturnUri    A [Uri] containing the Android App Link to use for app switching
      */
     @JvmOverloads
     constructor(
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
@@ -89,7 +89,10 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      * @param authorization   The tokenization key or client token to use. If an invalid
      * authorization is provided, a [BraintreeException] will be returned via callback.
      * @param returnUrlScheme A custom return url to use for browser and app switching
-     * @param appLinkReturnUri A [Uri] containing the Android App Link website associated with your application to be used to return to your app from browser or app switch based payment flows. This Android App Link will only be used for payment flows where `useAppLinkReturn` is set to `true`.
+     * @param appLinkReturnUri A [Uri] containing the Android App Link website associated with your
+     * application to be used to return to your app from browser or app switch based payment flows.
+     * This Android App Link will only be used for payment flows where `useAppLinkReturn` is set to
+     * `true`.
      */
     @JvmOverloads
     constructor (
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
@@ -7,7 +7,6 @@ import android.net.Uri
 import androidx.annotation.RestrictTo
 import androidx.annotation.VisibleForTesting
 import androidx.fragment.app.FragmentActivity
-import com.braintreepayments.api.IntegrationType.Integration
 
 /**
  * Core Braintree class that handles network requests.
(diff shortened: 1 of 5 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -126,44 +115,20 @@ open class BraintreeClient @VisibleForTesting internal constructor(
...
     ) : this(
         BraintreeOptions(
             context = context,
             clientTokenProvider = clientTokenProvider,
-            sessionId = sessionId,
-            integrationType = integrationType,
-        )
-    )
-
-    internal constructor(
```
tdchow, line 156: Removed internal constructor that was only referenced in a unit test.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -126,44 +115,20 @@ open class BraintreeClient @VisibleForTesting internal constructor(
...
-        returnUrlScheme: String
-    ) : this(
-        BraintreeOptions(
-            context = context,
-            clientTokenProvider = clientTokenProvider,
-            returnUrlScheme = returnUrlScheme
-        )
-    )
-
-    internal constructor(
```
tdchow, line 142: Removed internal constructor that was only referenced in a unit test.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -126,44 +115,20 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      * @param clientTokenProvider An implementation of [ClientTokenProvider] that [BraintreeClient]
      * will use to fetch a client token on demand.
      * @param returnUrlScheme     A custom return url to use for browser and app switching
+     * @param appLinkReturnUri TODO
      */
+    @JvmOverloads
     constructor(
         context: Context,
         clientTokenProvider: ClientTokenProvider,
-        returnUrlScheme: String
```
tdchow, line 133: The signature of the constructors would be changing, but it wouldn't be a breaking change as the `returnUrlScheme` param goes from non-null to nullable. It's possible that a merchant can now pass in null, but this shouldn't break anything as the internal default value was set to null.

I feel that this greatly reduced the number of constructors we have to manage, along with making it much easier to add additional params like the new `appLinkReturnUri`.

```kotlin
    constructor (
        context: Context,
        authorization: String,
        returnUrlScheme: String
    )
```
would now be
```kotlin
    constructor (
        context: Context,
        authorization: String,
        returnUrlScheme: String?
    )
```

Is there anything that I'm overlooking here? I want to ensure this is a viable change before moving forward with additional app link changes.

sarahkoop: What does this look like for a Java integration? I think that would be my only point of concern - if they were previously using `BraintreeClient(context, authorization)`, do they now need to pass `(context, authorization, null)` ?

tdchow: The Java integration that uses `BraintreeClient(context, authorization)` would be unchanged. Null does not need to be passed because of the default value in Kotlin.

Here's the decompiled Java code that's generated:

```Java
   @JvmOverloads
   public BraintreeClient(@NotNull Context context, @NotNull String authorization, @Nullable String returnUrlScheme, @Nullable Uri appLinkReturnUri) {
      Intrinsics.checkNotNullParameter(context, "context");
      Intrinsics.checkNotNullParameter(authorization, "authorization");
      Object var5 = null;
      Object var6 = null;
      this(new BraintreeOptions(context, (String)null, returnUrlScheme, appLinkReturnUri, authorization, (ClientTokenProvider)var6, (String)var5, 98, (DefaultConstructorMarker)null));
   }

   @JvmOverloads
   public BraintreeClient(@NotNull Context context, @NotNull String authorization, @Nullable String returnUrlScheme) {
      this(context, (String)authorization, returnUrlScheme, (Uri)null, 8, (DefaultConstructorMarker)null);
   }

   @JvmOverloads
   public BraintreeClient(@NotNull Context context, @NotNull String authorization) {
      this(context, (String)authorization, (String)null, (Uri)null, 12, (DefaultConstructorMarker)null);
   }
```

The only gotcha for Java integrations is if a merchant wants to pass in an `appLinkReturnUri`, null must be passed in for the `returnUrlScheme` param. It doesn't look like `BraintreeClient(@NotNull Context context, @NotNull String authorization, @Nullable Uri appLinkReturnUri)` is generated. Would this be okay?

The only other solution I see is to double the number of public constructors which would take us from the existing 4 to 8 😟

sarahkoop: That seems ok to me 👍

sshropshire (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
tdchow: Yep, I agree! I believe the `BraintreeClient` will be hidden from the merchant, so we'll have to add these URLs to the respective request objects for v5.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -84,7 +89,7 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      * @param authorization   The tokenization key or client token to use. If an invalid
      * authorization is provided, a [BraintreeException] will be returned via callback.
      * @param returnUrlScheme A custom return url to use for browser and app switching
-     * @param appLinkReturnUri TODO
+     * @param appLinkReturnUri A [Uri] containing the Android App Link to use for app switching
```
sarahkoop, line 92:
```suggestion
     * @param appLinkReturnUri A [Uri] containing the Android App Link website associated with your application to be used to return to your app from browser or app switch based payment flows. This Android App Link will only be used for payment flows where `useAppLinkReturn` is set to `true`.
```

Can workshop the doc strings but maybe we should be a little more specific since the integration isn't completely intuitive.

sarahkoop:
Let's also add the ongoing CHANGELOG entry for the BraintreeClient changes (unless you want to do all CHANGELOG at the end)

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
tdchow: I like what you suggested here!

tdchow:
> Let's also add the ongoing CHANGELOG entry for the BraintreeClient changes (unless you want to do all CHANGELOG at the end)

Good catch! I completely forgot about the changelog. Making the update now.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -126,44 +120,20 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      * @param clientTokenProvider An implementation of [ClientTokenProvider] that [BraintreeClient]
      * will use to fetch a client token on demand.
      * @param returnUrlScheme     A custom return url to use for browser and app switching
+     * @param appLinkReturnUri    A [Uri] containing the Android App Link to use for app switching
```
sarahkoop, line 123:
```suggestion
     * @param appLinkReturnUri A [Uri] containing the Android App Link website associated with your application to be used to return to your app from browser or app switch based payment flows. This Android App Link will only be used for payment flows where `useAppLinkReturn` is set to `true`.
```

Same docs change

PR #895, opened 2024-01-23
Paypal Messaging Implementation

jaxdesmarais:
This work is going into to `paypal-messaging-feature`

### Summary of changes

- Add `PayPalMessagingRequest`, `PayPalMessagingColor`, `PayPalMessagingLogoType`, `PayPalMessagingOfferType`, `PayPalMessagingPlacement`, `PayPalMessagingTextAlignment`, and `PayPalMessagingListener`
- Add `PayPalMessagingView` as the main entry point for merchants

Follow up PRs will include:
- Testing
- Analytics
- Demo App

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

PR #981, opened 2024-04-29
Bump PayPal Messages to v1.0.0 develop.2

richherrera:
### Summary of changes

- Bump PayPalMessages to `1.0.0-develop.2`
- Rename `PayPalMessagingPlacement` to `PayPalMessagingPageType`
   - Update  `PayPalMessagingPageType. rawValue `
- Rename `PayPalMessagingLogoType` enum case `.simple` to: `.none`
- Set up PayPal Messages Analytics

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors

- @richherrera 
- @jwarmkessel


BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -354,6 +354,14 @@ open class BraintreeClient @VisibleForTesting internal constructor(
         return browserSwitchClient.getResultFromCache(context)
     }
 
+    /**
+     * @suppress
+     */
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    fun getVersionSDK(): String {
```
tdchow, line 361: Instead of adding a getter to the BraintreeClient, would it be cleaner to add the build config field to the PayPal messaging module? 

example: https://github.com/braintree/braintree_android/blob/6b5d8c62977c7c9db5404488d37cc0d31b940f28/GooglePay/build.gradle#L16

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingPageType.kt
```diff
@@ -0,0 +1,56 @@
...
+     * Checkout view
+     */
+    CHECKOUT,
+
+    /**
+     * Search results
+     */
+    SEARCH_RESULTS;
+
+    internal val rawValue: PayPalMessagePageType
```
tdchow, line 46: Is this value to map to an internal `PayPalMessagePageType` type in the messaging SDK? If it is, what are your thoughts on renaming to something like `internalValue`? I wasn't quite sure what `rawValue` meant in this case.

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingRequest.kt
```diff
@@ -17,10 +17,10 @@ package com.braintreepayments.api
  */
 data class PayPalMessagingRequest(
     var amount: Double? = null,
```
sshropshire, line 19: random q: Is amount truly optional here? If yes, would a value like `0.0` work better? If no, I think it makes sense to remove the optional and require merchants to explicitly specify an amount.

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingRequest.kt
```diff
@@ -17,10 +17,10 @@ package com.braintreepayments.api
...
-    var placement: PayPalMessagingPlacement? = null,
+    var pageType: PayPalMessagingPageType? = null,
     var offerType: PayPalMessagingOfferType? = null,
```
sshropshire, line 21: Similar question with these values, if we can set default values for these properties instead of `null` that would make it a bit cleaner.

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingPageType.kt
```diff
@@ -0,0 +1,56 @@
+package com.braintreepayments.api
+
+import com.paypal.messages.config.PayPalMessagePageType
+
+/**
+ * Message location within an application
+ * Note: **This module is in beta. It's public API may change or be removed in future releases.**
+ */
+enum class PayPalMessagingPageType {
```
sarahkoop, line 9: Can we use the "@RequiresOptIn" annotation for all public beta classes in this module in addition to the doc strings notes?

sarahkoop: Ah I see there is a separate ticket for this - so doesn't need to be in this PR 👍

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
richherrera: Thanks!! updated here: [ab41ceb](https://github.com/braintree/braintree_android/pull/981/commits/ab41ceba0cc7d61fb87600567da9b1f3d0614e91)

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingPageType.kt
richherrera: Yes, both work for me, I think `rawValue` was used since it is the most common in iOS, but if `internalValue` is more Android, it's fine for me

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingRequest.kt
richherrera: I'm not sure, @jaxdesmarais. Could you help me clarify this, please? Happy to make any necessary changes to this :D

jaxdesmarais: The amount is optional and is displayed on the popup if provided, so we won't want to default to `0.0` or make it required. We are defaulting to `null` here to match the [underlying SDK](https://github.com/paypal/paypal-messages-android/blob/d3293c07ed758724700d48642f1159e576c748e3/library/src/main/java/com/paypal/messages/config/message/PayPalMessageData.kt#L17) default.

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingPageType.kt
jaxdesmarais: `rawValue` was suggested by @sshropshire [here](https://github.com/braintree/braintree_android/pull/895#discussion_r1465292799). I am not tied to it, so if there is a better representation for this on Android we should use it!

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingRequest.kt
jaxdesmarais: We are defaulting to `null` here as well because that's what the [underlying SDK does](https://github.com/paypal/paypal-messages-android/blob/d3293c07ed758724700d48642f1159e576c748e3/library/src/main/java/com/paypal/messages/config/message/PayPalMessageData.kt#L19-L20). I don't think we want to provide defaults here since the underlying SDK defaults to null.

Demo/src/main/java/com/braintreepayments/demo/PayPalMessagingFragment.kt
```diff
@@ -26,12 +26,11 @@ class PayPalMessagingFragment : BaseFragment(), PayPalMessagingListener {
 
         val payPalMessagingRequest = PayPalMessagingRequest(
             amount = 2.0,
-            placement = null,
+            pageType = null,
             offerType = PayPalMessagingOfferType.PAY_LATER_LONG_TERM,
             buyerCountry = "US",
             logoType = PayPalMessagingLogoType.PRIMARY,
-            textAlignment = PayPalMessagingTextAlignment.CENTER,
-            color = null
```
jaxdesmarais, line 34: Is `color` no longer needed?

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingRequest.kt
sshropshire: Yeah part of wrapping the SDK though is to allow us to enforce API best practices where it makes sense, and null values aren't the best when a default value is possible. Though if `null` does have semantic meaning I'd agree we should keep. Just making sure we aren't moving on autopilot here.

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingPageType.kt
sshropshire: That suggestion was to remove a redundant prefix. I could see `internalValue` making more since on Android. Raw Value is more of a Swift enum concept brought over from iOS.

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingRequest.kt
jaxdesmarais: For sure - yeah it seems like `null` does have a semantic meaning for the messaging SDK so we can roll with that as well vs forcing a default on merchants. We wouldn't want to specify a default page type as any default may not actually be accurate for where a merchant may place this notice, so null makes sense there. Same with offer type we wouldn't want to specify that if a merchant doesn't want to use a default as that would determine what offer type their customer is shown.

PayPalMessaging/src/main/java/com/braintreepayments/api/PayPalMessagingPageType.kt
richherrera: Updated here: e5b283ddf75275e451fbd15034045484b8af9312

Demo/src/main/java/com/braintreepayments/demo/PayPalMessagingFragment.kt
richherrera: `Color` is still `required`; in fact, it's now `required`. I had previously removed it so it would `default` to the value specified in the `PayPalMessagingRequest` constructor. I've added it back

tdchow (approved):

PR #983, opened 2024-04-30
Update PayPalInternalClient to use Pairing ID as Client Metadata ID

sshropshire:
### Summary of changes

 - Update PayPalInternalClient to set Client Metadata ID equal to Pairing ID internally to gain feature parity with iOS 

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

@sshropshire

### DIFF
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
@@ -75,7 +75,8 @@ class PayPalInternalClient {
                                                     String clientMetadataId = payPalRequest.getRiskCorrelationId();
                                                     if (clientMetadataId == null) {
                                                         PayPalDataCollectorInternalRequest dataCollectorRequest =
-                                                                new PayPalDataCollectorInternalRequest(payPalRequest.hasUserLocationConsent());
+                                                                new PayPalDataCollectorInternalRequest(payPalRequest.hasUserLocationConsent())
+                                                                        .setApplicationGuid(payPalDataCollector.getPayPalInstallationGUID(context));
                                                         dataCollectorRequest.setRiskCorrelationId(pairingId);
 
                                                         clientMetadataId = payPalDataCollector.getClientMetadataId(context, dataCollectorRequest, configuration);
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
@@ -71,8 +71,15 @@ class PayPalInternalClient {
 
                                                     String pairingIdKey = isBillingAgreement ? "ba_token" : "token";
                                                     String pairingId = parsedRedirectUri.getQueryParameter(pairingIdKey);
-                                                    String clientMetadataId = payPalRequest.getRiskCorrelationId() != null
-                                                            ? payPalRequest.getRiskCorrelationId() : payPalDataCollector.getClientMetadataId(context, configuration, payPalRequest.hasUserLocationConsent());
+
+                                                    String clientMetadataId = payPalRequest.getRiskCorrelationId();
+                                                    if (clientMetadataId == null) {
+                                                        PayPalDataCollectorInternalRequest dataCollectorRequest =
+                                                                new PayPalDataCollectorInternalRequest(payPalRequest.hasUserLocationConsent());
+                                                        dataCollectorRequest.setRiskCorrelationId(pairingId);
+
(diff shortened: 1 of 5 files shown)


PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
```diff
@@ -122,4 +130,12 @@ public void onResult(JSONObject tokenizationResponse, Exception exception) {
             }
         });
     }
+
+    private String findPairingId(Uri redirectUri) {
+        String pairingId = redirectUri.getQueryParameter("ba_token");
+        if (pairingId == null) {
+            pairingId = redirectUri.getQueryParameter("token");
+        }
+        return pairingId;
```
sshropshire, line 139: [Similar to iOS](https://github.com/braintree/braintree_ios/blob/83a5709a06bca8ceb3b48f842ebefd4016294047/Sources/BraintreePayPal/BTPayPalClient.swift#L326): we check for `ba_token` and fallback to `token` if possible.

PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
```diff
@@ -68,19 +68,27 @@ public void onResult(String responseBody, Exception httpError) {
...
+                                                            dataCollectorRequest.setRiskCorrelationId(pairingId);
+                                                        }
+                                                        clientMetadataId = payPalDataCollector.getClientMetadataId(context, dataCollectorRequest, configuration);
+                                                    }
 
                                                     if (pairingId != null) {
-                                                        payPalResponse
-                                                                .pairingId(pairingId)
-                                                                .clientMetadataId(clientMetadataId);
+                                                        payPalResponse.pairingId(pairingId);
```
sshropshire, line 86: This change resolves some testing errors. Logically it seems odd to set two fields in the `PayPalResponse` based on one value being `null`.

PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutInternalClient.java
```diff
@@ -62,10 +62,9 @@ void sendRequest(final Context context, final PayPalNativeRequest payPalRequest,
                                         String clientMetadataId = payPalRequest.getRiskCorrelationId() != null
                                                 ? payPalRequest.getRiskCorrelationId() : payPalDataCollector.getClientMetadataId(context, configuration, payPalRequest.hasUserLocationConsent());
 
+                                        payPalResponse.clientMetadataId(clientMetadataId);
                                         if (pairingId != null) {
-                                            payPalResponse
-                                                    .pairingId(pairingId)
-                                                    .clientMetadataId(clientMetadataId);
+                                            payPalResponse.pairingId(pairingId);
```
sshropshire, line 67: PayPalWeb and PayPalNative share the same text fixtures (and same implementation for the most part). This change mirrors the one [made in PayPalWeb](https://github.com/braintree/braintree_android/pull/983/files#r1588466622).

PayPalNativeCheckout/src/test/java/com/braintreepayments/api/PayPalNativeCheckoutInternalClientUnitTest.java
```diff
@@ -450,7 +450,7 @@ public void sendRequest_withPayPalCheckoutRequest_callsBackPayPalResponseOnSucce
         assertFalse(payPalResponse.isBillingAgreement());
         assertEquals("authorize", payPalResponse.getIntent());
         assertEquals("sample-merchant-account-id", payPalResponse.getMerchantAccountId());
-        assertEquals("EC-HERMES-SANDBOX-EC-TOKEN", payPalResponse.getPairingId());
+        assertNull(payPalResponse.getPairingId());
```
sshropshire, line 453: This helped clear the test, and also makes me wonder if we should be collecting data for MXO flows? The pairing ID relies on the presence of an app switch URL. It seems like we're collecting data for Native PayPal only because the implementation of `PayPalNativeCheckoutInternalClient` is identical to web.

PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
saperi22: What changed here?
I see the chained call being separated into another chained call. Is the order of calling things important here?

sshropshire: Shouldn't be the chained calls are only setters. The main difference now is `clientMetadataId` is always set.

saperi22 (approved):

sarahkoop (approved):

PR #984, opened 2024-04-30
Add useAppLinkReturn to PayPalRequest

tdchow:
### Summary of changes

 - Add `useAppLinkReturn` to `PayPalRequest` for Android App Link support

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### DIFF
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
@@ -71,6 +71,7 @@ public abstract class PayPalRequest implements Parcelable {
     private String riskCorrelationId;
     private final ArrayList<PayPalLineItem> lineItems;
     private final boolean hasUserLocationConsent;
+    private boolean useAppLinkReturn;
 
     /**
      * Deprecated. Use {@link PayPalRequest#PayPalRequest(boolean)} instead.
@@ -228,6 +229,20 @@ public abstract class PayPalRequest implements Parcelable {
         this.lineItems.addAll(lineItems);
     }
 
+    /**
+     * Optional: When set to true, the Android App Link website associated with your application
+     * will be used to return to your app from browser or app switch based payment flows. When set
+     * to false, the Set the deep link return URL will be used.
+     *
+     * Set the App Link value on `appLinkReturnUri` parameter in the {@link BraintreeClient}
+     * constructor.
+     *
+     * @param useAppLinkReturn indicates whether to use the set Android App Link
+     */
+    public void setUseAppLinkReturn(boolean useAppLinkReturn) {
+        this.useAppLinkReturn = useAppLinkReturn;
+    }
+
     @Nullable
     public String getLocaleCode() {
         return localeCode;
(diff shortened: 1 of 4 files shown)


PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
```diff
@@ -281,6 +296,10 @@ public boolean hasUserLocationConsent() {
         return hasUserLocationConsent;
     }
 
+    public boolean isUseAppLinkReturn() {
```
tdchow, line 299: Thoughts on renaming the variable to make the getter more readable? Maybe `appLinkEnabled` (`setAppLinkEnabled()` and `isAppLinkEnabled`)?

sarahkoop: +1 on `appLinkEnabled` naming is more clear

sshropshire: quick question: Could we infer this value internally by checking if `BraintreeClient` is configured to use app links? 

e.g.
```kotlin
val hasAppLinkReturnUri = (braintreeClient.appLinkReturnUri != null)
```

tdchow: Hmm that's a good call. If a merchant were to set a URI on the BraintreeClient, they'd probably want to use App Links. Is there a benefit in having them opt-in at the request level?

sarahkoop: My concern would be that since we are adding app link support only for PayPal currently, when we add support for future payment methods, merchants would be defaulted into using app links if they had already set them on their BraintreeClient (for PayPal specifically), and they might not be set up to use app links for the other payment flows (i.e. maybe they want to return to a different activity etc)

tdchow: In those cases would merchants be creating a new `BraintreeClient` instance where they would omit the app link URI?

sarahkoop: Yes they could, so there would be a work around - I guess my concern would be merchants that have a single `BraintreeClient`, if they set up for app links for PayPal, and in a future release we add app link support for another feature, then without any changes just upgrading to the latest version, the merchant would be opted into app links (which might be a breaking change for their integration)

tdchow: That makes sense. I'm good with keeping the enabled boolean in each of the request objects. These would most likely go away in v5 since we can just check if the URI is set on each of the request objects.

sshropshire: Yeah it definitely makes sense in v4 to have two-step enablement for app links. We should consider moving `returnUrlScheme` and `appLinkReturnUri` properties into the request objects in v5 to make it a one-step integration in the future. A single source of truth for enabling app links should make the integration less error prone.

PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
```diff
@@ -228,6 +229,20 @@ public void setLineItems(@NonNull Collection<PayPalLineItem> lineItems) {
         this.lineItems.addAll(lineItems);
     }
 
+    /**
+     * Optional: When set to true, the Android App Link website associated with your application
+     * will be used to return to your app from browser or app switch based payment flows. When set
+     * to false, the Set the deep link return URL will be used.
```
sarahkoop, line 235:
```suggestion
     * to false, the default or set deep link return URL will be used.
```

sarahkoop (approved):

KunJeongPark (approved):

PR #985, opened 2024-04-30
Add ExperimentalBetaApi annotation

saperi22:
### Summary of changes

 - Add UnstableApi annotation.

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

@saperi22

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/ExperimentalBetaApi.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/ExperimentalBetaApi.kt
@@ -1,9 +1,7 @@
 package com.braintreepayments.api
 
 /**
- * This annotation signals that the annotated class is an experimental feature that is in beta.
- * No guarantees are made on the stability of the beta API. They can be modified or removed in a
- * future release.
+ * The annotated method or class is in beta. It's public API may change or be removed in future releases.
  */
 @RequiresOptIn
 annotation class ExperimentalBetaApi
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/ExperimentalBetaApi.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/ExperimentalBetaApi.kt
@@ -1,4 +1,4 @@
 package com.braintreepayments.api
 
 @RequiresOptIn
-annotation class UnstableApi
+annotation class ExperimentalBetaApi
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/ExperimentalBetaApi.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/ExperimentalBetaApi.kt
@@ -1,4 +1,9 @@
 package com.braintreepayments.api
 
+/**
+ * This annotation signals that the annotated class is an experimental feature that is in beta.
+ * No guarantees are made on the stability of the beta API. They can be modified or removed in a
+ * future release.
+ */
 @RequiresOptIn
 annotation class ExperimentalBetaApi
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/UnstableApi.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/UnstableApi.kt
@@ -1,0 +1,4 @@
+package com.braintreepayments.api
+
+@RequiresOptIn
+annotation class UnstableApi
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
@@ -3,7 +3,7 @@ package com.braintreepayments.api
 import com.braintreepayments.api.EligiblePaymentsApiRequest.Companion.toJson
 import org.json.JSONException
 
-@UnstableApi
+@ExperimentalBetaApi
 internal class EligiblePaymentsApi(
     private val braintreeClient: BraintreeClient
 ) {
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
@@ -3,7 +3,6 @@ package com.braintreepayments.api
 import com.braintreepayments.api.EligiblePaymentsApiRequest.Companion.toJson
 import org.json.JSONException
 
-@OptIn(ExperimentalBetaApi::class)
 internal class EligiblePaymentsApi(
     private val braintreeClient: BraintreeClient
 ) {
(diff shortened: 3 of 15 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/UnstableApi.kt
```diff
@@ -0,0 +1,4 @@
+package com.braintreepayments.api
+
+@RequiresOptIn
+annotation class UnstableApi
```
sarahkoop, line 4: Should we maybe say "ExperimentalApi" or "BetaFeature"? I think we should establish a convention with this annotation that we use throughout others

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -18,6 +17,7 @@ import java.lang.Exception
  *
  * Note: **This feature is in beta. It's public API may change in future releases.**
  */
+@UnstableApi
```
jaxdesmarais, line 20: Does this add the annotation to all of the public methods in this class such as `getRecommendedPaymentMethods`? I think we'd also want to add the annotation to `ShopperInsightsRequest` and `ShopperInsightsCallback` correct?

BraintreeCore/src/main/java/com/braintreepayments/api/UnstableApi.kt
saperi22: I wanted to put up a PR to get ideas on what we might want to call this.
I'm open to suggestions.

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
saperi22: Ref: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-requires-opt-in/

Class level opt in makes everything in the class opt in.
Screenshot 2024-05-01 at 9 52 49 AM

Otherwise, each of the methods requires opt-in.
Screenshot 2024-05-01 at 9 53 51 AM

Screenshot 2024-05-01 at 9 53 59 AM

Screenshot 2024-05-01 at 9 53 40 AM

BraintreeCore/src/main/java/com/braintreepayments/api/UnstableApi.kt
warmkesselj: How about just `@Beta` or `@Unstable`? `@Experimental`sounds like A/B testing.

jaxdesmarais: Can we do a 2nd pass about this naming in stand? Would be great to hammer down what we want to call this since we will need to use it across several upcoming features.

jaxdesmarais: Would `BetaExperiementalApi` or `ExperiementalBetaApi` be too verbose?

tdchow: I like `ExperiementalBetaApi`. I think most people will use autocomplete in their IDE to add the annotation.

BraintreeCore/src/main/java/com/braintreepayments/api/ExperimentalBetaApi.kt
```diff
@@ -0,0 +1,4 @@
+package com.braintreepayments.api
+
+@RequiresOptIn
+annotation class ExperimentalBetaApi
```
sarahkoop, line 4: Can this annotation include docstrings? Will they be generated/published  like any other class? If so, let's explain what ExperimentalBetaApi means in our SDK in the doc strings

ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
```diff
@@ -3,6 +3,7 @@ package com.braintreepayments.api
 import com.braintreepayments.api.EligiblePaymentsApiRequest.Companion.toJson
 import org.json.JSONException
 
+@ExperimentalBetaApi
```
tdchow, line 6: Is there a benefit in marking these internal classes with the annotation?

saperi22: It is required by the compiler.

A class that depends on a @ExperimentalBetaApi annotated class can either be annotated with @ExperimentalBetaApi annotation or @OptIn(ExperimentalBetaApi::class).

Maybe in this case, the internal class can be annotated with @OptIn instead. Thoughts?

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsResult.kt
```diff
@@ -3,6 +3,7 @@ package com.braintreepayments.api
 /**
  * The result object returned when insights about a shopper is requested.
  */
+@ExperimentalBetaApi
```
saperi22, line 6: @tdchow I'm inclined to changing this to OptIn instead of the ExperimentalBetaApi annotation.
The result class isn't likely to change. The contents of `ShopperInsightsInfo` are likely to change, but we denote that by annotating that class itself.

BraintreeCore/src/main/java/com/braintreepayments/api/ExperimentalBetaApi.kt
saperi22: @sarahkoop please take a look at the language.

sarahkoop: Maybe let's use the same comment that's currently on the beta classes:

`The annotated method or class is in beta. It's public API may change or be removed in future releases.`

ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
tdchow: Would the depednency go the other way for the internal classes? The `EligiblePaymentsApi` class would not depend on the `ShopperInsightsClient` class. I'm wondering if we can just mark the externally facing classes with the annotation. Also not sure if missing something 😄

sarahkoop (approved):

jaxdesmarais (approved):

ShopperInsights/src/main/java/com/braintreepayments/api/EligiblePaymentsApi.kt
saperi22: I've made the internal classes and tests OptIn to the functionality.
Even among the classes that are exposed, only those that I think can change are being marked with the ExperimentalBetaApi annotation. For eg.: The `ShopperInsightsResult` class now opts in to the experimental api and does _not_ propagate it to its usage location.

tdchow: got it. that sounds good!

tdchow (approved):

PR #986, opened 2024-04-30
Send app link URLs to PayPal API

tdchow:
### Summary of changes

 - Send app link URLs to the PayPal API when `AppLinkReturnUri` is set on `BraintreeClient` and `appLinkEnabled` is set to true on the `PayPalRequest`

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### DIFF
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
@@ -14,9 +14,6 @@ class PayPalInternalClient {
     private static final String CREATE_SINGLE_PAYMENT_ENDPOINT = "paypal_hermes/create_payment_resource";
     private static final String SETUP_BILLING_AGREEMENT_ENDPOINT = "paypal_hermes/setup_billing_agreement";
 
-    private final String cancelUrl;
-    private final String successUrl;
-
     private final BraintreeClient braintreeClient;
     private final PayPalDataCollector payPalDataCollector;
     private final ApiClient apiClient;
@@ -30,9 +27,6 @@ class PayPalInternalClient {
         this.braintreeClient = braintreeClient;
         this.payPalDataCollector = payPalDataCollector;
         this.apiClient = apiClient;
-
-        this.cancelUrl = String.format("%s://onetouch/v1/cancel", braintreeClient.getReturnUrlScheme());
-        this.successUrl = String.format("%s://onetouch/v1/success", braintreeClient.getReturnUrlScheme());
     }
 
     void sendRequest(final Context context, final PayPalRequest payPalRequest, final PayPalInternalClientCallback callback) {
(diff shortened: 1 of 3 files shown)


PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
```diff
@@ -53,6 +47,17 @@ public void onResult(@Nullable final Configuration configuration, @Nullable Exce
...
 
+                                String cancelUrl;
+                                String successUrl;
+                                if (payPalRequest.isAppLinkEnabled() && braintreeClient.getAppLinkReturnUri() != null) {
+                                    String appLink = braintreeClient.getAppLinkReturnUri().toString();
+                                    cancelUrl = String.format("%s/cancel", appLink);
+                                    successUrl = String.format("%s/success", appLink);
+                                } else {
+                                    cancelUrl = String.format("%s://onetouch/v1/cancel", braintreeClient.getReturnUrlScheme());
+                                    successUrl = String.format("%s://onetouch/v1/success", braintreeClient.getReturnUrlScheme());
```
tdchow, line 58: Is there anything special about these `onetouch/v1` paths in the deep link URLs? Would we need these in the app link URLs or can we just append `/cancel` and `/success`?

scannillo: I don't think the onetouch naming is doing anything, it works with any random string in there.

PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
```diff
@@ -53,6 +47,17 @@ public void onResult(@Nullable final Configuration configuration, @Nullable Exce
                                         ? SETUP_BILLING_AGREEMENT_ENDPOINT : CREATE_SINGLE_PAYMENT_ENDPOINT;
                                 String url = String.format("/v1/%s", endpoint);
 
+                                String cancelUrl;
+                                String successUrl;
+                                if (payPalRequest.isAppLinkEnabled() && braintreeClient.getAppLinkReturnUri() != null) {
+                                    String appLink = braintreeClient.getAppLinkReturnUri().toString();
+                                    cancelUrl = String.format("%s/cancel", appLink);
+                                    successUrl = String.format("%s/success", appLink);
```
scannillo, line 55: Since we're appending to the merchants URL we'll have to make sure that they wildcard the URL they give us. Either that, or we ask them for explicit cancel and success full paths.

With the app switch project, we opted for asking them for a wildcarded path, specific to Braintree payments.

scannillo (approved):

PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
tdchow: Is there a wildcard format that we're currently using for the app switch integration? Is there a downside in just appending the cancel and success paths to the app link URL the merchant passes in?

The merchant is currently passing in a `Uri` type - I'm not sure if any URL validation is done when creating the Uri instance 🤔

PayPal/src/main/java/com/braintreepayments/api/PayPalInternalClient.java
tdchow: The gateway team confirmed that the `onetouch/v1` path is an SDK specific detail. So we're safe to omit it from the app link URL 👍

saperi22 (approved):

PR #991, opened 2024-05-03
Add `venmo_installed` Analytics tag

KunJeongPark:
### Summary of changes

 - Add `venmo_installed` analytics tag in `AnalyticsEvent`
 - Modify` sendEvent` to fetch isVenmoInstalled value from `DeviceInspector`
 - Modify AnalyticsDatabase to reflect change to `AnalyticsEvent`

### Checklist

 - [] Added a changelog entry

@KunJeongPark

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
@@ -19,13 +19,15 @@ internal class AnalyticsClient @VisibleForTesting constructor(
     private val httpClient: BraintreeHttpClient,
     private val analyticsDatabase: AnalyticsDatabase,
     private val workManager: WorkManager,
-    private val deviceInspector: DeviceInspector
+    private val deviceInspector: DeviceInspector,
+    private val isVenmoInstalled: Boolean
 ) {
     constructor(context: Context) : this(
         BraintreeHttpClient(),
         getInstance(context.applicationContext),
         WorkManager.getInstance(context.applicationContext),
-        DeviceInspector()
+        DeviceInspector(),
+        DeviceInspector().isVenmoInstalled(context.applicationContext)
     )
 
     fun sendEvent(
@@ -35,7 +37,8 @@ internal class AnalyticsClient @VisibleForTesting constructor(
         integration: String?,
         authorization: Authorization
     ): UUID {
-        scheduleAnalyticsWrite(event, authorization)
+        val updatedEvent = AnalyticsEvent(event.name, event.payPalContextId, event.linkType, event.timestamp, isVenmoInstalled)
+        scheduleAnalyticsWrite(updatedEvent, authorization)
         return scheduleAnalyticsUpload(configuration, authorization, sessionId, integration)
     }
 
@@ -68,7 +71,7 @@ internal class AnalyticsClient @VisibleForTesting constructor(
         return if (eventName == null || timestamp == INVALID_TIMESTAMP) {
             ListenableWorker.Result.failure()
         } else {
-            val event = AnalyticsEvent(eventName, payPalContextId, linkType, timestamp)
+            val event = AnalyticsEvent(eventName, payPalContextId, linkType, timestamp, venmoInstalled = isVenmoInstalled)
             val analyticsEventDao = analyticsDatabase.analyticsEventDao()
             analyticsEventDao.insertEvent(event)
             ListenableWorker.Result.success()
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
@@ -19,15 +19,13 @@ internal class AnalyticsClient @VisibleForTesting constructor(
     private val httpClient: BraintreeHttpClient,
     private val analyticsDatabase: AnalyticsDatabase,
     private val workManager: WorkManager,
-    private val deviceInspector: DeviceInspector,
-    private val isVenmoInstalled: Boolean
+    private val deviceInspector: DeviceInspector
 ) {
     constructor(context: Context) : this(
         BraintreeHttpClient(),
         getInstance(context.applicationContext),
         WorkManager.getInstance(context.applicationContext),
-        DeviceInspector(),
-        DeviceInspector().isVenmoInstalled(context.applicationContext)
+        DeviceInspector()
     )
 
     fun sendEvent(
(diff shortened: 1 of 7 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.kt
```diff
@@ -8,11 +8,14 @@ import androidx.room.RoomDatabase
...
 // Ref: https://developer.android.com/training/data-storage/room/migrating-db-versions
 @Database(
-        version = 3,
+        version = 4,
         entities = [AnalyticsEvent::class],
         autoMigrations = [
             AutoMigration(from = 1, to = 3),
-            AutoMigration(from = 2, to = 3)
+            AutoMigration(from = 2, to = 3),
+            AutoMigration(from = 1, to = 4),
```
saperi22, line 16: Since we already have migration from 1 to 2 and 2 to 3, please only add 3 to 4. Otherwise, the increase in auto-migrations we have to provide would increase non-linearly.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
```diff
@@ -19,13 +19,15 @@ internal class AnalyticsClient @VisibleForTesting constructor(
...
+    private val deviceInspector: DeviceInspector,
+    private val isVenmoInstalled: Boolean
 ) {
     constructor(context: Context) : this(
         BraintreeHttpClient(),
         getInstance(context.applicationContext),
         WorkManager.getInstance(context.applicationContext),
-        DeviceInspector()
+        DeviceInspector(),
+        DeviceInspector().isVenmoInstalled(context.applicationContext)
```
saperi22, line 30: This creates a new instance of device inspector, do we really want it? Can we reuse the device inspector we already have?

KunJeongPark: I couldn't figure out a way to do iOS equivalent of static function for device install check that I implemented (https://github.com/braintree/braintree_ios/pull/1289) 
I will try `companion object`

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.kt
sshropshire: I'm thinking we might want to consider using SQL to store the analytics JSON as a string instead of typed values to make adding analytics tags easier in the future. Do y'all think it makes sense to do this now?

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
sshropshire: We could add the `deviceInspector.isVenmoInstalled()` check [here](https://github.com/braintree/braintree_android/blob/426544a92c4729fe97d299f9166a14602b0eb112/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt#L222). Then we wouldn't need to create an `updatedEvent` internally.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.kt
sshropshire: It does seem like [more events are coming](https://github.com/braintree/braintree_ios/pull/1292) to track latency.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
KunJeongPark: Nice!

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.kt
KunJeongPark: I think that would be a separate ticket?

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
KunJeongPark: one issue I ran into is getting `venmoInstalled` in `writeAnalytics` function in AnalyticsClient without `isVenmoInstalled` 
property on AnalyticsClient.

KunJeongPark: Yes, without i`sVenmoInstalled` property on `AnalyticsClient`, I ran into issue of accessing `venmoInstalled` value in instantiating `AnalyticEvent` in `writeAnalytics` function in `AnalyticsClient`.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -219,7 +219,9 @@ open class BraintreeClient @VisibleForTesting internal constructor(
         getAuthorization { authorization, _ ->
             if (authorization != null) {
                 getConfiguration { configuration, _ ->
-                    val event = AnalyticsEvent(eventName, payPalContextId, linkType)
+                    val deviceInspector = DeviceInspector()
```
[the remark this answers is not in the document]
KunJeongPark: Thank you Steven and thanks for help on Friday. I didn't do the duet commit right.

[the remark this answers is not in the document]
sshropshire: It's cool. Getting it across the finish line is more important.

jaxdesmarais (commented):
Were you able to see the new events in FPTI for both installed and not installed cases?

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.kt
jaxdesmarais: Yes it would be a separate task. I created DTMOBILES-619 for exploring different storage options.

KunJeongPark: Thank you Jax!

jaxdesmarais (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.kt
```diff
@@ -8,11 +8,12 @@ import androidx.room.RoomDatabase
 
 // Ref: https://developer.android.com/training/data-storage/room/migrating-db-versions
 @Database(
-        version = 3,
+        version = 4,
         entities = [AnalyticsEvent::class],
         autoMigrations = [
             AutoMigration(from = 1, to = 3),
```
sshropshire, line 14:
```suggestion
            AutoMigration(from = 1, to = 2),
```
I noticed this may be a pre-existing bug. The migrations are supposed to [count up in incremental pairs](https://medium.com/androiddevelopers/room-auto-migrations-d5370b0ca6eb).

richherrera (approved):
🥇

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.kt
KunJeongPark: so it should be 1->2, 2->3, 3->4?

sshropshire: Yup yeah that's it.

PR #996, opened 2024-05-07
Add New Demo Tokenization Key for PayPal

sarahkoop:
### Summary of changes

 - Add tokenization key to demo app that is associated with merchant account that is configured for new PayPal checkout experience.
 - Requires settings "Tokenization Key", "PayPal Checkout Experience" and "Intent Type = Sale" to be set in demo app to launch new checkout experience from PayPal checkout flow
 - Prod tokenization key does not exist for this yet, but set up demo app at allow for one to be added easily

### Checklist

### DIFF
--- a/Demo/src/main/res/xml/settings.xml
+++ b/Demo/src/main/res/xml/settings.xml
@@ -92,7 +92,7 @@
         android:title="@string/paypal">
 
         <CheckBoxPreference
-            android:key="enable_manual_browser_switching"
+            android:key="show_checkout_experience"
             android:title="@string/paypal_checkout_experience"
             android:summary="@string/paypal_checkout_experience_summary"
             android:defaultValue="false" />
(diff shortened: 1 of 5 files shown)


Demo/src/main/res/xml/settings.xml
```diff
@@ -91,6 +91,12 @@
     <PreferenceCategory
         android:title="@string/paypal">
 
+        <CheckBoxPreference
```
tdchow, line 94: What are your thoughts on adding a setting where we have a dropdown of tokenization keys? We could even have an option to input a tokenization key in the demo app settings?

I'm thinking that the "new PayPal checkout experience" will eventually be rolled out and become the normal flow and we'll need to remove this preference when that happens.

sshropshire (approved):

tdchow (approved):
I had a suggestion for reducing some potential future tech debt - but I'm good with merging this in as is since it's needed for testing.

Demo/src/main/res/xml/settings.xml
sarahkoop: Yes I think that's a great idea - can you make a ticket for that?

PR #1000, opened 2024-05-13
Handle empty email address

tdchow:
### Summary of changes

 - Check for empty email address - prevent passing the value to the API
 - Provide a postal code in the demo app's `PayPalRequestFactory` which now seems to be required

Screenshot_20240513_152120

### Checklist

 - [ ] Added a changelog entry
 - [X] Relevant test coverage

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
@@ -90,6 +90,7 @@ public class PayPalRequestFactory {
             shippingAddress.setLocality("San Francisco");
             shippingAddress.setRegion("CA");
             shippingAddress.setCountryCodeAlpha2("US");
+            shippingAddress.setPostalCode("94103");
 
             request.setShippingAddressOverride(shippingAddress);
         }
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalCheckoutRequest.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalCheckoutRequest.java
@@ -208,7 +208,7 @@ public class PayPalCheckoutRequest extends PayPalRequest implements Parcelable {
         }
 
         if (userAuthenticationEmail != null && !userAuthenticationEmail.isEmpty()) {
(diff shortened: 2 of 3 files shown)


PayPal/src/main/java/com/braintreepayments/api/PayPalCheckoutRequest.java
```diff
@@ -207,7 +207,9 @@ String createRequestBody(Configuration configuration, Authorization authorizatio
             parameters.put(BILLING_AGREEMENT_DETAILS_KEY, details);
         }
 
-        parameters.putOpt(PAYER_EMAIL_KEY, userAuthenticationEmail);
+        if (userAuthenticationEmail != null && !userAuthenticationEmail.isEmpty()) {
+            parameters.putOpt(PAYER_EMAIL_KEY, userAuthenticationEmail);
```
saperi22, line 211: You could just do `put` instead of `putOpt` at this point.

saperi22 (approved):

Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
```diff
@@ -90,6 +90,7 @@ public static PayPalCheckoutRequest createPayPalCheckoutRequest(
             shippingAddress.setLocality("San Francisco");
             shippingAddress.setRegion("CA");
             shippingAddress.setCountryCodeAlpha2("US");
+            shippingAddress.setPostalCode("94103");
```
jaxdesmarais, line 93: Just curious if the postal code is needed for something else

jaxdesmarais (approved):

PayPal/src/main/java/com/braintreepayments/api/PayPalCheckoutRequest.java
tdchow: Good catch. let me update this.

Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
tdchow: It looks like it's dependent on the country code provided in the request. This might be new validation that the backend is doing for US 🤔

PR #1003, opened 2024-05-14
Add buyer phone number to PayPal request

tdchow:
### Summary of changes

 - Add `userPhoneNumber` field to `PayPalRequest`
 - Update demo app to accept phone number

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragment.kt
+++ b/Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragment.kt
@@ -139,7 +139,7 @@ class ShopperInsightsFragment : BaseFragment(), PayPalListener, VenmoListener {
         shopperInsightsClient.sendPayPalSelectedEvent()
         payPalClient.tokenizePayPalAccount(
             requireActivity(),
-            PayPalRequestFactory.createPayPalVaultRequest(activity, emailInput.editText?.text.toString())
+            PayPalRequestFactory.createPayPalVaultRequest(activity, emailInput.editText?.text.toString(), "")
         )
     }
 
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
@@ -27,6 +27,7 @@ public abstract class PayPalRequest implements Parcelable {
     static final String BILLING_AGREEMENT_DETAILS_KEY = "billing_agreement_details";
     static final String DESCRIPTION_KEY = "description";
     static final String PAYER_EMAIL_KEY = "payer_email";
+    static final String PHONE = "phone";
     static final String AUTHORIZATION_FINGERPRINT_KEY = "authorization_fingerprint";
     static final String TOKENIZATION_KEY = "client_key";
     static final String RETURN_URL_KEY = "return_url";
@@ -72,6 +73,7 @@ public abstract class PayPalRequest implements Parcelable {
     private final ArrayList<PayPalLineItem> lineItems;
     private final boolean hasUserLocationConsent;
     protected String userAuthenticationEmail;
+    protected String userPhoneNumber;
 
     /**
      * Deprecated. Use {@link PayPalRequest#PayPalRequest(boolean)} instead.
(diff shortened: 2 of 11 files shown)


PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
```diff
@@ -72,6 +73,7 @@ public abstract class PayPalRequest implements Parcelable {
     private final ArrayList<PayPalLineItem> lineItems;
     private final boolean hasUserLocationConsent;
     protected String userAuthenticationEmail;
+    protected String userPhoneNumber;
```
tdchow, line 76: @agedd - what are your thoughts on this field name? should we align across platforms?

Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragment.kt
```diff
@@ -139,7 +139,7 @@ class ShopperInsightsFragment : BaseFragment(), PayPalListener, VenmoListener {
         shopperInsightsClient.sendPayPalSelectedEvent()
         payPalClient.tokenizePayPalAccount(
             requireActivity(),
-            PayPalRequestFactory.createPayPalVaultRequest(activity, emailInput.editText?.text.toString())
+            PayPalRequestFactory.createPayPalVaultRequest(activity, emailInput.editText?.text.toString(), "")
```
sarahkoop, line 142: should we pass the phone input edit text value here instead of ""?

tdchow: This is in the ShopperInsightsFragment in the demo. Would we want to have an edit text on that screen as well as the PayPalFragment?

sarahkoop: Oh no sorry missed that this was in shopper insights!

sarahkoop (approved):

PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
```diff
@@ -297,6 +299,21 @@ public String getUserAuthenticationEmail() {
         return this.userAuthenticationEmail;
     }
 
+    /**
+     * Optional: User phone number to initiate a quicker authentication flow in cases where the user
+     * has a PayPal Account with the same phone number.
+     *
+     * @param userPhoneNumber - phone number of the payer
+     */
+    public void setUserPhoneNumber(@Nullable String userPhoneNumber) {
```
jaxdesmarais, line 308: Should we include some formatting guidelines, for example would `1112223333`, `(111) 222-3333`, `+1 (111)222 - 3333`, etc all be accepted?

PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
agedd: @tdchow `userPhoneNumber` sounds good to me, will make sure to align on iOS as well!

PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
```diff
@@ -27,6 +27,7 @@ public abstract class PayPalRequest implements Parcelable {
     static final String BILLING_AGREEMENT_DETAILS_KEY = "billing_agreement_details";
     static final String DESCRIPTION_KEY = "description";
     static final String PAYER_EMAIL_KEY = "payer_email";
+    static final String PHONE = "phone";
```
sshropshire, line 30:
```suggestion
    static final String PHONE_KEY = "phone";
```
Adding key suffix to match convention.

sshropshire (approved):

PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
tdchow: Good catch! just updated the name

PayPal/src/main/java/com/braintreepayments/api/PayPalRequest.java
tdchow: The Gateway will be stripping out any non-numeric character. But it still might be nice to give merchants some guidance to remove any potential confusion on the format of the phone number.

Maybe we suggest merchants to only pass in numeric characters? `1112223333` What do y'all think?

cc @agedd

sarahkoop: My two cents - we could either say `any non numeric characters will be removed from the input` to make it easier for merchants so they don't need to do any format changing, or we could just request input in all numeric characters to avoid confusion. Either is good with me

tdchow:
Converting the PR to a draft as API requirements are not yet finalized.

tdchow:
Closing PR as this work was pushed out.

PR #1004, opened 2024-05-17
Send `is_vault` FPTI tag

richherrera:
### Summary of changes

 - Send `is_vault` in `event_params` to PayPal's analytics service (FPTI)
 - Add `isVaultRequest` analytics tag in `AnalyticsEvent`
 - Modify AnalyticsDatabase to reflect change to `AnalyticsEvent`

Screenshot 2024-05-21 at 10 55 19 a m

### Checklist

 - [X] Added a changelog entry
 - [X] Relevant test coverage

### Authors

- @richherrera

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
@@ -216,13 +216,14 @@ open class BraintreeClient @VisibleForTesting internal constructor(
     fun sendAnalyticsEvent(
             eventName: String,
             payPalContextId: String? = null,
-            linkType: String? = null
+            linkType: String? = null,
+            isVaultRequest: Boolean = false
     ) {
         getAuthorization { authorization, _ ->
             if (authorization != null) {
                 getConfiguration { configuration, _ ->
                     val isVenmoInstalled = deviceInspector.isVenmoInstalled(applicationContext)
-                    val event = AnalyticsEvent(eventName, payPalContextId, linkType, venmoInstalled = isVenmoInstalled)
+                    val event = AnalyticsEvent(eventName, payPalContextId, linkType, venmoInstalled = isVenmoInstalled, isVaultRequest = isVaultRequest)
                     sendAnalyticsEvent(event, configuration, authorization)
                 }
             }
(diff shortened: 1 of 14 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -216,13 +216,20 @@ open class BraintreeClient @VisibleForTesting internal constructor(
     fun sendAnalyticsEvent(
```
saperi22, line 216: Looks like this method signature is growing on a weekly basis.
Would it make sense to extract all the payload into an object of its own so we can keep the method signature same?

saperi22 (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
sarahkoop: +1 There is an open ticket to investigate moving to a SQL DB since it keeps changing - but maybe a quick win to add an `AnalyticsParams` object or something like that so we don't have to update so many classes/tests every time

richherrera: Of course, happy to work on this and create an object with the parameters, I would just like to work on it in another PR, since several classes/unit tests will be modified

CHANGELOG.md
```diff
@@ -1,9 +1,10 @@
 # Braintree Android SDK Release Notes
 
 ## unreleased
-
+* PayPal
+  * Send `is_vault` in `event_params` to PayPal's analytics service (FPTI)s
 * Venmo
-  * Send `link_type` in `event_params` to PayPal's analytics service (FPTI)
+  * Send `link_type` and `is_vault` in `event_params` to PayPal's analytics service (FPTI)
```
sarahkoop, line 7: Can we update these to just say send `link_type` and `is_vault` analytics? - we don't need to state internal service naming.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
sarahkoop: Yes agreed - lets just make a ticket for that work and can address it at lower priority

CHANGELOG.md
richherrera: Addressed: [023cad0](https://github.com/braintree/braintree_android/pull/1004/commits/023cad0c49be7e885155fc1cee7c4351f9b99694)

jaxdesmarais (approved):

PR #1022, opened 2024-06-04
Update CHANGELOG with Missing Timestamps

sshropshire:
### Summary of changes

- Quick chore to update timestamps via `git log -1 --format=%ai "$VERSION"` command for each version that didn't have a timestamp
- Not sure if this will easily merge into `v5`, if it becomes a headache we can just close it

### Checklist

 - [x] Added a changelog entry

### Authors
@sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -583,7 +583,7 @@
 
 **Note:** Includes all changes in [4.0.0-beta1](#400-beta1), [4.0.0-beta2](#400-beta2), and [4.0.0-beta3](#400-beta3)
 
-## 4.0.0-beta3
+## 4.0.0-beta3 (2021-05-13)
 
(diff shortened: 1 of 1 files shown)


jaxdesmarais (commented):
🔥

tdchow (approved):
Looks great! Is there a way we can update the release script to automatically add the timestamp?

jaxdesmarais:
> Looks great! Is there a way we can update the release script to automatically add the timestamp?

@tdchow It exist already [here](https://github.com/braintree/braintree_android/blob/95758a0b5fece9131f7386209492134c7525bb1f/build.gradle#L202) but only added it moving forward for version 4.37.0+. This PR adds the date for previous versions before that change was made, so this should be the only time we have to backdate it manaually.

scannillo (approved):

PR #1030, opened 2024-06-07
[V4] Use TLS 1.3

scannillo:
### Background

Total time spent on GW networking is a top concern recently. 

Our SDK is using HTTP 1.1 and TLS 1.2, while the latest available are HTTP 2 and TLS 1.3. I conducted the below timing analysis to see where we would benefit from updating to the latest versions of these protocols.

### Local Testing 

I used [this script](https://github.com/chentda/curl-average-timings) to execute a cURL request 100 times and collect the average total time. 

I used the following URL: `https://api.braintreegateway.com/heartbeat.json`

| TLS Version | HTTP Version | Avg Time (seconds) | 
|-------------|--------------|--------------------|
| TLS 1.2     | HTTP 1.1     | 0.22898            |
| TLS 1.3     | HTTP 1.1     | 0.02098            |
| TLS 1.2     | HTTP 2       | 0.23487         |
| TLS 1.3     | HTTP 2       | 0.01953            |

#### Takeaways

* The improvement from TLS 1.2 --> 1.3 across both HTTP is significant, averaging **211.5 ms faster** 🎉🎉. 
* The savings from HTTP 1.1 --> HTTP 2 are less significant (0.145 ms), and seem to be coupled with using TLS 1.3.
    * _Note_: Android's `HTTPURLConnection` library, which we use for networking, only supports up to HTTP 1.1. So, if we did want to move to HTTP 2 at some point, we would probably have to migrate to OkHttp, which supports HTTP 2. Based on these timing results, it doesn't seem to be worth it at this time.

### Checklist

 - [X] Added a changelog entry

### Authors
@scannillo


sshropshire (approved):

SharedUtils/src/main/java/com/braintreepayments/api/TLSSocketFactory.java
```diff
@@ -118,7 +118,7 @@ private Socket enableTLSOnSocket(Socket socket) {
         if (socket instanceof SSLSocket) {
             ArrayList<String> supportedProtocols =
                     new ArrayList<>(Arrays.asList(((SSLSocket) socket).getSupportedProtocols()));
-            supportedProtocols.retainAll(Collections.singletonList("TLSv1.2"));
+            supportedProtocols.retainAll(Collections.singletonList("TLSv1.3"));
```
sshropshire, line 121: We might still need TLS v1.2 as a fallback. It looks like v1.3 is [available in API 29+](https://developer.android.com/reference/javax/net/ssl/SSLSocket#protocols) (Android 10 and above).

saperi22 (approved):

SharedUtils/src/main/java/com/braintreepayments/api/TLSSocketFactory.java
scannillo: Great catch, I updated to include both 1.2 & 1.3 in 8416f42967ba92ab11131f3c9f649e103f52a9a2

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
   * Fix `PayPalAccountNonce` Null Pointer Exception by ensuring that all `@NonNull` values are initialized with a non-null value.
 * PayPalNativeCheckout
   * Fix `PayPalNativeCheckoutAccountNonce` Null Pointer Exception by ensuring that all `@NonNull` values are initialized with a non-null value.
+* BraintreeCore
+ * Use TLS 1.3 for all HTTP requests
```
scannillo, line 10:
```suggestion
  * Use TLS 1.3 for all HTTP requests
```

warmkesselj (approved):

PR #1031, opened 2024-06-07
Add Analytics Event Params object

richherrera:
### Summary of changes

 - Extracts all the parameters that are necessary to send in events to an object (`AnalyticsEventParams`)

Screenshot 2024-06-07 at 12 44 56 p m

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### Authors

- @richherrera


BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventParams.kt
```diff
@@ -0,0 +1,18 @@
+package com.braintreepayments.api
+
+public data class AnalyticsEventParams private constructor(
```
tdchow, line 3:
```suggestion
data class AnalyticsEventParams private constructor(
```

The default modifier is public, so no need to specify `public` here.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventParams.kt
```diff
@@ -0,0 +1,18 @@
...
+    val payPalContextId: String?,
+    val linkType: String?,
+    val isVaultRequest: Boolean,
+) {
+    class Builder {
```
tdchow, line 8: Is this inner builder class needed? Could we instead use the `AnalyticsEventParams`'s constructor wherever we need to create new instances?

I feel that the builder pattern is more of a Java pattern where there isn't any support for named parameters or default values.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -188,20 +188,18 @@ open class BraintreeClient @VisibleForTesting internal constructor(
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
     fun sendAnalyticsEvent(
             eventName: String,
-            payPalContextId: String? = null,
-            linkType: String? = null,
-            isVaultRequest: Boolean = false
+            params: AnalyticsEventParams = AnalyticsEventParams.Builder().build()
```
tdchow, line 191: If we were to remove the builder, we could have params nullable here and default to null.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventParams.kt
```diff
@@ -0,0 +1,18 @@
+package com.braintreepayments.api
+
```
tdchow, line 2: We should add the `RestrictTo` annotation here to keep this class internal to the BT SDK:

```kotlin
```

richherrera: Addressed: [229c16b](https://github.com/braintree/braintree_android/pull/1031/commits/229c16bdaae38efcec18775a69b1c753c5e8715a)

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventParams.kt
richherrera: Thank you, addressed: [229c16b](https://github.com/braintree/braintree_android/pull/1031/commits/229c16bdaae38efcec18775a69b1c753c5e8715a)

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventParams.kt
richherrera: Got it, and I understand that the goal is to migrate everything to Kotlin. However, creating this object is to avoid having to change too many files whenever we add a new parameter to the events. The problem we have now is that we have both Java and Kotlin. This design pattern would help us create an instance without having to specify all the parameters in Java.

For example, in Java, when I only need to specify isVaultRequest, I have to do something like this:
`AnalyticsEventParams(null, null, true)`

With builder patter:
`AnalyticsEventParams.Builder.setIsVaultRequest`

The purpose of all this is that if we increase the number of parameters, we won't have to modify so many classes.

KunJeongPark (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventParams.kt
richherrera: Updated here: [930bef6](https://github.com/braintree/braintree_android/pull/1031/commits/930bef6fd4fef8eb137784283e87d9ac294b6b3b)

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
richherrera: Addressed: [930bef6](https://github.com/braintree/braintree_android/pull/1031/commits/930bef6fd4fef8eb137784283e87d9ac294b6b3b)

warmkesselj (approved):
LGTM

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventParams.kt
```diff
@@ -0,0 +1,12 @@
...
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+data class AnalyticsEventParams constructor(
+    var payPalContextId: String?,
+    var linkType: String?,
+    var isVaultRequest: Boolean,
+) {
+  constructor() : this(null, null, false)
```
sshropshire, line 11:
```suggestion
  // TODO: this is a convenience constructor for Java; remove after Kotlin migration is complete
  constructor() : this(null, null, false)
```
Adding a note to remove this in the future.

sshropshire (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventParams.kt
saperi22: @richherrera IMO, using default values would be better for this use-case.
please check https://kotlinlang.org/docs/functions.html#default-arguments.
>fun foo(
    bar: Int = 0,
    baz: Int,
) { /*...*/ }

>foo(baz = 1) // The default value bar = 0 is used

```
data class AnalyticsEventParams constructor(
    var payPalContextId: String? = null,
    var linkType: String? = null,
    var isVaultRequest: Boolean = false,
)
```
You could use any combination of the parameters and it'd work.
Add @JvmOverloads for java compatibility.

sshropshire: @saperi22 we looked into this there was one hiccup. Since there are no named params in Java the String param becomes ambiguous when it's `(payPalContextId, isVaultRequest)` vs. `(linkType, isVaultRequest)`. The compiler complains because it can't determine if the `String` in the `AnalyticsEventParams(String, boolean)` parameter tuple is a `payPalContextId` or a `linkType` without additional information.

PR #1032, opened 2024-06-07
[QL] Measure API Request Latency 

richherrera:
### Summary of changes

 - Send `start_time`, `end_time`, and `endpoint` to `FPTI` for tracking API request latency
- Add new `CoreAnalytics` object
- Add new `HttpResponse` for sending start/end timing from `HttpClient` to `BraintreeClient`
- Add new `NetworkResponseCallback` to pass `HttpResponse` from `HttpClient` to `BraintreeClient`

Previously, `SynchronousHttpClient` returned the body of the response. 
However, this is where we are now obtaining `start_time` and `end_time` to analyze the response time of the services. This led to the creation of the `HttpResponse` object. This object is sent in the new `NetworkResponseCallback` callback.

The previous `HttpResponseCallback` callback was sent from `BraintreeHttpClient` to the clients (i.e. PayPalClient). 
By creating a new callback with the `HttpResponse` object, we avoid making any changes at the payment method client level. 

In the `BraintreeClient`, we receive the `NetworkResponseCallback` callback from `HttpClient`, send the necessary `event`, and return the required information with the original `HttpResponseCallback` callback.

| Before | Now |
|----------------------------------------|---------------------------------------|
| Screenshot 2024-06-18 at 3 31 53 p m | Screenshot 2024-06-18 at 3 32 00 p m |

Screenshot 2024-06-12 at 11 14 10 a m

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### Authors

- @richherrera


BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
```diff
@@ -164,7 +173,13 @@ internal class AnalyticsClient @VisibleForTesting constructor(
...
         }
         val metadata = deviceInspector.getDeviceMetadata(context, configuration, sessionId, integration)
-        val event = AnalyticsEvent("android.crash", null, null, timestamp)
+        val event = AnalyticsEvent(
+                "android.crash",
+                null,
+                null,
+                false,
+                false,
+                timestamp = timestamp)
```
scannillo, line 182:
```suggestion
        val event = AnalyticsEvent(
            "android.crash",
            null,
            null,
            false,
            false,
            timestamp = timestamp
        )
```

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventParams.kt
```diff
@@ -4,10 +4,13 @@ import androidx.annotation.RestrictTo
...
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 data class AnalyticsEventParams constructor(
-    var payPalContextId: String?,
-    var linkType: String?,
-    var isVaultRequest: Boolean,
+    var payPalContextId: String? = null,
+    var linkType: String? = null,
+    var isVaultRequest: Boolean = false,
+    var startTime: Long = -1,
+    var endTime: Long = -1,
```
scannillo, line 11: Can we make this nullable and default this to `null` instead of `-1`? Similar to the `linkType`?

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -263,9 +289,30 @@ open class BraintreeClient @VisibleForTesting internal constructor(
...
+                                    responseCallback.onResult(it.body, null)
+                                } catch (jsonException: JSONException) {
+                                    responseCallback.onResult(null, jsonException)
+                                }
+                            } ?: httpError?.let { error ->
+                                val errorMessageFormat = "Request has failed: %s"
+                                val errorMessage = String.format(errorMessageFormat, error.message)
+                                val exception = BraintreeException(errorMessage, error)
+                                responseCallback.onResult(null, exception)
+                            }
```
scannillo, line 314: I don't think we need to handle this error. If this analytics call fails - we should still proceed

SharedUtils/src/main/java/com/braintreepayments/api/BTHttpResponse.kt
```diff
@@ -0,0 +1,7 @@
+package com.braintreepayments.api
+
+class BTHttpResponse constructor(
```
sshropshire, line 3:
```suggestion
class HttpResponse constructor(
```
We should get rid of the "BT" prefixes. Packages in both Java and Kotlin remove the need for us to prefix our class names.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -160,7 +161,7 @@ open class BraintreeClient @VisibleForTesting internal constructor(
     open fun getConfiguration(callback: ConfigurationCallback) {
         getAuthorization { authorization, authError ->
             if (authorization != null) {
-                configurationLoader.loadConfiguration(authorization) { configuration, configError ->
+                configurationLoader.loadConfiguration(authorization, this) { configuration, configError ->
```
sshropshire, line 164: **Random Thought**

We should consider delaying this configuration fetch. On Android, `sendAnalytics()` sends the event data to a SQLite store and schedules it to upload 30 seconds later. Fetching the configuration right before the scheduled upload would be an improvement on our current design.

BraintreeCore/src/main/java/com/braintreepayments/api/BTAPITiming.kt
```diff
@@ -0,0 +1,5 @@
+package com.braintreepayments.api
+
+fun interface BTAPITiming {
```
jaxdesmarais, line 3: I think we can remove `BT` from these names, that's an iOS specific thing for Obj-C interoperability.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -198,8 +199,11 @@ open class BraintreeClient @VisibleForTesting internal constructor(
                             eventName,
                             params.payPalContextId,
                             params.linkType,
-                            venmoInstalled = isVenmoInstalled,
-                            isVaultRequest = params.isVaultRequest
+                            isVenmoInstalled,
```
jaxdesmarais, line 202: General style question for @saperi22 and @sshropshire, do we prefer the `param = param` syntax for Kotlin in v5 or just `param`?

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -232,7 +236,29 @@ open class BraintreeClient @VisibleForTesting internal constructor(
...
                 getConfiguration { configuration, configError ->
                     if (configuration != null) {
-                        httpClient.get(url, configuration, authorization, responseCallback)
+                        httpClient.get(url, configuration, authorization) { response, httpError ->
+                            response?.let {
+                                try {
+                                    val params = AnalyticsEventParams(
+                                        startTime = it.startTime,
+                                        endTime = it.endTime,
+                                        endpoint = url
```
jaxdesmarais, line 245: Are we stripping out the `merchants/merchant_id/client_api` like on iOS ([here](https://github.com/braintree/braintree_ios/blob/dbe62b831090eb57ffc909634643bb30df11fb3f/Sources/BraintreeCore/BTAPIClient.swift#L428))

BraintreeCore/src/test/java/com/braintreepayments/api/BraintreeClientUnitTest.kt
```diff
@@ -156,16 +156,19 @@ class BraintreeClientUnitTest {
         val params = createDefaultParams(configurationLoader, authorizationLoader)
         val sut = BraintreeClient(params)
         val httpResponseCallback = mockk<HttpResponseCallback>(relaxed = true)
+        val btHttpResponseCallbackSlot = slot<BTHttpResponseCallback>()
```
jaxdesmarais, line 159:
```suggestion
        val httpResponseCallbackSlot = slot<HttpResponseCallback>()
```

SharedUtils/src/main/java/com/braintreepayments/api/BTHttpResponse.kt
jaxdesmarais: ➕ 1

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
richherrera: In Android, using `param = param` is necessary to specify the name of the parameter we are assigning a value to. Previously, between `linkType` and `venmoInstalled`, we had a parameter called `timestamp`, which had a default value. I moved this parameter to the end. Now, since all other parameters require a value assignment, it is no longer necessary to specify the name and value

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventParams.kt
richherrera: Yes we can, addressed here, let me know what do you think: [1a32ee9](https://github.com/braintree/braintree_android/pull/1032/commits/1a32ee9f0998d5de4fd6e5c4ae5a539ceb554c49)

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
richherrera: This error comes from HttpClient.notifyErrorOnMainThread, let me know what do you think: [ee093d0] (https://github.com/braintree/braintree_android/pull/1032/commits/ee093d0f4e6042000a09d9b09bb905210f7abd18)

BraintreeCore/src/main/java/com/braintreepayments/api/BTAPITiming.kt
richherrera: Thanks for your feedback. Addressed: [8bb8aae](https://github.com/braintree/braintree_android/pull/1032/commits/8bb8aae8ee7df8da642079f1ecb6c6f1e05f067b)

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
richherrera: Nice catch, added here: [ea509f1](https://github.com/braintree/braintree_android/pull/1032/commits/ea509f1a5a548753e744baf415ddc94aa3973605)

SharedUtils/src/main/java/com/braintreepayments/api/BTHttpResponse.kt
richherrera: Fix: [15c60c1](https://github.com/braintree/braintree_android/pull/1032/commits/15c60c180911a84ad08427c09531b8c5221056b4)

SharedUtils/src/main/java/com/braintreepayments/api/BTHttpResponse.kt
```diff
@@ -0,0 +1,7 @@
...
+    var startTime: Long = 0,
+    var endTime: Long = 0,
```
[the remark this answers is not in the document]
richherrera: Fix it

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
scannillo: This is a good suggestion. We implemented it on iOS here - https://github.com/braintree/braintree_ios/pull/1337

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -288,9 +335,31 @@ open class BraintreeClient @VisibleForTesting internal constructor(
...
                             payload,
                             configuration,
-                            authorization,
-                            responseCallback
-                        )
+                            authorization
+                        ) { response, httpError ->
+                            response?.let {
+                                try {
+                                    val query = getSubstringAfterKey(payload.toString(), "query")
```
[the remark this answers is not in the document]
scannillo: Ah, I misread this code! It does account for both query & mutation requests in GraphQL.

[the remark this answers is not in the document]
scannillo: However this sends the _full_ query as the `endpoint`, for example:

* `mutation CreateVenmoPaymentContext($input: CreateVenmoPaymentContextInput!) { createVenmoPaymentContext(input: $input) { venmoPaymentContext { id } } }`
* `query PaymentContext($id: ID!) { node(id: $id) { ... on VenmoPaymentContext { paymentMethodId userName payerInfo { firstName lastName phoneNumber email externalId userName shippingAddress { fullName addressLine1 addressLine2 adminArea1 adminArea2 postalCode countryCode } billingAddress { fullName addressLine1 addressLine2 adminArea1 adminArea2 postalCode countryCode } } } } }`

Could we instead just send the type & name, like:

* `mutation CreateVenmoPaymentContext`
* `query PaymentContext`

(Also we have not yet done this on iOS, so we can mirror whatever we decide there)

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
jaxdesmarais: But we can still use it if we want correct? I was just curious if we had a style guide established yet for a preference since I think we are currently doing a mix of both.

BraintreeCore/src/main/java/com/braintreepayments/api/HttpNoTimingResponse.kt
```diff
@@ -0,0 +1,7 @@
+package com.braintreepayments.api
+
+internal class HttpNoTimingResponse : HttpTimingResponseCallback {
```
jaxdesmarais, line 3: Are we using this anywhere?

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
richherrera: The combination occurs when one of the parameters with a default value is not assigned a value. By leaving the default value, Android needs to know the next parameter we are configuring. That’s why the compiler automatically adds the parameter name.

jaxdesmarais: Yeah I understand that piece, was moreso curious if we have decided on a general style on the Android side for Kotlin features since looking through the SDK it's a bit all over the place. The following is also valid syntax:
```kotlin
val event = AnalyticsEvent(
    name = eventName,
    payPalContextId = params.payPalContextId,
    linkType = params.linkType,
    venmoInstalled = isVenmoInstalled,
    isVaultRequest = params.isVaultRequest,
    startTime = params.startTime,
    endTime = params.endTime,
    endpoint = params.endpoint
)
```
This isn't a blocker and may not be decided by the Android folks yet so just wanted to bring it up. It does not seem like our pattern is to only used when assigning a value, we are instead kinda all over the place. 🙈

richherrera: Yes, thanks for the feedback. I'd like to know what the rest of the team thinks about all of this

BraintreeCore/src/main/java/com/braintreepayments/api/HttpNoTimingResponse.kt
richherrera: Yep, Even Steven asked if we could make the callback optional, but it's something I need to investigate.
https://github.com/braintree/braintree_android/blob/d6ec39e79bf546e5ce57bd9cb460947e4acfe7e1/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt#L192

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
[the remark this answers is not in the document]
richherrera: Sending `mutation mutationName` [d6ec39e](https://github.com/braintree/braintree_android/pull/1032/commits/d6ec39e79bf546e5ce57bd9cb460947e4acfe7e1)

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -47,7 +48,7 @@ open class BraintreeClient @VisibleForTesting internal constructor(
      */
     @get:RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
     val appLinkReturnUri: Uri?,
-) {
+) : APITiming {
```
sarahkoop, line 51: I think this is _technically_ a breaking change. We could make the case for it, but it also seems like maybe the API timing should live in the Analytics client? Could the `sendEvent` with timestamps method be moved there and be referenced from both `BraintreeClient` and `ConfigurationLoader`?

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -512,12 +512,19 @@ open class BraintreeClient @VisibleForTesting internal constructor(
     }
 
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-    fun getSubstringAfterKey(jsonString: String, key: String): String {
-        val keyWithQuotes = "\"$key\":\""
-        val startIndex = jsonString.indexOf(keyWithQuotes) + keyWithQuotes.length
-        val endIndex = jsonString.indexOf("\",", startIndex)
-        return if (startIndex != -1 && endIndex != -1) {
-            jsonString.substring(startIndex, endIndex)
+    fun getGraphQLMutationName(jsonString: String): String {
```
KunJeongPark, line 515: We have ticket item for this in iOS, or is that something different?

richherrera: Yep, we have a ticket `DTMOBILES`-683, last sprint, last sprint, I checked this out a bit, but I think there's some kind of redirection that's stopping us from getting the query. I'll need to dig into it more

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
```diff
@@ -173,7 +189,7 @@ internal class AnalyticsClient @VisibleForTesting constructor(
                 data = analyticsRequest.toString(),
                 configuration = null,
                 authorization = authorization,
-                callback = HttpNoResponse()
+                callback = HttpNoTimingResponse()
```
sshropshire, line 192: I think we can make callback `null` by default. On iOS I think it's needed because we're using generics. Here we can make it optional and guard it's invocation with [an optional unwrap operator](https://stackoverflow.com/a/48185699).

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -466,4 +510,31 @@ open class BraintreeClient @VisibleForTesting internal constructor(
     open fun launchesBrowserSwitchAsNewTask(launchesBrowserSwitchAsNewTask: Boolean) {
         this.launchesBrowserSwitchAsNewTask = launchesBrowserSwitchAsNewTask
     }
+
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    fun getGraphQLMutationName(jsonString: String): String {
+        val regex = """"query"\s*:\s*"(.*?)"""".toRegex()
+        val matchResult = regex.find(jsonString)
+
+        return if (matchResult != null) {
```
sshropshire, line 519: q: Can we get the value from the GraphQL POST body instead since it's already in JSON?

SharedUtils/src/main/java/com/braintreepayments/api/SynchronousHttpClient.java
```diff
@@ -75,9 +77,13 @@ String request(HttpRequest httpRequest) throws Exception {
 
         try {
             int responseCode = connection.getResponseCode();
-            return parser.parse(responseCode, connection);
+            long endTime = System.currentTimeMillis();
+
+            String responseBody = parser.parse(responseCode, connection);
+
+            return new HttpResponse(startTime, endTime, responseBody);
```
sshropshire, line 84: Random thought: We should circle back and preserve the response headers here as well. I could see us surfacing PayPal debug IDs in error messages in the near future.

BraintreeCore/src/main/java/com/braintreepayments/api/APITiming.kt
```diff
@@ -0,0 +1,5 @@
+package com.braintreepayments.api
+
+fun interface APITiming {
+    fun sendEvent(startTime: Long, endTime: Long, endpoint: String)
```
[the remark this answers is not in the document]
richherrera: This interface is used to communicate `ConfigurationLoader` and `BraintreeClient`

BraintreeCore/src/main/java/com/braintreepayments/api/HttpNoTimingResponse.kt
richherrera: Removed [fa734e9](https://github.com/braintree/braintree_android/pull/1032/commits/fa734e964296b2dda10156edc4ccc0d271284fc1)

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
richherrera: Ready, thanks for your comments and sharing documentation [741e0aa](https://github.com/braintree/braintree_android/pull/1032/commits/741e0aa71b124b3dc3f355a4c4b35ec0d56183c3)

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
richherrera: Actually, the body is of type `string`, any ideas? do you have any references I could use?

sshropshire: There may be some refactoring we can do to access the JSON directly. We can also clean this up in a separate PR.

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -466,4 +513,35 @@ open class BraintreeClient @VisibleForTesting internal constructor(
     open fun launchesBrowserSwitchAsNewTask(launchesBrowserSwitchAsNewTask: Boolean) {
         this.launchesBrowserSwitchAsNewTask = launchesBrowserSwitchAsNewTask
     }
+
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    fun getGraphQLMutationName(jsonString: String): String {
```
sshropshire, line 518: I think we can remove this method now since we're getting the query name directly from the JSON.

SharedUtils/src/main/java/com/braintreepayments/api/NetworkResponseCallback.kt
```diff
@@ -0,0 +1,13 @@
+package com.braintreepayments.api
+
+import androidx.annotation.MainThread
+import androidx.annotation.RestrictTo
+
+/**
+ * @suppress
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+fun interface NetworkResponseCallback {
```
sshropshire, line 10: Any chance we could rename this to `HttpResponseCallback`? The `Http` prefix is helpful to call out the grouping of other HTTP related files.

sshropshire (approved):
🔥

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
richherrera: Thank you, removed here: [6d06cfe](https://github.com/braintree/braintree_android/pull/1032/commits/6d06cfeded1574f3c06e6d8c2104d261f224a297)

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
```diff
@@ -280,17 +307,37 @@ open class BraintreeClient @VisibleForTesting internal constructor(
...
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-    fun sendGraphQLPOST(payload: String?, responseCallback: HttpResponseCallback) {
+    fun sendGraphQLPOST(json: JSONObject?, responseCallback: HttpResponseCallback) {
         getAuthorization { authorization, authError ->
             if (authorization != null) {
                 getConfiguration { configuration, configError ->
                     if (configuration != null) {
                         graphQLClient.post(
-                            payload,
+                            json.toString(),
```
sarahkoop, line 316:
```suggestion
                            json?.toString(),
```

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
```diff
@@ -91,19 +91,19 @@ internal class BraintreeHttpClient(
         configuration: Configuration?,
         authorization: Authorization?,
         additionalHeaders: Map<String, String> = emptyMap(),
-        callback: HttpResponseCallback
+        callback: NetworkResponseCallback?
```
sarahkoop, line 94: Does this need to be a nullable callback now? This leaves the possibility that we will just return with no callback at all from this post request

sshropshire: I actually suggested this one. We originally had `HttpNoResponseCallback` that would no-op, but a nullable callback conveys the same meaning. It doesn't have a default arg so an explicit `null` will need to be passed. Maybe there's a better way does it make sense to revert to `HttpNoResponseCallback`?

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeClient.kt
richherrera: Updated: [0436d22](https://github.com/braintree/braintree_android/pull/1032/commits/0436d22662af20121c31291140896afb105083b5)

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
sarahkoop: Ahh ok I think I was missing that this is callback is only null for analytics requests where it would be ok if we had a no-op with no response. I think it's good as is then

sarahkoop (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
richherrera: Thanks to both!!

Issue #1020, opened 2024-06-03
NullPointerException: com.braintreepayments.api.TLSCertificatePinning.getCertInputStream (TLSCertificatePinning.java:719)


kimrtaylor:
### Braintree SDK Version

4.45.0

### Environment

Production

### Android Version & Device

Multiple

### Braintree dependencies

com.braintreepayments.api:card:4.45.0

### Describe the bug

We're seeing a new crash affecting some of our customers, since updating from 4.44.0 to 4.45.0. It occurs during initialisation of BraintreeClient

### To reproduce

```
Fatal Exception: java.lang.NullPointerException
Attempt to invoke virtual method 'byte[] java.lang.String.getBytes(java.nio.charset.Charset)' on a null object reference
com.braintreepayments.api.TLSCertificatePinning.getCertInputStream (TLSCertificatePinning.java:719)
com.braintreepayments.api.BraintreeHttpClient$Companion.createDefaultHttpClient (BraintreeHttpClient.java:178)
com.braintreepayments.api.BraintreeHttpClient$Companion.access$createDefaultHttpClient (BraintreeHttpClient.java:172)
com.braintreepayments.api.BraintreeHttpClient.<init> (BraintreeHttpClient.kt:12)
com.braintreepayments.api.AnalyticsClient.<init> (AnalyticsClient.kt:25)
com.braintreepayments.api.BraintreeClientParams.<init> (BraintreeClientParams.kt:14)
com.braintreepayments.api.BraintreeClientParams.<init> (BraintreeClientParams.kt:22)
com.braintreepayments.api.BraintreeClient.<init> (BraintreeClient.kt:70)
com.braintreepayments.api.BraintreeClient.<init> (BraintreeClient.kt:80)
```

### Expected behavior

exception handled

sshropshire:
Hey @kimrtaylor thanks for using the Braintree SDK for Android. I'd like to help pinpoint the issue and hopefully find a fix.

One thing that seems odd to me is the stack trace mentions the exception originating form `TLSCertificatePinning.java` in our SDK. [TLSCetificatePinning](https://github.com/braintree/braintree_android/blob/main/BraintreeCore/src/main/java/com/braintreepayments/api/TLSCertificatePinning.kt) in our project is written in Kotlin.

Can you help me to answer a few questions:

- Have you been able to reproduce the issue?
- Is your app native Android or React Native?
[...]

PR #1036, opened 2024-06-11
Refactor TLSCertificatePinning `certInputStream` Accessor

sshropshire:
### Summary of changes

 - Reference Issue: https://github.com/braintree/braintree_android/issues/1020
 - There's an occasional NPE crash in `TLSCertificatePinning.getCertInputStream()`
 - This PR replaces `TLSCertificatePinning.certInputStream` with an explicit `TLSCertificatePinning.createCertificateInputStream` factory method
 - Looking to see if refactoring to Kotlin constant `val` helps resolve the issue by converting the certificate stream to a byte array only once–instead of once every time a certificate input stream is needed

### Checklist

 - [x] Added a changelog entry

@sshropshire


BraintreeCore/src/main/java/com/braintreepayments/api/BraintreeHttpClient.kt
```diff
@@ -178,6 +179,7 @@ internal class BraintreeHttpClient(
         private const val USER_AGENT_HEADER = "User-Agent"
         private const val CLIENT_KEY_HEADER = "Client-Key"
 
+        @Throws(SSLException::class)
```
sshropshire, line 182: On the other hand, `TLSSocketFactory` constructors may throw an `SSLException`.

saperi22 (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/TLSCertificatePinning.kt
```diff
@@ -714,7 +712,7 @@ CkPa1mnI2g7vVq2d66WZmpsDtY1EwUya7vmZTxw7BFlntV7lrVxPLhUbq1wJGIbT
 uEnXgUGZxnLYZfCNsC/UAQxvKKpxBLWus/n1InM/97HfA31WJsvQ
 -----END CERTIFICATE-----"""
 
-    @get:Throws(SSLException::class)
-    val certInputStream: InputStream
-        get() = ByteArrayInputStream(CERTIFICATES.toByteArray(StandardCharsets.UTF_8))
+    private val certificateByteArray = CERTIFICATES.toByteArray(StandardCharsets.UTF_8)
+
+    fun createCertificateInputStream() = ByteArrayInputStream(certificateByteArray)
```
sshropshire, line 717: Through debugging, I found that each networking client needs its own input stream for certificate pinning. When `BraintreeHttpClient` and `BraintreeGraphQLClient` share an input stream, an `SSLException` is thrown. I believe the first networking client to acquire the shared InputStream succeeds, but subsequent clients will fail because the singleton InputStream is closed.

Since a new `InputStream` is required to create each networking client, I decided to make its intent more explicit by refactoring the getter into a `createCertificateInputStream()` factory method.

agedd (approved):

PR #1038, opened 2024-06-14
Fixups for main merge into v5

saperi22:
### Summary of changes

 - Making changes to the demo app to make it work with v5 changes.

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/PayPalMessagingFragment.kt
+++ b/Demo/src/main/java/com/braintreepayments/demo/PayPalMessagingFragment.kt
@@ -13,7 +13,6 @@ import com.braintreepayments.api.PayPalMessagingOfferType
 import com.braintreepayments.api.PayPalMessagingRequest
 import com.braintreepayments.api.PayPalMessagingTextAlignment
 import com.braintreepayments.api.PayPalMessagingView
-import com.braintreepayments.api.core.BraintreeClient
 
 @OptIn(ExperimentalBetaApi::class)
 class PayPalMessagingFragment : BaseFragment(), PayPalMessagingListener {
@@ -38,9 +37,7 @@ class PayPalMessagingFragment : BaseFragment(), PayPalMessagingListener {
             color = PayPalMessagingColor.BLACK
         )
 
-        // TODO: Refactor BraintreeClient out of PayPalMessagingView
-        val braintreeClient = BraintreeClient(requireContext(), authStringArg)
-        val payPalMessagingView = PayPalMessagingView(braintreeClient, requireContext())
+        val payPalMessagingView = PayPalMessagingView(requireContext(), authStringArg)
         payPalMessagingView.setListener(this)
         payPalMessagingView.start(payPalMessagingRequest)
         payPalMessagingView.layoutParams = ViewGroup.LayoutParams(
--- a/Demo/src/main/java/com/braintreepayments/demo/PayPalMessagingFragment.kt
+++ b/Demo/src/main/java/com/braintreepayments/demo/PayPalMessagingFragment.kt
@@ -38,6 +38,7 @@ class PayPalMessagingFragment : BaseFragment(), PayPalMessagingListener {
             color = PayPalMessagingColor.BLACK
         )
 
+        // TODO: Refactor BraintreeClient out of PayPalMessagingView
         val braintreeClient = BraintreeClient(requireContext(), authStringArg)
         val payPalMessagingView = PayPalMessagingView(braintreeClient, requireContext())
         payPalMessagingView.setListener(this)
--- a/Demo/src/main/java/com/braintreepayments/demo/PayPalMessagingFragment.kt
+++ b/Demo/src/main/java/com/braintreepayments/demo/PayPalMessagingFragment.kt
@@ -4,6 +4,7 @@ import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import android.widget.LinearLayout
 import com.braintreepayments.api.ExperimentalBetaApi
 import com.braintreepayments.api.PayPalMessagingColor
 import com.braintreepayments.api.PayPalMessagingListener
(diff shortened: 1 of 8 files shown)


saperi22:
@sarahkoop I'm not sure I got the createPaymentAuthRequest/launch/handleReturnToApp/tokenize correct. It feels a bit convoluted to work with. Is what I'm doing correct?
We should update the v5 migration guide with the sample after we have working code.

Demo/src/main/java/com/braintreepayments/demo/PayPalMessagingFragment.kt
```diff
@@ -35,17 +38,17 @@ class PayPalMessagingFragment : BaseFragment(), PayPalMessagingListener {
...
-//        payPalMessagingView.setListener(this)
-//        payPalMessagingView.start(payPalMessagingRequest)
-//        payPalMessagingView.layoutParams = ViewGroup.LayoutParams(
-//            ViewGroup.LayoutParams.MATCH_PARENT,
-//            ViewGroup.LayoutParams.MATCH_PARENT
-//        )
-//
-//        val messagingView: LinearLayout = view.findViewById(R.id.content)
-//        messagingView.addView(payPalMessagingView)
+        val braintreeClient = BraintreeClient(requireContext(), authStringArg)
```
sarahkoop, line 41: It can be a separate TODO but I think we want to encapsulate the `BraintreeClient` within the `PayPalMessagingView` - we made that change with other feature clients across v5 (merchants no longer need to be aware of `BraintreeClient` at all

Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragment.kt
```diff
@@ -31,142 +53,224 @@ class ShopperInsightsFragment : BaseFragment() {
...
-//        venmoClient = VenmoClient(this, braintreeClient)
-//        venmoClient.setListener(this)
-//
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View? {
+        braintreeClient = BraintreeClient(requireContext(), authStringArg)
+        shopperInsightsClient = ShopperInsightsClient(braintreeClient)
```
sarahkoop, line 69: Same thing here - let's instantiate `BraintreeClient` under the hood in `ShopperInsightsClient`

Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragment.kt
```diff
@@ -31,142 +53,224 @@ class ShopperInsightsFragment : BaseFragment() {
...
+
+        venmoClient = VenmoClient(requireContext(), super.getAuthStringArg())
+
+        // TODO: What does manual browser switch do?
```
sarahkoop, line 73: Manual browser switching was a pattern added to v4 for use cases when encapsulating browser switching didn't work for merchant architecture - in v5 all of browser switching is "manual" (the merchants have to handle getting results) so separating the patterns isn't needed

Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragment.kt
```diff
@@ -31,142 +53,224 @@ class ShopperInsightsFragment : BaseFragment() {
...
+            )
+        ) {
+            if (it == null) return@createPaymentAuthRequest
+            when (it) {
+                is PayPalPaymentAuthRequest.Failure -> {
+                    handleError(it.error)
+                }
+
+                is PayPalPaymentAuthRequest.ReadyToLaunch -> {
+                    paypalPendingRequest = paypalLauncher.launch(requireActivity(), it)
```
sarahkoop, line 247: I think we'd want to handle the `Started` or `Failure` pending request here - if it's `Started`, store it, if it's a failure display the error message. Same goes for Venmo, then we don't need to handle error scenarios in `handleReturnToApp` at that point it should only be a successfully started request

Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragment.kt
```diff
@@ -31,142 +53,224 @@ class ShopperInsightsFragment : BaseFragment() {
...
+                                    val action =
+                                        ShopperInsightsFragmentDirections
+                                            .actionShopperInsightsFragmentToDisplayNonceFragment(
+                                            it.nonce
+                                        )
+                                    NavHostFragment.findNavController(this).navigate(action)
+                                }
+
+                                is PayPalResult.Failure -> {}
+                                is PayPalResult.Cancel -> {}
```
jaxdesmarais, line 130: Should we handle these states?

Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragment.kt
```diff
@@ -31,142 +53,224 @@ class ShopperInsightsFragment : BaseFragment() {
...
+                        Toast.makeText(
+                            requireContext(),
+                            "Failed to authenticate user.",
+                            Toast.LENGTH_LONG
+                        )
+                            .show()
+                    }
+                }
+
+                is PayPalPendingRequest.Failure -> {}
```
jaxdesmarais, line 144: Same question here and below on Venmo, not sure if we should be handling these states in some way

Demo/src/main/java/com/braintreepayments/demo/PayPalMessagingFragment.kt
```diff
@@ -35,17 +38,17 @@ class PayPalMessagingFragment : BaseFragment(), PayPalMessagingListener {
             color = PayPalMessagingColor.BLACK
         )
 
-        // TODO: Refactor BraintreeClient out of PayPalMessagingView
```
jaxdesmarais, line 38: I think we can leave this TODO for now

Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragment.kt
```diff
@@ -31,142 +53,233 @@ class ShopperInsightsFragment : BaseFragment() {
...
+                        is PayPalResult.Success -> {
+                            val action =
+                                ShopperInsightsFragmentDirections
+                                    .actionShopperInsightsFragmentToDisplayNonceFragment(
+                                        it.nonce
+                                    )
+                            NavHostFragment.findNavController(this).navigate(action)
+                        }
+
+                        is PayPalResult.Failure -> {}
```
sarahkoop, line 127: Let's show error messages for failure or cancel message for both PayPal and Venmo in these methods

jaxdesmarais (approved):

sshropshire (approved):

PR #1044, opened 2024-06-18
Add custom fields to threedsecure request

warmkesselj:
### Summary of changes

Add custom fields to 3D Secure request for verifyCards.

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### Authors
@agedd 
@jwarmkessel 
@warmkesselj 
@KunJeongPark


ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
```diff
@@ -82,6 +84,15 @@ public void setNonce(@Nullable String nonce) {
         this.nonce = nonce;
     }
 
+    /**
+     * Set Custom Fields
+     *
+     * @param customFields /// Object where each key is the name of a custom field which has been configured in the Control Panel. In the Control Panel you can configure 3D Secure Rules which trigger on certain values.
+     */
+    public void setCustomFields(@Nullable HashMap<String, Object> customFields) {
```
scannillo, line 92: We talked about this being a `JSONObject` type, which is commonly used in Android development to model JSON

sshropshire: Using the interface `Map` type is a best practice.
```suggestion
    public void setCustomFields(@Nullable Map<String, Object> customFields) {
```

sshropshire: On the other hand, I'm not sure if we have a set pattern of using `JSONObject`. It's hard to say which is better.

If the JSON itself single layered, an alternative could be to offer a key-value setter like `setCustomField(String key, String value)`. But if the `customFields` property supports nested JSON schemas then it gets more complex.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
```diff
@@ -56,6 +57,7 @@ public class ThreeDSecureRequest implements Parcelable {
 
     private String nonce;
     private String amount;
+    private HashMap<String,Object> customFields;
```
sshropshire, line 60:
```suggestion
    private Map<String,Object> customFields;
```
Same best practice here for the internal variable declaration.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
warmkesselj: I'm getting more information about what is expected in terms of custom values. Nested/layered, n values or just two maximum.

sshropshire: Ok cool yeah. Honestly if it's just for merchants it might make sense to make it a `String` and let them encode / decode the value. We're doing something similar on Browser Switch with opaque values to reduce the SDK's responsibility.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
```diff
@@ -82,6 +85,15 @@ public void setNonce(@Nullable String nonce) {
         this.nonce = nonce;
     }
 
+    /**
+     * Set Custom Fields
+     *
+     * @param customFields /// Object where each key is the name of a custom field which has been configured in the Control Panel. In the Control Panel you can configure 3D Secure Rules which trigger on certain values.
```
tdchow, line 91:
```suggestion
     * @param customFields Object where each key is the name of a custom field which has been configured in the Control Panel. In the Control Panel you can configure 3D Secure Rules which trigger on certain values.
```

I think this might be swift comment syntax 😄

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
sarahkoop: I see the decision was made to go with a Map - can we document the reasoning behind that? To me it still makes more sense to just have the merchant pass in a JSONObject since that's what we turn it into under the hood

KunJeongPark: I just went with the latest comments here saying that we don't have a pattern set of using JSONObject. 
Just following the thread and what was previously done. I'm open to changing and figuring out how to do that.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
```diff
@@ -56,6 +58,7 @@ public class ThreeDSecureRequest implements Parcelable {
 
     private String nonce;
     private String amount;
+    private Map<String,Object> customFields;
```
sshropshire, line 61:
```suggestion
    private Map<String, String> customFields;
```
Can we change this to align with iOS? `Object` in Java is like `Any` in Swift. Changing to Map<String, String> would align with iOS's [[String: String]](https://github.com/braintree/braintree_ios/blob/f1444f07489bbb306a3adf2e18c6c493036ff55a/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift#L16) interface better.

ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureRequest.java
```diff
@@ -429,6 +446,15 @@ public void writeToParcel(Parcel dest, int flags) {
...
         dest.writeParcelable(v1UiCustomization, flags);
         dest.writeString(accountType);
+        if (customFields != null) {
+            dest.writeInt(customFields.size());
+            for (Map.Entry<String, Object> entry: customFields.entrySet()) {
+                dest.writeString(entry.getKey());
+                dest.writeValue(entry.getValue());
+            }
+        } else {
+            dest.writeInt(0);
```
sshropshire, line 456:
```suggestion
            // set custom fields size == 0 so we know not to create a customFields map when deserializing
            dest.writeInt(0);
```
Adding a comment to callout that this is related to custom fields.

tdchow (approved):
Looks good! Do we have a story to update the demo app to test the custom fields param?

KunJeongPark:
> Looks good! Do we have a story to update the demo app to test the custom fields param?

Thank you! That's a good call. I don't see one in backlog. I will create one.

sshropshire (approved):

PR #1049, opened 2024-06-27
Update v5 migration guide and make callback request objects non null

tdchow:
### Summary of changes

 - Resolve issues in v5 migration code snippets
 - Update callback result objects to non-null

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/v5_MIGRATION_GUIDE.md
+++ b/v5_MIGRATION_GUIDE.md
@@ -64,10 +64,12 @@ callback)`, where `riskCorrelationId` is an optional client metadata ID.
 
 ```kotlin
 val dataCollector = DataCollector(context, authorization)
-dataCollector.collectDeviceData(context) { result ->
+val dataCollectoRequest = DataCollectorRequest(hasUserLocationConsent)
+
+dataCollector.collectDeviceData(context, dataCollectoRequest) { result ->
     if (result is DataCollectorResult.Success) {
         // send result.deviceData to your server
-    } 
+    }
 }
 ```
 
@@ -92,6 +94,8 @@ The result handling of fetching American Express rewards balance has been update
 `AmericanExpressGetRewardsBalanceCallback` returns a single `AmericanExpressResult` object
 
 ```kotlin
+val americanExpressClient = AmericanExpressClient(context, "TOKENIZATION_KEY_OR_CLIENT_TOKEN")
+
 americanExpressClient.getRewardsBalance(nonce, currencyCode) { result ->
     when(result) {
         is AmericanExpressResult.Success -> { /* handle result.rewardsBalance */ }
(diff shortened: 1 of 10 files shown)


v5_MIGRATION_GUIDE.md
```diff
@@ -130,16 +143,16 @@ class MyActivity : FragmentActivity() {
     private lateinit var venmoClient: VenmoClient
     
     override fun onCreate(savedInstanceState: Bundle?) {
-+       // can initialize clients outside of onCreate if desired
--       initializeClients()
-+       venmoLauncher = VenmoLauncher()  
++       // can initialize Venmo classes outside of onCreate if desired
+       initializeVenmo()
```
sshropshire, line 147:
```suggestion
        initializeVenmo()
```
nit: minor spacing alignment issue.

sshropshire (approved):

v5_MIGRATION_GUIDE.md
```diff
@@ -149,41 +162,43 @@ class MyActivity : FragmentActivity() {
     
     // ALL OTHER ACTIVITY LAUNCH MODES 
     override fun onResume() {
-+       handleReturnToAppFromBrowser(requireActivity().intent)
++       handleReturnToAppFromBrowser(intent)
```
sarahkoop, line 165: Does this work in Kotlin? I thought you had to get the intent from the activity

sarahkoop: Oh maybe that's only in a fragment implementation nvm 🤦‍♀️

v5_MIGRATION_GUIDE.md
```diff
@@ -223,7 +238,7 @@ class MyActivity : FragmentActivity() {
     
     override fun onCreate(savedInstanceState: Bundle?) {
 +       // can initialize clients outside of onCreate if desired
```
sarahkoop, line 240: The `GooglePayLauncher` uses the android activity result API under the hood (registers for activity results from Google Pay SDK) so it actually _does_ need to be instantiated in `onCreate`(the client doesn't need to be though) - idk if calling out that the others dont is helpful or confuses things. What do you think?

v5_MIGRATION_GUIDE.md
tdchow: The example class we currently have is an Activity:
```kotlin
class MyActivity : FragmentActivity() {
```

Would we want to make this a Fragment?

sarahkoop: Nah it doesn't need to be a fragment - it actually seems like with compose the Android ecosystem is now moving away from fragments

v5_MIGRATION_GUIDE.md
tdchow: Ah I see. Yeah maybe calling out that the client doesn't need to be created in `onCreate` would cause more confusion. Were there merchants who needed to instantiate Client classes outside of 
onCreate?

sarahkoop: Yes one of the problems in v4 is that the clients need to be instantiated in `OnCreate` because they use that activity result API, but they also need authorization, which merchants often fetch async. So breaking them into two classes, one that needs auth and can be instantiated after `OnCreate`, and one that doesn't need auth but does need to be instantiated in `OnCreate` was the solution.

tdchow: Got it! What do you think about adding comments for both launcher and client creation for each integration? That could help remove confusion.

sarahkoop: Yes sounds good!

sarahkoop (approved):

PR #1051, opened 2024-07-01
Store Analytics Events as JSON Blobs in Room

sshropshire:
### Summary of changes

 - This PR refactors analytics event storage in room to store a single piece of JSON
 - This allows us to add new Analytics fields without having to perform Room schema migrations

### Checklist

@sshropshire


BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
```diff
@@ -16,14 +15,16 @@ import java.util.concurrent.TimeUnit
 
 @Suppress("SwallowedException", "TooGenericExceptionCaught")
 internal class AnalyticsClient @VisibleForTesting constructor(
```
tdchow, line 17: Outside of this PR's scope: What are your thoughts on using default values in the constructor instead of defining multiple constructors? 

This entire class is internal, so we wouldn't be leaking anything to the merchant. We could also remove the `@VisibleForTesting` annotation.

This might be a solution to the DI issue I ran into in my other PR.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.kt
```diff
@@ -3,24 +3,30 @@ package com.braintreepayments.api
...
-        ]
+    version = 7,
+    entities = [AnalyticsEventBlob::class],
+    autoMigrations = [
+        AutoMigration(from = 1, to = 2),
+        AutoMigration(from = 2, to = 3),
+        AutoMigration(from = 3, to = 4),
+        AutoMigration(from = 4, to = 5),
+        AutoMigration(from = 5, to = 6),
+        AutoMigration(from = 6, to = 7, spec = AnalyticsDatabase.DeleteAnalyticsEventTableAutoMigration::class)
```
tdchow, line 21: This should be the last migration we'll ever need to do ❤️

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventBlob.kt
```diff
@@ -0,0 +1,17 @@
+package com.braintreepayments.api
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+
+// Ref: https://www.reddit.com/r/javascript/comments/10st04/what_is_a_json_blob/
```
tdchow, line 7: Should we remove URL references in code? One concern is that the reddit post is removed at a later date.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventBlob.kt
```diff
@@ -0,0 +1,17 @@
...
+@Entity(tableName = "analytics_event_blob")
+data class AnalyticsEventBlob(
+    @ColumnInfo(name = "json_string") val jsonString: String
+) {
+    // NOTE: default 0 value for _id is replaced when key is auto generated
+    @JvmField
+    @PrimaryKey(autoGenerate = true)
+    @ColumnInfo(name = "_id")
+    var id: Long = 0L
```
tdchow, line 16: Is there a reason why the `id` property cannot live in the main constructor and be a `val`?

https://developer.android.com/training/data-storage/room/defining-data

tdchow:
Will we need to make this same change on the v5 branch?

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
sshropshire: We could yeah. The only reason I've done this in the past is to discourage use of the `@VisibleForTesting` constructor. But we could also enforce this by convention. I agree it's cleaner with default parameter values.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsDatabase.kt
sshropshire: 🤞 yeah once we moved from BT analytics to FPTI the requirements changed. Since there could be new analytics fields in the future, it makes sense to store the analytics request body as a string in the DB.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventBlob.kt
sshropshire: Yeah we can. I wish we could find a better name that feels more intuitive than "blob." I may replace this with a comment to call out that we're enforcing schema at the JSON level.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventBlob.kt
sshropshire: That's a good call I'll update this. I can't remember an exact reason why this is `var`.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEvent.kt
```diff
@@ -1,44 +1,13 @@
...
-    open val endpoint: String? = null,
-
-    open val timestamp: Long = System.currentTimeMillis()
-) {
-    @JvmField
-    @PrimaryKey(autoGenerate = true)
-    @ColumnInfo(name = "_id")
-    var id: Long = 0L
-}
+internal data class AnalyticsEvent constructor(
```
sarahkoop, line 3: _technically_ a breaking change, but since no merchant should be using this class I am ok with it - what do others think?

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventBlob.kt
sshropshire: I tried changing it to `val` and the build errors out. It may have to do with the version of Room we're using.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
```diff
@@ -64,38 +57,19 @@ internal class AnalyticsClient @VisibleForTesting constructor(
...
-        val venmoInstalled = inputData.getBoolean(WORK_INPUT_KEY_VENMO_INSTALLED, false)
-        val isVaultRequest = inputData.getBoolean(WORK_INPUT_KEY_IS_VAULT_REQUEST, false)
-        val startTime = inputData.getLong(WORK_INPUT_KEY_START_TIME, INVALID_TIMESTAMP)
-        val endTime = inputData.getLong(WORK_INPUT_KEY_END_TIME, INVALID_TIMESTAMP)
-        val endpoint = inputData.getString(WORK_INPUT_KEY_ENDPOINT)
-
-        return if (eventName == null || timestamp == INVALID_TIMESTAMP) {
+    fun performAnalyticsWrite(inputData: Data): ListenableWorker.Result {
+        val analyticsJSON = inputData.getString(WORK_INPUT_KEY_ANALYTICS_JSON)
+        return if (analyticsJSON == null) {
```
sarahkoop, line 62: We are losing the INVALID_TIMESTAMP check here - is that intentional?

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEventBlob.kt
sshropshire: Ok I found [this example](https://github.com/android/compose-samples/blob/c844e536deffe17799f73431271cfa2cddc7e0b0/Jetcaster/core/data/src/main/java/com/example/jetcaster/core/data/database/model/Category.kt#L24). I'll move the `id` declaration into the constructor 👍

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt
sshropshire: Yeah I think we needed `INVALID_TIMESTAMP` because of WorkManager's `Data` class. This PR increases our alignment with the [iOS implementation](https://github.com/braintree/braintree_ios/blob/d93bd064ad5b333b128d50bbf397ef45a42a8ab1/Sources/BraintreeCore/Analytics/FPTIBatchData.swift#L52).

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEvent.kt
sshropshire: Yeah it felt like the right call to break it. I'll add a `@RestrictTo` annotation to call out that this is intended for SDK use only. I doubt anyone's using / extending these types. If they are I think it's better for us to break those integrations to discourage its use.

sshropshire:
> Will we need to make this same change on the v5 branch?

@tdchow Yeah there are a few things that need to be migrated. For the most part though we should be able to cherry-pick some commits or checkout the individual files to ensure that all changes are captured.

BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsEvent.kt
tdchow: i'm also good with this breaking change. we shouldn't need the restrict to annotation since the entire class is marked as internal.

sshropshire: Oh true yeah. I'll remove.

sarahkoop:
@sshropshire were you able to validate that the data is making it to FPTI correctly formatted with this change?

sshropshire:
@sarahkoop I still need to do that. I can follow up.

sarahkoop (approved):

tdchow (approved):

PR #1056, opened 2024-07-05
Add returnUrlScheme to feature clients

tdchow:
### Summary of changes

 - Add `returnUrlScheme` param to Local Payment, SEPA, and Venmo clients

Note - I was unable to find a good way to unit test these changes without refactoring how dependency injection currently works. If anyone has other ideas on how to test the returnUrlScheme field, please let me know!

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentClient.java
+++ b/LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentClient.java
@@ -42,9 +42,14 @@ public class LocalPaymentClient {
      *
      * @param context an Android Context
      * @param authorization a Tokenization Key or Client Token used to authenticate
+     * @param returnUrlScheme a custom return url to use for browser and app switching
      */
-    public LocalPaymentClient(@NonNull Context context, @NonNull String authorization) {
-        this(new BraintreeClient(context, authorization));
+    public LocalPaymentClient(
+        @NonNull Context context,
+        @NonNull String authorization,
+        String returnUrlScheme
+    ) {
+        this(new BraintreeClient(context, authorization, returnUrlScheme));
     }
 
    @VisibleForTesting
--- a/LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentClient.java
+++ b/LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentClient.java
@@ -5,6 +5,7 @@ import android.content.Intent;
 import android.net.Uri;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
 
 import com.braintreepayments.api.BrowserSwitchFinalResult;
@@ -47,7 +48,7 @@ public class LocalPaymentClient {
     public LocalPaymentClient(
         @NonNull Context context,
         @NonNull String authorization,
-        String returnUrlScheme
+        @Nullable String returnUrlScheme
     ) {
         this(new BraintreeClient(context, authorization, returnUrlScheme));
     }
--- a/SEPADirectDebit/src/main/java/com/braintreepayments/api/sepadirectdebit/SEPADirectDebitClient.java
+++ b/SEPADirectDebit/src/main/java/com/braintreepayments/api/sepadirectdebit/SEPADirectDebitClient.java
@@ -7,6 +7,7 @@ import android.webkit.URLUtil;
 
 import androidx.activity.ComponentActivity;
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import androidx.annotation.VisibleForTesting;
 
 import com.braintreepayments.api.BrowserSwitchFinalResult;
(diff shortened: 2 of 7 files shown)


LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentClient.java
```diff
@@ -42,9 +42,14 @@ public class LocalPaymentClient {
...
      * @param context an Android Context
      * @param authorization a Tokenization Key or Client Token used to authenticate
+     * @param returnUrlScheme a custom return url to use for browser and app switching
      */
-    public LocalPaymentClient(@NonNull Context context, @NonNull String authorization) {
-        this(new BraintreeClient(context, authorization));
+    public LocalPaymentClient(
+        @NonNull Context context,
+        @NonNull String authorization,
+        String returnUrlScheme
```
sshropshire, line 50:
```suggestion
        @Nullable String returnUrlScheme
```

sshropshire (approved):

sshropshire:
> Note - I was unable to find a good way to unit test these changes without refactoring how dependency injection currently works. If anyone has other ideas on how to test the returnUrlScheme field, please let me know!

This is a great callout. In `v5` it seems like `BraintreeClient` may no longer be needed. Its counterpart in `v3` is [BraintreeFragment.java](https://github.com/braintree/braintree_android/blob/3.x/Braintree/src/main/java/com/braintreepayments/api/BraintreeFragment.java), but since we refactored `v4` for increased modularity we shouldn't need this class. Also PPCP Android has no concept of a core client.

I'd personally like to remove BraintreeClient–doing so would resolve the DI ambiguity we see here.

KunJeongPark:
Why do we need this? Is this an android specific feature? 
For app switching in iOS, I think we hard code the base url components

SEPADirectDebit/src/main/java/com/braintreepayments/api/sepadirectdebit/SEPADirectDebitClient.java
```diff
@@ -36,9 +36,14 @@ public class SEPADirectDebitClient {
...
      * @param context an Android Context
      * @param authorization a Tokenization Key or Client Token used to authenticate
+     * @param returnUrlScheme a custom return url to use for browser and app switching
      */
-    public SEPADirectDebitClient(@NonNull Context context, @NonNull String authorization) {
-       this(new BraintreeClient(context, authorization));
+    public SEPADirectDebitClient(
+        @NonNull Context context,
+        @NonNull String authorization,
+        String returnUrlScheme
```
jaxdesmarais, line 44: If we use Stevens suggestion above we should update to the same here

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.java
```diff
@@ -60,9 +60,14 @@ private VenmoClient(BraintreeClient braintreeClient, ApiClient apiClient) {
...
      * @param context an Android Context
      * @param authorization a Tokenization Key or Client Token used to authenticate
+     * @param returnUrlScheme a custom return url to use for browser and app switching
      */
-    public VenmoClient(@NonNull Context context, @NonNull String authorization) {
-        this(new BraintreeClient(context, authorization));
+    public VenmoClient(
+        @NonNull Context context,
+        @NonNull String authorization,
+        String returnUrlScheme
```
jaxdesmarais, line 68: And here as well

jaxdesmarais (approved):

tdchow:
> > Note - I was unable to find a good way to unit test these changes without refactoring how dependency injection currently works. If anyone has other ideas on how to test the returnUrlScheme field, please let me know!
> 
> This is a great callout. In `v5` it seems like `BraintreeClient` may no longer be needed. Its counterpart in `v3` is [BraintreeFragment.java](https://github.com/braintree/braintree_android/blob/3.x/Braintree/src/main/java/com/braintreepayments/api/BraintreeFragment.java), but since we refactored `v4` for increased modularity we shouldn't need this class. Also PPCP Android has no concept of a core client.
> 
> I'd personally like to remove BraintreeClient–doing so would resolve the DI ambiguity we see here.

I agree! I feel like we can refactor out BraintreeClient from the feature clients. Luckily it won't be a breaking change and we can hopefully tackle this once v5 is released.

tdchow:
> Why do we need this? Is this an android specific feature? For app switching in iOS, I think we hard code the base url components

This is a specific change for v5. When merging main into the v5 branch the `returnUrlScheme` param was set on the `BraintreeClient`, which is no longer implemented by the merchant.

sshropshire:
> Why do we need this? Is this an android specific feature? For app switching in iOS, I think we hard code the base url components

Yes it's Android specific. It allows merchants to explicitly declare a deep link scheme.

One common use case for a merchant providing their own custom url scheme is when their application ID contains underscores. The default `AndroidManifest.xml` [placeholder](https://github.com/braintree/braintree_android/blob/5a187beb148cb72645f5f62d3848f1371525fbf1/Demo/src/main/AndroidManifest.xml#L28) we recommend would produce an invalid scheme in this case, since URL schemes cannot contain [underscores](https://stackoverflow.com/a/9142331).

PR #1057, opened 2024-07-09
Add Error when Tokenization Key is Used for `ShopperInsights`

jaxdesmarais:
### Summary of changes

 - Add error when tokenization key is used for `ShopperInsights`
 - iOS PR with this change: https://github.com/braintree/braintree_ios/pull/1350

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors

@jaxdesmarais


ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -43,6 +42,14 @@ class ShopperInsightsClient @VisibleForTesting internal constructor(
     ) {
         braintreeClient.sendAnalyticsEvent(GET_RECOMMENDED_PAYMENTS_STARTED)
 
+        if (isTokenizationKey()) {
+            callbackFailure(
+                callback = callback,
+                error = BraintreeException("Invalid authorization. This feature can only be used with a client token.")
```
KunJeongPark, line 48: Do we have in our documentation that tokenization token is not allowed with ShopperInsights module?

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -102,6 +109,19 @@ class ShopperInsightsClient @VisibleForTesting internal constructor(
         }
     }
 
+    private fun isTokenizationKey(): Boolean {
+        var isTokenizationKey = false
+        braintreeClient.getAuthorization { authorization, _ ->
```
KunJeongPark, line 114: Not sure about Kotlin, but it looks like getAuthorization is in like completion handler and the boolean is being returned right away. Is this okay?

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
```diff
@@ -102,6 +109,19 @@ class ShopperInsightsClient @VisibleForTesting internal constructor(
         }
     }
 
+    private fun isTokenizationKey(): Boolean {
+        var isTokenizationKey = false
```
sarahkoop, line 113: Would it make more sense for this to return a `Boolean?` and set it to `null` initially? Edge case, but in the event that there is an error returning auth, we are setting `isTokenizationKey` to `false`, which would then allow the shopper insights API request to proceed without error.

KunJeongPark: I think there is a problem to proceed with tokenizationkey and we want to send error if it is.

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
jaxdesmarais: Yep, in the `note` section here: https://developer.paypal.com/braintree/docs/guides/shopper-insights/ios/v6#integration

ShopperInsights/src/main/java/com/braintreepayments/api/ShopperInsightsClient.kt
jaxdesmarais: Yes, it can be tested in the demo app to see the functionality by switching between client token and tokenization key

tdchow: Yep, I think we might have a bug here where the function returns before the lambda executes and sets `isTokenizationKey` to true.

One solution is to move the code below the tokenization key check in `getRecommendedPaymentMethods` to inside the lambda to ensure execution.

jaxdesmarais: Yeah I have a change locally for it but it's requiring some more major test changes because we use `lateinit` all over in our client tests currently. I may bug you a bit about it later if I can't get it sorted but have the check moved into the callback locally.

sshropshire: Yeah won't need to do this in v5. It's only weird here because of the `AuthorizationLoader` pattern.

sshropshire: We can implement a workaround though if we want to make a synchronous `BraintreeClient.hasTokenizationKeyAuth` property or something that's one possibility. Otherwise like Tim mentioned it could be done within the `getRecommendedPaymentMethods()` call.

jaxdesmarais: Updated here: https://github.com/braintree/braintree_android/pull/1057/commits/0e48fea6bfd932c504f0db759e6000d475dbae3f and https://github.com/braintree/braintree_android/pull/1057/commits/ba4553005e3d42e0ceb22b80057d3c7b1c4974eb. Thanks to @tdchow as always for being willing to answer my questions about how to do things in Kotlin with reguards to tests.

KunJeongPark (approved):

sshropshire (approved):

PR #1058, opened 2024-07-09
Remove unneeded open modifier

tdchow:
### Summary of changes

 - Remove unneeded open modifier on various classes and properties
 - Fix a small typo in property name

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/UserCanceledException.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/UserCanceledException.kt
@@ -5,7 +5,7 @@ import androidx.annotation.RestrictTo
 /**
  * Error class thrown when a user cancels a payment flow
  *
- * @property isExplicitCancelation whether or not the user explicitly canceled the payment flow.
+ * @property isExplicitCancellation whether or not the user explicitly canceled the payment flow.
  *
  * This value will be true if the user manually confirms cancellation of the payment flow.
  *
@@ -14,7 +14,7 @@ import androidx.annotation.RestrictTo
  * could mean the user returned to the app through multi-tasking without completing the flow,
  * the user closed the browser tab, or the user pressed the back button.
  */
-open class UserCanceledException @JvmOverloads @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP) constructor(
+class UserCanceledException @JvmOverloads @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP) constructor(
     message: String?,
-    val isExplicitCancelation: Boolean = false
+    val isExplicitCancellation: Boolean = false
 ) : BraintreeException(message)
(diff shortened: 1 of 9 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/core/UserCanceledException.kt
```diff
@@ -14,7 +14,7 @@ import androidx.annotation.RestrictTo
  * could mean the user returned to the app through multi-tasking without completing the flow,
  * the user closed the browser tab, or the user pressed the back button.
  */
-open class UserCanceledException @JvmOverloads @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP) constructor(
+class UserCanceledException @JvmOverloads @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP) constructor(
     message: String?,
-    val isExplicitCancelation: Boolean = false
+    val isExplicitCancellation: Boolean = false
```
sshropshire, line 19:
```suggestion
    val isExplicitCancellation: Boolean = false
```
We actually use the [single 'L' variant](https://www.merriam-webster.com/grammar/canceled-or-cancelled) because 🇺🇸 .

jaxdesmarais, line 19: The single `l` is intentional so we will want to revert this

sshropshire: Oh but wait with "cancellation" it's double 'L' in [both languages](https://www.scribbr.com/us-vs-uk/cancelled-or-canceled/#cancelation)? 🤯

sshropshire (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/UserCanceledException.kt
sarahkoop, line 19: Maybe a separate PR but do we even need this property anymore? I think we only return `UserCanceledException` for explicit cancellations in v5 right?

sshropshire: Ah good callout in v5 we `:kill_it_with_fire:` this is the best option lol.

tdchow: the Android Studio spell check actually flagged it as a typo which is why I made the change. I was also seeing recommendations to avoid using the single L even in the US 🤔 

in any case, I think we can remove this property altogether!

tdchow: Similar thread here: https://github.com/braintree/braintree_android/pull/1058#discussion_r1671204729

jaxdesmarais: Ah, sorry Steven and I apparently posted the same thing at the same time!

tdchow: I'll remove this property in a follow up PR 👍

tdchow: Reverted for this PR!

jaxdesmarais (approved):

PR #1059, opened 2024-07-09
Fix PayPal Pay Later Parsing Error

sshropshire:
### Summary of changes

 - Fix issue that causes a JSON parsing error when Pay Later is selected during checkout.

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

@sshropshire

### DIFF
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalCreditFinancingAmount.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalCreditFinancingAmount.java
@@ -22,8 +22,8 @@ public class PayPalCreditFinancingAmount implements Parcelable {
     static PayPalCreditFinancingAmount fromJson(JSONObject amount) {
         PayPalCreditFinancingAmount result = new PayPalCreditFinancingAmount();
 
-        result.currency = Json.optString(amount, CURRENCY_KEY, "");
-        result.value = Json.optString(amount, VALUE_KEY, "");
+        result.currency = Json.optString(amount, CURRENCY_KEY, null);
+        result.value = Json.optString(amount, VALUE_KEY, null);
 
         return result;
     }
--- a/PayPal/src/main/java/com/braintreepayments/api/PayPalCreditFinancingAmount.java
+++ b/PayPal/src/main/java/com/braintreepayments/api/PayPalCreditFinancingAmount.java
@@ -22,6 +22,10 @@ public class PayPalCreditFinancingAmount implements Parcelable {
     static PayPalCreditFinancingAmount fromJson(JSONObject amount) {
         PayPalCreditFinancingAmount result = new PayPalCreditFinancingAmount();
 
+        if (amount == null) {
+            return result;
+        }
+
         result.currency = Json.optString(amount, CURRENCY_KEY, null);
         result.value = Json.optString(amount, VALUE_KEY, null);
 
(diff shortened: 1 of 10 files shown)


PayPal/src/main/java/com/braintreepayments/api/PayPalCreditFinancingAmount.java
```diff
@@ -22,12 +22,8 @@ private PayPalCreditFinancingAmount() {}
     static PayPalCreditFinancingAmount fromJson(JSONObject amount) {
         PayPalCreditFinancingAmount result = new PayPalCreditFinancingAmount();
 
-        if (amount == null) {
-            return result;
-        }
-
-        result.currency = Json.optString(amount, CURRENCY_KEY, null);
-        result.value = Json.optString(amount, VALUE_KEY, null);
```
sarahkoop, line 30: Are these changes needed (and in native checkout too)? It looks like this logic would already tolerate the fields not existing

sshropshire: The change we made falls back to empty strings when possible. I'm not sure if it's needed or not for MXO, but the change was easy enough to prevent a merchant filing an inbound and having to tell the merchant it's deprecated etc.

sarahkoop: My concern with this is that now when the amount is null, the `PayPalCreditFinancingAmount` object returned will have the `currency` and `value` properties set to `""` rather than not set. That seems like unintended behavior

sshropshire: Ok I see what you mean. I'll refactor the tests to expect `null` for those values.

sshropshire: Made the update here: [8c95edc](https://github.com/braintree/braintree_android/pull/1059/commits/8c95edce819f3187c248d375bc0a9c41322e5013)

PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutCreditFinancingAmount.java
```diff
@@ -25,10 +25,6 @@ private PayPalNativeCheckoutCreditFinancingAmount() {}
     static PayPalNativeCheckoutCreditFinancingAmount fromJson(JSONObject amount) {
         PayPalNativeCheckoutCreditFinancingAmount result = new PayPalNativeCheckoutCreditFinancingAmount();
 
-        if (amount == null) {
-            return result;
-        }
-
```
sarahkoop, line 31: Let's revert this too

sshropshire: Yeah we do own `Json.optString` though and it performs this check internally. I'm not sure why this guard exists, but I'm down to keep.

sshropshire: Done: [bb94c54](https://github.com/braintree/braintree_android/pull/1059/commits/bb94c54728c389230f80a07d57728e4265b1aebb)

sarahkoop (approved):

jaxdesmarais (approved):

PR #1067, opened 2024-07-15
Jetpack compose v5

tdchow:
### Summary of changes

 - Convert MainFragment to Kotlin
 - Convert MainFragment UI to Jetpack Compose

Screenshot_20240715_092434

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


Demo/src/main/java/com/braintreepayments/demo/MainFragment.kt
```diff
@@ -0,0 +1,192 @@
...
+                    }
+
+                    Card(
+                        modifier = Modifier.padding(12.dp),
+                    ) {
+                        Text(
+                            modifier = Modifier
+                                .align(Alignment.CenterHorizontally)
+                                .padding(top = 8.dp),
+                            text = "Other",
```
tdchow, line 66: I tried to group the PayPal modules since "Shopper Insights" is clearly defined as a PayPal module.

I put everything else in the same "Other" category. Is there a better way to group the remaining modules?

jaxdesmarais: I think either makes sense. Or if we don't find the groupings helpful we can remove then and just list PayPal first? I don't feel strongly either way.

PR #1076, opened 2024-07-22
Fix Publish Release Signing Issue

tdchow:
### Summary of changes

 - Add signing configs for each module to fix maven publish
 
Successful test publish: https://github.com/braintree/braintree_android/actions/runs/10043185768/job/27755633475

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


tdchow:
The changes are more verbose but I feel it makes more sense to have each module define it's signing config vs having it defined in the demo app module which doesn't need to be signed.

AmericanExpress/build.gradle
```diff
@@ -15,6 +15,15 @@ android {
         testInstrumentationRunner 'androidx.test.runner.AndroidJUnitRunner'
     }
 
+    signingConfigs {
```
jaxdesmarais, line 18: Not sure if there is a way to do this, but is there a way we can define this like how we have `compileSdkVersion` defined in the top level `build.gradle` that then every other `build.gradle` references? Just thinking about forgetting 1 of these files if we ever need to make an update and would be nice to make the change in 1 place that everything else references.

PayPalMessaging/build.gradle
```diff
@@ -27,12 +36,6 @@ android {
         }
     }
 
-    buildTypes {
-        release {
-            minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
-        }
-    }
```
sarahkoop, line 35: Was this causing problems too?

AmericanExpress/build.gradle
tdchow: That's a good call. I believe we're only defining simple properties to share across our gradle files. So defining the signing configs block would be different. I'm no groovy expert, but let me see if I can find the syntax for it.

PayPalMessaging/build.gradle
tdchow: It wasn't causing problems, but I don't think it's needed. None of the other modules have this defined and we don't have a proguard rules file for this module.

sshropshire (approved):
🌊

AmericanExpress/build.gradle
tdchow: Adding the `android { ... }` blcok to the gradle-publish.gradle file ended up working!

sarahkoop (approved):

jaxdesmarais (approved):

PR #1077, opened 2024-07-22
Add Java Setup to all release jobs

tdchow:
### Summary of changes

 - Add Setup Java steps to release workflow
 - Add Temp Release Workflow for running `bump_version` and `create_github_release` jobs for `5.0.0-beta1` release. This workflow will be removed once it runs.

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/.github/workflows/temp_release.yml
+++ b/.github/workflows/temp_release.yml
@@ -1,0 +1,59 @@
+name: Temp Release Workflow
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to release'
+        required: true
+env:
+  SIGNING_KEY_FILE_PATH: /home/runner/secretKey.gpg
+jobs:
+  bump_version:
+    name: Bump Version
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+      - name: Setup Java
+        uses: ./.github/actions/setup
+      - name: Set GitHub User
+        uses: ./.github/actions/set_github_user
+      - name: Update Version
+        run: |
+          ./gradlew -PversionParam=${{ github.event.inputs.version }} changeGradleReleaseVersion
+          ./gradlew -PversionParam=${{ github.event.inputs.version }} changeREADMEVersion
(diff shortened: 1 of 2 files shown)


.github/workflows/temp_release.yml
```diff
@@ -0,0 +1,59 @@
+name: Temp Release Workflow
```
jaxdesmarais, line 1: We can also run these steps manually vs creating a removing a new workflow. That's what we've typically done previously on iOS at least unless some of these steps cannot be done manually.

tdchow: We don't currently have that option with how the workflow is set up. Right now each job requires the previous one: https://github.com/braintree/braintree_android/blob/e5500e0f270b7fd14e66eee19ec461c262ad1cef/.github/workflows/release.yml#L249

I feel that there's a benefit in restricting the running of individual jobs and enforcing the entire workflow to be run for a release. I worry that it'd open up the possibility of publishing something to maven central by accident.

jaxdesmarais: iOS does not rely on the same settings, really the [iOS releases](https://github.com/braintree/braintree_ios/blob/main/.github/workflows/release.yml) as a whole are pretty different than what we do on android (for example we don't run all the tests again). It is nice in that say cocoapods or publishing reference docs fails we typically just run those 1-2 steps manually. We do rely on the order of things but can pick up where the release failed if something goes sideways.

sarahkoop (approved):

.github/workflows/temp_release.yml
tdchow: Gotcha. I think the Android release workflow has been stable up until v5, so hopefully we won't have any of these CI issues until the next major version where our gradle version and build tools are bumped.

Are you okay with merging in the temp workflow, then following up with another PR to remove it?

scannillo (approved):
LGTM! Maybe leaving a comment in the new temp workflow to eventually remove will help us remember

jaxdesmarais (approved):

tdchow:
> LGTM! Maybe leaving a comment in the new temp workflow to eventually remove will help us remember

I'll plan to open another PR for removing the workflow once it runs successfully. So it shouldn't exist for very long!

PR #1079, opened 2024-07-23
Add Recurring Billing Agreement to PayPalVaultRequest

sarahkoop:
### Summary of changes

 - This initiative allows merchants to provide more transparency to their users regarding the details of their Recurring Billing Agreement (RBA) including billing schedule, dates, amounts, free trail details, etc. These details will be displayed in a redesigned XO experience with a streamlined, modernized UI.
 - Add new public types to collect recurring billing details, to set on `PayPalVaultRequest`
 - Format new params into appropriate JSON POST format for BT GW (untested until feature is available)

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors
@sarahkoop


PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalVaultRequestUnitTest.java
```diff
@@ -1,26 +1,26 @@
...
-import static junit.framework.Assert.assertEquals;
-import static junit.framework.Assert.assertFalse;
-import static junit.framework.Assert.assertNull;
-import static junit.framework.Assert.assertTrue;
-import static org.mockito.Mockito.mock;
+import java.util.List;
+import java.util.Objects;
 
 @RunWith(RobolectricTestRunner.class)
 public class PayPalVaultRequestUnitTest {
```
sarahkoop, line 26: We don't currently have robust testing of our JSON parameter formatting - we rely on demo app to test that the request is properly formatted. Is there a good way to test the JSON body of these requests or should we wait until GW sandbox verification via the demo app (future ticket)?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingCycle.kt
```diff
@@ -0,0 +1,80 @@
...
+ * If provided and the merchant wants the billing cycle to start at the time of checkout, provide
+ * the current time. Otherwise the [startDate] can be in future.
+ * @property isTrial The tenure type of the billing cycle. In case of a plan having trial cycle,
+ * only 2 trial cycles are allowed per plan.
+ * @property pricing The active pricing scheme for this billing cycle. Required if [isTrial] is
+ * false. Optional if [isTrial] is true.
+ */
+@Parcelize
+data class PayPalBillingCycle(
+    var interval: PayPalBillingInterval,
```
tdchow, line 35: Do we need to keep these properties as `var`s for java setter support?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingCycle.kt
```diff
@@ -0,0 +1,80 @@
...
+) : Parcelable {
+    constructor(
+        interval: PayPalBillingInterval,
+        intervalCount: Int,
+        numberOfExecutions: Int
+    ) : this(
+        interval = interval,
+        intervalCount = intervalCount,
+        numberOfExecutions = numberOfExecutions,
+        null,
```
tdchow, line 51: Could we combine this constructor with default values with the main constructor? We could use the `@JvmOverloads` annotation to generate the java constructors: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-overloads/

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingCycle.kt
```diff
@@ -0,0 +1,80 @@
...
+        null,
+        false,
+        null
+    )
+
+    companion object {
+        fun PayPalBillingCycle.toJson(): String {
```
tdchow, line 58: Is there a benefit to having the `toJson()` defined as an extension function? Could we have it as a regular function within the class?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingPricing.kt
```diff
@@ -0,0 +1,39 @@
...
+data class PayPalBillingPricing(
+    var pricingModel: PayPalPricingModel,
+    var amount: String,
+    var reloadThresholdAmount: String?
+) : Parcelable {
+
+    constructor(
+        pricingModel: PayPalPricingModel,
+        amount: String
+    ) : this(pricingModel = pricingModel, amount = amount, reloadThresholdAmount = null)
```
tdchow, line 24: Same constructor question here. Could we combine this with the main constructor?

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalVaultRequestUnitTest.java
tdchow: I think we could write a test around calling `toJson()` on a given `PayPalVaultRequest` and verifying the JSON string that's returned. Is there additional testing that we'd want to do?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingCycle.kt
sarahkoop: Ah yes - fixed!

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingCycle.kt
sarahkoop: The optional values need to be `var` for Java setter support, but I updated the required params to `val`. Open to re-considering if we want to continue providing Java setter support in v5 though, but that's likely a separate conversation

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingPricing.kt
sarahkoop: Done

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingCycle.kt
sarahkoop: Hmm no I don't think so - this was just modeled after how we set up the data classes in the `ShopperInsights` module. Updated to define the methods as regular class functions 👍

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalVaultRequestUnitTest.java
```diff
@@ -52,6 +52,27 @@ public void setsValuesCorrectly() {
         request.setLandingPageType(PayPalRequest.LANDING_PAGE_TYPE_LOGIN);
         request.setShouldOfferCredit(true);
         request.setAppLinkEnabled(true);
+        PayPalBillingInterval billingInterval = PayPalBillingInterval.MONTH;
```
tdchow, line 55: An aside: what are people's thoughts on not having to write unit tests for Kotlin data class properties?

I feel that there isn't any real logic that we're testing unless properties have custom setters/getters.

tdchow (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* PayPal
+  * Add `PayPalRecurringBillingDetails` and `ayPalRecurringBillingPlanType` opt-in request objects. Including these details will provide transparency to users on their billing schedule, dates, and amounts, as well as launch a modernized checkout UI.
```
scannillo, line 6:
```suggestion
  * Add `PayPalRecurringBillingDetails` and `PayPalRecurringBillingPlanType` opt-in request objects. Including these details will provide transparency to users on their billing schedule, dates, and amounts, as well as launch a modernized checkout UI.
```

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingPricing.kt
```diff
@@ -0,0 +1,36 @@
...
+    fun toJson(): String {
+        return JSONObject().apply {
+            put(KEY_PRICING_MODEL, pricingModel.name)
+            put(KEY_AMOUNT, amount)
+            putOpt(KEY_RELOAD_THRESHOLD_AMOUNT, reloadThresholdAmount)
+        }.toString()
+    }
+
+    companion object {
+
```
scannillo, line 30:
```suggestion
```

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalVaultRequestUnitTest.java
scannillo: I agree - there is no value to testing that properties were set. Like you suggested earlier, we should assert that the JSON is properly formatted from a given Data class set of inputs.

sarahkoop: I think the only value currently is confirming Java integration doesn't change (i.e. sanity check that the setter pattern is working as expected).

Update to include a JSON formatting unit test too 👍

scannillo (approved):

PR #1080, opened 2024-07-24
Update 3DS and GraphQL Endpoint Sent to FPTI (v5)

jaxdesmarais:
### Summary of changes

- 3D Secure: remove token/unique ID from path
- GraphQL: remove reliance on `operationName` since this is only present for card, causing us not to send the correct path for the Venmo flow
- Both endpoints sending as expected have been verified in FPTI and via breakpoints, additionally all endpoint have been checked to ensure we are sending the expected paths now
- iOS PR: https://github.com/braintree/braintree_ios/pull/1375

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
@@ -357,21 +357,12 @@ class BraintreeClient @VisibleForTesting internal constructor(
     }
 
     private fun sendAnalyticsTimingEvent(endpoint: String, timing: HttpResponseTiming) {
-        val cleanedPath = endpoint.replace(Regex("/merchants/([A-Za-z0-9]+)/client_api"), "")
+        var cleanedPath = endpoint.replace(Regex("/merchants/([A-Za-z0-9]+)/client_api"), "")
+        cleanedPath = cleanedPath.replace(
+            Regex("payment_methods/.*/three_d_secure"), "payment_methods/three_d_secure"
+        )
 
-        if (cleanedPath.contains("three_d_secure")) {
-            val threeDSecurePath = cleanedPath.replace(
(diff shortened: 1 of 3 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
```diff
@@ -355,15 +357,21 @@ class BraintreeClient @VisibleForTesting internal constructor(
...
-                startTime = timing.startTime,
-                endTime = timing.endTime,
-                endpoint = cleanedPath
-            )
+        var cleanedPath = endpoint.replace(Regex("/merchants/([A-Za-z0-9]+)/client_api"), "")
+        cleanedPath = cleanedPath.replace(
+            Regex("payment_methods/.*/three_d_secure"), "payment_methods/three_d_secure"
         )
+
+        if (cleanedPath != "/v1/tracking/batch/events") {
```
sarahkoop, line 365: What is the context on this part of the change? What is that path used for that isn't sent to analytics?

jaxdesmarais: On iOS it was added to ensure that we don't send analytics events basically saying that we sent analytics since that's the FPTI path ([here](https://github.com/braintree/braintree_ios/blob/7b81129478fb3302b2ffacdbb4ade80e561faef7/Sources/BraintreeCore/BTAPIClient.swift#L406)). Though adding breakpoints on Android it seems like we use a totally different post method just for analytics so this is likely not needed currently for Android.

jaxdesmarais: Updated: https://github.com/braintree/braintree_android/pull/1080/commits/a72b21a6a679cb8b204bcbb3ff3c65cff2ae25e8

sarahkoop (approved):

agedd (approved):

PR #1087, opened 2024-07-29
Parcelize Nonce Classes

tdchow:
### Summary of changes

 - Convert Nonce classes to Kotlin
 - Implement `Parcelable` using the `@Parcelize` annotation

Note that the `fromJSON()` functions in each nonce class was converted directly from Java. There's some clean up we can do, but to keep the scope down on this PR, I left it as is.

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentNonce.kt
```diff
@@ -0,0 +1,121 @@
...
+    override val isDefault: Boolean,
+    val clientMetadataId: String?,
+    val billingAddress: PostalAddress,
+    val shippingAddress: PostalAddress,
+    val givenName: String,
+    val surname: String,
+    val phone: String,
+    val email: String?,
+    val payerId: String,
+    private val type: String?
```
tdchow, line 37: This value was not used or exposed. Should we remove it?

sshropshire: Yeah it doesn't seem like it's needed. I think it's a good idea to remove.

detekt/detekt-config.yml
```diff
@@ -119,6 +119,7 @@ complexity:
     ignoreDefaultParameters: true
     ignoreDataClasses: true
     ignoreAnnotated: []
+    excludes: ['**/*Nonce.kt']
```
sshropshire, line 122: curious: Does converting the nonces to data classes prevent lint errors?

sshropshire (approved):

detekt/detekt-config.yml
tdchow: great catch! yep, converting them to data classes removes the detekt error. although, `CardNonce` needs the suppress annotation since it's an open class and can't also be defined as a data class. i just pushed an update for this!

sshropshire: Nice! We may be able to convert CardNonce one day to a data class. The `open` modifiers can be removed also, the only reason we added them in the past is to prevent breaking changes since Java classes are non-final by default.

scannillo (approved):

PR #1088, opened 2024-07-30
Paypal Module Kotlin Conversion

tdchow:
### Summary of changes

- Convert various classes in the PayPal module from Java to Kotlin
- There will be follow up PRs converting the remaining classes

### Checklist

 - [x] Added a changelog entry
 - [ ] Relevant test coverage


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCreditFinancing.kt
```diff
@@ -0,0 +1,67 @@
...
+ * during the lifetime of the loan.
+ * @property totalInterest Estimated interest or fees amount the payer will have to pay during the
+ * lifetime of the loan.
+ * @property term Length of financing terms in months.
+ */
+@Parcelize
+data class PayPalCreditFinancing internal constructor(
+    val isCardAmountImmutable: Boolean = false,
+    val monthlyPayment: PayPalCreditFinancingAmount? = null,
+    val payerAcceptance: Boolean = false,
```
sshropshire, line 27:
```suggestion
    val hasPayerAcceptance: Boolean = false,
```
🧶 : Any opposition to keeping the `has-` prefix? I really like this portion of the [Swift naming guidelines](https://www.swift.org/documentation/api-design-guidelines/#boolean-assertions):

> Uses of Boolean methods and properties should read as assertions about the receiver when the use is nonmutating, e.g. x.isEmpty, line1.intersects(line2).

I understand it might be extra work if the compiler generates `getHasPayerAcceptance()` though. In that case, we'd need to override the generated property accessor with a custom `get set` declaration.

sshropshire (approved):

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCreditFinancing.kt
tdchow: Good callout. I tried to find a solution that would scale for all of the booleans we have in our request and response objects.

```kotlin
class Test(
    @JvmField // removes the generated java getter function
    val hasPayerAcceptance: Boolean = false,
) {
    
    fun hasPayerAcceptance() = hasPayerAcceptance
}
```

One downside to this is that in Kotlin there will be 2 ways to get the boolean value:
```kotlin
val test = Test(true)
test.hasPayerAcceptance()
test.hasPayerAcceptance
```

This pattern also will not work for abstract classes containing booleans, as the `@JvmStatic` annotation requires the property to be final.

Would we be okay with the generated `getHasPayerAcceptance()` for java integrations? or would we want to keep the naming as `payerAcceptance`?

I'm leaning towards just exposing `getHasPayerAcceptance()` given the adoption of Kotlin and that devs would still see the correct function in IDE autocomplete.

jaxdesmarais (commented):
Not sure if this is a me issue, but I am running into errors with both the checkout and vault flows:
- Checkout: after checkout getting an error that it cannot get the path to return to merchant in CCT
- Vault: failing immediately with "Parsing error response failed"

Not sure if you're running into the same in the demo app or this is just a me problem?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalPaymentIntent.kt
```diff
@@ -0,0 +1,29 @@
+package com.braintreepayments.api.paypal
+
+/**
+ * The payment intent in the PayPal Checkout flow
+ */
+enum class PayPalPaymentIntent(val stringValue: String) {
```
jaxdesmarais, line 6: Do we need to call this out in the changelog?

tdchow: Good catch! let me add this.

tdchow:
> Not sure if this is a me issue, but I am running into errors with both the checkout and vault flows:
> 
> * Checkout: after checkout getting an error that it cannot get the path to return to merchant in CCT
> * Vault: failing immediately with "Parsing error response failed"
> 
> Not sure if you're running into the same in the demo app or this is just a me problem?

Vault seems to be working for me. But I'm seeing the "Parsing error response failed" for checkout. I switched over to the main branch and I'm seeing the same errors 😟 I'll look into this.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCreditFinancing.kt
sshropshire: I see yeah I figured maintainability would be a concern. I'd lean towards `getHasPayerAcceptance()` as the past of least resistance. Another option could be to refactor the property to something like `isPayerAccepted`, but i'm being super nit picky. I trust your judgement this is a take-or-leave for me.

tdchow:
@jaxdesmarais After some testing, I've found that the parsing error is not related to the kotlin conversion changes. After changing intent type from order to sale, I was able to successfully complete the checkout flow.

jaxdesmarais (approved):

PR #1089, opened 2024-08-01
Convert PayPal Classes to Kotlin

tdchow:
### Summary of changes

 - Convert PayPalAccount, PayPalCheckoutRequest, PayPalRequest, and PayPalVaultRequest to Kotlin

### Checklist

 - [X] Added a changelog entry
 - [ ] Relevant test coverage


sshropshire (approved):

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalAccount.kt
```diff
@@ -0,0 +1,65 @@
...
+ * request to create a PayPal account. Response data will be merged into the payment method json on
+ * [buildJSON].
+ * @property intent Used by PayPal wrappers to construct a request to create a PayPal account.
+ * @property merchantAccountId String merchant account id
+ * @property paymentType Payment type from original PayPal request. Either "billing-agreement" or
+ * "single-payment"
+ */
+internal data class PayPalAccount(
+    val clientMetadataId: String?,
+    val urlResponseData: JSONObject,
```
jaxdesmarais, line 23: General style question, do we want to consider having required properties at the top followed by optional?

jaxdesmarais (approved):

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalRequest.kt
```diff
@@ -0,0 +1,135 @@
...
+    open var shippingAddressOverride: PostalAddress? = null,
+    open var landingPageType: PayPalLandingPageType? = null,
+    open var displayName: String? = null,
+    open var merchantAccountId: String? = null,
+    open var riskCorrelationId: String? = null,
+    open var userAuthenticationEmail: String? = null,
+    open var lineItems: List<PayPalLineItem> = emptyList()
+) : Parcelable {
+
+    enum class PayPalLandingPageType(val stringValue: String) {
```
sarahkoop, line 89: Do we want to declare a preference over enums living in their own classes or not? For the RBA metadata work, I started putting all the enums in their own files - but we can convert to this style if it's preferred

sarahkoop:
Note to also update `rba-metadata-feature` branch when this is merged

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalAccount.kt
tdchow: Yep! I think having all required fields at the top makes sense. In this case, all of these fields are required as there are no default values defined. Although most of them are marked as nullable.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalRequest.kt
tdchow: Good callout. I was thinking the same. I'll move them out to their own classes/files.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalAccount.kt
tdchow: This is actually an internal class, so we'd be the ones to invoke the constructor 😄 Going to merge the PR as is for now but happy to keep the convo for this open!

PR #1091, opened 2024-08-02
Convert PayPal module classes to Kotlin

tdchow:
### Summary of changes

 - Convert PayPalClient, PayPalInternalClient, and PayPalLineItem to Kotlin.
 - This completes the Kotlin conversion for all of the PayPal module classes.

### Checklist

 - [X] Added a changelog entry
 - [ ] Relevant test coverage


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalLineItemKind.kt
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * The type of PayPal line item.
+ *
+ * [KIND_CREDIT] A line item that is a credit.
+ * [KIND_DEBIT] A line item that debits.
+ */
+enum class PayPalLineItemKind(val stringValue: String) {
+    KIND_CREDIT("credit"),
+    KIND_DEBIT("debit")
+}
```
jaxdesmarais, line 12: Is duplicating the word "kind" generally a best practice in Kotlin/Java or could we drop it from either the class name or enum name?

jaxdesmarais (approved):

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -0,0 +1,352 @@
...
+            sendVaultRequest(context, payPalRequest, callback)
+        }
+    }
+
+    private fun sendCheckoutRequest(
+        context: Context,
+        payPalCheckoutRequest: PayPalCheckoutRequest,
+        callback: PayPalPaymentAuthCallback
+    ) {
+        braintreeClient.sendAnalyticsEvent("paypal.single-payment.selected", analyticsParams)
```
sarahkoop, line 83: These might have accidentally made it into v5 - I think all analytics events should be in the `PayPalAnalytics` enum now. Should these be removed?

tdchow: Good catch! Mind if I make this change in a follow-up PR? I feel we should reduce the amount of changes in the Kotlin conversion PR.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalLineItem.kt
```diff
@@ -0,0 +1,76 @@
...
+ * @property url The URL to product information.
+ */
+
+@Parcelize
+data class PayPalLineItem
+@JvmOverloads constructor(
+    val kind: PayPalLineItemKind,
+    val name: String,
+    val quantity: String,
+    val unitAmount: String?,
```
sarahkoop, line 34:
```suggestion
    val unitAmount: String,
```

I think this value is required

sarahkoop (approved):

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalLineItem.kt
tdchow: Whoops, I removed the default value but kept the type as nullable. Just updated this.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
tdchow: Opened https://github.com/braintree/braintree_android/pull/1093

PR #1092, opened 2024-08-02
[App Switch] Construct Uri to launch PayPal App

richherrera:
### Summary of changes

[DTMOBILES-898]
 - Update the response model by adding the new parameter, which contains the Uri to launch the PayPal app.
 - Build the redirect URI with the necessary query items (`source` and `switch_initiated_time`).
 - Validate if the token exists, otherwise, an exception is thrown.
 
**Note:** The flow will be tested once we have a build of the app. [DTMOBILES-951]

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### Authors

- @richherrera

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
@@ -389,4 +389,8 @@ class BraintreeClient @VisibleForTesting internal constructor(
         val message = "Valid authorization required. See $clientSDKSetupURL for more info."
         return BraintreeException(message)
     }
+
+    fun isPayPalInstalled(): Boolean {
+        return deviceInspector.isPayPalInstalled(applicationContext)
+    }
 }
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
@@ -61,6 +61,9 @@ class BraintreeClient @VisibleForTesting internal constructor(
     private var launchesBrowserSwitchAsNewTask: Boolean = false
     private val deviceInspector: DeviceInspector
 
+    var isPayPalInstalled: Boolean = false
+        get() = deviceInspector.isPayPalInstalled(applicationContext)
+
     // NOTE: this constructor is used to make dependency injection easy
     internal constructor(params: BraintreeClientParams) : this(
         applicationContext = params.applicationContext,
@@ -388,9 +391,5 @@ class BraintreeClient @VisibleForTesting internal constructor(
             "https://developer.paypal.com/braintree/docs/guides/client-sdk/setup/android/v4#initialization"
         val message = "Valid authorization required. See $clientSDKSetupURL for more info."
         return BraintreeException(message)
-    }
-
-    fun isPayPalInstalled(): Boolean {
-        return deviceInspector.isPayPalInstalled(applicationContext)
     }
 }
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
@@ -61,9 +61,6 @@ class BraintreeClient @VisibleForTesting internal constructor(
     private var launchesBrowserSwitchAsNewTask: Boolean = false
     private val deviceInspector: DeviceInspector
 
-    var isPayPalInstalled: Boolean = false
-        get() = deviceInspector.isPayPalInstalled(applicationContext)
-
     // NOTE: this constructor is used to make dependency injection easy
     internal constructor(params: BraintreeClientParams) : this(
         applicationContext = params.applicationContext,
@@ -384,6 +381,10 @@ class BraintreeClient @VisibleForTesting internal constructor(
      */
     fun launchesBrowserSwitchAsNewTask(launchesBrowserSwitchAsNewTask: Boolean) {
         this.launchesBrowserSwitchAsNewTask = launchesBrowserSwitchAsNewTask
+    }
+
+    fun isPayPalInstalled(): Boolean {
+        return deviceInspector.isPayPalInstalled(applicationContext)
     }
 
     private fun createAuthError(): BraintreeException {
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/LinkType.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/LinkType.kt
@@ -5,7 +5,7 @@ package com.braintreepayments.api.core
  * Note: This enum is exposed for internal Braintree use only. Do not use.
  * It is not covered by Semantic Versioning and may change or be removed at any time.
  */
-public enum class LinkType(val stringValue: String)  {
+public enum class LinkType(val stringValue: String) {
     UNIVERSAL("universal"),
     DEEPLINK("deeplink")
 }
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/LinkType.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/LinkType.kt
@@ -1,10 +1,13 @@
 package com.braintreepayments.api.core
 
+import androidx.annotation.RestrictTo
+
 /**
  * Used to describe the link type for analytics
  * Note: This enum is exposed for internal Braintree use only. Do not use.
  * It is not covered by Semantic Versioning and may change or be removed at any time.
  */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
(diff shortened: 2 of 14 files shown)


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.java
```diff
@@ -59,6 +62,10 @@ void sendRequest(final Context context, final PayPalRequest payPalRequest,
                         ? SETUP_BILLING_AGREEMENT_ENDPOINT : CREATE_SINGLE_PAYMENT_ENDPOINT;
                 String url = String.format("/v1/%s", endpoint);
                 String appLinkReturn = isBillingAgreement ? appLink : null;
+                
+                final String linkType = (isBillingAgreement &&
+                                        ((PayPalVaultRequest) payPalRequest).getEnablePayPalAppSwitch() &&
+                                        braintreeClient.isPayPalInstalled()) ? "universal" : "deeplink";
```
sarahkoop, line 68: If this is only used internally can we call it "applink" instead of "universal"?

jaxdesmarais: We unfortunately decided these names on iOS for the venmo app switch project. We pass the same for that flow as well: https://github.com/braintree/braintree_android/blob/01c5f5237b51c0cf5dfa4fa2fe767000513f0b6b/Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.java#L38. We can consider changing it on android but would need to loop in the analytics team since they currently build their analytics around universal/deeplink as the expected values.

sarahkoop: Ahhh ok disregard then we can leave it "universal"

scannillo: This PR isn't actually sending the `link_type` analytics value. For that we should stick with `universal` or `deeplink`, like Jax called out.

But the code here is just internal and can be whatever names we want. The `universal` naming doesn't have to come into play until the analytics layer ([here](https://github.com/braintree/braintree_android/blob/01c5f5237b51c0cf5dfa4fa2fe767000513f0b6b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt#L224-L248))

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.java
```diff
@@ -102,8 +109,20 @@ void sendRequest(final Context context, final PayPalRequest payPalRequest,
...
-                                            .clientMetadataId(clientMetadataId)
-                                            .approvalUrl(parsedRedirectUri.toString());
+                                                .clientMetadataId(clientMetadataId);
+
+                                        if (linkType.equals("universal")) {
+                                            if (pairingId != null && !pairingId.isEmpty()) {
+                                                paymentAuthRequest
+                                                        .approvalUrl(createAppSwitchUri(parsedRedirectUri).toString());
+                                            } else {
+                                                callback.onResult(null, new Exception("Missing BA Token for PayPal App Switch."));
```
scannillo, line 119: 👏 Nice call using the same error details! This makes debugging much easier across platforms

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.java
```diff
@@ -59,6 +62,10 @@ void sendRequest(final Context context, final PayPalRequest payPalRequest,
                         ? SETUP_BILLING_AGREEMENT_ENDPOINT : CREATE_SINGLE_PAYMENT_ENDPOINT;
                 String url = String.format("/v1/%s", endpoint);
                 String appLinkReturn = isBillingAgreement ? appLink : null;
+                
+                final String linkType = (isBillingAgreement &&
```
scannillo, line 66: Nitpick - on iOS too, I think we want to make this linkType an enum at some point. We could do that here instead of switching on a String.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.java
richherrera: What Sammy mentioned is correct. In my mind, I have two options: one is to convert this to a boolean and have the same logic in `BTPayPalClient` (Where I really need the string with values _"universal"_ or _"deeplink"_), or as a second option, modify this method so that it allows me to pass the `linkType` as a parameter from `BTPayPalClient`.

Any other ideas? What do you all think?

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
```diff
@@ -389,4 +389,8 @@ class BraintreeClient @VisibleForTesting internal constructor(
         val message = "Valid authorization required. See $clientSDKSetupURL for more info."
         return BraintreeException(message)
     }
+
+    fun isPayPalInstalled(): Boolean {
+        return deviceInspector.isPayPalInstalled(applicationContext)
+    }
```
sshropshire, line 395:
```suggestion
    var isPayPalInstalled: Boolean
      get() = deviceInspector.isPayPalInstalled(applicationContext)
```
nit: shorthand to make it a property

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.java
sshropshire: I'd vote for creating an `enum` like [Sammy mentioned](https://github.com/braintree/braintree_android/pull/1092/files#r1702196392). With an enum the `if/else` logic can be converted to a Kotlin `when()` clause (similar to `switch` in Swift).

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.java
sshropshire, line 119:
```suggestion
                                                callback.onResult(null, new BraintreeException("Missing BA Token for PayPal App Switch."));
```
`BraintreeException` is the standard exception type we use in the SDK when an error happens.

TestUtils/src/main/java/com/braintreepayments/api/testutils/MockBraintreeClientBuilder.java
```diff
@@ -128,6 +135,7 @@ public BraintreeClient build() {
         when(braintreeClient.getManifestActivityInfo(any())).thenReturn(activityInfo);
         when(braintreeClient.launchesBrowserSwitchAsNewTask()).thenReturn(
                 launchesBrowserSwitchAsNewTask);
+        when(braintreeClient.isPayPalInstalled()).thenReturn(isPayPalInstalled);
```
sshropshire, line 138: Feel free to ignore [this comment](https://github.com/braintree/braintree_android/pull/1092/files#r1704227805) if it makes mocking more difficult. I'd rather get this merged in now, and we'll look into making the code more Kotlin-like in the future.

sshropshire (approved):

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.java
richherrera: Added: [8a76214](https://github.com/braintree/braintree_android/pull/1092/commits/8a762142cfb7d5940f49867c7afcbb6b92ee6142)

BraintreeCore/src/main/java/com/braintreepayments/api/core/LinkType.kt
```diff
@@ -0,0 +1,11 @@
+package com.braintreepayments.api.core
+
+/**
+ * Used to describe the link type for analytics
+ * Note: This enum is exposed for internal Braintree use only. Do not use.
+ * It is not covered by Semantic Versioning and may change or be removed at any time.
+ */
+public enum class LinkType(val stringValue: String) {
```
tdchow, line 8: Could we use the `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)` to scope this enum to internal use?

BraintreeCore/src/main/java/com/braintreepayments/api/core/LinkType.kt
```diff
@@ -0,0 +1,11 @@
...
+    UNIVERSAL("universal"),
```
saperi22, line 9: On Android, the other type of link is App Links. Not sure if we want to keep the iOS nomenclature for analytics or use Android specific one.

saperi22: I see that there was some discussion around this. Please dis-regard if the rest of the team has come to a conclusion.

tdchow (approved):

PR #1093, opened 2024-08-02
Remove old analytic events lingering from v4

tdchow:
### Summary of changes

 - Remove lingering analytic events from v4
 - Slight refactor in PayPalClient - removed a few private functions that did the same for vault and checkout flows

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
@@ -19,7 +19,6 @@ import org.json.JSONObject
 /**
  * Used to tokenize PayPal accounts. For more information see the [documentation](https://developer.paypal.com/braintree/docs/guides/paypal/overview/android/v4)
  */
-@Suppress("TooManyFunctions")
 class PayPalClient @VisibleForTesting internal constructor(
     private val braintreeClient: BraintreeClient,
     private val internalPayPalClient: PayPalInternalClient = PayPalInternalClient(braintreeClient),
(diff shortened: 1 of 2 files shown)


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -65,55 +64,9 @@ class PayPalClient @VisibleForTesting internal constructor(
...
-        callback: PayPalPaymentAuthCallback
-    ) {
-        braintreeClient.sendAnalyticsEvent("paypal.billing-agreement.selected", analyticsParams)
-        if (payPalVaultRequest.shouldOfferCredit) {
-            braintreeClient.sendAnalyticsEvent(
-                "paypal.billing-agreement.credit.offered",
-                analyticsParams
-            )
-        }
+        braintreeClient.sendAnalyticsEvent(PayPalAnalytics.TOKENIZATION_STARTED, analyticsParams)
```
tdchow, line 69: Calling out that I moved this analytics call to after `isVaultRequest` is set. I also added the passing of `analyticsParams`.

We were previously omitting if the analytics event is a vault request.

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalClientUnitTest.java
```diff
@@ -100,7 +100,7 @@ public void createPaymentAuthRequest_callsBackPayPalResponse_sendsStartedAnalyti
 
         verify(braintreeClient).sendAnalyticsEvent(
             PayPalAnalytics.TOKENIZATION_STARTED,
-            new AnalyticsEventParams(null, null, false, null, null, null)
+            new AnalyticsEventParams(null, null, true, null, null, null)
```
tdchow, line 103: This test failed after making the change. The vault boolean defaulted to false before but the request in this test is a `PayPalVaultRequest`.

sarahkoop (approved):

PR #1096, opened 2024-08-05
Convert Card module classes to Kotlin

chinghlinn:
### Summary of changes

  - Convert AuthenticationInsight, BaseCard, BinData, Card, and CardClient to Kotlin
  - Parcelable interface implementation is changed to utilize kotlin-parcelize plugin(https://developer.android.com/kotlin/parcelize), therefore, the override functions from Parcelable and corresponding tests have been removed. The original property setters to set values null when the input is null or empty

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


BraintreeCore/src/main/java/com/braintreepayments/api/core/PaymentMethod.kt
```diff
@@ -25,79 +30,20 @@ abstract class PaymentMethod {
...
-    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-    constructor()
-
-    /**
-     * Sets the integration method associated with the tokenization call for analytics use.
-     * Defaults to custom and does not need to ever be set.
-     *
-     * @param integration the current integration style.
-     * @suppress
-     */
```
sshropshire, line 46: We'll want to move these docstrings into property annotations ([similar to here](https://github.com/braintree/braintree_android/blob/f13626a796ae46b21fe41d62b0092c7f73d3c036/BraintreeCore/src/main/java/com/braintreepayments/api/core/Authorization.kt#L8)).

Card/src/main/java/com/braintreepayments/api/card/AuthenticationInsight.kt
```diff
@@ -0,0 +1,40 @@
+package com.braintreepayments.api.card
+
+import android.os.Parcelable
+import com.braintreepayments.api.sharedutils.Json
+import kotlinx.parcelize.Parcelize
+import org.json.JSONObject
+
+@Parcelize
+data class AuthenticationInsight(val regulationEnvironment: String) : Parcelable {
```
sshropshire, line 9: Same here: there are some doc strings in [AuthenticationInsight.java](https://github.com/braintree/braintree_android/blob/f13626a796ae46b21fe41d62b0092c7f73d3c036/Card/src/main/java/com/braintreepayments/api/card/AuthenticationInsight.java#L48) that will need to be ported over.

Card/src/main/java/com/braintreepayments/api/card/BinData.kt
```diff
@@ -0,0 +1,147 @@
...
+    @StringDef(YES, NO, UNKNOWN)
+    internal annotation class BinType
+
+    /**
+     * @return Whether the card is a prepaid card. Possible [BinType] values are [BinData.YES], [BinData.NO] or [BinData.UNKNOWN]
+     */
+    @BinType
+    fun getPrepaid(): String {
+        return prepaid
+    }
```
sshropshire, line 36: random thought: We may be able to remove these accessors since it's now a Kotlin `data class`. We can migrate the docs strings into `@property` annotations as well.

Card/src/main/java/com/braintreepayments/api/card/BaseCard.kt
```diff
@@ -0,0 +1,116 @@
...
+
+import androidx.annotation.RestrictTo
+import com.braintreepayments.api.core.PaymentMethod
+import org.json.JSONException
+import org.json.JSONObject
+
+/**
+ * Base class used to build various types of cards
+ */
+abstract class BaseCard(
```
sshropshire, line 11: This is out of scope for this PR, but we should make a note to remove the `BaseCard` class. We removed `UnionPayCard`, so we should only need a single `Card` type in `v5`.

Card/src/main/java/com/braintreepayments/api/card/BaseCard.kt
```diff
@@ -0,0 +1,116 @@
...
+    open val cvv: String? = null,
+    open var expirationMonth: String? = null,
+    open var expirationYear: String? = null,
+    open val extendedAddress: String? = null,
+    open val firstName: String? = null,
+    open val lastName: String? = null,
+    open val locality: String? = null,
+    open val postalCode: String? = null,
+    open val region: String? = null,
+    open val streetAddress: String? = null,
```
tdchow, line 25: Should we make all of these `val`s `var`s instead? This way Java integrations are able to call the generated setter functions.

Card/src/main/java/com/braintreepayments/api/card/BaseCard.kt
```diff
@@ -0,0 +1,116 @@
...
+        const val LOCALITY_KEY: String = "locality"
+        const val NUMBER_KEY: String = "number"
+        const val POSTAL_CODE_KEY: String = "postalCode"
+        const val REGION_KEY: String = "region"
+        const val STREET_ADDRESS_KEY: String = "streetAddress"
+    }
+
+    var expirationDate: String?
+        get() {
+            if (expirationMonth?.isNotEmpty() == true && expirationYear?.isNotEmpty() == null) {
```
tdchow, line 54:
```suggestion
            if (expirationMonth?.isNotEmpty() == true && expirationYear?.isNotEmpty() == true) {
```

Should we check for true in both cases?

Card/src/main/java/com/braintreepayments/api/card/BinData.kt
```diff
@@ -0,0 +1,147 @@
...
+    private var commercial: String = UNKNOWN,
+    private var payroll: String = UNKNOWN,
+    private var issuingBank: String = UNKNOWN,
+    private var countryOfIssuance: String = UNKNOWN,
+    private var productId: String = UNKNOWN
+) : Parcelable {
+
+    @Retention(AnnotationRetention.SOURCE)
+    @StringDef(YES, NO, UNKNOWN)
+    internal annotation class BinType
```
tdchow, line 28: Could we convert `BinType` to an enum type?

Card/src/main/java/com/braintreepayments/api/card/CardClient.kt
```diff
@@ -0,0 +1,141 @@
...
+    )
+
+    @VisibleForTesting
+    internal constructor(
+        braintreeClient: BraintreeClient,
+        apiClient: ApiClient
+    ) {
+        this.braintreeClient = braintreeClient
+        this.apiClient = apiClient
+    }
```
tdchow, line 49: We might be able to consolidate these internal constructors and move `braintreeClient` and `apiClient` to constructor properties.

See PayPalClient as an example: https://github.com/braintree/braintree_android/blob/f13626a796ae46b21fe41d62b0092c7f73d3c036/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt#L22

Card/src/main/java/com/braintreepayments/api/card/CardClient.kt
```diff
@@ -0,0 +1,141 @@
...
+     */
+    fun tokenize(card: Card, callback: CardTokenizeCallback) {
+        braintreeClient.sendAnalyticsEvent(CardAnalytics.CARD_TOKENIZE_STARTED)
+        braintreeClient.getConfiguration { configuration: Configuration?, error: Exception? ->
+            if (error != null) {
+                callbackFailure(callback, CardResult.Failure(error))
+                return@getConfiguration
+            }
+            val shouldTokenizeViaGraphQL =
+                configuration!!.isGraphQLFeatureEnabled(
```
tdchow, line 85: Can we handle the case when `configuration` is null? 

We return an error in PayPalClient: https://github.com/braintree/braintree_android/blob/f13626a796ae46b21fe41d62b0092c7f73d3c036/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt#L75

Card/src/main/java/com/braintreepayments/api/card/BaseCard.kt
chinghlinn: Thought the same way, PaymentMethod will also be removed.

Card/src/main/java/com/braintreepayments/api/card/BaseCard.kt
chinghlinn: Thank you for catching the error, this will be addressed after removing super classes from Card.

Card/src/main/java/com/braintreepayments/api/card/BaseCard.kt
chinghlinn: One is for Month and one is for Year.

Card/src/main/java/com/braintreepayments/api/card/BinData.kt
sshropshire: Definitely yeah. We should convert all `@StringDef` and `@IntDef`, etc. types to `enum`. These are holdovers going way back to `v3`. At the time enums were considered memory-heavy in Android. I believe they fixed this at some point and it's no longer a concern.

Card/src/main/java/com/braintreepayments/api/card/BaseCard.kt
tdchow: Sorry I should have made it more clear. For the month condition we're checking if `expirationMonth?.isNotEmpty()` is true. But for the year condition we're checking if `expirationYear?.isNotEmpty()` is null. Should these conditions align and check fo true instead of null?

sshropshire: Yeah for some reason the second boolean equality check is checking against `null` and it should be `true`. Might be an existing bug that was never detected on the Java side.

BraintreeCore/src/main/java/com/braintreepayments/api/core/PaymentMethod.kt
chinghlinn: I will rollback PaymentMethod and remove super classes from Card.

Card/src/main/java/com/braintreepayments/api/card/BinData.kt
chinghlinn: Will change to enum

Card/src/main/java/com/braintreepayments/api/card/BaseCard.kt
sshropshire: Actually, I'd be up for removing this `expirationDate` convenience accessor. The `MM/YY` format of the string it produces is somewhat opinionated.

One of our goals should be to make the SDK less opinionated in general, and removing this property is in line with that goal.

Card/src/main/java/com/braintreepayments/api/card/CardClient.kt
sshropshire: I Agree. We can finally leverage Kotlin default parameters 🙌 .

Card/src/main/java/com/braintreepayments/api/card/CardClient.kt
sshropshire: Yeah it would be awesome if we can remove state from all feature clients. There's potential for `CardClient.configuration` to become stale in an app that hosts multiple merchants e.g. a marketplace app.

Card/src/main/java/com/braintreepayments/api/card/BaseCard.kt
tdchow: @sshropshire - that's a great call! it should be up to the merchant on how they wish to display their formatted date. I'm in favor of removing this accessor.

BraintreeCore/src/main/java/com/braintreepayments/api/core/PaymentMethod.kt
```diff
@@ -1,103 +1,38 @@
...
-    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-    fun setIntegration(integration: String) {
-        _integration = integration
-    }
-
-    /**
-     * Sets the source associated with the tokenization call for analytics use. Set automatically.
```
sshropshire, line 53: We should migrate keep these docstrings as `@property` annotations if possible.

BraintreeCore/src/main/java/com/braintreepayments/api/core/PaymentMethod.kt
```diff
@@ -1,103 +1,38 @@
...
     fun buildMetadataJSON(): JSONObject {
         return MetadataBuilder()
-            .sessionId(_sessionId)
-            .source(_source)
-            .integration(_integration)
+            .sessionId(sessionId)
+            .source(source)
+            .integration(integration)
             .build()
     }
```
sshropshire, line 29: It might be better to duplicate this logic in each class that implements `PaymentMethod` for now. I'm not sure if Java fully supports default implementation [just yet](https://stackoverflow.com/a/49933841). Even if there is support for it, I think ultimately we'll want PaymentMethod types to become Plain Data objects.

sshropshire:
@chinghlinn there's a lint error we'll have to resolve. Running `./gradlew detekt` in the root dir of this repo on your local machine runs the linter. I think it's minor from what I can see in the GitHub Actions output.

BraintreeCore/src/main/java/com/braintreepayments/api/core/PaymentMethod.kt
```diff
@@ -1,102 +1,35 @@
...
 
-    /**
-     * @suppress
-     */
-    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-    companion object {
+    var sessionId: String?
+    var source: String?
+    var integration: IntegrationType?
+    val apiPath: String
```
tdchow, line 23: Should `apiPath` also be a var?

Card/src/main/java/com/braintreepayments/api/card/BinData.kt
```diff
@@ -0,0 +1,85 @@
...
+ * @property commercial Whether the card type is a commercial card and is capable of processing
+ * Level 2 transactions.
+ * @property payroll Whether the card is a payroll card.
+ * @property issuingBank The bank that issued the credit card.
+ * @property countryOfIssuance The country that issued the credit card.
+ * @property productId The code for the product type of the card
+ * (e.g. `D` (Visa Signature Preferred), `G` (Visa Business)).
+ */
+@Parcelize
+data class BinData(
```
tdchow, line 26: Is this class only instantiated by the SDK? If so, could we define the constructor as internal or use the `RestrictTo` annotation if the constructor is used in a different module.

Card/src/main/java/com/braintreepayments/api/card/BinData.kt
```diff
@@ -0,0 +1,85 @@
...
+    val debit: BinType = BinType.Unknown,
+    val durbinRegulated: BinType = BinType.Unknown,
+    val commercial: BinType = BinType.Unknown,
+    val payroll: BinType = BinType.Unknown,
+    val issuingBank: String = BinType.Unknown.name,
+    val countryOfIssuance: String = BinType.Unknown.name,
+    val productId: String = BinType.Unknown.name
+) : Parcelable {
+
+    enum class BinType {
```
tdchow, line 38: Should we move this enum class to a separate file? We had followed this pattern in other modules.

Card/src/main/java/com/braintreepayments/api/card/CardClient.kt
```diff
@@ -0,0 +1,134 @@
...
+     */
+    constructor(context: Context, authorization: String) : this(
+        BraintreeClient(
+            context,
+            authorization
+        )
+    )
+
+    @VisibleForTesting
+    internal constructor(braintreeClient: BraintreeClient) : this(
```
tdchow, line 37: Could we consolidate this constructor by setting a default value for `apiClient` in the primary constructor?

BraintreeCore/src/main/java/com/braintreepayments/api/core/PaymentMethod.kt
chinghlinn: The value for apiPath is fixed for the type of PaymentMethod, for example, in VenmoAccount, the implementation is 
    public String getApiPath() {
        return "venmo_accounts";
    }

tdchow (approved):

PR #1098, opened 2024-08-06
Add PayPalVaultEditResult.Success properties

tdchow:
### Summary of changes

 - Add properties to `PayPalVaultEditResult.Success`

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/vaultedit/PayPalVaultEditResult.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/vaultedit/PayPalVaultEditResult.kt
@@ -9,7 +9,8 @@ sealed class PayPalVaultEditResult {
     /**
      * The PayPal vault edit flow completed successfully.
      *
-     * @property clientMetadataId client metadata id
+     * @property clientMetadataId This ID is used to link subsequent retry attempts if payment is
+     * declined.
      * @property payerId ID of the payer
      * @property email email address of the payer
      * @property firstName first name of the payer
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/vaultedit/PayPalVaultEditResult.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/vaultedit/PayPalVaultEditResult.kt
@@ -1,6 +1,7 @@
 package com.braintreepayments.api.paypal.vaultedit
 
 import com.braintreepayments.api.ExperimentalBetaApi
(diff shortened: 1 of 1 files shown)


jaxdesmarais (approved):

PayPal/src/main/java/com/braintreepayments/api/paypal/vaultedit/PayPalVaultEditResult.kt
```diff
@@ -1,22 +1,29 @@
...
 @ExperimentalBetaApi
 sealed class PayPalVaultEditResult {
 
     /**
      * The PayPal vault edit flow completed successfully.
      */
-    @Suppress("EmptyDefaultConstructor")
-    class Success(
-        // TODO: add FI details
+    class Success internal constructor(
```
sarahkoop, line 12: Should we have doc strings for these params?

tdchow: Ah yeah we should. Let me update this.

PayPal/src/main/java/com/braintreepayments/api/paypal/vaultedit/PayPalVaultEditResult.kt
```diff
@@ -1,22 +1,38 @@
...
 import com.braintreepayments.api.ExperimentalBetaApi
+import com.braintreepayments.api.core.PostalAddress
 
 @ExperimentalBetaApi
 sealed class PayPalVaultEditResult {
 
     /**
      * The PayPal vault edit flow completed successfully.
+     *
+     * @property clientMetadataId client metadata id
```
tdchow, line 12: @jaxdesmarais do you know if there's anything specific we need to call out for the `clientMetadataId` property?

jaxdesmarais: It will be pertinent to error handling flow, so maybe we can add something like: "used to link subsequent retry attempts if payment is declined"? Essentially this value acts as a state management and will be passed into the `edit` method after the first attempt if their is a failure. What do you think, does that make it clear?

tdchow: Oh that's right, the error handling flow! I think that description is clear, let me update the PR.

sshropshire (approved):

PR #1101, opened 2024-08-07
Updates to Core Module

tdchow:
### Summary of changes

 - Move ExperimentalBetaApi file to the core package
 - Update visibility modifiers
 - Clean up class constructors
 - Remove unused classes
 - Convert `BraintreeRequestCodes` and `IntegrationType` to enums

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


BraintreeCore/src/main/java/com/braintreepayments/api/core/ApiClient.kt
```diff
@@ -14,17 +13,15 @@ class ApiClient(private val braintreeClient: BraintreeClient) {
...
-                    parseResponseToJSON(responseBody)?.let { json ->
-                        sendAnalyticsEvent("card.graphql.tokenization.success")
-                        callback.onResult(json, null)
-                    } ?: httpError?.let { error ->
-                        sendAnalyticsEvent("card.graphql.tokenization.failure")
-                        callback.onResult(null, error)
-                    }
+            sendGraphQLPOST(tokenizePayload) { responseBody, httpError ->
+                parseResponseToJSON(responseBody)?.let { json ->
+                    sendAnalyticsEvent("card.graphql.tokenization.success")
```
sarahkoop, line 18: Are these also leftover events from v4?

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
```diff
@@ -29,7 +28,7 @@ class BraintreeClient @VisibleForTesting internal constructor(
      * @suppress
      */
     @get:RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-    val integrationType: String,
+    val integrationType: IntegrationType,
```
sarahkoop, line 31: Can this whole class be internal now?

tdchow: Unfortunately, the feature clients in other modules reference this `integrationType` property, so we wouldn't be able to make it internal without additional refactoring.

sarahkoop: Ah sorry I meant the whole `BraintreeClient` class - can we restrict it to library scope or no?

BraintreeCore/src/main/java/com/braintreepayments/api/core/ApiClient.kt
tdchow: 🤔 I'm not sure if the v5 core events were ever updated. I'm only seeing a latency event defined in the `CoreAnalytics` class:

```kotlin
internal object CoreAnalytics {
    const val apiRequestLatency = "core:api-request-latency"
}
```

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
tdchow: Oh! BraintreeClient already has the library scope annotation. We can remove them from the individual properties and functions. I'll make an update for that.

BraintreeCore/src/main/java/com/braintreepayments/api/core/ApiClient.kt
sarahkoop: Ok I think in an effort to reduce analytics noise we should go ahead and remove any events not defined in our shared iOS/Android events list (which should all existing in the Analytics enums now)

tdchow: Sounds good. I'll remove these events in this PR and follow up with an audit of all events sent.

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
tdchow: Updated: 05273983696d48e6c98e1ee507a193543d4c678b

BraintreeCore/src/main/java/com/braintreepayments/api/core/ApiClient.kt
tdchow: Updated: 0d389595336d6656ab98c0148f80cab36db5307b

sarahkoop (approved):

PR #1108, opened 2024-08-09
Add Version Catalog

tdchow:
### Summary of changes

 - This PR adds a `libs.versions.toml` file that contains all versions, libraries, and plugins.
 
https://developer.android.com/build/migrate-to-catalogs#groovy

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/Demo/build.gradle
+++ b/Demo/build.gradle
@@ -105,6 +105,7 @@ dependencies {
     implementation libs.androidx.legacy.support.v4
     implementation libs.androidx.lifecycle.extensions
     implementation libs.androidx.lifecycle.viewmodel.ktx
+    implementation libs.bt.card.form
 
     def composeBom = platform(libs.androidx.compose.bom)
     implementation composeBom
--- a/Demo/build.gradle
+++ b/Demo/build.gradle
@@ -75,7 +75,7 @@ android {
 
 dependencies {
     implementation 'androidx.preference:preference:1.2.1'
-    implementation deps.kotlinStdLib
+    implementation libs.kotlin.stdlib
 
     implementation('com.squareup.retrofit:retrofit:1.9.0') {
         exclude module: 'com.google.gson'
@@ -96,7 +96,7 @@ dependencies {
 
     implementation 'com.google.code.gson:gson:2.10.1'
     implementation 'de.greenrobot:eventbus:2.4.1'
-    implementation deps.appCompat
+    implementation libs.androidx.appcompat
     implementation 'androidx.preference:preference:1.2.1'
     implementation 'com.google.android.material:material:1.10.0'
     implementation 'androidx.cardview:cardview:1.0.0'
--- a/Demo/build.gradle
+++ b/Demo/build.gradle
@@ -74,10 +74,9 @@ android {
 }
 
 dependencies {
-    implementation 'androidx.preference:preference:1.2.1'
     implementation libs.kotlin.stdlib
 
-    implementation('com.squareup.retrofit:retrofit:1.9.0') {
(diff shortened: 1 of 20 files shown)


Demo/build.gradle
```diff
@@ -52,7 +52,7 @@ android {
     }
 
     composeOptions {
-        kotlinCompilerExtensionVersion = "1.4.1"
+        kotlinCompilerExtensionVersion = "1.5.3"
```
tdchow, line 55: This needed to be updated to match the Kotlin version.

Demo/build.gradle
```diff
@@ -94,34 +93,29 @@ dependencies {
...
-    implementation 'com.google.code.gson:gson:2.10.1'
-    implementation 'de.greenrobot:eventbus:2.4.1'
-    implementation deps.appCompat
-    implementation 'androidx.preference:preference:1.2.1'
-    implementation 'com.google.android.material:material:1.10.0'
-    implementation 'androidx.cardview:cardview:1.0.0'
-    implementation 'androidx.multidex:multidex:2.0.1'
-
-    implementation 'io.card:android-sdk:5.5.1'
-    implementation 'com.braintreepayments:card-form:5.4.0'
```
tdchow, line 106: These weren't referenced anywhere. Are they safe to remove?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,107 @@
...
+gradleNexusPublish = "1.1.0"
+kotlin = "1.9.10"
+coreKtx = "1.12.0"
+androidxLifecycle = "2.6.2"
+androidxRoom = "2.6.1"
+androidxTest = "1.5.0"
+androidxAppcompat = "1.6.0"
+androidxAnnotation = "1.7.0"
+androidxWork = "2.8.1"
+androidxJunit = "1.1.5"
```
tdchow, line 14: Calling out that versions are camel case, while library and plugin definitions are kabab case. According to the docs, kebab case is used to help with autocomplete in gradle files.

settings.gradle
```diff
@@ -1,3 +1,17 @@
+pluginManagement {
```
tdchow, line 1: This block is required to be able to resolve plugins in the root build.gradle plugins block.

Demo/build.gradle
jaxdesmarais: Should be fine to remove. `io.card` is the old snap a photo autofill credit card thing used in the Drop-in and `card-form` is used by the Drop-in as well. Neither should be needed here.

jaxdesmarais (approved):

settings.gradle
sshropshire: curious: does gradle sync fail without this block?

tdchow: Yep there were failures due to this plugins block in the root build.gradle file:

```groovy
plugins {
    alias libs.plugins.dokka
    alias libs.plugins.gradle.nexus.publish
    alias libs.plugins.detekt

    alias libs.plugins.android.application apply false
    alias libs.plugins.android.library apply false
    alias libs.plugins.kotlin.android apply false
    alias libs.plugins.kotlin.parcelize apply false
    alias libs.plugins.kotlin.kapt apply false
    alias libs.plugins.androidx.safeargs apply false
}
```

Demo/build.gradle
tdchow: Hmm, I'm now seeing errors when removing `card-form` 🤔 there are references in the demo app. I'll leave it in for now, but it sounds like we could refactor it out at some point.

PR #1110, opened 2024-08-12
[App Switch] Send `paypal_installed` FPTI tag

richherrera:
### Summary of changes

- Send `paypal_installed` in `batch_params` to PayPal's analytics service (FPTI)
- Fix PayPal fragment BA token message

Screenshot 2024-08-12 at 3 12 12 p m
Screenshot 2024-08-12 at 3 10 48 p m

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### Authors

- @richherrera


BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
```diff
@@ -210,6 +210,7 @@ internal class AnalyticsClient constructor(
             .put(FPTI_KEY_EVENT_NAME, event.name)
             .put(FPTI_KEY_TIMESTAMP, event.timestamp)
             .put(FPTI_KEY_VENMO_INSTALLED, event.venmoInstalled)
+            .put(FPTI_KEY_PAYPAL_INSTALLED, event.isPayPalInstalled)
```
sshropshire, line 213: Hey Rich. Some of this logic has changed recently. Here's [the PR](https://github.com/braintree/braintree_android/pull/1104) with the changes. We decided to start putting `venmo_installed` and `paypal_installed` into batch params now since they aren't tied to any individual event.

richherrera: 😲 Thanks for the heads up!!!

sshropshire: Yeah no prob it just got merged in. It's a lot cleaner now! We were able to remove the `DeviceInspector` reference from `BraintreeClient` in the updated implementation.

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
```diff
@@ -53,6 +53,7 @@ class BraintreeClient @VisibleForTesting internal constructor(
 
     private val crashReporter: CrashReporter
     private var launchesBrowserSwitchAsNewTask: Boolean = false
+    private val deviceInspector: DeviceInspector
```
richherrera, line 56: I need the deviceInspector here because the client (PayPalClient) needs to know if PayPal is installed

https://github.com/braintree/braintree_android/blob/06fac45120d563630c55bf9ff95ff97230892e88/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt#L46

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
richherrera: Updated!!! I had to add the `DeviceInspector` to `BraintreeClient` 😔 because I need it to check if the PayPal app is installed. This way, we can determine the AppSwitch flow. Let me know if there's any other approach we can take.

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
sshropshire: Oh got it yeah. We could also move the `DeviceInspector` reference into `PayPalInternalClient` if that works? I'm hoping we'll be able to phase out `BraintreeClient` over time.

agedd (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/DeviceInspector.kt
```diff
@@ -54,10 +56,16 @@ internal class DeviceInspector(
         return isVenmoIntentAvailable && isVenmoSignatureValid
     }
 
+    /**
+     * @suppress
+     */
```
sshropshire, line 61: A single `@suppress` annotation at the `class DeviceInspector` level may be enough to keep the docs tool from generating docs for this class.

sshropshire (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/DeviceInspector.kt
richherrera: 😲 Updated: [b3f9a33](https://github.com/braintree/braintree_android/pull/1110/commits/b3f9a33b2d2f38491f68c1c45d717e1b30cbb0e5)

PR #1111, opened 2024-08-13
[App Switch] Add analytic events

richherrera:
### Summary of changes

 - [DTMOBILES-173] & [DTMOBILES-899] Add relevant events for the flow

 

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### Authors

- @richherrera


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -93,8 +99,20 @@ class PayPalClient internal constructor(
             error: Exception? ->
             if (payPalResponse != null) {
                 payPalContextId = payPalResponse.pairingId
+                linkType = if (internalPayPalClient.isAppSwitchEnabled(payPalRequest) &&
+                    internalPayPalClient.isPayPalInstalled(context)) {
+                    LinkType.UNIVERSAL
+                } else {
+                    LinkType.DEEPLINK
```
sarahkoop, line 106: For PayPal in v5 all links are going to be universal/app links - we no longer support deep links

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -93,8 +99,20 @@ class PayPalClient internal constructor(
...
+                }
+
                 try {
                     payPalResponse.browserSwitchOptions = buildBrowserSwitchOptions(payPalResponse)
+
+                    if (internalPayPalClient.isAppSwitchEnabled(payPalRequest)) {
+                        braintreeClient.sendAnalyticsEvent(PayPalAnalytics.APP_SWITCH_STARTED, analyticsParams)
```
sarahkoop, line 113: Should this analytic event be moved up to right after the `TOKENIZE_STARTED` event? I assume we want to trigger this immediately at the start of the flow

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -193,20 +212,28 @@ class PayPalClient internal constructor(
 
             internalPayPalClient.tokenize(payPalAccount) { payPalAccountNonce: PayPalAccountNonce?, error: Exception? ->
                 if (payPalAccountNonce != null) {
+                    if (isAppSwitchFlow) {
+                        braintreeClient.sendAnalyticsEvent(
+                            PayPalAnalytics.HANDLE_RETURN_STARTED,
+                            analyticsParams
+                        )
+                    }
```
sarahkoop, line 220: I think this should also move to right at the start of this method (line 189)

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -258,32 +285,51 @@ class PayPalClient internal constructor(
...
     private fun callbackBrowserSwitchCancel(
         callback: PayPalTokenizeCallback,
-        cancel: PayPalResult.Cancel
+        cancel: PayPalResult.Cancel,
+        isAppSwitchFlow: Boolean
     ) {
         braintreeClient.sendAnalyticsEvent(PayPalAnalytics.BROWSER_LOGIN_CANCELED, analyticsParams)
+
+        if (isAppSwitchFlow) {
+            braintreeClient.sendAnalyticsEvent(PayPalAnalytics.APP_SWITCH_FAILED, analyticsParams)
```
tdchow, line 294: Would we want to track the canceled flow separately from `APP_SWITCH_FAILED`?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
jaxdesmarais: I think this actually indicates if we fall back to the legacy (CCT) flow or are using the app switch flow

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
jaxdesmarais: I think we want to trigger this event right before we launch the Venice app

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -153,12 +153,12 @@ internal class PayPalInternalClient(
             .build()
     }
 
-    private fun isAppSwitchEnabled(payPalRequest: PayPalRequest): Boolean {
+    fun isAppSwitchEnabled(payPalRequest: PayPalRequest): Boolean {
```
jaxdesmarais, line 156: Should these remain public or be internal at least? Not sure if we want to make them public.

richherrera: These methods are from `PayPalInternalClient`, which is now `internal`. I changed them to `public` because I need access to them for UTs. `PayPalInternalClient` and `PayPalClient` need them to define certain flow criteria, any suggestions? UTs are requiring these methods to be public

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
richherrera: Exactly, this event needs to be called before launching the Venice app. The callback that comes afterward (`callback.onPayPalPaymentAuthRequest`) is the one that tells the fragment to launch the Venice app

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
richherrera: That's correct, this is just for analytics purposes. In this case, we're setting the `linkType` to differentiate the type of flow that's happening. If the user is eligible for AppSwitch, we send the parameter as `UNIVERSAL`,  otherwise, we send it as `DEEPLINK`

tdchow: Can we internally rename these to `APP_LINK` and `APP_SWITCH`? If we need to send certain values for analytics purposes, we could have a string value for each of these. I think it'd make things a bit clearer inside the SDK.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
richherrera: Agree, but for now, I think we're only requiring the events that are currently added. We can always iterate later. What do you think?

tdchow: I think we should try to have our events in a final state before we do an SDK release. It'll be hard to sift through the data if we end up changing this to a unique Cancel event in a later release. Can we sync with the larger team to see if we can add a cancel event for app switch?

richherrera: Thanks you, updated: [ae501ff](https://github.com/braintree/braintree_android/pull/1111/commits/ae501ff361186528495f31308b1ba36b012147d1)

tdchow (approved):

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
richherrera: Updated: [209b57c](https://github.com/braintree/braintree_android/pull/1111/commits/209b57c44328035385fc5095f0077f5cdeaf59d4)

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -93,8 +99,21 @@ class PayPalClient internal constructor(
             error: Exception? ->
             if (payPalResponse != null) {
                 payPalContextId = payPalResponse.pairingId
+                val isAppSwitchFlow = internalPayPalClient.isAppSwitchEnabled(payPalRequest) &&
+                    internalPayPalClient.isPayPalInstalled(context)
+                linkType = if (isAppSwitchFlow) {
+                    LinkType.APP_SWITCH
+                } else {
+                    LinkType.APP_LINK
+                }
```
jaxdesmarais, line 108:
```suggestion
                linkType = if (isAppSwitchFlow) LinkType.APP_SWITCH else LinkType.APP_LINK
```

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
jaxdesmarais: 🤔 not sure if @tdchow has a suggestion on this, ideally we'd like to keep them private. If we are going to make them public temporarily we should likely add some docstrings saying it's not covered by semantic versioning so we can update in the future.

tdchow: I believe @richherrera is correct. Even though we removed the `private` modifiers, since this entire class is `internal` these functions will be hidden from merchants. We're also using these functions in `PayPalClient`. So we'll need to keep them as internal.

jaxdesmarais: Nice! Thank you both for the explanation.

jaxdesmarais (approved):

PR #1114, opened 2024-08-13
Googlepay kotlin conversion

chinghlinn:
### Summary of changes

 - Convert BraintreeGooglePayWalletConstants, GooglePayActivity, GooglePayActivityResultContract, GooglePayCapabilities,  GooglePayClient, GooglePayException, GooglePayInternalClient, GooglePayLauncher, GooglePayPaymentAuthRequestParams, GooglePayPaymentAuthResult, GooglePayRequest, and ReadyForGooglePayRequest to Kotlin.
 - GooglePayRequest remains Parcelable pattern to implement Parcelable interface because the HashMap properties have values of JSONObject type which does not implement Parcelable interface and cannot utilize the kotlin-parcelize plugin.
 - Suppress warnings SwallowedException, LongMethod, MaximumLineLength, TooGenericExceptionCaught, CyclomaticComplexMethod, NestedBlockDepth, MaxLineLength as refactoring is required.

### Checklist

 - [x] Added a changelog entry
 - [ ] Relevant test coverage


GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayActivity.kt
```diff
@@ -0,0 +1,62 @@
...
+        val paymentsClient = Wallet.getPaymentsClient(
+            this, WalletOptions.Builder()
+                .setEnvironment(
+                    intent.getIntExtra(
+                        EXTRA_ENVIRONMENT,
+                        WalletConstants.ENVIRONMENT_TEST
+                    )
+                )
+                .build()
+        )
```
sshropshire, line 29: random q: Did this code auto format this way? The Java variant was about 3 lines. I feel like Kotlin would be about the same since it's less verbose than Java.

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayClient.kt
```diff
@@ -0,0 +1,636 @@
...
+
+                if (request != null) {
+                    json.put(
+                        "existingPaymentMethodRequired",
+                        request.isExistingPaymentMethodRequired
+                    )
+                }
+            } catch (ignored: JSONException) {
+            }
+            val request1 = IsReadyToPayRequest.fromJson(json.toString())
```
sshropshire, line 134:
```suggestion
            val readyToPayRequest = IsReadyToPayRequest.fromJson(json.toString())
```

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayActivity.kt
chinghlinn: Yes, it was auto formatted. The number of parameters are the same as Java version, it looks like the line is broken by function calls and parameters to stay within the line length warning.

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayClient.kt
```diff
@@ -0,0 +1,636 @@
...
+
+        braintreeClient.getConfiguration { configuration: Configuration?, configError: Exception? ->
+            if (configuration == null && configError != null) {
+                callbackPaymentRequestFailure(
+                    GooglePayPaymentAuthRequest.Failure(configError),
+                    callback
+                )
+                return@getConfiguration
+            }
+            if (!configuration!!.isGooglePayEnabled) {
```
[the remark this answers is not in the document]
chinghlinn: I overlooked that the configuration variable is possible to be null here because when configuration is null && configError is null. I will probably have to re-think this block.

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayClient.kt
```diff
@@ -263,14 +246,14 @@ class GooglePayClient @VisibleForTesting internal constructor(
      * @param callback    [GooglePayTokenizeCallback]
      */
     @SuppressWarnings("TooGenericExceptionCaught")
-    fun tokenize(paymentData: PaymentData?, callback: GooglePayTokenizeCallback) {
+    fun tokenize(paymentData: PaymentData, callback: GooglePayTokenizeCallback) {
```
tdchow, line 249: I think making `PaymentData` non-null would be a breaking change for merchants. Should we keep it as nullable and return an error for this case? Or should we keep this change and update the changelog?

chinghlinn: It looks like it's also an overload function to fun tokenize(paymentAuthResult: GooglePayPaymentAuthResult, callback: GooglePayTokenizeCallback), which checks non-null before calling this function tokenize(paymentData: PaymentData, callback: GooglePayTokenizeCallback). Inside this function, each try-catch block tries to access data paymentData until the final JSONException catch. When the passed parameter paymentData is null, it will always cause exception. We can only handle exception when the data in paymentData is invalid by making it non-null.

tdchow: Ahh I see. I missed that!

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayInternalClient.kt
```diff
@@ -0,0 +1,48 @@
...
+
+internal class GooglePayInternalClient {
+    fun isReadyToPay(
+        context: Context?,
+        configuration: Configuration,
+        isReadyToPayRequest: IsReadyToPayRequest?,
+        callback: GooglePayIsReadyToPayCallback
+    ) {
+        val paymentsClient = Wallet.getPaymentsClient(
+            context!!,
```
tdchow, line 21: Can we handle the null case here?

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayInternalClient.kt
```diff
@@ -0,0 +1,48 @@
...
+            WalletOptions.Builder()
+                .setEnvironment(getGooglePayEnvironment(configuration))
+                .build()
+        )
+        paymentsClient.isReadyToPay(isReadyToPayRequest!!)
```
tdchow, line 26: Same here, can we handle when `isReadyToPayRequest` is null?

chinghlinn: It looks like we can simply make the function parameters non-null, I will push another update shortly.

tdchow (approved):
Looks great - thanks for making the updates!

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayClient.kt
```diff
@@ -0,0 +1,620 @@
...
+
+    /**
+     * Start the Google Pay payment flow. This will return [GooglePayPaymentAuthRequestParams] that are
+     * used to present Google Pay payment sheet in
+     * [GooglePayLauncher.launch]
+     *
+     * @param request  The [GooglePayRequest] containing options for the transaction.
+     * @param callback [GooglePayPaymentAuthRequestCallback]
+     */
+    @SuppressWarnings("LongMethod", "MaximumLineLength")
```
sarahkoop, line 172: Which line is causing the max line length error? I think we should be able to resolve that instead of suppress

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayRequest.kt
```diff
@@ -0,0 +1,437 @@
...
+import org.json.JSONArray
+import org.json.JSONException
+import org.json.JSONObject
+import java.util.Locale
+
+/**
+ * Represents the parameters that are needed to use the Google Pay API.
+ */
+@SuppressWarnings("TooManyFunctions")
+class GooglePayRequest : Parcelable {
```
sarahkoop, line 19: Can we use the parcelize annotation instead of Parcelable here? then we shouldn't need the implementation code

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayRequest.kt
```diff
@@ -0,0 +1,437 @@
...
+    // JSON object under the hood
+    var transactionInfo: TransactionInfo? = null
+    /**
+     * @return If the buyer's email address is required to be returned.
+     */
+    /**
+     * Optional.
+     *
+     * @param emailRequired `true` if the buyer's email address is required to be returned, `false` otherwise.
+     */
```
sarahkoop, line 40: The docs strings on these look like they didn't format correctly for Kotlin - I think we can remove all of the return statements

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayRequest.kt
chinghlinn: I did not use Parcelize annotation was because the internal HashMap properties have values with type of JSONObject and JSONArray, they do not implement Parcelable. However, I just found they are not write to parcel either, are they intended to be omitted?

sshropshire: The `:GooglePay` module definitely has some tech debt associated with it. We found a bug a while back where a critical property [wasn't being parceled properly](https://github.com/braintree/braintree_android/pull/705/files), so it wouldn't surprise me if the parcelable implementation is lacking.

The error-prone nature of maintaining our own `Parcelable` implementation highlights the huge benefit we gain by moving to `@Parcelize`. I don't know if it comes in a separate PR, but if we need to refactor `GooglePayRequest` I think we should.

chinghlinn: Without making a major change (keep the public functions signatures unchanged), I am thinking to change the HashMap value type from JSONObject/JSONArray to String. Then there will be a data conversion in the get and set functions. For example, 

   private val allowedPaymentMethods = HashMap<String, String>()

    fun getAllowedPaymentMethod(type: String): JSONObject? {
        return jsonObjectParser.parse(allowedPaymentMethods[type])
    }

    fun setAllowedPaymentMethod(paymentMethodType: String, parameters: JSONObject) {
        allowedPaymentMethods[paymentMethodType] = parameters.toString()
    }

Let me know what you think.

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayLauncher.kt
```diff
@@ -0,0 +1,73 @@
...
+    callback: GooglePayLauncherCallback
+) {
+
+    private val activityLauncher: ActivityResultLauncher<GooglePayPaymentAuthRequestParams> = registry.register(
+        GOOGLE_PAY_RESULT, lifecycleOwner,
+        GooglePayActivityResultContract()
+    ) { googlePayPaymentAuthResult: GooglePayPaymentAuthResult? ->
+        callback.onGooglePayLauncherResult(
+            googlePayPaymentAuthResult
+        )
```
sshropshire, line 25:
```suggestion
        callback.onGooglePayLauncherResult(googlePayPaymentAuthResult)
```
nit: It just feels like the linter is very aggressively auto formatting. It actually has a reverse effect of making the code less readable IMO.

sshropshire (approved):

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayRequest.kt
sshropshire: I'm not sure immediately. Personally I'd prefer to get the Kotlin changes in and maybe we tackle in a separate PR? We're on a tight timeline to move out of beta and into GA. There may be an opportunity to for us to refactor in a minor version bump.

chinghlinn: I made a change in the last commit already, https://github.com/braintree/braintree_android/pull/1114/commits/4b22de09555202a480a92ad211254fd7c0642b15

sshropshire: Oh that's fine. We may want to revisit after still. This is the only merchant facing API that I'm aware of that takes `JSONObject` parameters. It would be nice to fully encapsulate our JSON serialization preference in case we decide to change to Gson one day.

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayCapabilities.kt
```diff
@@ -0,0 +1,32 @@
...
+            Class.forName(Wallet::class.java.name)
+
+            return configuration.isGooglePayEnabled && GoogleApiAvailability.getInstance()
+                .isGooglePlayServicesAvailable(context) ==
+                    ConnectionResult.SUCCESS
+        } catch (e: ClassNotFoundException) {
+            return false
+        } catch (e: NoClassDefFoundError) {
+            return false
+        }
```
tdchow, line 30:
```suggestion
        return try {
            Class.forName(Wallet::class.java.name)

            configuration.isGooglePayEnabled && GoogleApiAvailability.getInstance()
                .isGooglePlayServicesAvailable(context) ==
                    ConnectionResult.SUCCESS
        } catch (e: ClassNotFoundException) {
            false
        } catch (e: NoClassDefFoundError) {
            false
        }
```

nit - we can lift the return statement outside of the try/catch.

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayRequest.kt
```diff
@@ -0,0 +1,316 @@
...
+        return environment
+    }
+
+    /**
+     * Simple wrapper to assign given parameters to specified paymentMethod
+     *
+     * @param paymentMethodType The paymentMethod to add to
+     * @param parameters        Parameters to assign to the paymentMethod
+     */
+    fun setAllowedPaymentMethod(paymentMethodType: String, parameters: JSONObject) {
```
tdchow, line 103: Do these setter functions match the original Java interface? Would this be a breaking change for merchants?

chinghlinn: The function signature does not change for this function.

PR #1115, opened 2024-08-14
Convert DataCollector module classes to Kotlin

chinghlinn:
### Summary of changes

 - Convert DataCollectorInternalRequest and MagnesInternalClient to Kotlin.

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


DataCollector/src/main/java/com/braintreepayments/api/datacollector/DataCollectorInternalRequest.kt
```diff
@@ -0,0 +1,62 @@
+package com.braintreepayments.api.datacollector
+
+import androidx.annotation.RestrictTo
+import kotlin.math.min
+
+/**
+ * Used to configuration the PayPalDataCollector request
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class DataCollectorInternalRequest(val hasUserLocationConsent: Boolean) {
```
sshropshire, line 10: Is it possible to make this a data class?

chinghlinn: > Is it possible to make this a data class?

It would be confusing to make it a data class because a data class must have properties in it primary constructor and those primary constructor properties cannot have custom setters or getters. We have one property clientMetadataId that requires custom setter and can be declared in the body only.

chinghlinn: I was thinking the data class features(equals() and hashCode()) by adding all the properties to its primary constructor, but since we are not going to compare the objects and the class is not Parcelable, probably it's fine to make DataCollectorInternalRequest a data class. I updated it already.

sshropshire: Yeah we aren't using those InternalRequest objects as keys in a Map or anything so it should be fine. I've been leaning toward data classes more in our SDK because it seems like the value objects have too much responsibility at times e.g. `toJson()`. Ideally in our SDK we should have a set of pure data types and classes that leverage those data types to make HTTP requests.

chinghlinn: I agree, I also like the design of data class (a POJO with auto-generated equals) and the business logic could be implemented on the data consumer. When looking at the Google Pay tutorial(https://developers.google.com/pay/api/android/guides/tutorial), they also use JSONObject, JSONArray to wrap data(other modules are using those, for example, Card). I wonder maybe it's more suitable for payment api, sdk implementation, or the tutorial was written a while back.

sshropshire: Yeah they switched to JSONObject a while back. For our SDK I think it makes sense to have strongly typed objects as a way to interact with our API. One of the main responsibilities of our Core SDK is to provide an interface to merchants to make payments that doesn't require them to think about Networking/HTTP request logic.

DataCollector/src/main/java/com/braintreepayments/api/datacollector/DataCollector.kt
```diff
@@ -100,10 +99,11 @@ class DataCollector @VisibleForTesting internal constructor(
...
                 val deviceData = JSONObject()
                 try {
-                    val internalRequest = DataCollectorInternalRequest(request.hasUserLocationConsent)
-                        .setApplicationGuid(getPayPalInstallationGUID(context))
+                    val internalRequest = DataCollectorInternalRequest(request.hasUserLocationConsent).apply {
+                        applicationGuid = getPayPalInstallationGUID(context)
+                    }
                     if (request.riskCorrelationId != null) {
-                        internalRequest.setRiskCorrelationId(request.riskCorrelationId)
+                        internalRequest.clientMetadataId = request.riskCorrelationId
```
sshropshire, line 106: qq: Was the goal with renaming this property from `riskCorrelationId` to `clientMetadataId` to make things clearer?

DataCollector/src/main/java/com/braintreepayments/api/datacollector/DataCollectorInternalRequest.kt
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @property hasUserLocationConsent Whether the request has user location consent
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+data class DataCollectorInternalRequest(val hasUserLocationConsent: Boolean) {
+
+    /**
+     * @Property additionalData Additional data that should be associated with the data collection.
+     */
+    var additionalData: HashMap<String, String>? = null
```
sshropshire, line 16: Can these defaulted properties be moved into the constructor?

DataCollector/src/main/java/com/braintreepayments/api/datacollector/DataCollectorInternalRequest.kt
```diff
@@ -0,0 +1,46 @@
...
+    var clientMetadataId: String? = null
+        set(value) {
+            field = value?.let {
+                if (it.length > RISK_CORRELATION_ID_MAX_LENGTH) {
+                    it.substring(0, RISK_CORRELATION_ID_MAX_LENGTH)
+                } else {
+                    it.substring(0, it.length)
+                }
+            }
+        }
```
sshropshire, line 36: I'm wondering if this client metadata id sanitation should belong in another class? May be out of scope for this PR but worth bringing up.

sshropshire (approved):
LGTM overall. A few nit picks 🧶  but they're mostly style, architecture related.

DataCollector/src/main/java/com/braintreepayments/api/datacollector/DataCollectorInternalRequest.kt
chinghlinn: That could be the next major refactoring after all the Kotlin conversion.

DataCollector/src/main/java/com/braintreepayments/api/datacollector/DataCollector.kt
chinghlinn: In the original implementation of DataCollectorInternalRequest, the setter's parameter has a different name from the internal parameter's, however, the getter returns the internal parameter's name. Therefore, I used the internal parameter's name as the Kotlin property's name and kept the function caller's parameters the same.

public DataCollectorInternalRequest setRiskCorrelationId(@NonNull String riskCorrelationId) {
        this.clientMetadataId = riskCorrelationId.substring(0, Math.min(riskCorrelationId.length(), 32));

        return this;
    }
    
    String getClientMetadataId() {
        return clientMetadataId;
    }

sshropshire: Ok yeah that makes sense. There's a lot of discrepancy with those synonymous terms. I agree with `clientMetadataId` though it seems to work well here.

tdchow: Can we call out this change in the changelog? this would be a breaking change for merchants.

tdchow (approved):
looks great! we may want to update the changelog, calling out the breaking change.

PR #1025, opened 2024-06-05
Fix 3DS Black Screen Bug on Android 14

sshropshire:
### Summary of changes

* Fix issue that causes a black screen to display after successful 3DS validation
* Make `ThreeDSecureActivity` transparency explicit via Activity theme 

#### Before

https://github.com/braintree/braintree_android/assets/58225613/0e4e2967-f19b-49c5-bb2a-e0077cdcf738

#### After

https://github.com/braintree/braintree_android/assets/58225613/fa3dc5e6-3bf5-411d-9f41-60ca4dbe1864

### Checklist

- [x] Added a changelog entry 
~- [ ] Relevant test coverage~

@sshropshire

PR #1116, opened 2024-08-15
Support Android 15 Edge-to-Edge Display And Fix 3DS Activity Bug

sshropshire:
### Summary of changes

 - This PR adjusts the bounds of the Demo app to support edge-to-edge displays (the default in Android 15)
 - This PR also effectively hides the action bar title for the `ThreeDSecureActivity` that encapsulates interaction with the Cardinal SDK

### Checklist

- [x] Added a changelog entry

- @sshropshire

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## unreleased
 * BraintreeCore
   * Update `endpoint` syntax sent to FPTI for 3D Secure and Venmo flows
+* ThreeDSecure
+  * Update `ThreeDSecureActivity` theme attributes to prevent the Action Bar title from displaying and enforce transparency properly with AppCompat theme attributes
 * Breaking Changes
     * PayPal
         * Remove `appLinkEnabled` from `PayPalRequest` as Android app links are now required
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Braintree Android SDK Release Notes
 
 ## unreleased
+* All Modules
+  * Android 15 Support
+    * Upgrade `compileSdkVersion` and `targetSdkVersion` to API 35
 * BraintreeCore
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -82,11 +82,11 @@ allprojects {
 version '5.0.0-beta2-SNAPSHOT'
 group 'com.braintreepayments'
 ext {
-    compileSdkVersion = 34
+    compileSdkVersion = 35
```
jaxdesmarais, line 85: Do we need to call this out in the changelog at all?

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## unreleased
 * BraintreeCore
   * Update `endpoint` syntax sent to FPTI for 3D Secure and Venmo flows
+* ThreeDSecure
+  * Update `ThreeDSecureActivity` theme attributes to prevent the Action Bar title from displaying and enforce transparency properly with AppCompat theme attributes
```
jaxdesmarais, line 7: Just curious if we had an inbound about this that we should link here or if this was just a general bug we noticed?

jaxdesmarais (approved):

build.gradle
sshropshire: Oh yeah def I zoned out on the test portion. I'll add it in.

CHANGELOG.md
sshropshire: There was mention of it somewhere I think the bug was closed but resurfaced in Android 15. I'll try and find it.

sshropshire: This PR is the only thing I could find: https://github.com/braintree/braintree_android/pull/1025.

sshropshire: I don't think an actual issue was filed, but I could be wrong.

jaxdesmarais: Ah cool, if we don't have one that's fine, was just curious more than anything else.

tdchow (approved):
🎉

PR #1117, opened 2024-08-16
[Tech Debt] Remove BraintreeHttpClient Convenience Methods

sshropshire:
### Summary of changes

 - This PR removes http-verb-named convenience methods on `BraintreeHttpClient` and replaces them with two methods:
   - `sendRequest()`
   - `sendRequestSync()`
 - Both new methods take a `BraintreeHttpRequest` object to hold HTTP verb, target path, data, etc. parameters
 - This will make code more DRY and provide a foundation for a clean refactor of `ConfigurationLoader` and Http multithreading logic in the Core SDK

### Checklist

 - [x] Relevant test coverage

- @sshropshire

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
@@ -113,12 +113,13 @@ internal class AnalyticsClient(
                         IntegrationType.fromString(integration)
                     )
                     val analyticsRequest = createFPTIPayload(authorization, eventBlobs, metadata)
-                    httpClient.post(
-                        FPTI_ANALYTICS_URL,
-                        analyticsRequest.toString(),
-                        configuration,
-                        authorization
+
+                    val request = BraintreeHttpRequest(
+                        method = "POST",
+                        path = FPTI_ANALYTICS_URL,
+                        data = analyticsRequest.toString()
                     )
+                    httpClient.sendRequestSync(request, configuration, authorization)
                     analyticsEventBlobDao.deleteEventBlobs(eventBlobs)
                 }
                 ListenableWorker.Result.success()
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
@@ -114,7 +114,7 @@ internal class AnalyticsClient(
                     )
                     val analyticsRequest = createFPTIPayload(authorization, eventBlobs, metadata)
 
-                    val request = BraintreeHttpRequest(
+                    val request = InternalHttpRequest(
                         method = "POST",
                         path = FPTI_ANALYTICS_URL,
                         data = analyticsRequest.toString()
@@ -165,7 +165,7 @@ internal class AnalyticsClient(
         val eventBlobs = listOf(AnalyticsEventBlob(eventJSON))
         try {
             val analyticsRequest = createFPTIPayload(authorization, eventBlobs, metadata)
-            val request = BraintreeHttpRequest(
+            val request = InternalHttpRequest(
                 method = "POST",
                 path = FPTI_ANALYTICS_URL,
                 data = analyticsRequest.toString()
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
@@ -7,6 +7,7 @@ import androidx.work.ExistingWorkPolicy
 import androidx.work.ListenableWorker
 import androidx.work.OneTimeWorkRequest
 import androidx.work.WorkManager
+import com.braintreepayments.api.sharedutils.HttpMethod
 import org.json.JSONArray
 import org.json.JSONException
 import org.json.JSONObject
@@ -115,7 +116,7 @@ internal class AnalyticsClient(
                     val analyticsRequest = createFPTIPayload(authorization, eventBlobs, metadata)
 
                     val request = InternalHttpRequest(
-                        method = "POST",
+                        method = HttpMethod.POST,
                         path = FPTI_ANALYTICS_URL,
                         data = analyticsRequest.toString()
                     )
@@ -166,7 +167,7 @@ internal class AnalyticsClient(
         try {
             val analyticsRequest = createFPTIPayload(authorization, eventBlobs, metadata)
             val request = InternalHttpRequest(
-                method = "POST",
+                method = HttpMethod.POST,
                 path = FPTI_ANALYTICS_URL,
                 data = analyticsRequest.toString()
             )
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
@@ -164,13 +164,12 @@ internal class AnalyticsClient(
         val eventBlobs = listOf(AnalyticsEventBlob(eventJSON))
         try {
             val analyticsRequest = createFPTIPayload(authorization, eventBlobs, metadata)
-            httpClient.post(
+            val request = BraintreeHttpRequest(
+                method = "POST",
                 path = FPTI_ANALYTICS_URL,
-                data = analyticsRequest.toString(),
-                configuration = null,
-                authorization = authorization,
-                callback = null
+                data = analyticsRequest.toString()
             )
+            httpClient.sendRequest(request, authorization = authorization, callback = null)
         } catch (e: JSONException) { /* ignored */
         }
     }
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
@@ -182,7 +182,11 @@ class BraintreeClient @VisibleForTesting internal constructor(
         }
         getConfiguration { configuration, configError ->
             if (configuration != null) {
-                httpClient.get(url, configuration, authorization) { response, httpError ->
+                val request = BraintreeHttpRequest(
+                    method = "GET",
+                    path = url,
+                )
+                httpClient.sendRequest(request, configuration, authorization) { response, httpError ->
                     response?.let {
                         try {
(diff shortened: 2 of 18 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
```diff
@@ -113,12 +113,13 @@ internal class AnalyticsClient(
                         IntegrationType.fromString(integration)
                     )
                     val analyticsRequest = createFPTIPayload(authorization, eventBlobs, metadata)
-                    httpClient.post(
-                        FPTI_ANALYTICS_URL,
-                        analyticsRequest.toString(),
-                        configuration,
-                        authorization
+
+                    val request = BraintreeHttpRequest(
```
jaxdesmarais, line 117: nit but do we need to call this `BraintreeHttpRequest` or can we just use `HttpRequest`? It feels a bit odd especially here since the URL is technically a PayPal URL, so we aren't actually building a BT request: https://github.com/braintree/braintree_android/blob/9bcdb877cfc22c07ce0128a071a6691f1c76509b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt#L247

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
```diff
@@ -176,28 +176,8 @@ class BraintreeClient @VisibleForTesting internal constructor(
...
-                        }
-                    } ?: httpError?.let { error ->
-                        responseCallback.onResult(null, error)
-                    }
-                }
-            } else {
-                responseCallback.onResult(null, configError)
-            }
-        }
+        val request = BraintreeHttpRequest(method = "GET", path = url)
```
jaxdesmarais, line 179: Can we extract the method names into an enum?

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeHttpClient.kt
```diff
@@ -18,164 +16,99 @@ internal class BraintreeHttpClient(
...
     /**
-     * Make a HTTP GET request to Braintree using the base url, path and authorization provided.
-     * If the path is a full url, it will be used instead of the previously provided url.
-     * @param path The path or url to request from the server via GET
-     * @param configuration configuration for the Braintree Android SDK.
-     * @param authorization
-     * @param callback [NetworkResponseCallback]
+     * Make an asynchronous Braintree authenticated HTTP request.
+     *
+     * @param request the braintree http request.
```
jaxdesmarais, line 21: Update since this needs to handle requests being built for BT and PP APIs.
```suggestion
     * @param request the http request.
```

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeHttpClient.kt
```diff
@@ -18,164 +16,99 @@ internal class BraintreeHttpClient(
...
+        configuration: Configuration? = null,
+        authorization: Authorization? = null,
+        callback: NetworkResponseCallback?
+    ) {
+        try {
+            val httpRequest = buildHttpRequest(request, configuration, authorization)
+            httpClient.sendRequest(httpRequest, request.retryStrategy, callback)
+        } catch (e: Exception) {
+            // forward errors
+            callback?.onResult(null, e)
```
jaxdesmarais, line 36: Take it or leave it but not sure if we want to make these more verbose as we move to Kotlin, not sure what the best practice is here:
```suggestion
        } catch (exception: Exception) {
            // forward errors
            callback?.onResult(null, exception)
```

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
sshropshire: So we do already have an `HttpRequest`. We can iterate on the best name, but the main purpose of `BraintreeHttpRequest` is to allow feature clients to make HTTP-like calls without having to know how the requests will be authenticated. `BraintreeGatewayRequest` could work too or something like it.

sshropshire: > It feels a bit odd especially here since the URL is technically a PayPal URL, so we aren't actually building a BT request

This is a great callout, but may also be a separate issue. We're still authenticating FPTI calls with BT since those requests used to get routed to Arachne. Does it still work this way on iOS?

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
sshropshire: Yeah def. I can add that part in.

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeHttpClient.kt
sshropshire: I think this still holds for `BraintreeHttpClient`. For PP requests we should probably use a different HttpClient that doesn't append authentication headers to the request.

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeHttpClient.kt
sshropshire: Yeah we could. We sometimes go with `error` I can put that in.

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeHttpRequest.kt
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * @suppress
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+data class BraintreeHttpRequest(
+    val method: String,
+    val path: String,
+    val data: String? = null,
+    val additionalHeaders: Map<String, String> = emptyMap(),
+    @RetryStrategy val retryStrategy: Int = HttpClient.NO_RETRY
```
tdchow, line 16: Not sure if it's in scope for this PR - should we update this to an enum type?

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
jaxdesmarais: Maybe we could use `InternalHttpRequest` to capture both PayPal and BT HTTP requests?

> We're still authenticating FPTI calls with BT since those requests used to get routed to Arachne. Does it still work this way on iOS?

We shouldn't need to do this, we don't do anything with auth to the gateway for FPTI requests.

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeHttpRequest.kt
sshropshire: Yeah actually I was thinking of removing this enum all together in a follow up PR. The `ConfigurationLoader` is the only class that needs retry logic, so the end-goal is to fully encapsulate retries within the config loader.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
sshropshire: Internal could work. The main reason it's `BraintreeHttpRequest` though is because they get sent through `BraintreeHttpClient`. The `BraintreeHttp` prefix is mostly to highlight the compound BT/HTTP domain that the request and the client both belong to.

Also we are still authenticated FPTI calls [on iOS](https://github.com/braintree/braintree_ios/blob/main/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift#L82). I think it's something that we haven't discussed since moving to FPTI. The requirements changed for Analytics as a whole, but on both platforms we still use BT patterns to communicate with that endpoint.

jaxdesmarais: I think I am confused by what you mean by "authenticating FPTI". We have logic built around passing in the custom URL which in this case is `https://api.paypal.com/v1/tracking/batch/events`, but we do not hit the BT GW before FPTI, PayPal APIs have their own authentication mechanism separate from Braintree.

On iOS we use 1 pattern for all requests irrespective of BT vs PP APIs when building the request but initialize the `BTHTTP` class in a different way depending on the API we are hitting.

sshropshire: Oh ok there's an if statement baked [into BTHTTP here](https://github.com/braintree/braintree_ios/blob/c17642dad8f8a45f1b2d51c0c762873d35e17f6e/Sources/BraintreeCore/BTHTTP.swift#L190). That's my bad I hadn't looked into it deeper.

For us `BraintreeClient` fetches configuration ([similar to APIClient](https://github.com/braintree/braintree_ios/blob/main/Sources/BraintreeCore/BTAPIClient.swift) on iOS). I think we've found a similar piece of technical debt between the two platforms. Does the name for `BTHTTP` make sense if it's also making requests to PayPal endpoints?

I think we can find architectural parity between iOS and Android for this. It might be out of scope for this PR, but it's definitely something we should consider ironing out.

jaxdesmarais: Nah no worries, it's complicated for sure. The `BT` prefix only exists because Obj-C requires a 2-3 letter prefix for all classes, but it's not BT specific at all. In the future now that the entire SDK is Swift we can drop the prefix and add a specific annotation just for the Obj-C extension, but it does not necessary denote BT specifics, just a requirement for Obj-C interoperability.

Agree there is some code smell that we can handle in a separate PR. For this PR maybe we can run with `Internal` for now?

sshropshire: Yeah that makes sense for now. I'm down to go with Internal.

tdchow (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeHttpClient.kt
sshropshire: I'll commit this one as well.

jaxdesmarais (approved):

PR #1118, opened 2024-08-20
Support Android 15: Bump JDK Version to 17 and Fix Core Integration Tests

sshropshire:
### Summary of changes

 - This PR bumps the JDK Version to 17 (along with the Gradle wrapper version) to resolve compilation issues related to Braintree Core SDK's integration tests
 - This PR also converts the Braintree Core SDK integration tests to Kotlin to fix compilation errors caused by Kotlin/Java interop

### Checklist

 - [x] Added a changelog entry

- @sshropshire


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalAccount.kt
```diff
@@ -18,7 +19,8 @@ import org.json.JSONObject
  * @property paymentType Payment type from original PayPal request. Either "billing-agreement" or
  * "single-payment"
  */
-internal data class PayPalAccount(
```
tdchow, line 21: Curious - Was there an access issue in the tests when converting to Kotlin?

tdchow (approved):
Looks good! Are there other changes we want to bundle in the `v__android_15` branch? Or is it safe to merge this into `main`?

gradle/libs.versions.toml
```diff
@@ -1,5 +1,5 @@
 [versions]
-gradle = "8.1.4"
+gradle = "8.5.2"
```
jaxdesmarais, line 2: Do we need to call out this gradle bump at all?

jaxdesmarais (approved):

sshropshire:
> Looks good! Are there other changes we want to bundle in the v__android_15 branch? Or is it safe to merge this into main?

@tdchow Thanks! We may be good to merge but I do want to double check to make sure.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalAccount.kt
sshropshire: Yeah there's an integration test in Core that uses PayPalAccount for some reason. We may be able to audit right before we go GA to make sure.

gradle/libs.versions.toml
sshropshire: We could I noticed we've been mentioning it recently. I'll add it to the CHANGELOG.

sshropshire:
There are a couple other modules that need updates. I'll merge this in and add those in a follow up PR.

PR #1120, opened 2024-08-20
ThreeDS Module Kotlin Converesion

tdchow:
### Summary of changes

 - Convert `ThreeDSecureInfo`, `ThreeDSecureLookup`, `ThreeDSecureParams`, `ThreeDSecureActivity`, `ThreeDSecureActivityResultContract`, `ThreeDSecurePaymentAuthResult`, and `ThreeDSecureShippingMethod` to Kotlin

### Checklist

 - [X] Added a changelog entry
 - [ ] Relevant test coverage


sshropshire (approved):

CHANGELOG.md
```diff
@@ -15,6 +15,12 @@
         * Change `PayPalLineItemKind` to an enum
         * Rename `PayPalLineItemKind.KIND_CREDIT` to `CREDIT`
         * Rename `PayPalLineItemKind.KIND_DEBIT` to `DEBIT`
+        * Update `ThreeDSSecureInfo.isLiabilityShifted()` to `ThreeDSSecureInfo.getLiabilityShifted()`
+        * Update `ThreeDSSecureInfo.isLiabilityShiftPossible()` to `ThreeDSSecureInfo.getLiabilityShiftPossible()`
+        * Update `ThreeDSSecureInfo.wasVerified()` to `ThreeDSSecureInfo.getWasVerified()`
```
sarahkoop, line 20: I might have missed the context on this - are we changing these for better Kotlin interop or consistency across modules?

tdchow: Unfortunately, to utilize the `@Parcelize` annotation, all properties must be declared in the primary constructor. We'd need to add a specific getter for java integrations.

Here's a related thread on a previous PR: https://github.com/braintree/braintree_android/pull/1088#discussion_r1698935948

sarahkoop: Got it - thanks 👍

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureLookup.kt
```diff
@@ -0,0 +1,59 @@
...
+         * @throws JSONException when parsing fails.
+         */
+        @JvmStatic
+        @Throws(JSONException::class)
+        fun fromJson(jsonString: String): ThreeDSecureLookup {
+            val json = JSONObject(jsonString)
+            return ThreeDSecureLookup(
+                acsUrl = Json.optString(json, ACS_URL_KEY, null),
+                md = Json.optString(json, MD_KEY, ""),
+                termUrl = Json.optString(json, TERM_URL_KEY, ""),
```
sarahkoop, line 44: I notice these changing from get to opt - are we ok with losing the JSONException when they are missing? I don't know if they ever should be and assume an error would surface down the line instead?

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureClient.java
```diff
@@ -256,9 +256,9 @@ public void tokenize(ThreeDSecurePaymentAuthResult paymentAuthResult,
         if (threeDSecureError != null) {
             callbackTokenizeFailure(callback, new ThreeDSecureResult.Failure(threeDSecureError, null));
         } else {
-            ThreeDSecureParams threeDSecureParams = paymentAuthResult.getThreeSecureResult();
```
sarahkoop, line 259: Is there going to be a separate PR for converting the external classes - this one and the `ThreeDSecureRequest` and `Launcher`?)

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureLookup.kt
tdchow: I updated from get to opt to be consistent with all the properties. Hmm yeah something would probably fail further in the flow, but since these fields are all non-null, I think failing fast here makes the most sense. I'll update these all from opt to get.

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureClient.java
tdchow: Yep! I'll have a few more follow-up PRs for the remaining 3DS classes.

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureLookup.kt
tdchow: I ended up leaving optString for `acsUrl` since it's the only nullable field. The others have been updated to use `getString`.

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecurePaymentAuthResult.kt
```diff
@@ -0,0 +1,10 @@
+package com.braintreepayments.api.threedsecure
+
+import com.cardinalcommerce.cardinalmobilesdk.models.ValidateResponse
+
+data class ThreeDSecurePaymentAuthResult internal constructor(
```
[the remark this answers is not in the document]
tdchow: Good catch, let me add the kdocs.

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureLookup.kt
tdchow: It looks like the previous behavior is that the pareq value is set to an empty string if it's missing from the response. Do you know if the `pareq` value is required for the 3DS flow? If it is, are we still good with throwing an exception if it's missing?

Unit test that checks that `pareq` is an empty string when it's missing from the response.
```java
    @Test
    public void fromJson_whenPareqMissing_parsesCorrectly() throws JSONException {
        JSONObject lookupV2MissingPareq = new JSONObject(
                Fixtures.THREE_D_SECURE_V2_LOOKUP_RESPONSE_MISSING_PAREQ).getJSONObject("lookup");
        ThreeDSecureLookup sut = ThreeDSecureLookup.fromJson(lookupV2MissingPareq.toString());
        assertEquals("", sut.getPareq());
    }
```

tdchow: I ended up keeping the existing behavior. If the `pareq` value from the response is missing or null, an empty string is set.

tdchow: Going to merge the PR, but open to updating this is folks feel that we should handle the null case differently.

PR #1122, opened 2024-08-21
Convert LocalPayment classes to Kotlin

ashivashankars:
### Summary of changes

Convert LocalPaymentClientTest, LocalPaymentApi, LocalPaymentAuthRequestParams, LocalPaymentAuthResultInfo, LocalPaymentClient, LocalPaymentRequest to Kotlin 

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


LocalPayment/src/androidTest/java/com/braintreepayments/api/localpayment/LocalPaymentClientTest.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.util.concurrent.CountDownLatch
+
+@RunWith(AndroidJUnit4ClassRunner::class)
+class LocalPaymentClientTest {
+    private var countDownLatch: CountDownLatch? = null
+    private var braintreeClient: BraintreeClient? = null
```
tdchow, line 17: Optional - these `vars` could be defined as `lateinit` non null. This way we don't need to handle the null case in the tests.

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentApi.kt
```diff
@@ -0,0 +1,84 @@
...
+        callback: LocalPaymentInternalAuthRequestCallback
+    ) {
+        val returnUrl =
+            braintreeClient.getReturnUrlScheme() + "://" + LocalPaymentClient.LOCAL_PAYMENT_SUCCESS
+        val cancel =
+            braintreeClient.getReturnUrlScheme() + "://" + LocalPaymentClient.LOCAL_PAYMENT_CANCEL
+
+        val url = "/v1/local_payments/create"
+        braintreeClient.sendPOST(url = url, data = request.build(returnUrl, cancel),
+            responseCallback = { responseBody: String?, httpError: Exception? ->
```
tdchow, line 20: We should be able to move the lambda callback outside of the `sendPost()` parenthesis since the callback parameter is the last in the list: https://kotlinlang.org/docs/lambdas.html#passing-trailing-lambdas

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentApi.kt
```diff
@@ -0,0 +1,84 @@
...
+                        callback.onLocalPaymentInternalAuthResult(null, e)
+                    }
+                } else {
+                    callback.onLocalPaymentInternalAuthResult(null, httpError)
+                }
+            })
+    }
+
+    fun tokenize(
+        merchantAccountId: String?, responseString: String?, clientMetadataID: String?,
```
tdchow, line 42: I'm not sure if detekt will flag having multiple params on a single line. It may help with readability to have them all on separate lines.

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentApi.kt
```diff
@@ -0,0 +1,84 @@
...
+                .put("source", "client")
+                .put("integration", braintreeClient.integrationType.stringValue)
+                .put("sessionId", braintreeClient.sessionId)
+            payload.put("_meta", metaData)
+
+            val url = "/v1/payment_methods/paypal_accounts"
+            braintreeClient.sendPOST(
+                url = url,
+                data = payload.toString(),
+                responseCallback = { responseBody: String?, httpError: Exception? ->
```
tdchow, line 68: Same here, can we make this a trailing lambda?

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentAuthRequestParams.kt
```diff
@@ -0,0 +1,24 @@
+package com.braintreepayments.api.localpayment
+
+import com.braintreepayments.api.BrowserSwitchOptions
+
+/**
+ * Local payment result information.
+ */
+class LocalPaymentAuthRequestParams internal constructor(
+    /**
+     * @return The original request used to create the local payment transaction.
```
tdchow, line 10: Can we move the doc for each property into the class doc, using the `@property` annotation in the comment? 

example: https://github.com/braintree/braintree_android/blob/474febc875cac373c3d6823b59903bb8e674dd41/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt#L55

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentAuthResultInfo.kt
```diff
@@ -0,0 +1,13 @@
+package com.braintreepayments.api.localpayment
+
+import com.braintreepayments.api.BrowserSwitchFinalResult
+
+/**
+ * Details of a [LocalPaymentAuthResult.Success]
+ */
+class LocalPaymentAuthResultInfo internal constructor
```
tdchow, line 8: I believe detekt will enforce having the open parenthesis on this line. 

You can run the `./ci lint` command locally to run all of the lint checks.

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentClient.kt
```diff
@@ -0,0 +1,258 @@
...
+     * @param returnUrlScheme a custom return url to use for browser and app switching
+     */
+    constructor(
+        context: Context,
+        authorization: String,
+        returnUrlScheme: String?
+    ) : this(BraintreeClient(context, authorization, returnUrlScheme))
+
+    @VisibleForTesting
+    internal constructor(braintreeClient: BraintreeClient) : this(
```
tdchow, line 48: Can we collapse this constructor into the primary constructor by setting default values for `dataCollector` and `localPaymentApi`?

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentClient.kt
```diff
@@ -0,0 +1,258 @@
...
+                                "Local payments are not enabled for this " +
+                                        "merchant."
+                            ),
+                            callback
+                        )
+                        return@getConfiguration
+                    }
+
+                    localPaymentApi.createPaymentMethod(
+                        request!!
```
tdchow, line 98: Can we handle the null case here? We should try to avoid crashing the SDK.

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentClient.kt
```diff
@@ -0,0 +1,258 @@
...
+                        }
+                    }
+                } else if (error != null) {
+                    authRequestFailure(error, callback)
+                }
+            }
+        }
+    }
+
+    fun buildBrowserSwitchOptions(
```
tdchow, line 127: Does this function need to be public? It seems to be package-private in the java file. I'm not sure where this is being called, but if it's only called in the same module, we could mark it as `internal`. If other modules are invoking it, we can add the `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)` annotation to prevent merchants from calling it.

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentLauncher.kt
```diff
@@ -33,7 +33,7 @@ class LocalPaymentLauncher internal constructor(private val browserSwitchClient:
     ): LocalPaymentPendingRequest {
         val params = localPaymentAuthRequest.requestParams
         val browserSwitchPendingRequest =
-            browserSwitchClient.start(activity, params.browserSwitchOptions)
+            browserSwitchClient.start(activity, params.browserSwitchOptions!!)
```
tdchow, line 36: Can we handle the null case here?

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentRequest.kt
```diff
@@ -0,0 +1,171 @@
...
+import com.braintreepayments.api.core.PostalAddress
+import org.json.JSONException
+import org.json.JSONObject
+
+/**
+ * Builder used to construct an local payment request.
+ */
+class LocalPaymentRequest
+/**
+ * @param hasUserLocationConsent is an optional parameter that informs the SDK
```
tdchow, line 12: Can we combine this comment block with the main class kdoc?

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentRequest.kt
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * @see [User Data policies for the Google Play Developer Program ](https://support.google.com/googleplay/android-developer/answer/10144311.personal-sensitive)
+ *
+ * @see [Examples of prominent in-app disclosures](https://support.google.com/googleplay/android-developer/answer/9799150?hl=en.Prominent%20in-app%20disclosure)
+ */(private val hasUserLocationConsent: Boolean) {
+    /**
+     * @param address Optional - The address of the customer. An error will occur if this address is not valid.
+     */
+    @JvmField
+    var address: PostalAddress? = null
```
tdchow, line 25: Could we move these fields into the primary constructor? This way we'd be able to make it a data class.

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentRequest.kt
```diff
@@ -0,0 +1,171 @@
...
+     */
+    var isShippingAddressRequired: Boolean = false
+
+    /**
+     * @param surname Optional - Surname (last name) of the customer.
+     */
+    @JvmField
+    var surname: String? = null
+
+    fun hasUserLocationConsent(): Boolean {
```
tdchow, line 105: We should be able to remove this getter function. It'll be a breaking change that we can call out in the changelog. 

similar change: https://github.com/braintree/braintree_android/blob/474febc875cac373c3d6823b59903bb8e674dd41/CHANGELOG.md?plain=1#L13

LocalPayment/src/test/java/com/braintreepayments/api/localpayment/LocalPaymentApiUnitTest.kt
```diff
@@ -0,0 +1,309 @@
...
+import org.junit.runner.RunWith
+import org.mockito.ArgumentCaptor
+import org.mockito.ArgumentMatchers
+import org.mockito.Mockito
+import org.robolectric.RobolectricTestRunner
+import org.skyscreamer.jsonassert.JSONAssert
+
+@RunWith(RobolectricTestRunner::class)
+class LocalPaymentApiUnitTest {
+    private var localPaymentInternalAuthRequestCallback: LocalPaymentInternalAuthRequestCallback? =
```
tdchow, line 26: Nit - Similar change here, we could define this var as `lateinit` and make it non-null.

LocalPayment/src/test/java/com/braintreepayments/api/localpayment/LocalPaymentClientUnitTest.kt
```diff
@@ -0,0 +1,958 @@
...
+
+@RunWith(RobolectricTestRunner::class)
+class LocalPaymentClientUnitTest {
+    private var activity: FragmentActivity? = null
+    private var localPaymentAuthCallback: LocalPaymentAuthCallback? = null
+    private var localPaymentTokenizeCallback: LocalPaymentTokenizeCallback? = null
+    private var braintreeClient: BraintreeClient? = null
+    private var dataCollector: DataCollector? = null
+    private var localPaymentApi: LocalPaymentApi? = null
+    private var localPaymentAuthRequestParams: LocalPaymentAuthRequestParams? = null
```
tdchow, line 44: Same here - we could mark these as `lateinit` and non-null.

Demo/src/main/java/com/braintreepayments/demo/LocalPaymentFragment.java
```diff
@@ -72,16 +72,16 @@ public void launchIdeal(View v) {
...
-        request.setGivenName("Test");
-        request.setSurname("Buyer");
+        request.paymentType = "ideal";
+        request.amount = "1.10";
+        request.address = address;
+        request.phone = "207215300";
+        request.email = "android-test-buyer@paypal.com";
+        request.givenName = "Test";
+        request.surname = "Buyer";
         request.setShippingAddressRequired(true);
```
sarahkoop, line 82: Are we planning to keep support for the `.setProperty` syntax for Java merchants or remove it with v5? I think we should have consistency across all modules/properties but not sure what was done for other conversions (cc @sshropshire @tdchow @chinghlinn)

tdchow: Yep, I agree. We should keep the setter functions for java integrations. I think we'll need to move all the properties in the LocalPaymentRequest into the constructor for the setters to be generated.

tdchow: Actually, I think the getter/setter functions aren't generated due to the `@JvmField` annotations on each of the properties.

sshropshire: Yeah the `set` prefix is standard for Java. It's probably safe to assume our Java merchants are comfortable with the verbosity if they haven't migrated to Kotlin by now. If we remove the `@JvmField` annotations we can leverage the default Kotlin / Java interop behavior and reduce our maintenance costs with fewer lines-of-code.

ashivashankars: I agree with removing the @JvmField annotations. Leveraging Kotlin’s default interop behavior will simplify our code and reduce maintenance by eliminating unnecessary boilerplate. I have checked all relevant directories, and these annotations are not used elsewhere.

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentClient.kt
```diff
@@ -0,0 +1,258 @@
...
+
+        val deepLinkUri: Uri = browserSwitchResult.returnUrl
+        if (deepLinkUri == null) {
+            tokenizeFailure(
+                BraintreeException(
+                    "LocalPayment encountered an error, return URL is " +
+                            "invalid."
+                ),
+                callback
+            )
```
sshropshire, line 201:
```suggestion
            val errorMessage = "LocalPayment encountered an error, return URL is invalid."
            tokenizeFailure(BraintreeException(errorMessage), callback)
```
nit pick: I feel like this string could read more cleanly as a one liner.

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentRequest.kt
sshropshire: +1 kdoc property annotations are great for readability.

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentApi.kt
```diff
@@ -0,0 +1,88 @@
...
+
+internal class LocalPaymentApi(private val braintreeClient: BraintreeClient) {
+    fun createPaymentMethod(
+        request: LocalPaymentRequest,
+        callback: LocalPaymentInternalAuthRequestCallback
+    ) {
+        val returnUrl =
+            braintreeClient.getReturnUrlScheme() + "://" + LocalPaymentClient.LOCAL_PAYMENT_SUCCESS
+        val cancel =
+            braintreeClient.getReturnUrlScheme() + "://" + LocalPaymentClient.LOCAL_PAYMENT_CANCEL
```
sarahkoop, line 16: Can we use the Kotlin $ string interpolation here?

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentAuthRequestParams.kt
```diff
@@ -0,0 +1,19 @@
...
+ * @property approvalUrl used for payment approval.
+ * @property paymentId of the local payment after creation.
+ */
+
+class LocalPaymentAuthRequestParams internal constructor(
+    val request: LocalPaymentRequest,
+    val approvalUrl: String,
+    val paymentId: String
+) {
+    var browserSwitchOptions: BrowserSwitchOptions? = null
```
sarahkoop, line 18: I think this class can be a data class and `browserSwitchOptions` should be restricted to library scope or private if possible

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentResult.kt
```diff
@@ -18,5 +18,5 @@ sealed class LocalPaymentResult {
     /**
      * The user canceled the local payment flow.
      */
-    object Cancel : LocalPaymentResult()
+    data object Cancel : LocalPaymentResult()
```
sarahkoop, line 21: Is this required to be a data object? What does it change?

ashivashankars: Studio suggested converting the sealed sub-object to a data object, but this change neither improved functionality nor caused any regression issues. Since data object is not a valid Kotlin construct, I will revert the sub-object back to an object.

BraintreeCore/src/main/java/com/braintreepayments/api/core/PostalAddress.kt
```diff
@@ -9,18 +9,16 @@ import android.os.Parcelable
...
-    var phoneNumber: String?,
-    var streetAddress: String?,
-    var extendedAddress: String?,
-    var locality: String?,
-    var region: String?,
-    var postalCode: String?,
-    var sortingCode: String?,
-    var countryCodeAlpha2: String?) : Parcelable {
-
-    constructor() : this(null, null, null, null, null, null, null, null, null)
```
sshropshire, line 22: It'll be great if we don't need this constructor any more. Usually we include this one when we want Java interop without the `@JvmOverloads` annotation.

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentApi.kt
sshropshire: +1. The Kotlin way can help to improve readability.

sshropshire (approved):

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentAuthResult.kt
```diff
@@ -22,5 +22,5 @@ sealed class LocalPaymentAuthResult {
      * browser to cancel the payment flow, or returns to the app without completing the
      * authentication flow.
      */
-    object NoResult : LocalPaymentAuthResult()
+    data object NoResult : LocalPaymentAuthResult()
```
sarahkoop, line 25: Same change here let's remove `data` if we aren't using it across other result objects

BraintreeCore/src/main/java/com/braintreepayments/api/core/ConfigurationException.kt
```diff
@@ -5,7 +5,8 @@ import androidx.annotation.RestrictTo
 /**
  * Error class thrown when a configuration value is invalid
  */
-class ConfigurationException @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP) @JvmOverloads constructor(
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class ConfigurationException @JvmOverloads constructor(
```
tdchow, line 9: I believe merchants reference this class. We should move the RestrictTo annotation to the constructor.

sarahkoop: Updated!

PR #1124, opened 2024-08-22
American express kotlin conversion

ashivashankars:
### Summary of changes

Convert  AmericanExpressClient, AmericanExpressRewardsBalance, AmericanExpressRewardsBalanceUnitTest.
cleaned up unit test cases on  AmericanExpressClientUnitTest

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


AmericanExpress/src/main/java/com/braintreepayments/api/americanexpress/AmericanExpressRewardsBalance.kt
```diff
@@ -0,0 +1,144 @@
...
+import android.os.Parcelable
+import androidx.annotation.RestrictTo
+import com.braintreepayments.api.sharedutils.Json
+import org.json.JSONException
+import org.json.JSONObject
+
+/**
+ * Class to parse and contain American Express rewards balance
+ */
+class AmericanExpressRewardsBalance : Parcelable {
```
sarahkoop, line 13: Can we use the @Parcelize annotation here instead? I think that should remove the need to define the parcelable implementation in this class

AmericanExpress/src/main/java/com/braintreepayments/api/americanexpress/AmericanExpressClient.kt
```diff
@@ -0,0 +1,87 @@
...
+import androidx.annotation.RestrictTo
+import com.braintreepayments.api.core.ApiClient.Companion.versionedPath
+import com.braintreepayments.api.core.BraintreeClient
+import org.json.JSONException
+
+/**
+ * Used to integrate with Braintree's American Express API
+ */
+class AmericanExpressClient {
+    private val braintreeClient: BraintreeClient
```
tdchow, line 14: Could we move this property into the primary constructor of the class? This way we could remove this constructor:

```kotlin
    constructor(braintreeClient: BraintreeClient) {
        this.braintreeClient = braintreeClient
    }
```

AmericanExpress/src/main/java/com/braintreepayments/api/americanexpress/AmericanExpressRewardsBalance.kt
```diff
@@ -0,0 +1,144 @@
...
+         *
+         * @param jsonString The json response from the Braintree Gateway American Express rewards balance route.
+         * @return The [AmericanExpressRewardsBalance] with rewards balance data.
+         * @throws JSONException when parsing fails.
+         */
+        @JvmStatic
+        @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+        @Throws(JSONException::class)
+        fun fromJson(jsonString: String?): AmericanExpressRewardsBalance {
+            val json = JSONObject(jsonString!!)
```
tdchow, line 112: Can we handle the null case here?

AmericanExpress/src/main/java/com/braintreepayments/api/americanexpress/AmericanExpressRewardsBalance.kt
sshropshire: +1

AmericanExpress/src/main/java/com/braintreepayments/api/americanexpress/AmericanExpressRewardsBalance.kt
sshropshire: I'm noticing the original code may not have handled `null` either. Let's return `null` here when `jsonString` is `null`.

sshropshire (approved):
LGTM. A single constructor on `AmericanExpressClient` and `@Parcelize` on `AmericanExpressRewardsBalance` should be the only changes left.

tdchow (approved):

PR #1126, opened 2024-08-22
3DS Module Conversion Part 2

tdchow:
### Summary of changes

 - Convert ThreeDSecureAccountType, ThreeDSecureAdditionalInformation, ThreeDSecureRenderType, ThreeDSecureRequestedExemptionType, ThreeDSecureUiType, ThreeDSecurePostalAddress, and ThreeDSecureRequest to Kotlin
 
There are a few more classes to convert following this PR.

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureRequestedExemptionType.kt
```diff
@@ -0,0 +1,8 @@
+package com.braintreepayments.api.threedsecure
+
+enum class ThreeDSecureRequestedExemptionType(val stringValue: String) {
```
chinghlinn, line 3: I wonder if property `name` for the string value is compatible with the legacy code, maybe there is a chance we don't need addition values.

tdchow: I think the only place where the stringValue is used is where the enum type is converted to json, so there shouldn't be any legacy references to the actual string values. I can double-check to see if we can reduce the visibility modifier of the `stringValue` as merchants wouldn't need to reference it.

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureRenderType.kt
```diff
@@ -0,0 +1,9 @@
+package com.braintreepayments.api.threedsecure
+
+enum class ThreeDSecureRenderType {
```
sarahkoop, line 3: Were there doc strings for this?

sarahkoop (approved):

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureRenderType.kt
tdchow: There was not :( It looks like these values map to types in the Cardinal SDK. I can do some digging and see if I can come up with some docs for the enum.

sarahkoop: If they didn't exist before that can be a later TODO 👍

tdchow: Sounds good. After a quick search, I wasn't able to find Cardinal documentation on these :(

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/CardinalClient.java
```diff
@@ -78,27 +78,24 @@ private void configureCardinal(Context context, Configuration configuration,
...
         switch (request.getUiType()) {
-            case ThreeDSecureRequest.NATIVE:
-                cardinalConfigurationParameters.setUiType(CardinalUiType.NATIVE);
-            case ThreeDSecureRequest.HTML:
-                cardinalConfigurationParameters.setUiType(CardinalUiType.HTML);
-            case ThreeDSecureRequest.BOTH:
-                cardinalConfigurationParameters.setUiType(CardinalUiType.BOTH);
+            case NATIVE: cardinalConfigurationParameters.setUiType(CardinalUiType.NATIVE);
+            case HTML: cardinalConfigurationParameters.setUiType(CardinalUiType.HTML);
+            case BOTH: cardinalConfigurationParameters.setUiType(CardinalUiType.BOTH);
```
sshropshire, line 83: Not a huge concern since the values are distinct, but I just noticed these have always fallen through since we don't use break statements.

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/CardinalClient.java
```diff
@@ -78,27 +78,24 @@ private void configureCardinal(Context context, Configuration configuration,
...
         }
 
         if (request.getRenderTypes() != null) {
             JSONArray renderTypes = new JSONArray();
 
-            for (Integer renderType : request.getRenderTypes()) {
-                if (renderType.equals(ThreeDSecureRequest.OTP)) {
+            for (ThreeDSecureRenderType renderType : request.getRenderTypes()) {
+                if (renderType.equals(ThreeDSecureRenderType.OTP)) {
```
sshropshire, line 90: nit: we could technically `switch` here but the behavior would be the same.

sshropshire (approved):

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/CardinalClient.java
tdchow: I tried utilizing a `switch` here, but it seems that Kotlin sealed classes don't play well with java `switch`es :( 

I'll be converting this entire class to Kotlin. So I'll be sure to convert this to a when statement.

sshropshire: Oh cool yeah that makes sense 👍

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/CardinalClient.java
tdchow: Good catch! I'll be converting this file to Kotlin in another PR. This issue should be resolved by converting to a when statment.

PR #1129, opened 2024-08-23
[Fix ConfigurationLoader] Improve ConfigurationCache Encapsulation

sshropshire:
**Note:** This PR is Part 3 in a series of refactoring PRs to eliminate `ConfigurationLoader` race conditions.
Previous PR: [#1123 [Fix ConfigurationLoader] Remove HttpClient.RetryStrategy](https://github.com/braintree/braintree_android/pull/1123)

### Summary of changes

 - This PR moves Configuration caching logic from `ConfigurationLoader` into `ConfigurationCache` where it makes more sense.
 - This refactor is mostly to enforce proper encapsulation. It will also make the PR to fix the race condition more clear.

### Checklist

- @sshropshire


BraintreeCore/src/test/java/com/braintreepayments/api/core/ConfigurationCacheUnitTest.kt
```diff
@@ -1,62 +1,72 @@
...
+            sut.getConfiguration(authorization, configUrl, TimeUnit.MINUTES.toMillis(20))
+        )
     }
 
     @Test
     fun getConfiguration_returnsConfigurationFromSharedPrefs() {
-        val configuration = fromJson(Fixtures.CONFIGURATION_WITHOUT_ACCESS_TOKEN)
-        every { braintreeSharedPreferences.containsKey("cacheKey_timestamp") } returns true
-        every { braintreeSharedPreferences.getLong("cacheKey_timestamp") } returns 0L
-        every { braintreeSharedPreferences.getString("cacheKey", "") } returns configuration.toJson()
```
sshropshire, line 39: Also replacing `mockk` `BraintreeSharedPreferences` with a real instance to avoid leaking implementation details of the cache key's string format.

sarahkoop (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/ConfigurationCache.kt
```diff
@@ -1,37 +1,45 @@
...
 
-    fun getConfiguration(cacheKey: String): String? {
-        return getConfiguration(cacheKey, System.currentTimeMillis())
-    }
-
-    fun getConfiguration(cacheKey: String, currentTimeMillis: Long): String? {
+    fun getConfiguration(
+        authorization: Authorization,
+        configUrl: String,
+        currentTimeMillis: Long = System.currentTimeMillis()
```
tdchow, line 16: What are your thoughts on injecting a class that can handle getting `System.currentTimeMillis()`? This way we could inject that class into `ConfigurationCache`'s primary constructor. We could then remove the parameter from both `getConfiguration` and `putConfiguration`.

tdchow: Coming up with a name is the hardest part 😂 What about `Time.getCurrentTime()`? I could see it also handling anywhere where we deal with timestamps (possibly timestamps in analytic events).

This isn't a current pattern, but thinking in MVVM, I could see it being it's own use case: `GetCurrentTimeUseCase`.

sshropshire: > Coming up with a name is the hardest part 😂 

Forreal! I'm about to get this as a tattoo lol 😭. I like Time! I'm gonna update the PR. It does clean up both the src and test files which is awesome 😮‍💨

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/Time.java
```diff
@@ -0,0 +1,11 @@
+package com.braintreepayments.api.sharedutils;
+
+import androidx.annotation.RestrictTo;
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public class Time {
```
tdchow, line 6: Nit: any reason why we can't make this a Kotlin class?

tdchow (approved):

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/Time.java
sshropshire: Good call! I initially thought the shared utils module was all Java. I'll refactor and push up the change.

PR #1130, opened 2024-08-23
Add error handling to the demo app and convert demo app classes to Kotlin

tdchow:
### Summary of changes

 - Convert a few demo app classes to Kotlin
 - Update `BraintreeAuthorizationResult` and `FetchClientTokenResult` to a sealed class that contains a Success and Error type - this resolved the 2 crashes I was seeing when running into emulator certificate errors.

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


Demo/src/main/java/com/braintreepayments/demo/TransactionRequest.kt
```diff
@@ -0,0 +1,8 @@
+package com.braintreepayments.demo
+
+data class TransactionRequest(
+    val amount: String,
+    val paymentMethodNonce: String,
+    val merchantAccountId: String?,
+    val threeDSecureRequired: Boolean
```
jaxdesmarais, line 7: Take it or leave it, not sure if we want to group required vs optional together and flop the order here

sshropshire (approved):
🌊

Demo/src/main/java/com/braintreepayments/demo/TransactionRequest.kt
tdchow: All the parameters are required to be passed into the constructor as there are no default values. Even though `merchantAccountId` is marked as nullable, an explicit `null` needs to be passed in.

I kept the ordering of the old Java constructor to keep the call sites the same:

```java
    public TransactionRequest(String amount, String paymentMethodNonce, String merchantAccountId, Boolean threeDSecureRequired) {
        ...
    }
```

PR #1131, opened 2024-08-23
Convert SEPADirectDebit module classes to Kotlin

chinghlinn:
### Summary of changes

 - Convert CreateMandateResult, SEPADirectDebitApi, SEPADirectDebitClient, SEPADirectDebitMandateType, SEPADirectDebitPaymentAuthRequestParams, SEPADirectDebitPaymentAuthResultInfo, SEPADirectDebitRequest to Kotlin
 - Convert `SEPADirectDebitMandateType.java` to `SEPADirectDebitMandateType.kt` and utilize Kotlin enum class property name and function valueOf(value: String) for value converstion between string and enum
  - Update `interface SEPADirectDebitInternalTokenizeCallback` to `fun interface SEPADirectDebitInternalTokenizeCallback` so that it can be implemented with Kotlin lambda and maintain backward compatible
  - Fix format in tests from other PR

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


SEPADirectDebit/src/main/java/com/braintreepayments/api/sepadirectdebit/CreateMandateResult.kt
```diff
@@ -0,0 +1,9 @@
+package com.braintreepayments.api.sepadirectdebit
+
+data class CreateMandateResult internal constructor(
+    val approvalUrl: String,
+    val ibanLastFour: String,
+    val customerId: String,
+    val bankReferenceToken: String,
+    val mandateType: SEPADirectDebitMandateType?
```
sarahkoop, line 8: Does this have to be optional? I don't think it should ever be null

SEPADirectDebit/src/main/java/com/braintreepayments/api/sepadirectdebit/CreateMandateResult.kt
```diff
@@ -0,0 +1,9 @@
+package com.braintreepayments.api.sepadirectdebit
+
+data class CreateMandateResult internal constructor(
+    val approvalUrl: String,
+    val ibanLastFour: String,
+    val customerId: String,
+    val bankReferenceToken: String,
+    val mandateType: SEPADirectDebitMandateType
```
sshropshire, line 8: In [CreateMandateResult.java](https://github.com/braintree/braintree_android/blob/31b2377a47b570c69ccb6096efc306a8a80983e3/SEPADirectDebit/src/main/java/com/braintreepayments/api/sepadirectdebit/CreateMandateResult.java#L17) the constructor takes all strings. We should do the same thing here in Kotlin to maintain feature parity.

sshropshire (approved):
One small nit. Overall LGTM 👍

SEPADirectDebit/src/main/java/com/braintreepayments/api/sepadirectdebit/CreateMandateResult.kt
tdchow: Since this is an internal constructor that the SDK invokes, should we update the call site to pass in the `SEPADirectDebitMandateType` enum type? This way we won't have to manage an extra constructor that's called internally.

SEPADirectDebit/src/main/java/com/braintreepayments/api/sepadirectdebit/SEPADirectDebitInternalTokenizeCallback.kt
```diff
@@ -1,5 +1,5 @@
 package com.braintreepayments.api.sepadirectdebit
 
-internal interface SEPADirectDebitInternalTokenizeCallback {
+fun interface SEPADirectDebitInternalTokenizeCallback {
```
tdchow, line 3: Can we keep this interface marked with `internal`?

SEPADirectDebit/src/main/java/com/braintreepayments/api/sepadirectdebit/CreateMandateResult.kt
chinghlinn: @sshropshire In the original implementation, the constructor takes string as input but returns enum for getter for the mandateType property. Using enum is easier to maintain data integrity, so I picked enum for this property during conversion.

@tdchow Yes, the call side has to convert the input data to enum type, for example https://github.com/braintree/braintree_android/blob/15c6e6b2c2ea8212f522ffe38bbc831e1589a387/SEPADirectDebit/src/main/java/com/braintreepayments/api/sepadirectdebit/SEPADirectDebitApi.kt#L108

tdchow (approved):

PR #1135, opened 2024-08-28
Remove throwing exceptions in LocalPaymentNonce when fields are missing

tdchow:
### Summary of changes

 - `JSONException("givenName is null")` was getting thrown in the demo app since the response didn't contain a givenName.
 - Insetead of throwing an exception, the non-null fields are set to an empty string.

### Checklist

 - [ ] Added a changelog entry
 - [X] Relevant test coverage

### DIFF
--- a/LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentNonce.kt
+++ b/LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentNonce.kt
@@ -106,11 +106,11 @@ data class LocalPaymentNonce internal constructor(
                 clientMetadataId = clientMetadataId,
                 billingAddress = billingAddress,
                 shippingAddress = shippingAddress,
-                givenName = givenName ?: throw JSONException("givenName is null"),
-                surname = surname ?: throw JSONException("surname is null"),
-                phone = phone ?: throw JSONException("phone is null"),
+                givenName = givenName ?: "",
+                surname = surname ?: "",
+                phone = phone ?: "",
                 email = email,
-                payerId = payerId ?: throw JSONException("payerId is null"),
+                payerId = payerId ?: "",
             )
         }
     }
--- a/LocalPayment/src/test/java/com/braintreepayments/api/localpayment/LocalPaymentNonceUnitTest.java
+++ b/LocalPayment/src/test/java/com/braintreepayments/api/localpayment/LocalPaymentNonceUnitTest.java
@@ -68,25 +68,19 @@ public class LocalPaymentNonceUnitTest {
             new JSONObject(Fixtures.PAYMENT_METHODS_LOCAL_PAYMENT_MISSING_FIELDS_RESPONSE)
         );
 
-        Parcel parcel = Parcel.obtain();
-        result.writeToParcel(parcel, 0);
-        parcel.setDataPosition(0);
-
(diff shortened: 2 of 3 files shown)


LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentNonce.kt
```diff
@@ -106,11 +106,11 @@ data class LocalPaymentNonce internal constructor(
                 clientMetadataId = clientMetadataId,
                 billingAddress = billingAddress,
                 shippingAddress = shippingAddress,
-                givenName = givenName ?: throw JSONException("givenName is null"),
-                surname = surname ?: throw JSONException("surname is null"),
-                phone = phone ?: throw JSONException("phone is null"),
+                givenName = givenName ?: "",
+                surname = surname ?: "",
```
ashivashankars, line 110: Would it be good to throw exception for the user by initialize fields to empty string  var givenName: String? = ""

sshropshire (approved):

LocalPayment/src/test/java/com/braintreepayments/api/localpayment/LocalPaymentNonceUnitTest.java
```diff
@@ -60,4 +61,32 @@ public void parcelsCorrectly() throws JSONException {
...
+    public void when_fromJSON_is_called_missing_fields_default_to_an_empty_string() throws JSONException {
+        LocalPaymentNonce result = LocalPaymentNonce.fromJSON(
+            new JSONObject(Fixtures.PAYMENT_METHODS_LOCAL_PAYMENT_MISSING_FIELDS_RESPONSE)
+        );
+
+        Parcel parcel = Parcel.obtain();
+        result.writeToParcel(parcel, 0);
+        parcel.setDataPosition(0);
+
+        LocalPaymentNonce parceled = LocalPaymentNonce.CREATOR.createFromParcel(parcel);
```
sshropshire, line 75: 🧶: we can remove this portion of the test since we no longer own the `Parcelable` implementation.

LocalPayment/src/main/java/com/braintreepayments/api/localpayment/LocalPaymentNonce.kt
tdchow: The demo app seems to have a successful flow when these fields are empty, so I'm not sure if we'd want to throw any exceptions here. The constructor is internal to the SDK as the merchant does not instantiate this class.

LocalPayment/src/test/java/com/braintreepayments/api/localpayment/LocalPaymentNonceUnitTest.java
ashivashankars: I have made changes for this file, will push those changes soon.

ashivashankars (approved):

sarahkoop (approved):

LocalPayment/src/test/java/com/braintreepayments/api/localpayment/LocalPaymentNonceUnitTest.java
tdchow: Good call. I've removed the parcel logic from this test.

@ashivashankars - it should make things easier when merging this fix into your branch

PR #1137, opened 2024-08-28
3DS Kotlin Conversion Part 4

tdchow:
### Summary of changes

 - Convert CardinalClient, ThreeDSecureAPI, ThreeDSecureClient, and ThreeDSecureLauncher to Kotlin
 
### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureAPI.kt
```diff
@@ -0,0 +1,75 @@
...
+import com.braintreepayments.api.threedsecure.ThreeDSecureParams.Companion.fromJson
+import org.json.JSONException
+import org.json.JSONObject
+
+internal class ThreeDSecureAPI(
+    private val braintreeClient: BraintreeClient
+) {
+
+    fun performLookup(
+        request: ThreeDSecureRequest, cardinalConsumerSessionId: String?,
```
sarahkoop, line 15: Nit can we newline `cardinalConsumerSessionId`?

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureAPI.kt
```diff
@@ -0,0 +1,75 @@
...
+                }
+            } else {
+                callback.onThreeDSecureResult(null, httpError)
+            }
+        }
+    }
+
+    fun authenticateCardinalJWT(
+        threeDSecureParams: ThreeDSecureParams,
+        cardinalJWT: String?,
```
sarahkoop, line 39: Same thing here I don't think this should be optional

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureClient.java
```diff
@@ -1,324 +1,324 @@
...
-        braintreeClient.sendAnalyticsEvent(ThreeDSecureAnalytics.VERIFY_SUCCEEDED);
-        callback.onThreeDSecureResult(result);
-    }
-
-    private void callbackCancel(ThreeDSecureTokenizeCallback callback) {
-        braintreeClient.sendAnalyticsEvent(ThreeDSecureAnalytics.VERIFY_CANCELED);
-        callback.onThreeDSecureResult(ThreeDSecureResult.Cancel.INSTANCE);
-    }
-}
+//package com.braintreepayments.api.threedsecure;
```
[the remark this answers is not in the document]
tdchow: Whoops! Yep this should have been removed 😅

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureAPI.kt
tdchow: 👍 I'll make this parameter non null and add a null check prior to the call site.

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/CardinalClient.kt
```diff
@@ -0,0 +1,136 @@
...
+        } catch (e: RuntimeException) {
+            throw BraintreeException("Cardinal SDK init Error.", e)
+        }
+    }
+
+    @Throws(BraintreeException::class)
+    @Suppress("TooGenericExceptionCaught")
+    fun continueLookup(
+        threeDSecureParams: ThreeDSecureParams,
+        challengeObserver: CardinalChallengeObserver?
```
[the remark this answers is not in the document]
sshropshire: Is it possible to make the `challengeObserver` param non-optional?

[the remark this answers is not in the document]
tdchow: I originally tried to make this parameter non-null, but then the null check needed to be moved to `ThreeDSecureActivity` since the observer is set in the Activity's onCreate. I felt that it made more sense to keep the logic out of the Activity which made it much easier to add a unit test covering this path. But happy to revisit if folks feel differently!

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureClient.kt
```diff
@@ -0,0 +1,426 @@
...
+            val failure = when {
+                configuration == null -> {
+                    error ?: BraintreeException("Configuration is null")
+                }
+
+                !configuration.isThreeDSecureEnabled -> {
+                    BraintreeException(
+                        "Three D Secure is not enabled for this account. " +
+                            "Please contact Braintree Support for assistance."
+                    )
```
sshropshire, line 84:
```suggestion
                    val message = "Three D Secure is not enabled for this account. Please contact Braintree Support for assistance."
                    BraintreeException(message)
```
suuuper nit: It isn't much of a performance boost since it's not frequently called in a loop or anything, but string concatenation does have a runtime cost. The linter may be influencing the implementation here as well.

Also definitely out of scope but I wonder if we should consider externalizing our error strings into XML in the future? We may be able to inject some documentation links to help merchants who get stuck.

sshropshire (approved):

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/CardinalClient.kt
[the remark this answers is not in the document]
sshropshire: Ok got it 👍 that makes sense.

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureClient.kt
tdchow: This is a great callout. I agree the string concatenation here is neglible, but we should find a long term solution for handling longer strings. Maybe some sort of StringLoader class that either grabs strings from XML or has them defined internally.

sarahkoop (approved):

PR #1138, opened 2024-08-29
Venmo kotlin conversion

ashivashankars:
### Summary of changes

 - Convert VenmoApi, VenmoClient, VenmoPaymentAuthRequestParams, VenmoPaymentAuthResultInfo, VenmoSharedPrefsWriter to Kotlin
 - Updated the VenmoClientUnitTest


Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoLineItem.kt
```diff
@@ -0,0 +1,74 @@
...
+) : Parcelable {
+    /**
+     * The type of Venmo line item.
+     *
+     * [.KIND_CREDIT] A line item that is a credit. [.KIND_DEBIT] A line item that
+     * debits.
+     */
+    @Retention(AnnotationRetention.SOURCE)
+    @StringDef(KIND_CREDIT, KIND_DEBIT)
+    internal annotation class VenmoLineItemKind
```
sarahkoop, line 42: This should probably be it's own enum in a separate file

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoPaymentAuthRequestParams.kt
```diff
@@ -0,0 +1,10 @@
+package com.braintreepayments.api.venmo
+
+import com.braintreepayments.api.BrowserSwitchOptions
+
+/**
+ * Used to request Venmo authentication via [VenmoLauncher.launch] )}
+ */
+class VenmoPaymentAuthRequestParams internal constructor(
```
sarahkoop, line 8: Seems like we don't have consistency across modules on these `PaymentAuthRequestParams` classes now. I created a separate ticket to align them, but maybe we can start the discussion on the approach here. Should these all be data classes? (cc @tdchow) I also think `browserSwitchOptions` can be library scope on all of them

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoPaymentMethodUsage.kt
```diff
@@ -0,0 +1,15 @@
...
+ * Usage type for the tokenized Venmo account.
+ */
+@Retention(AnnotationRetention.SOURCE)
+@IntDef(VenmoPaymentMethodUsage.SINGLE_USE, VenmoPaymentMethodUsage.MULTI_USE)
+annotation class VenmoPaymentMethodUsage {
+    companion object {
+        const val SINGLE_USE: Int = 1
+        const val MULTI_USE: Int = 2
+    }
+}
```
sarahkoop, line 15: Let's make this an enum

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoPaymentAuthResultInfo.kt
```diff
@@ -0,0 +1,13 @@
+package com.braintreepayments.api.venmo
+
+import com.braintreepayments.api.BrowserSwitchFinalResult
+
+/**
+ * Details of a [VenmoPaymentAuthResult.Success]
+ */
+class VenmoPaymentAuthResultInfo internal constructor
```
sarahkoop, line 8: Same here - this should follow the same format as PayPal: https://github.com/braintree/braintree_android/blob/main/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalPaymentAuthResultInfo.kt

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoApi.kt
```diff
@@ -0,0 +1,185 @@
...
+    fun createPaymentContext(
+        request: VenmoRequest,
+        venmoProfileId: String?,
+        callback: VenmoApiCallback
+    ) {
+        val params = JSONObject()
+        try {
+            params.put("query", "mutation CreateVenmoPaymentContext(\$input: " +
+                        "CreateVenmoPaymentContextInput!) { createVenmoPaymentContext" +
+                        "(input: \$input) { venmoPaymentContext { id } } }"
```
tdchow, line 27: We could use a kotlin multiline string here to properly format the graphql query: https://kotlinlang.org/docs/java-to-kotlin-idioms-strings.html#use-multiline-strings

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoApi.kt
```diff
@@ -0,0 +1,185 @@
...
+        try {
+            params.put(
+                "query",
+                "query PaymentContext(\$id: ID!) { node(id: \$id) " +
+                        "{ ... on VenmoPaymentContext { paymentMethodId userName payerInfo " +
+                        "{ firstName lastName phoneNumber email externalId userName " +
+                        "shippingAddress { fullName addressLine1 addressLine2 " +
+                        "adminArea1 adminArea2 postalCode countryCode } " +
+                        "billingAddress { fullName addressLine1 addressLine2 " +
+                        "adminArea1 adminArea2 postalCode countryCode } } } } }"
```
tdchow, line 125: Same here - we could use a multiline string.

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.kt
```diff
@@ -0,0 +1,352 @@
...
+
+    /**
+     * True if `tokenize()` was called with a Vault request object type
+     */
+    private var isVaultRequest = false
+
+    private constructor(braintreeClient: BraintreeClient, apiClient: ApiClient) : this(
+        braintreeClient, VenmoApi(braintreeClient, apiClient),
+        VenmoSharedPrefsWriter()
+    )
```
tdchow, line 46: Can we use default values in the primary constructor to remove this constructor?

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.kt
```diff
@@ -0,0 +1,352 @@
...
+        context: Context,
+        authorization: String,
+        returnUrlScheme: String?
+    ) : this(BraintreeClient(context, authorization, returnUrlScheme))
+
+    @VisibleForTesting
+    internal constructor(braintreeClient: BraintreeClient) : this(
+        braintreeClient,
+        ApiClient(braintreeClient)
+    )
```
tdchow, line 65: We should be able to remove this constructor as well with adding default values to the primary constructor.

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.kt
```diff
@@ -0,0 +1,352 @@
...
+                    } else if (error != null) {
+                        callbackTokenizeFailure(callback, VenmoResult.Failure(error))
+                    }
+                }
+            } else {
+                val venmoAccountNonce = VenmoAccountNonce(
+                    paymentMethodNonce, false, null, null,
+                    null, null, null, username,
+                    null, null
+                )
```
tdchow, line 297: This constructor might read a bit better if all arguments are on different lines and if param names are added for each.

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoLineItem.kt
```diff
@@ -0,0 +1,74 @@
...
+                .putOpt(NAME_KEY, name)
+                .putOpt(PRODUCT_CODE_KEY, productCode)
+                .putOpt(QUANTITY_KEY, quantity)
+                .putOpt(UNIT_AMOUNT_KEY, unitAmount)
+                .putOpt(UNIT_TAX_AMOUNT_KEY, unitTaxAmount)
+                .putOpt(URL_KEY, url)
+        } catch (ignored: JSONException) {
+        }
+
+        return JSONObject()
```
tdchow, line 58: Nit - we could combine the return statement with the try to simplify things

```suggestion
        return try {
            JSONObject()
                .putOpt(DESCRIPTION_KEY, description)
                .putOpt(KIND_KEY, kind)
                .putOpt(NAME_KEY, name)
                .putOpt(PRODUCT_CODE_KEY, productCode)
                .putOpt(QUANTITY_KEY, quantity)
                .putOpt(UNIT_AMOUNT_KEY, unitAmount)
                .putOpt(UNIT_TAX_AMOUNT_KEY, unitTaxAmount)
                .putOpt(URL_KEY, url)
        } catch (ignored: JSONException) {
            JSONObject()
        }
```

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoPaymentAuthRequestParams.kt
tdchow: Great catch! I think we'll want to match the other modules in making this a sealed class.

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoPaymentAuthResultInfo.kt
```diff
@@ -0,0 +1,13 @@
...
+    (private val browserSwitchSuccess: BrowserSwitchFinalResult.Success) {
```
tdchow, line 9: Can we make this a public property and remove the getter?

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoPaymentMethodUsage.kt
```diff
@@ -0,0 +1,15 @@
+package com.braintreepayments.api.venmo
+
+import androidx.annotation.IntDef
+
+/**
+ * Usage type for the tokenized Venmo account.
+ */
+@Retention(AnnotationRetention.SOURCE)
+@IntDef(VenmoPaymentMethodUsage.SINGLE_USE, VenmoPaymentMethodUsage.MULTI_USE)
+annotation class VenmoPaymentMethodUsage {
```
tdchow, line 10: Are we able to convert this to an enum?

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoRequest.kt
```diff
@@ -0,0 +1,123 @@
...
+    var collectCustomerBillingAddress: Boolean = false,
+    var totalAmount: String? = null,
+    var subTotalAmount: String? = null,
+    var discountAmount: String? = null,
+    var taxAmount: String? = null,
+    var shippingAmount: String? = null,
+    var isFinalAmount: Boolean = false
+) : Parcelable {
+
+    val paymentMethodUsageAsString: String?
```
tdchow, line 84: If we convert `VenmoPaymentMethodUsage` to an enum, we should be able to use `SINGLE_USE`.name and `MULTI_USE`.name to get the string value of the enum.

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoRequest.kt
```diff
@@ -0,0 +1,123 @@
...
+     */
+    fun setVenmoLineItem(lineItems: Collection<VenmoLineItem>) {
+        this.lineItems.clear()
+        this.lineItems.addAll(lineItems)
+    }
+
+    /**
+     * @return The line items for this transaction. Can include up to 249 line items.
+     */
+    fun getVenmoLineItem(): ArrayList<VenmoLineItem> {
```
tdchow, line 105: Do we need this getter function? It looks like the `lineItems` property is public.

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoRequest.kt
```diff
@@ -0,0 +1,123 @@
...
+    fun getCollectCustomerBillingAddressAsString(): String {
+        return collectCustomerBillingAddress.toString()
+    }
+
+    /**
+     * @return Whether or not the purchase amount is the final amount as a string value.
+     */
+    fun getIsFinalAmountAsString(): String {
+        return isFinalAmount.toString()
+    }
```
tdchow, line 122: I think these were existing functions - should we look into removing them and having the callsites invoke the `toString()` functions? 

I feel that `VenmoRequest` is a simple data class that shouldn't contain any logic and that the places where we use the properties should be where we format it.

Demo/src/main/java/com/braintreepayments/demo/VenmoFragment.java
```diff
@@ -102,7 +102,7 @@ public void launchVenmo(View v) {
         ArrayList<VenmoLineItem> lineItems = new ArrayList<>();
         lineItems.add(new VenmoLineItem(VenmoLineItem.KIND_CREDIT, "Some Item", 1, "2"));
         lineItems.add(new VenmoLineItem(VenmoLineItem.KIND_DEBIT, "Two Items", 2, "10"));
-        venmoRequest.setLineItems(lineItems);
+        venmoRequest.setVenmoLineItem(lineItems);
```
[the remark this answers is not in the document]
ashivashankars: The method name was changed from setLineItem to setVenmoLineItems to more accurately reflect its functionality, and found this name more relatable. The previous name was ambiguous because it did not specify the type of line items being set. In reality, the method was responsible for handling a collection of VenmoLineItem.

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.kt
sarahkoop: Updated to put args on different lines - I'm not a fan of comments for param names since Android Studio provides them for you but can add if it would make it more clear

sarahkoop: Ah ok - added 👍

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoRequest.kt
sarahkoop: Our logic around this was kind of complex - I updated to make this an optional parameter, which removes the getter/setter but requires a force unwrap in where it's used in the API but I think it's ok. Let me know what you think

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoRequest.kt
sarahkoop: Done

Venmo/src/test/java/com/braintreepayments/api/venmo/VenmoApiUnitTest.java
```diff
@@ -67,9 +67,6 @@ public void createPaymentContext_createsPaymentContextViaGraphQL() throws JSONEx
         verify(braintreeClient).sendGraphQLPOST(captor.capture(), any(HttpResponseCallback.class));
 
         JSONObject graphQLJSON = captor.getValue();
-        String expectedQuery =
```
sarahkoop, line 70: These fail with the Kotlin multiline string indentation. I think since they are set string values in the code it doesn't really add value to include them in the unit tests. The alternative would be to convert the whole test file to Kotlin to change these to use the multiline spacing - what do you think?

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoPaymentAuthRequestParams.kt
```diff
@@ -0,0 +1,12 @@
+package com.braintreepayments.api.venmo
+
+import com.braintreepayments.api.BrowserSwitchOptions
+
+/**
+ * Used to request Venmo authentication via [VenmoLauncher.launch] )}
+ */
+data class VenmoPaymentAuthRequestParams internal constructor(
+    // TODO: this should be internal and restricted to library scope but can't be until
+    //  VenmoClientUnitTest is converted to Kotlin
```
sarahkoop, line 10: @tdchow I think ideally all the request params class would follow this structure and have the `browserSwitchOptions` param internal with library visibility. This breaks the unit tests, so we can't do that unless also converting the `VenmoClientUnitTest` class. Can I leave this as is for your ticket to unify all the auth request classes and then we can decide if it's worth converting/updating?

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoRequest.kt
tdchow: Yep, that sounds good to me 👍

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoPaymentAuthResultInfo.kt
sarahkoop: I think we want this to be internal - merchants shouldn't need to access it. Removed the separate getter and used the restrict annotation though. Feel free to change the way this is set up to align with other modules in the alignment ticket work

Venmo/src/test/java/com/braintreepayments/api/venmo/VenmoApiUnitTest.java
tdchow: Ah I see. I think we can leave it as a follow up todo.

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoPaymentAuthRequestParams.kt
tdchow: Gotcha. Yeah I can take a look as a part of the the other interface story.

sarahkoop:
Updated this PR and is ready for re-review

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.kt
tdchow: Ah I didn't realize that `VenmoAccountNonce` was a Java class where we can't use named parameters.

tdchow (approved):

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.kt
sarahkoop: Ahhh it's not - I misunderstood this comment 🤦‍♀️. Updated to use named params!

Demo/src/main/java/com/braintreepayments/demo/VenmoFragment.java
```diff
@@ -1,5 +1,6 @@
 package com.braintreepayments.demo;
 
+
```
jaxdesmarais, line 3:
```suggestion
```

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoPaymentMethodUsage.kt
```diff
@@ -0,0 +1,9 @@
+package com.braintreepayments.api.venmo
+
+/**
+ * Usage type for the tokenized Venmo account.
+ */
+enum class VenmoPaymentMethodUsage {
+    SINGLE_USE,
+    MULTI_USE
```
jaxdesmarais, line 8: Do we need any docstrings here?

jaxdesmarais (approved):

PR #1133, opened 2024-08-27
[Fix Configuration Loader] Prevent Race Condition

sshropshire:
**Note:** This PR is Part 4 in a series of refactoring PRs to eliminate `ConfigurationLoader` race conditions.
Previous PR: [#1129 [Fix ConfigurationLoader] Improve ConfigurationCache Encapsulation](https://github.com/braintree/braintree_android/pull/1129)

### Summary of changes

 - This PR moves multi-threading logic into `ConfigurationLoader` to guard against race conditions related to `ConfigurationCache` access
 - We create a background thread to perform synchronous HTTP requests and write to the cache in one atomic cycle
 - Internally we use a single-thread thread pool to behave as a serial dispatch queue
 - Only one request for `Configuration` is processed at a time. Subsequent requests are blocked until the currently active request finishes
 - This results in a single HTTP config request that is cached on success before responding to other config requests 

### Checklist

 - [ ] ~Added a changelog entry~
 - [x] Relevant test coverage

- @sshropshire

PR #1139, opened 2024-09-03
[Fix Configuration Loader] Migrate Thread Scheduling Logic into BraintreeClient

sshropshire:
**Note:** This PR is Part 5 in a series of refactoring PRs to eliminate `ConfigurationLoader` race conditions.
Previous PR: [#1133 [Fix Configuration Loader] Prevent Race Condition](https://github.com/braintree/braintree_android/pull/1133)

### Summary of changes

 - This PR migrates multi-threading logic into BraintreeClient to set up an eventual easy migration to coroutines
 - In the future, we can leverage co-routines to reduce the SDK's responsibility to create background threads.
 - Upcoming Tasks to finish Fix Configuration Loader PR series:
   - Remove `HttpClient` and make all networking clients thread synchronous by default
   - Implement retry logic in `ConfigurationLoader`

### Checklist

 - [x] Relevant test coverage

- @sshropshire

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
@@ -5,12 +5,14 @@ import android.content.pm.ActivityInfo
 import android.net.Uri
 import androidx.annotation.RestrictTo
 import androidx.annotation.VisibleForTesting
+import androidx.annotation.WorkerThread
 import com.braintreepayments.api.sharedutils.HttpMethod
 import com.braintreepayments.api.sharedutils.HttpResponseCallback
 import com.braintreepayments.api.sharedutils.HttpResponseTiming
 import com.braintreepayments.api.sharedutils.ManifestValidator
-import org.json.JSONException
+import com.braintreepayments.api.sharedutils.Scheduler
 import org.json.JSONObject
+import java.lang.ref.WeakReference
 
 /**
  * Core Braintree class that handles network requests.
@@ -50,6 +52,7 @@ class BraintreeClient @VisibleForTesting internal constructor(
      * @suppress
      */
     val appLinkReturnUri: Uri?,
+    private val threadScheduler: Scheduler
 ) {
 
     private val crashReporter: CrashReporter
@@ -68,7 +71,8 @@ class BraintreeClient @VisibleForTesting internal constructor(
         manifestValidator = params.manifestValidator,
         returnUrlScheme = params.returnUrlScheme,
         braintreeDeepLinkReturnUrlScheme = params.braintreeReturnUrlScheme,
-        appLinkReturnUri = params.appLinkReturnUri
+        appLinkReturnUri = params.appLinkReturnUri,
+        threadScheduler = params.threadScheduler
     )
 
     /**
@@ -132,7 +136,7 @@ class BraintreeClient @VisibleForTesting internal constructor(
             } else {
                 callback.onResult(null, response.error)
             }
-            response.timing?.let { sendAnalyticsTimingEvent("/v1/configuration", it) }
+            response.timing?.let { sendRESTTimingEvent("/v1/configuration", it) }
         }
     }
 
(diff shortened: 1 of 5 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
```diff
@@ -233,6 +228,19 @@ class BraintreeClient @VisibleForTesting internal constructor(
         }
     }
 
+    @Suppress("TooGenericExceptionCaught")
+    @WorkerThread
+    private fun sendHttpRequestSync(
+        request: InternalHttpRequest,
+        configuration: Configuration
+    ): Pair<String?, Exception?> = try {
```
tdchow, line 236: Instead of returning a `Pair<String?, Exception?>`, what are your thoughts in returning some sort of `Result` sealed class? This way we can better handle the result with a when statement instead of having to check for null?

It would also remove 2 of the possible cases that should never happen - when both String and Exception are not null and when they are both null.

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
```diff
@@ -278,6 +264,22 @@ class BraintreeClient @VisibleForTesting internal constructor(
...
     }
 
+    /**
+     * @suppress
+     */
+    @Suppress("TooGenericExceptionCaught")
+    private fun sendGraphQLPOSTSync(
+        json: JSONObject?,
+        configuration: Configuration
+    ): Pair<String?, Exception?> = try {
```
tdchow, line 274: Same here - we might be able to use the same sealed class type here.

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
sshropshire: I 100% agree the `try/catch` has a big time Java feel. It would move us closer towards the coroutine direction as well.

I do have two follow up PRs in mind, one to remove `HttpClient` because the threading logic is no longer needed, and the other is to implement the retry logic in `ConfigurationLoader`. I wonder if this would make sense in the `HttpClient` clean up PR? Or would you prefer we go for it here? I'm cool with either approach.

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
sshropshire: Yup totally agree. This is a great callout.

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
tdchow: I'm good either way, whichever is easier on your end!

tdchow (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
sshropshire: Ok cool yeah. I can clean this up in the next one! It will make a clean diff. 🙏

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
```diff
@@ -278,6 +264,22 @@ class BraintreeClient @VisibleForTesting internal constructor(
...
+        val data = json?.toString()
+        val response = graphQLClient.post("", data, configuration, authorization)
```
sarahkoop, line 276: Nit: we have an overload graphQLClient `post` method that doesn't take in `path` and sends empty string in the method - can we use that `post` method here instead?

jaxdesmarais:
@tdchow is this PR and the PRs in the base branch something we still need to get into v5? It's been inactive for a bit so wasn't sure if this still needed another 👍 and for us to open a PR from base to main.

tdchow:
> @tdchow is this PR and the PRs in the base branch something we still need to get into v5? It's been inactive for a bit so wasn't sure if this still needed another 👍 and for us to open a PR from base to main.

@sshropshire is this part 5 the final changes needed for the configuration loader refactor? would we be good merging the feature branch into main after this PR?

sshropshire:
It's definitely close. There's retry logic that needs to be re-implemented for this to be fully complete. It used to happen in [HttpClient](https://github.com/braintree/braintree_android/blob/c53317c8cf5cd6530e1c49aaa02f6cccd3392089/SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/HttpClient.java#L73) but it makes more sense for it to live in `ConfigurationLoader`.

saperi22:
@sshropshire @tdchow what state is this PR in? Does it need more work or are we good to merge?

tdchow:
@saperi22 - We decided to put the ConfigurationLoader work on hold for now. I'll go ahead and close this PR.

PR #1142, opened 2024-09-06
PayPal Account Locked Error

ashivashankars:
### Summary of changes

 - Error : iOS received a proper USER_ACCOUNT_LOCKED error message, but on Android received a more generic "Parsing error response failed" message.
 - Handled error similar to iOS.
 - Used LoggingUtils to track exceptions
 - Return null if it's bad json.
 - Unit test cases for the existing fixture json response.

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/ErrorWithResponse.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/ErrorWithResponse.kt
@@ -1,8 +1,10 @@
 package com.braintreepayments.api.core
 
 import android.os.Parcelable
+import android.util.Log
 import androidx.annotation.RestrictTo
 import com.braintreepayments.api.core.GraphQLConstants.ErrorMessages
+import com.braintreepayments.api.sharedutils.LoggingUtils
 import kotlinx.parcelize.Parcelize
 import org.json.JSONException
 import org.json.JSONObject
@@ -31,7 +33,7 @@ data class ErrorWithResponse internal constructor(
         try {
             parseJson(errorResponse)
         } catch (e: JSONException) {
-            message = "Parsing error response failed"
+            Log.d(LoggingUtils.TAG, e.message.toString())
             fieldErrors = ArrayList()
         }
     }
@@ -39,7 +41,11 @@ data class ErrorWithResponse internal constructor(
     @Throws(JSONException::class)
     private fun parseJson(jsonString: String?) {
         jsonString?.let { JSONObject(it) }?.let { json ->
-            message = json.getJSONObject(ERROR_KEY).getString(MESSAGE_KEY)
+            message = json.optJSONObject(ERROR_KEY)
+                ?.let { jsonObject ->
+                    jsonObject.optString(MESSAGE_KEY, "").takeIf { it.isNotEmpty() }
+                        ?: jsonObject.optString(DEV_MESSAGE_KEY, "")
+                } ?: throw JSONException("Error key not found in JSON")
             fieldErrors = BraintreeError.fromJsonArray(json.optJSONArray(FIELD_ERRORS_KEY))
         }
     }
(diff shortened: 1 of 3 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/core/ErrorWithResponse.kt
```diff
@@ -98,18 +105,19 @@ data class ErrorWithResponse internal constructor(
             try {
                 val errors = json
                     ?.let { JSONObject(it) }
-                    ?.getJSONArray(GraphQLConstants.Keys.ERRORS)
+                    ?.optJSONArray(GraphQLConstants.Keys.ERRORS)
+                    ?: throw JSONException("Errors key not found in JSON")
```
sshropshire, line 109: We don't typically throw `JSONException` in the SDK. Is there a big advantage of using `optJSONArray` instead of `getJSONArray` as before?

ashivashankars: -> Returns null: Instead of throwing an exception, optJSONArray() will return null if the key is not found or the value is not a valid JSONArray.
-> Graceful handling: It provides a more flexible way to access optional or uncertain data. This can be useful when you are dealing with data that may not always include the key or if the key's value may not be a JSONArray but you don't want to handle exceptions.
->Use optJSONArray() if we prefer more flexible code, where the absence of the key or the wrong type doesn't cause an error but can be handled in a more graceful way

sshropshire: > If we don't typically throw JSONException in the SDK, then is okay to return null or empty string?

This might be tricky. I'd defer to whatever iOS does in that case. iOS doesn't use exceptions but I believe they do have nullability checks for either of the fields to determine how the error response is parsed.

Overall this comment is a take-or-leave nitpick. The only reason i'm calling it out is it does feel somewhat odd to throw an exception type we don't own.

BraintreeCore/src/main/java/com/braintreepayments/api/core/ErrorWithResponse.kt
```diff
@@ -98,18 +105,19 @@ data class ErrorWithResponse internal constructor(
...
                 val fieldErrorsEmpty = errorWithResponse.fieldErrors?.isEmpty() ?: true
                 errorWithResponse.message = if (fieldErrorsEmpty) {
-                    errors?.getJSONObject(0)?.getString(GraphQLConstants.Keys.MESSAGE)
+                    errors.getJSONObject(0)?.getString(GraphQLConstants.Keys.MESSAGE)
                 } else {
                     ErrorMessages.USER
                 }
             } catch (e: JSONException) {
                 errorWithResponse.apply {
-                    message = "Parsing error response failed"
```
sshropshire, line 112: Does this message no longer need to be set?

ashivashankars: Yes, I think so because USER like to know more specific error messages.
Also, i think we should 
->Check API Responses to ensure that the API or server responses are in the expected format and properly structured and those keys are matching.
->Improve Error Handling: Add better error handling to provide more specific error messages instead of generic ones like "Parsing error response failed."
->Log Errors for Debugging: Use logging to capture the exact issue when parsing fails to help diagnose and fix the problem from development side.

BraintreeCore/src/main/java/com/braintreepayments/api/core/ErrorWithResponse.kt
ashivashankars: +1 
I agree, to handling as iOS is doing!!

PR #1150, opened 2024-09-11
Align Merchant API Surface

sarahkoop:
### Summary of changes

 - Make `PaymentAuthRequestParams` classes restricted to library scope across modules
 - Remove `PaymentAuthResultInfo` classes that were just wrapping browser switch results and make the references to the browser switch result classes internal
 - Make plain result objects `data object` as suggested by Android Studio lint
 - Convert a few test classes to Kotlin that were required to make parameters internal (Java interop doesn't seem to recognize the module scope/internal kotlin params)
 
### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors

@sarahkoop


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalPaymentAuthRequest.kt
```diff
@@ -8,6 +8,8 @@ sealed class PayPalPaymentAuthRequest {
     /**
      * The request was successfully created and is ready to be launched by [PayPalLauncher]
      */
+    // TODO: requestParams should be internal but can't be until PayPalClientUnitTest is converted
+    //  to Kotlin (same for all modules with ReadyToLaunch request)
```
sarahkoop, line 12: Converting all of the client unit tests to Kotlin is definitely the end goal but it is a large task. Are there any other ways to mark these params as internal until we convert all the tests? Maybe doc strings would be ok for now? @tdchow

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayLauncherCallback.kt
```diff
@@ -6,5 +6,5 @@ package com.braintreepayments.api.googlepay
  */
 fun interface GooglePayLauncherCallback {
 
-    fun onGooglePayLauncherResult(googlePayPaymentAuthResult: GooglePayPaymentAuthResult?)
```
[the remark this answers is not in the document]
sshropshire: That is odd. Do we notify errors with the same callback?

sshropshire (approved):

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayLauncherCallback.kt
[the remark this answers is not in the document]
sarahkoop: There are no "errors" returned in this scenario - it's from the ActivityResultAPI which either returns a result or nothing at all. In my opinion if nothing is returned from that activity result API, we shouldn't return anything in the callback (see [use])(https://github.com/braintree/braintree_android/blob/main/GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayLauncher.kt#L22)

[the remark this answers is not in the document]
sshropshire: That might be a typo in our code the docs do [mark it as non-null](https://developer.android.com/reference/androidx/activity/result/ActivityResultRegistry#register(kotlin.String,androidx.lifecycle.LifecycleOwner,androidx.activity.result.contract.ActivityResultContract,androidx.activity.result.ActivityResultCallback)). The code may still compile if we make it a non-optional.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalPaymentAuthRequest.kt
tdchow: Are we seeing compile errors caused by the internal Kotlin scopes? I was seeing these errors in Java unit tests but they were still able to compile and run.

tdchow (approved):

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalPaymentAuthRequest.kt
sarahkoop: Actual compiler errors - "cannot find symbol `getRequestParams`"

tdchow: Gotcha, yeah I think using doc strings is a good idea.

PR #1152, opened 2024-09-11
Batch Analytics Events by Session ID

tdchow:
### Summary of changes

 - Add `sessionId` column to `AnalyticsEventBlob` table
 - Increase room DB to 8
 - Update query to fetch events by session ID
 - Append sessionId to the `uniqueWorkName` when calling `WorkManager.enqueueUniqueWork()` to send separate API calls for each unique session ID

### Checklist

 - [ ] Added a changelog entry
 - [X] Relevant test coverage


BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
```diff
@@ -74,10 +78,11 @@ internal class AnalyticsClient(
         authorization: Authorization,
         integration: IntegrationType?
     ): UUID {
+        val sessionId = analyticsParamRepository.sessionId
```
sarahkoop, line 81: If the issue was that this value might change do we need to make a copy of it here or is this enough to make it immutable in the inputData?

sarahkoop:
Did we validate that the data is getting to FPTI with sessionID tracked across modules?

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
tdchow: It's a very small edge case where this value can change in between creating the input data and enqueuing the work. But yep, creating a copy here ensures that the session ID in the `uniqueWorkName` and the session ID in the inputData match.

tdchow:
> Did we validate that the data is getting to FPTI with sessionID tracked across modules?

I verified the requests FPTI API calls from the SDK but will do a quick check on FPTI to ensure the logs are correct.

tdchow:
@sarahkoop the expected events are showing up in FPTI with the correct session IDs!

sarahkoop (approved):

sshropshire (approved):
A couple questions but overall LGTM.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
```diff
@@ -161,7 +171,12 @@ internal class AnalyticsClient(
         )
         val event = AnalyticsEvent(name = "crash", timestamp = timestamp)
         val eventJSON = mapAnalyticsEventToFPTIEventJSON(event)
-        val eventBlobs = listOf(AnalyticsEventBlob(eventJSON))
+        val eventBlobs = listOf(
+            AnalyticsEventBlob(
+                jsonString = eventJSON,
+                sessionId = ""
```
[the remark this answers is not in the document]
tdchow: The reportCrash function doesn't end up using the `sessionId` in `AnalyticsEventBlob` as it uses the `sessionId` in the deviceMetadata. But unfortunately `AnalyticsEventBlob.sessionId` is a required field.

I think the more correct approach here would be to remove the use of `AnalyticsEventBlob` when reporting a crash since we're not interacting with the DB. We could create a separate data model for the FPTI request. But I didn't want to introduce to big of a change leading up to the release of v5, but open to other thoughts!

PR #1159, opened 2024-09-18
[QL] Update `interval` and `intervalCount`

jaxdesmarais:
### Summary of changes

- End to end testing revealed that there are cases where `billing_frequency` and `billing_frequency_unit` are not required for `unscheduled` cases for RBA metadata - updated and tested that these fields can be omitted without error when the `recurringBillingPlanType` is `.unscheduled` and the `pricingModel` is `.autoReload` per the test case
- Updated docstrings

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingCycle.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingCycle.kt
@@ -31,10 +31,10 @@ import org.json.JSONObject
  */
 @Parcelize
 data class PayPalBillingCycle @JvmOverloads constructor(
-    var isTrial: Boolean,
+    val isTrial: Boolean,
     val numberOfExecutions: Int,
-    val interval: PayPalBillingInterval? = null,
-    val intervalCount: Int? = null,
+    var interval: PayPalBillingInterval? = null,
+    var intervalCount: Int? = null,
     var sequence: Int? = null,
     var startDate: String? = null,
     var pricing: PayPalBillingPricing? = null
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingCycle.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingCycle.kt
@@ -7,16 +7,18 @@ import org.json.JSONObject
 /**
  * PayPal recurring billing cycle details.
  *
+ * @property isTrial The tenure type of the billing cycle. In case of a plan having trial cycle,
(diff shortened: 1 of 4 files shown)


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingCycle.kt
```diff
@@ -7,47 +7,47 @@ import org.json.JSONObject
...
- * @property isTrial The tenure type of the billing cycle. In case of a plan having trial cycle,
- * only 2 trial cycles are allowed per plan.
  * @property pricing The active pricing scheme for this billing cycle. Required if [isTrial] is
  * false. Optional if [isTrial] is true.
  */
 @Parcelize
 data class PayPalBillingCycle @JvmOverloads constructor(
-    val interval: PayPalBillingInterval,
-    val intervalCount: Int,
+    var isTrial: Boolean,
```
sarahkoop, line 34:
```suggestion
    var isTrial: Boolean = false,
```
I think this should still default to false right?

jaxdesmarais: That's a great question, I had updated this on android to match iOS where `isTrial` is required, but am not sure which one is correct. Let me see if I can find some documentation on which is expected.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingCycle.kt
```diff
@@ -7,47 +7,47 @@ import org.json.JSONObject
...
     val numberOfExecutions: Int,
+    val interval: PayPalBillingInterval? = null,
+    val intervalCount: Int? = null,
```
tdchow, line 37:
```suggestion
    var interval: PayPalBillingInterval? = null,
    var intervalCount: Int? = null,
```

Since these properties are no longer required, we may want to make them `var`s instead of `vals`. This way Java integrations can create a `PayPalBillingCycle` instance, and then call setter functions to set `interval` and `intervalCount` values.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalBillingCycle.kt
sarahkoop: If it is required by the API and we don't really know what the most common value is then we can just leave it required!

sarahkoop (approved):

tdchow (approved):

PR #1161, opened 2024-09-19
Fix 3DS Cancel

sarahkoop:
### Summary of changes

 - Fix logic that was returning a JWT is null error on 3DS cancel

### Checklist

 - [x] Relevant test coverage

### Authors
- @sarahkoop


tdchow (approved):

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureAPI.kt
```diff
@@ -36,12 +37,16 @@ internal class ThreeDSecureAPI(
...
-        threeDSecureParams: ThreeDSecureParams,
-        cardinalJWT: String,
+        threeDSecureParams: ThreeDSecureParams?,
+        cardinalJWT: String?,
         callback: ThreeDSecureResultCallback
     ) {
-        val lookupCardNonce = threeDSecureParams.threeDSecureNonce
-        val lookupNonce = threeDSecureParams.threeDSecureNonce?.string
+        if (threeDSecureParams == null || cardinalJWT == null) {
+            callback.onThreeDSecureResult(null, BraintreeException("An unexpected error occurred"))
```
jaxdesmarais, line 45: Do we want to reuse the previous error "threeDSecureParams or jwt is null" here?

sarahkoop: I don't know that felt like leaking internal implementation details - the merchant shouldn't know about that params or JWT object. What do you think?

sarahkoop: Ah ok - put it back to the params/jwt null error message for troubleshooting!

jaxdesmarais (approved):

richherrera (approved):

PR #1162, opened 2024-09-19
Fix `os_version` and `enablePayPalAppSwitch`

richherrera:
### Summary of changes

 - Fix the `os_version` so it’s sent as a String instead of an integer. This is to follow the same approach as iOS.
 - Fix the overwriting of `payPalVaultRequest.enablePayPalAppSwitch`. Previously, it only checked if it was a vault request, and if the PayPal app was installed, the value was being overwritten, causing incorrect behavior.

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### Authors

* @richherrera

### DIFF
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
@@ -44,8 +44,8 @@ internal class PayPalInternalClient(
                 val url = "/v1/$endpoint"
                 val appLinkReturn = if (isBillingAgreement) appLink else null
 
-                if (isBillingAgreement) {
-                    (payPalRequest as PayPalVaultRequest).enablePayPalAppSwitch = isPayPalInstalled(context)
+                if (isBillingAgreement && (payPalRequest as PayPalVaultRequest).enablePayPalAppSwitch) {
+                    payPalRequest.enablePayPalAppSwitch = isPayPalInstalled(context)
                 }
 
                 val requestBody = payPalRequest.createRequestBody(
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalVaultRequest.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalVaultRequest.kt
@@ -89,7 +89,7 @@ class PayPalVaultRequest
 
         if (enablePayPalAppSwitch && !appLink.isNullOrEmpty() && !userAuthenticationEmail.isNullOrEmpty()) {
             parameters.put(ENABLE_APP_SWITCH_KEY, enablePayPalAppSwitch)
-            parameters.put(OS_VERSION_KEY, Build.VERSION.RELEASE)
+            parameters.put(OS_VERSION_KEY, Build.VERSION.SDK_INT.toString())
             parameters.put(OS_TYPE_KEY, "Android")
             parameters.put(MERCHANT_APP_RETURN_URL_KEY, appLink)
         }
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalVaultRequest.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalVaultRequest.kt
@@ -89,7 +89,7 @@ class PayPalVaultRequest
 
         if (enablePayPalAppSwitch && !appLink.isNullOrEmpty() && !userAuthenticationEmail.isNullOrEmpty()) {
             parameters.put(ENABLE_APP_SWITCH_KEY, enablePayPalAppSwitch)
-            parameters.put(OS_VERSION_KEY, Build.VERSION.SDK_INT)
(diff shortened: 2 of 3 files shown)


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -44,8 +44,8 @@ internal class PayPalInternalClient(
                 val url = "/v1/$endpoint"
                 val appLinkReturn = if (isBillingAgreement) appLink else null
 
-                if (isBillingAgreement) {
-                    (payPalRequest as PayPalVaultRequest).enablePayPalAppSwitch = isPayPalInstalled(context)
+                if (isBillingAgreement && (payPalRequest as PayPalVaultRequest).enablePayPalAppSwitch) {
```
richherrera, line 47: the feature could be turned off in settings, but during this validation, we would set `enablePayPalAppSwitch` to true if the Venice app was installed, which caused an odd flow

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalVaultRequest.kt
```diff
@@ -89,7 +89,7 @@ class PayPalVaultRequest
 
         if (enablePayPalAppSwitch && !appLink.isNullOrEmpty() && !userAuthenticationEmail.isNullOrEmpty()) {
             parameters.put(ENABLE_APP_SWITCH_KEY, enablePayPalAppSwitch)
-            parameters.put(OS_VERSION_KEY, Build.VERSION.SDK_INT)
+            parameters.put(OS_VERSION_KEY, Build.VERSION.RELEASE)
```
stechiu, line 92: Does setting this to `.RELEASE` cover all future major and minor versions too?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
jaxdesmarais: Should we also check if the email is present here or are we doing that check elsewhere?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalVaultRequest.kt
jaxdesmarais: ➕ 1 - I'm not sure what the difference between `SDK_INT` and `RELEASE` results in

richherrera: Previously, an integer was being sent instead of a string. Regarding your question, `SDK_INT` refers to the `API level`, while `RELEASE` refers to the `version number`, as shown in the table in this [link](https://en.wikipedia.org/wiki/Android_version_history). Do you think it's better to keep sending `SDK_INT` and just cast it to a string?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
richherrera: We check that [here](https://github.com/braintree/braintree_android/blob/2b7ff747e520c231bc57560702d000b4f920034b/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalVaultRequest.kt#L90)

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalVaultRequest.kt
tdchow: I think casting the existing `SDK_INT` value might make things easier on the reporting end since the values will all be whole integers.

Separate question - is this value sent to the API for reporting purposes?

richherrera: Done: [97363bb](https://github.com/braintree/braintree_android/pull/1162/commits/97363bbf1dbf1258c6cbbbaae23524cc75326c83)

tdchow (approved):

jaxdesmarais (approved):

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalVaultRequest.kt
jaxdesmarais: Oh actually I just noticed we don't pass the SDK version here on iOS, we pass `UIDevice.current.systemVersion` - so want to triple check which is the correct value.

richherrera: > Separate question - is this value sent to the API for reporting purposes?

For now, that’s correct, it’s just for reporting purposes.

> Oh actually I just noticed we don't pass the SDK version here on iOS, we pass UIDevice.current.systemVersion - so want to triple check which is the correct value.

After confirming, since it’s only used for reporting purposes for now, we’ll send the `SDK API Level`. This will allow us to have more granularity in the future if needed.

PR #1165, opened 2024-09-25
Edit fi v5 feature unit tests

warmkesselj:
### Summary of changes

Adding unit tests for editFI feature

### Checklist

 - [ ] Added a changelog entry
 - [X] Relevant test coverage


PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalVaultErrorHandlingEditRequestUnitTest.java
```diff
@@ -0,0 +1,25 @@
...
+import static junit.framework.Assert.assertEquals;
+import static junit.framework.Assert.assertFalse;
+import static junit.framework.Assert.assertNotNull;
+import static junit.framework.Assert.assertNull;
+
+import com.braintreepayments.api.paypal.vaultedit.PayPalVaultErrorHandlingEditRequest;
+
+import org.junit.Test;
+
+public class PayPalVaultErrorHandlingEditRequestUnitTest {
```
sarahkoop, line 12: What do folks think about not having unit tests for standard Kotlin data objects? @tdchow @saperi22

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -247,19 +248,24 @@ internal class PayPalInternalClient(
...
+            callback(correlationId, null)
+        } else {
+            braintreeClient.getConfiguration { configuration, error ->
+                if (error != null) {
+                    callback(null, BraintreeException("An unexpected error occurred"))
+                } else {
+                    val clientMetadataId = dataCollector.getClientMetadataId(
+                        context,
+                        configuration,
+                        false
```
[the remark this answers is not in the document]
KunJeongPark: yes, that's correct as far as I know. In iOS, we take in a` riskCorrelationID` for error handling flow and pass that into the dataCollector function and for non error handling flow, we generate one in dataCollector function. For both cases, we pass in the riskCorrelationID in the payload for the gateway generate editFI url API. 

For error handling flow, the dataCollector function just returns the same id passed in by the merchant, if you pass in nil into the dataCollector function as in the non-error handling flow, it generates a new ID.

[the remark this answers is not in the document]
jaxdesmarais: That's correct, we are missing this logic: https://github.com/braintree/braintree_ios/blob/88c99c12ea7ef9d3b40f8242c88ef50f8aab9030/Sources/BraintreePayPal/BTPayPalClient.swift#L498-L502

Then we pass it to the parameters/gateway here: https://github.com/braintree/braintree_ios/blob/88c99c12ea7ef9d3b40f8242c88ef50f8aab9030/Sources/BraintreePayPal/BTPayPalClient.swift#L507

Which is then returned through as the result: https://github.com/braintree/braintree_ios/blob/88c99c12ea7ef9d3b40f8242c88ef50f8aab9030/Sources/BraintreePayPal/BTPayPalClient.swift#L691

> For error handling flow, the dataCollector function just returns the same id passed in by the merchant

While it returns the same ID there is linking that is happening behind the scenes in the Magnes SDK that allows us to build a risk profile. That to say, we cannot simply return the ID, we do still need to call Magnes in these cases.

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalClientUnitTest.java
```diff
@@ -507,7 +514,43 @@ public void tokenize_whenPayPalInternalClientTokenizeResult_callsBackResult()
     }
 
     @Test
-<<<<<<< HEAD
+    public void edit_auth_payPalVaultEditRequest() {
```
jaxdesmarais, line 517: Lets rename these methods to match the examples already in this class. They all follow a similar pattern that these new tests do not follow. Seems like this one should be: `createEditAuthRequest_whenEditRequest_sendsPayPalEditRequestViaInternalClient`

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalVaultErrorHandlingEditRequestUnitTest.java
jaxdesmarais: I think maybe these are supposed to be more similar to `PayPalVaultRequestUnitTest` and `PayPalCheckoutRequestUnitTest` where we are instead checking that we are building the parameters as expected? It looks like for edit FI we are currently instead doing that in the `PayPalInternalClient` which maybe is not the pattern we want to follow, not sure though.

tdchow: I also agree - we shouldn't need to test the getting/setting of properties on data classes. However if the data class contains any logic, we should have coverage for that.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -172,14 +174,18 @@ internal class PayPalInternalClient(
             ?: redirectUri.getQueryParameter("token")
     }
 
+
```
jaxdesmarais, line 177:
```suggestion
```

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -247,19 +253,28 @@ internal class PayPalInternalClient(
...
             } else {
-                val clientMetadataId = dataCollector.getClientMetadataId(
-                    context,
-                    configuration,
+                val request = DataCollectorRequest(false, correlationId)
+
+                val dataCollectorRequest = DataCollectorInternalRequest(
                     false
+                ).apply {
+                    applicationGuid = dataCollector.getPayPalInstallationGUID(context)
```
jaxdesmarais, line 268: Do we need to be doing this apply on the `applicationGuid`? I am not sure the context of what this does on Android

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -247,19 +253,28 @@ internal class PayPalInternalClient(
...
+                    clientMetadataId = correlationId
+                }
+                val clientMetaDataId = dataCollector.getClientMetadataId(
```
jaxdesmarais, line 271:
```suggestion
                val clientMetadataId = dataCollector.getClientMetadataId(
```

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalClientUnitTest.java
```diff
@@ -506,6 +510,43 @@ public void tokenize_whenPayPalInternalClientTokenizeResult_callsBackResult()
         verify(braintreeClient).sendAnalyticsEvent(PayPalAnalytics.TOKENIZATION_SUCCEEDED, params);
     }
 
+    @Test
+    public void createEditAuthRequest_whenEditRequest_sendsPayPalEditRequestViaInternalClient() {
```
jaxdesmarais, line 514: It looks like we just have tests around the edit auth request. Do we also need tests around the public edit method similar to the tokenize tests in this file?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -247,16 +252,25 @@ internal class PayPalInternalClient(
...
         callback: (String?) -> Unit
     ) {
         braintreeClient.getConfiguration { configuration, error ->
             if (error != null) {
                 callback(error("No Client Metadata Id"))
             } else {
-                val clientMetadataId = dataCollector.getClientMetadataId(
-                    context,
-                    configuration,
+                val request = DataCollectorRequest(false, correlationId)
```
sarahkoop, line 262: I'm thinking we probably want to add the `hasUserLocationConsent` param to the request object and pass it through here. If we always just pass false, Magnes is limited in the data it collects. I'm not sure if it's needed in this case though

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalVaultEditRequestUnitTest.java
```diff
@@ -0,0 +1,22 @@
...
+import static junit.framework.Assert.assertEquals;
+import static junit.framework.Assert.assertFalse;
+import static junit.framework.Assert.assertNotNull;
+import static junit.framework.Assert.assertNull;
+
+import com.braintreepayments.api.paypal.vaultedit.PayPalVaultEditRequest;
+
+import org.junit.Test;
+
+public class PayPalVaultEditRequestUnitTest {
```
sarahkoop, line 12: I think this can be deleted too

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
warmkesselj: The existing PayPal is doing the same thing as we're doing here. Also, this function seems to be updating state for both the DataCollector and BraintreeSharedPreferences.

jaxdesmarais (approved):

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
sarahkoop, line 262: We can address whether the param should be true/false in a future PR, but let's still make sure the implementation is correct in this PR aside from that parameter. This `request` object isn't currently being used - we should pass it to `dataCollector.collectDeviceData` rather than calling `dataCollector.getClientMetadataId` directly

sarahkoop (approved):

PR #1169, opened 2024-09-26
Fix PendingRequestConstructors

sarahkoop:
### Summary of changes

 - Make `PendingRequest.Started` constructors public because they are required to reconstruct from stored string to pass to `handleReturnToApp`

### Checklist

### Authors
@sarahkoop


jaxdesmarais (approved):

sarahkoop:
> Great catch! I'm confused why we weren't seeing compile issues due to the constructor being internal 🤔
> 
> https://github.com/braintree/braintree_android/blob/ef22d3f0d4691f604cd5ba7e37d439fd2a53e387/Demo/src/main/java/com/braintreepayments/demo/PendingRequestStore.java#L53
> 
> I would assume that this line would cause an error.

Yeah I am also confused by that - it seems like Java interop maybe doesn't understand the `internal` keyword?

PR #1171, opened 2024-09-27
App link setup docs

sarahkoop:
### Summary of changes

 - Add guide for setting up app links since it is now required for the PayPal flow. 

### Checklist

### Authors
@sarahkoop

### DIFF
--- a/APP_LINK_SETUP.md
+++ b/APP_LINK_SETUP.md
@@ -1,6 +1,6 @@
 ## Set Up App Links
 
-In order to use the PayPal flow your application must be set up for [App Links](https://developer.android.com/training/app-links?authuser=2).
+In order to use the PayPal flow your application must be set up for [App Links](https://developer.android.com/training/app-links).
 
 ### Register App Link in Control Panel
 
--- a/APP_LINK_SETUP.md
+++ b/APP_LINK_SETUP.md
@@ -1,6 +1,6 @@
 ## Set Up App Links
 
-In order to use the PayPal App Switch flow your application must be set up for [App Links](https://developer.android.com/training/app-links?authuser=2).
+In order to use the PayPal flow your application must be set up for [App Links](https://developer.android.com/training/app-links?authuser=2).
 
 ### Register App Link in Control Panel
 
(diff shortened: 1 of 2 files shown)


v5_MIGRATION_GUIDE.md
```diff
@@ -386,7 +386,7 @@ before the `OnResume` method of your Activity or Fragment.
 do not need to be instantiated in `OnCreate`.
 
 The PayPal integration now requires an Android App link be configured to return to your app from the 
-PayPal flow. 
+PayPal flow. See the [App Link Setup Guide](APP_LINK_SETUP.md) for more information.
```
jaxdesmarais, line 389: Instead of linking to a new markdown file do we want to link to the dev docs for app switch? I know they are currently under the v4 section which we call out, but could update to the v5 links once the docs are updated.

sarahkoop: It seemed a little confusing to link directly to those since they reference app switch and app links are now required on Android for any PayPal flow, not just app switch. I think for the v5 dev docs we should have a separate page for app link setup that we can refer to from any of the pages (PayPal web or app switch)

APP_LINK_SETUP.md
```diff
@@ -0,0 +1,34 @@
+## Set Up App Links
+
+In order to use the PayPal App Switch flow your application must be set up for [App Links](https://developer.android.com/training/app-links?authuser=2).
```
jaxdesmarais, line 3: Should this just say PayPal flow vs App Switch?

jaxdesmarais (approved):

APP_LINK_SETUP.md
sarahkoop: Ah yes good catch - updated

saperi22 (approved):

APP_LINK_SETUP.md
```diff
@@ -0,0 +1,34 @@
+## Set Up App Links
+
+In order to use the PayPal flow your application must be set up for [App Links](https://developer.android.com/training/app-links?authuser=2).
```
tdchow, line 3:
```suggestion
In order to use the PayPal flow your application must be set up for [App Links](https://developer.android.com/training/app-links).
```

Not sure what the query param is used for, but the link works with it removed.

tdchow (approved):

v5_MIGRATION_GUIDE.md
scannillo: Also like the idea of having this live in the Docs vs the repo when we start Android V5 docs!

PR #1179, opened 2024-10-03
Edit fi v5 feature pp analytics

warmkesselj:

### Summary of changes

 - 

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
@@ -362,28 +362,31 @@ class PayPalClient internal constructor(
         braintreeClient.sendAnalyticsEvent(PayPalAnalytics.EDIT_FI_STARTED, analyticsParams)
 
         if (braintreeClient.authorization is TokenizationKey) {
-            callback.onPayPalVaultEditAuthRequest(
-                PayPalVaultEditAuthRequest.Failure(
-                    BraintreeException("Invalid authorization. This feature can only be used with a client token.")
+            callBackEditFIFailed(
+                callback,
+                BraintreeException(
+                    "Invalid authorization. This feature can only be used with a client token."
                 )
             )
+
             return
         }
 
         internalPayPalClient.sendVaultEditRequest(context, payPalVaultEditRequest) { result, error ->
             if (error != null) {
-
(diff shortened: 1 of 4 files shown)


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -364,11 +369,15 @@ class PayPalClient internal constructor(
...
 
+        braintreeClient.sendAnalyticsEvent(PayPalAnalytics.EDIT_FI_STARTED)
+
         internalPayPalClient.sendVaultEditRequest(context, payPalVaultEditRequest) { result, error ->
             if (error != null) {
+                braintreeClient.sendAnalyticsEvent(PayPalAnalytics.EDIT_FI_FAILED)
                 callback.onPayPalVaultEditAuthRequest(PayPalVaultEditAuthRequest.Failure(error))
             }
 
+            braintreeClient.sendAnalyticsEvent(PayPalAnalytics.EDIT_FI_SUCCEEDED)
```
jaxdesmarais, line 380: I think instead of firing these at each callback callsite we want to instead use a helper method like we do for the tokenize flow. Let's match a similar pattern to `callbackCreatePaymentAuthFailure`/`callbackBrowserSwitchCancel`/`callbackTokenizeFailure`/`callbackTokenizeSuccess` for edit FI.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -355,33 +377,67 @@ class PayPalClient internal constructor(
...
+    ) {
+        braintreeClient.sendAnalyticsEvent(PayPalAnalytics.EDIT_FI_FAILED, analyticsParams)
+
+        val failure = PayPalVaultEditAuthRequest.Failure(error)
+
+        callback.onPayPalVaultEditAuthRequest(failure)
+    }
+
+    @OptIn(ExperimentalBetaApi::class)
+    private fun callbackEditFiSucceeded(
```
sarahkoop, line 426: Nitpick can we be consistent if we are calling it `EditFi` or `EditFI`

jaxdesmarais (approved):

PR #1180, opened 2024-10-04
[open dev] Cleanup integration tests

sarahkoop:
### Summary of changes

 - Fix broken integration tests
 - Cleanup ignored tests

### Checklist

### Authors
@sarahkoop


Demo/src/androidTest/java/com/braintreepayments/demo/test/PayPalBrowserSwitchTest.java
```diff
@@ -1,54 +0,0 @@
...
-import static com.braintreepayments.AutomatorAction.click;
-import static com.braintreepayments.AutomatorAssertion.text;
-import static com.braintreepayments.DeviceAutomator.onDevice;
-import static com.braintreepayments.UiObjectMatcher.withText;
-import static com.braintreepayments.UiObjectMatcher.withTextStartingWith;
-import static org.hamcrest.CoreMatchers.containsString;
-import static org.hamcrest.CoreMatchers.endsWith;
-
-@RunWith(AndroidJUnit4ClassRunner.class)
-public class PayPalBrowserSwitchTest extends TestHelper {
```
sarahkoop, line 20: I opted to remove these tests since we no longer have a mocked sandbox PayPal flow so they don't work - the PayPal flow continues to change and there are many paths (ex: app switch) that aren't included in these, so I don't think they provide much value. What do others think?

Demo/src/androidTest/java/com/braintreepayments/demo/test/VenmoClientTest.java
```diff
@@ -1,43 +0,0 @@
...
-import static org.junit.Assume.assumeTrue;
-
-@RunWith(AndroidJUnit4ClassRunner.class)
-public class VenmoClientTest extends TestHelper {
-
-    private static final String VENMO_PACKAGE_NAME = "com.venmo";
-
-    @Ignore("Disable until physical device test automation implemented")
-    @Test(timeout = 120000)
-    public void tokenizesVenmo() {
```
sarahkoop, line 29: We pretty much rely on manual testing for Venmo - this test has been ignored for many years, I think at this point we can remove and revisit in the future if we ever do have device automation

tdchow (approved):
Thanks for updating these. I'm guilty of making changes that break instrumentation tests 😅 I wonder if we can add UI tests to our CI checks.

Demo/src/androidTest/java/com/braintreepayments/demo/test/PayPalBrowserSwitchTest.java
tdchow: I think removing them makes sense. We could hit the sandbox endpoints in the instrumentation tests for now. A long term solution might be to have API responses bundled in the tests for specific flows.

jaxdesmarais (approved):
Agree with Tim, we should figure out how to get these to run on CI. I am also guilty of completely forgetting they exist. 😅

PR #1187, opened 2024-10-14
Shopper insights button analytics -- add new params

saperi22:
### Summary of changes

 - Adding new parameters to be tracked for analytics. Addresses: DTMOBILES-1051

### Checklist

 - [x] Added a changelog entry
 - [ ] Relevant test coverage

- @saperi22 
- @richherrera was kind enough to get me screenshots from Jupyter notebooks.


ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/ShopperInsightsClient.kt
```diff
@@ -50,10 +51,14 @@ class ShopperInsightsClient internal constructor(
      */
     fun getRecommendedPaymentMethods(
         request: ShopperInsightsRequest,
+        experiment: String? = null,
```
tdchow, line 54: Can we update the kdoc to call out this new parameter?

ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/ShopperInsightsClient.kt
```diff
@@ -154,8 +159,14 @@ class ShopperInsightsClient internal constructor(
      * Call this method when the PayPal button has been successfully displayed to the buyer.
      * This method sends analytics to help improve the Shopper Insights feature experience.
      */
-    fun sendPayPalPresentedEvent() {
-        braintreeClient.sendAnalyticsEvent(PAYPAL_PRESENTED)
+    fun sendPayPalPresentedEvent(experiment: String? = null, buttonRank: Int? = null) {
```
tdchow, line 162: Same here, can we update the kdoc for the new parameters?

ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/ShopperInsightsClient.kt
```diff
@@ -170,8 +181,14 @@ class ShopperInsightsClient internal constructor(
      * Call this method when the Venmo button has been successfully displayed to the buyer.
      * This method sends analytics to help improve the Shopper Insights feature experience.
      */
-    fun sendVenmoPresentedEvent() {
-        braintreeClient.sendAnalyticsEvent(VENMO_PRESENTED)
+    fun sendVenmoPresentedEvent(experiment: String? = null, buttonRank: Int? = null) {
```
tdchow, line 184: Same here.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEvent.kt
```diff
@@ -8,5 +8,7 @@ internal data class AnalyticsEvent(
     val isVaultRequest: Boolean = false,
     val startTime: Long? = null,
     val endTime: Long? = null,
-    val endpoint: String? = null
+    val endpoint: String? = null,
+    val experiment: String? = null,
+    val buttonRank: Int? = null
```
tdchow, line 13: Curious - was there an issue with using delegation or inheritance to not have the base analytics class have module specific properties? 

also is there a good way we can document these properties even though it's internal to the SDK?

saperi22: > was there an issue with using delegation or inheritance to not have the base analytics class have module specific properties?

No, I didn't want to do that as part of this ticket. After looking at the way it is setup, I think grouping the events into modules within this `AnalyticsEvent` class should be sufficient. We use `putOpt` and only add the relevant parameters to the events, so there isn't extra bloat.
IMO The delegation pattern example you've shared would probably be overly verbose for what we are doing here. If there's a cleaner way and more concise way, I'd prefer that.

saperi22: > also is there a good way we can document these properties even though it's internal to the SDK?

I'd think we'd want to break the properties logically into module based properties. Then it'd probably be easier to document them.
Let me address these issues one by one.

Good callouts!

saperi22:
@tdchow I don't want to hold up this PR for the refactor since works and that'd would be considered an improvement. It'll give me time to try a couple different options to figure out which works best. Let me know if you are ok with me moving the refactor to a separate PR.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEvent.kt
saperi22: Actually, some of these events are shared by multiple modules. So, there isn't necessarily a proper separation. This is probably the best we'll do on this.

For the documentation though, `buttonRank` could be module specific, but `experiment`, though currently limited to one module, could be anything and could be used by any module. Not sure what the best course of action is. I'll probably just keep it as is.

jaxdesmarais, line 13: I think we moved forward with calling this `paymentMethodsDisplayed` as an array of strings: https://github.com/braintree/braintree_ios/blob/78c6c12030c94805602d2f1bdebf7c9b540152ef/Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift#L110. Then we compact that down as a string with the key `payment_methods_displayed` to FPTI.

tdchow:
> @tdchow I don't want to hold up this PR for the refactor since works and that'd would be considered an improvement. It'll give me time to try a couple different options to figure out which works best. Let me know if you are ok with me moving the refactor to a separate PR.

I think we can refactor later down the line when the number of analytic params increases. No need to hold up your PR or even continue with the refactor in a follow-up.

However, we still might want to document the new properties somehow. My only worry is that it'll add confusion for new folks who join the team who do not have context on this work.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEvent.kt
saperi22: I'll fix that up.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
```diff
@@ -2,12 +2,30 @@ package com.braintreepayments.api.core
 
 import androidx.annotation.RestrictTo
 
+/**
```
saperi22, line 5: @tdchow does this look good?

jaxdesmarais (approved):
Assuming we've verified things look good in FPTI, this looks great to me!

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
tdchow: looks good!

tdchow (approved):
Looks great! thanks for adding the docs.

PR #1193, opened 2024-10-23
Update CMID Property Passed to the Gateway

jaxdesmarais:
### Summary of changes

- Rename `risk_correlation_id` to `correlation_id` to match what should be passed to the gateway for Edit FI/Error Handling.
- Update unit tests

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
@@ -211,7 +211,7 @@ internal class PayPalInternalClient(
     ) {
         val params = editFiParameters(payPalVaultEditRequest.editPayPalVaultId).toMutableMap()
 
-        params["risk_correlation_id"] = riskCorrelationId
+        params["correlation_id"] = riskCorrelationId
 
         val jsonObject = JSONObject(params.toMap())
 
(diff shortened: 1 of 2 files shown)


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -211,7 +211,7 @@ internal class PayPalInternalClient(
     ) {
         val params = editFiParameters(payPalVaultEditRequest.editPayPalVaultId).toMutableMap()
 
-        params["risk_correlation_id"] = riskCorrelationId
+        params["correlation_id"] = riskCorrelationId
```
tdchow, line 214: Is there any risk of older versions still sending `risk_correlation_id` to the gateway? Was a server-side change recently made?

jaxdesmarais: This is for edit FI and what we are sending today as `risk_correlation_id` they are ignoring since the actual property name is supposed to be `correlation_id`. No changes were made, we just never were provided this property name in the contract. Not sure where this name was pulled from.

warmkesselj (approved):

tdchow (approved):

PR #1198, opened 2024-10-30
Update heroku server endpoints

warmkesselj:
### Summary of changes

Updating the url endpoints for the new Heroku apps ruby server.


tdchow (approved):
Looks good! Have we tested the demo app with this new URL?

warmkesselj:
> Looks good! Have we tested the demo app with this new URL?

I've run through on the iOS app but will be running through Android today. Would be nice to have another pair of eyes check it out though.

scannillo (approved):

jaxdesmarais (approved):

PR #1202, opened 2024-11-05
Add MerchantRepository for holding BraintreeClient properties

tdchow:
### Summary of changes

 - Create MerchantRepository which holds properties that were previously in BraintreeClient
 - MerchantRepository singleton that keeps fields in memory, allowing any class to access its values via dependency injection

### Checklist

 - [ ] Added a changelog entry
 - [X] Relevant test coverage

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/MerchantRepository.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/MerchantRepository.kt
@@ -4,6 +4,9 @@ import android.content.Context
 import android.net.Uri
 import androidx.annotation.RestrictTo
 
+/**
+ * An internal repository that holds properties set by the integrating merchant.
+ */
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 class MerchantRepository {
 
@@ -14,6 +17,10 @@ class MerchantRepository {
     var appLinkReturnUri: Uri? = null
 
     companion object {
+
+        /**
+         * Singleton instance of the MerchantRepository.
+         */
         val instance: MerchantRepository by lazy { MerchantRepository() }
     }
 }
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/MerchantRepository.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/MerchantRepository.kt
@@ -1,0 +1,19 @@
+package com.braintreepayments.api.core
+
+import android.content.Context
+import android.net.Uri
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class MerchantRepository {
+
+    lateinit var applicationContext: Context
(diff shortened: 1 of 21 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/core/MerchantRepository.kt
```diff
@@ -0,0 +1,19 @@
+package com.braintreepayments.api.core
+
+import android.content.Context
+import android.net.Uri
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class MerchantRepository {
```
tdchow, line 8: Not sure if there's a clearer name for this class.

sarahkoop: Maybe a doc string for our internal understanding of what this class is for?

sarahkoop (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/MerchantRepository.kt
jaxdesmarais: ➕ 1 I think some docstrings would help, I don't have any issues with the name since this naming seems like a common pattern on Android.

jaxdesmarais (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/MerchantRepository.kt
tdchow: does the `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)` prevent dokka from publishing to the generated docs?

richherrera (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/MerchantRepository.kt
tdchow: It looks like those internal classes do get published :(

https://braintree.github.io/braintree_android/BraintreeCore/com.braintreepayments.api.core/index.html

I'll look to see if there's a way to prevent this.

saperi22: > does the `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)` prevent dokka from publishing to the generated docs?

Not sure if it does that, but you could add `@suppress` in the docs to suppress it.
Also see: https://github.com/Kotlin/dokka/issues/2404 for a custom implementation we can do to suppress all `@RestrictTo` annotated methods.

tdchow: Dang, I thought there'd be support directly in Dokka for that. I've added some docs for this class.

I'll create a tech debt ticket for possibly adding a custom implementation to suppress based on the annotation.

PR #1203, opened 2024-11-07
Add SdkComponent and move configuration fetch out of BraintreeClient

tdchow:
### Summary of changes

 - Created `ConfigurationLoaderResult` to make it easier to handle `ConfigurationLoaderCallback.onResult()`
 - Move the fetching of configuration to `AnalyticsClient`
 - Move latency analytics call to configuration loader
 - Create SdkComponent for instantiating classes that require Context

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
@@ -27,17 +27,14 @@ internal class AnalyticsClient(
     private val applicationContext: Context
         get() = merchantRepository.applicationContext
 
-    fun sendEvent(
-        event: AnalyticsEvent,
-        integration: IntegrationType?,
-    ) {
+    fun sendEvent(event: AnalyticsEvent) {
         configurationLoader.loadConfiguration { result ->
             if (result is ConfigurationLoaderResult.Success) {
                 scheduleAnalyticsWriteInBackground(event, merchantRepository.authorization)
                 scheduleAnalyticsUploadInBackground(
                     configuration = result.configuration,
                     authorization = merchantRepository.authorization,
-                    integration = integration
+                    integration = merchantRepository.integrationType
                 )
             }
         }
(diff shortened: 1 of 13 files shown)


saperi22 (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
```diff
@@ -15,28 +15,29 @@ import java.util.concurrent.TimeUnit
...
-        authorization: Authorization
-    ): UUID {
-        scheduleAnalyticsWriteInBackground(event, authorization)
-        return scheduleAnalyticsUploadInBackground(
-            configuration,
-            authorization,
-            integration
-        )
+    fun sendEvent(event: AnalyticsEvent) {
+        configurationLoader.loadConfiguration { result ->
```
tdchow, line 31: Added the fetching of Configuration here

BraintreeCore/src/main/java/com/braintreepayments/api/core/SdkComponent.kt
```diff
@@ -0,0 +1,51 @@
...
+            return instance ?: SdkComponent(applicationContext).also { sdkComponent ->
+                instance = sdkComponent
+            }
+        }
+
+        /**
+         * Returns the instance of [SdkComponent]
+         */
+        fun getInstance(): SdkComponent {
+            return checkNotNull(instance)
```
jaxdesmarais, line 33: If this ever is null what happens at that point?

jaxdesmarais (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/SdkComponent.kt
tdchow: It would throw an `IllegalStateException`. However this should never happen since `create()` would be called before `getInstance()`, which would set the singleton instance.

saperi22 (approved):

PR #1211, opened 2024-11-15
Update paypal request

saperi22:
### Summary of changes

 - Changes for enabling app-switch for one time checkout flow.
 - Added TestParameterInjector for parameterized unit tests. Fewer tests, more coverage.

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

@saperi22

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,11 @@
 ## unreleased
 
 * PayPal
-  * Add `enablePayPalAppSwitch` property to `PayPalCheckoutRequest`
+    * Add PayPal App Switch for one time checkout flow (BETA)
+        * Add `enablePayPalAppSwitch` property to `PayPalCheckoutRequest` for App Switch support 
+        * Require `PayPalCheckoutRequest.userAuthenticationEmail` for App Switch support
+        * Require `PayPalClient.appLinkReturnUrl` for App Switch support
+        * **Note:** This feature is currently in beta and may change or be removed in future releases.
 
 ## 5.2.0 (2024-10-30)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* PayPal
+  * Add `enablePayPalAppSwitch` property to `PayPalCheckoutRequest`
+
 ## 5.2.0 (2024-10-30)
 
 * GooglePay
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalVaultRequest.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalVaultRequest.kt
@@ -37,7 +37,7 @@ class PayPalVaultRequest
     var shouldOfferCredit: Boolean = false,
     var recurringBillingDetails: PayPalRecurringBillingDetails? = null,
     var recurringBillingPlanType: PayPalRecurringBillingPlanType? = null,
-    var enablePayPalAppSwitch: Boolean = false,
+    override var enablePayPalAppSwitch: Boolean = false,
     override var localeCode: String? = null,
     override var billingAgreementDescription: String? = null,
     override var isShippingAddressRequired: Boolean = false,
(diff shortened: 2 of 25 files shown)


gradle/libs.versions.toml
```diff
@@ -19,7 +19,8 @@ cardinal = "2.2.7-5"
 navigationSafeArgsGradlePlugin = "2.5.0"
 playServices = "19.4.0"
 junit = "4.13.2"
-robolectric = "4.11.1"
+testParameterInjector = "1.18"
+robolectric = "4.14-beta-1"
```
saperi22, line 23: https://github.com/google/TestParameterInjector/issues/5

>The [4.14-beta-1](https://github.com/robolectric/robolectric/releases/tag/robolectric-4.14-beta-1) release of Robolectric includes the new RobolectricTestParameterInjector that integrates with TestParameterInjector using the newly exposed APIs in TestParameterInjector 1.18.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalVaultRequest.kt
```diff
@@ -37,7 +37,7 @@ class PayPalVaultRequest
     var shouldOfferCredit: Boolean = false,
     var recurringBillingDetails: PayPalRecurringBillingDetails? = null,
     var recurringBillingPlanType: PayPalRecurringBillingPlanType? = null,
-    var enablePayPalAppSwitch: Boolean = false,
+    override var enablePayPalAppSwitch: Boolean = false,
```
tdchow, line 40: Nit - I can't highlight the line, but should we remove the `enablePayPalAppSwitch` kdoc to align with the other overridden properties?

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalCheckoutRequestUnitTest.java
```diff
@@ -37,11 +38,13 @@ public void newPayPalCheckoutRequest_setsDefaultValues() {
...
         assertNull(request.getBillingAgreementDescription());
         assertFalse(request.getShouldOfferPayLater());
+        assertFalse(request.getEnablePayPalAppSwitch());
+        assertNull(request.getUserAuthenticationEmail());
         assertFalse(request.getHasUserLocationConsent());
     }
 
     @Test
-    public void setsValuesCorrectly() {
+    public void setsValuesCorrectly(@TestParameter boolean appSwitchEnabled) {
```
tdchow, line 47: I'm not familiar with Robolectric's TestParameter. Will it run the test with a random value? Or run the test multiple times with different values?

tdchow (approved):

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalCheckoutRequestUnitTest.java
saperi22: You'd usually have to provide which values to run.
In this case though, since the value is a boolean, it just runs for false and true.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalVaultRequest.kt
saperi22: done

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalCheckoutRequestUnitTest.java
saperi22: BTW, to be precise, `@TestParameter` is from the [TestParameterInjector](https://github.com/google/TestParameterInjector/) library.
We can't have two `@RunWith` classes; with 4.14-beta-1 robolectric now supports integration with `TestParameterInjector`, which IMO, has a better and concise way of doing parameterized tests. Robolectric used to do it differently.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* PayPal
+  * Add `enablePayPalAppSwitch` property to `PayPalCheckoutRequest`
```
jaxdesmarais, line 6: Can we add an entry similar to what exists in `5.0.0`? That way we can more easily indicate this is in beta for the checkout flow as well.

jaxdesmarais (commented):
We will also want to change the base branch to the feature branch vs main. If one hasn't been created yet feel free to make it (I think the name is in the ticket).

jaxdesmarais (approved):

PR #1215, opened 2024-11-15
Convert HttpClient to Kotlin

tdchow:
### Summary of changes

 - Convert `HttpClient` and `SynchronusHttpClient` to Kotlin

### Checklist

 - [ ] Added a changelog entry
 - [X] Relevant test coverage


SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/HttpClient.kt
```diff
@@ -0,0 +1,110 @@
...
+        request: HttpRequest,
+        retryStrategy: RetryStrategy,
+        callback: NetworkResponseCallback?
+    ) {
+        var url: URL? = null
+        try {
+            url = request.url
+        } catch (ignore: MalformedURLException) {
+        } catch (ignore: URISyntaxException) {
+        }
```
jaxdesmarais, line 72: Are we able to add a comment or something in there blocks to indicate why we ignore these? To me they seem like something we wouldn't want to ignore, but I see we were also doing this previously.

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/HttpClient.kt
```diff
@@ -0,0 +1,110 @@
...
+                retryCountMap[url] = retryCount + 1
+            } else {
+                val message = "Retry limit has been exceeded. Try again later."
+                val retryLimitException = HttpClientException(message)
+                notifyErrorOnMainThread(callback, retryLimitException)
+            }
+        }
+    }
+
+    private fun getNumRetriesSoFar(url: URL): Int {
```
jaxdesmarais, line 88: Not sure if `getCurrentRetryCount` would be a better name for this. What do you think?

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/HttpClient.kt
```diff
@@ -0,0 +1,110 @@
...
+        val retryCount = retryCountMap[url] ?: return 0
+        return retryCount
+    }
+
+    private fun resetRetryCount(request: HttpRequest) {
+        try {
+            request.url?.let { retryCountMap.remove(it) }
+        } catch (ignore: MalformedURLException) {
+        } catch (ignore: URISyntaxException) {
+        }
```
jaxdesmarais, line 98: Same question here as above, are we able to add a comment in these blocks for why we ignore these?

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/SynchronousHttpClient.kt
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * This class performs an http request on the calling thread. The external caller is
+ * responsible for thread scheduling to ensure that this is not called on the main thread.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+internal class SynchronousHttpClient(
+    private val socketFactory: SSLSocketFactory,
+    private val parser: HttpResponseParser
+) {
+
```
jaxdesmarais, line 17:
```suggestion
```

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/SynchronousHttpClient.kt
```diff
@@ -0,0 +1,66 @@
...
+    @Throws(Exception::class)
+    fun request(httpRequest: HttpRequest): HttpResponse {
+        requireNotNull(httpRequest.path) { "Path cannot be null" }
+        val url = httpRequest.url
+
+        val startTime = System.currentTimeMillis()
```
jaxdesmarais, line 23: Not sure if this spacing makes more sense, take it or leave it:
```suggestion
        requireNotNull(httpRequest.path) { "Path cannot be null" }

        val url = httpRequest.url
        val startTime = System.currentTimeMillis()
```

SharedUtils/src/test/java/com/braintreepayments/api/sharedutils/SynchronousHttpClientUnitTest.java
```diff
@@ -137,36 +115,6 @@ public void request_whenConnectionIsHttps_setsSSLSocketFactory() throws Exceptio
         verify(connection).setSSLSocketFactory(sslSocketFactory);
     }
 
-    @Test
-    public void request_whenConnectionIsHttps_andSSLSocketFactoryIsNull_throwsSSLException()
```
jaxdesmarais, line 141: Are these tests no longer relevant with the updated logic?

jaxdesmarais (approved):
Just a few general questions, but none of them blockers.

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/HttpClient.kt
tdchow: I'm not too sure why we're catching these 😅 but from a quick look, `HttpRequest.getURL()` can throw these exceptions and it feels like we don't want to crash due to those.

I think a lot of the networking code can be refactored to remove the ambiguity, but I wanted to keep the actual code changes to a minimum other than the java to kotlin conversion.

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/HttpClient.kt
tdchow: That makes sense to me! Although I feel the rename should be done in a separate PR since we wouldn't have git history of the change since we're converting from java to kotlin in this PR.

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/SynchronousHttpClient.kt
tdchow: I feel that the new line here helps with readability, adding some separation from the class's constructor and the contents of the class.

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/HttpClient.kt
jaxdesmarais: That sounds good, yeah it's not blocking was more for my general understanding.

PR #1218, opened 2024-11-18
Add app switch analytics calls for Venmo

tdchow:
### Summary of changes

 - Add `app-switch:started` and `handle-return:started` events for the Venmo integration
 - Add `url` analytics parameter for Venmo events. This value represents the `appSwitchUrl`

I also validated these changes in Lighthouse.

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
@@ -24,8 +24,13 @@ class AnalyticsClient internal constructor(
     private val analyticsParamRepository: AnalyticsParamRepository = AnalyticsParamRepository.instance,
     private val time: Time = Time(),
     private val configurationLoader: ConfigurationLoader = ConfigurationLoader.instance,
-    private val merchantRepository: MerchantRepository = MerchantRepository.instance
+    private val merchantRepository: MerchantRepository
 ) {
+
+    constructor() : this(
+        merchantRepository = MerchantRepository.instance
+    )
+
     private val applicationContext: Context
         get() = merchantRepository.applicationContext
 
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
@@ -44,7 +44,8 @@ class AnalyticsClient internal constructor(
             endTime = analyticsEventParams.endTime,
             endpoint = analyticsEventParams.endpoint,
             experiment = analyticsEventParams.experiment,
-            paymentMethodsDisplayed = analyticsEventParams.paymentMethodsDisplayed
+            paymentMethodsDisplayed = analyticsEventParams.paymentMethodsDisplayed,
+            appSwitchUrl = analyticsEventParams.appSwitchUrl
         )
         configurationLoader.loadConfiguration { result ->
             if (result is ConfigurationLoaderResult.Success) {
@@ -240,6 +241,7 @@ class AnalyticsClient internal constructor(
             .putOpt(FPTI_KEY_MERCHANT_EXPERIMENT, event.experiment)
             .putOpt(FPTI_KEY_MERCHANT_PAYMENT_METHODS_DISPLAYED,
                 event.paymentMethodsDisplayed.ifEmpty { null })
+            .putOpt(FPTI_KEY_URL, event.appSwitchUrl)
         return json.toString()
     }
 
@@ -292,6 +294,7 @@ class AnalyticsClient internal constructor(
         private const val FPTI_KEY_ENDPOINT = "endpoint"
         private const val FPTI_KEY_MERCHANT_EXPERIMENT = "experiment"
         private const val FPTI_KEY_MERCHANT_PAYMENT_METHODS_DISPLAYED = "payment_methods_displayed"
+        private const val FPTI_KEY_URL = "url"
 
         private const val FPTI_BATCH_KEY_VENMO_INSTALLED = "venmo_installed"
         private const val FPTI_BATCH_KEY_PAYPAL_INSTALLED = "paypal_installed"
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
@@ -24,12 +24,8 @@ class AnalyticsClient internal constructor(
     private val analyticsParamRepository: AnalyticsParamRepository = AnalyticsParamRepository.instance,
     private val time: Time = Time(),
     private val configurationLoader: ConfigurationLoader = ConfigurationLoader.instance,
-    private val merchantRepository: MerchantRepository
+    private val merchantRepository: MerchantRepository = MerchantRepository.instance
 ) {
-
-    constructor() : this(
-        merchantRepository = MerchantRepository.instance
-    )
 
     private val applicationContext: Context
         get() = merchantRepository.applicationContext
@@ -275,6 +271,9 @@ class AnalyticsClient internal constructor(
     }
 
     companion object {
+
+        val lazyInstance: Lazy<AnalyticsClient> = lazy { AnalyticsClient() }
+
         private const val FPTI_ANALYTICS_URL = "https://api-m.paypal.com/v1/tracking/batch/events"
 
         private const val FPTI_KEY_PAYPAL_CONTEXT_ID = "paypal_context_id"
--- a/Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoAnalytics.kt
+++ b/Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoAnalytics.kt
@@ -8,11 +8,14 @@ internal object VenmoAnalytics {
     const val TOKENIZE_SUCCEEDED = "venmo:tokenize:succeeded"
     const val APP_SWITCH_CANCELED = "venmo:tokenize:app-switch:canceled"
 
-    // Additional Detail Events
+    // Launching App Switch events
+    const val APP_SWITCH_STARTED = "venmo:tokenize:app-switch:started"
     const val APP_SWITCH_SUCCEEDED = "venmo:tokenize:app-switch:succeeded"
     const val APP_SWITCH_FAILED = "venmo:tokenize:app-switch:failed"
 
-    // App Switch events
-    const val APP_SWITCH_STARTED = "venmo:tokenize:app-switch:started"
+    // Handle return events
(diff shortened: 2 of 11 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsClient.kt
```diff
@@ -24,8 +24,13 @@ class AnalyticsClient internal constructor(
     private val analyticsParamRepository: AnalyticsParamRepository = AnalyticsParamRepository.instance,
     private val time: Time = Time(),
     private val configurationLoader: ConfigurationLoader = ConfigurationLoader.instance,
-    private val merchantRepository: MerchantRepository = MerchantRepository.instance
+    private val merchantRepository: MerchantRepository
 ) {
+
+    constructor() : this(
+        merchantRepository = MerchantRepository.instance
+    )
```
tdchow, line 32: The AnalyticsClient constructor needs to be defined as public (still scoped to the library) for the Launcher classes to instantiate them while keeping the default constructor internal to not leak internal classes. The only way I found to create an empty constructor is to pass one of the public params to the default constructor. Would love to hear if anyone has other ideas for solving this issue.

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoLauncher.kt
```diff
@@ -7,15 +7,23 @@ import com.braintreepayments.api.BrowserSwitchClient
...
+import com.braintreepayments.api.core.AnalyticsClient
 import com.braintreepayments.api.core.BraintreeException
 
 /**
  * Responsible for launching the Venmo app to authenticate users
  */
 class VenmoLauncher internal constructor(
-    private val browserSwitchClient: BrowserSwitchClient
+    private val browserSwitchClient: BrowserSwitchClient,
+    lazyAnalyticsClient: Lazy<AnalyticsClient>
```
tdchow, line 18: I ran into a crash where the SdkComponent was not yet initialized when the `VenmoLauncher` was created before the `VenmoClient`. This is because the creation of the `VenmoClient` initializes the `BraintreeClient` which creates the SdkComponent instance.

We'll need to remain flexible on the ordering of these classes.

The main issue is that `Context` is not passed into the SDK until the Client class is created. So any dependencies that reference context, such as these Launcher classes, would need to have those dependencies marked as lazy in this approach.

I think long term it may become confusing on which dependencies need to be lazy which is why I'm on the fence on this solution. But I wasn't able to find a cleaner solution that fixes this DI issue.

saperi22 (approved):

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.kt
```diff
@@ -335,7 +335,7 @@ class VenmoClient internal constructor(
 
     private val analyticsParams: AnalyticsEventParams
         get() {
-            val eventParameters = AnalyticsEventParams()
+            val eventParameters = AnalyticsEventParams(appSwitchUrl = merchantRepository.returnUrlScheme)
```
jaxdesmarais, line 338: This may be my understanding of venmo, but would this be the full URL we are attempting to open or just the scheme? I think we want the full URL, but may be misunderstanding this property on android.

jaxdesmarais (commented):
We did add a changelog entry for this on iOS: https://github.com/braintree/braintree_ios/blob/main/CHANGELOG.md#unreleased

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.kt
jaxdesmarais: Ah gotcha, I think what we actually want is this URL instead: https://github.com/braintree/braintree_android/blob/9966290f91eea97bbc27850055c98d91f7368126/Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.kt#L188 It doesn't look like we ever set the return URL as the base URL.

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoLauncher.kt
```diff
@@ -32,6 +44,10 @@ class VenmoLauncher internal constructor(
         activity: ComponentActivity,
         paymentAuthRequest: VenmoPaymentAuthRequest.ReadyToLaunch
     ): VenmoPendingRequest {
+        analyticsClient.sendEvent(
+            eventName = VenmoAnalytics.APP_SWITCH_STARTED,
+            analyticsEventParams = analyticsEventParams
+        )
```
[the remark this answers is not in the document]
tdchow: Are we treating `venmo:tokenize:app-switch:started` as the success event for when merchants call `launch()`?

I originally thought this event was tracking when merchants invoked `launch()` and attempted to app/CCT switch rather than if `VenmoPendingRequest.Started` was returned. I can update this if my understanding is incorrect!

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.kt
tdchow: Oh! I misinterpreted the AC in the story. So we want to log the URL that we send to the Venmo app/CCT. Thanks for calling this out - I'll make the change.

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoLauncher.kt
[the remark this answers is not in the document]
sarahkoop: Ahh hmm I guess it could be interpreted either way. I agree it is good to have an event triggered at the start of launch, but with only that we still have a gap if there is an start failure. Although the start failures should be something caught in development (invalid manifest) so maybe less necessary for analytics. We can leave it as is and add more detailed failure events if needed in the future.

sarahkoop (approved):

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoLauncher.kt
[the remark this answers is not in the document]
tdchow:
```kotlin
const val APP_SWITCH_SUCCEEDED = "venmo:tokenize:app-switch:succeeded"
const val APP_SWITCH_FAILED = "venmo:tokenize:app-switch:failed"
```

We actually have these constants already defined. Currently `APP_SWITCH_SUCCEEDED` is sent in the incorrect spot - when tokenize is called. And `APP_SWITCH_FAILED` is unused.

I also noticed that `const val TOKENIZE_STARTED = "venmo:tokenize:started"` is incorrectly sent when the merchant calls `createPaymentAuthRequest`. I'll make the fix for this.

What are your thoughts on sending `APP_SWITCH_SUCCEEDED` and `APP_SWITCH_FAILED` based on the returned `VenmoPendingRequest`?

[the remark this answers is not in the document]
sarahkoop: We use `TOKENIZE_STARTED` as the "start" event for all flows (aligns with iOS and v4), it doesn't make as much sense for v5, but I am hesitant to change it because conversion funnels might be based on that event. If we do rename it, we should do it across all payment methods and define and new conversion event in our list of FPTI events. 

I think we want to reserve `APP_SWITCH_SUCCEEDED` and `APP_SWITCH_FAILED` for the return from app switch (unless we have other events for those).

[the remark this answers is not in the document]
tdchow: Ah, I do see that `TOKENIZE_STARTED` is sent at the top of all the flows. I won't make any changes to this event.

| Event                               | Description                                                    |
|-------------------------------------|----------------------------------------------------------------|
| venmo:tokenize:app-switch:succeeded | Fired when we are successfully able to switch to the Venmo app |
| venmo:tokenize:app-switch:failed    | Fired when an error occurs when switching to the Venmo app     |

This is what's called out in the FPTI events doc for the app-switch succeeded and failed events. It sounds like these should be sent when the Venmo app or CCT is initially launched.

The new `const val HANDLE_RETURN_STARTED = "venmo:tokenize:handle-return:started"` event is sent when the merchant calls `VenmoLauncher.handleReturnToApp()`. We could add a handle-return succeeded and failed based on the returned `VenmoPaymentAuthResult`. Do you think it'd be worth adding these?

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.kt
tdchow: Updated in this commit: 1135c26b105250621b94fd5ce19ef6354abb427f

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoLauncher.kt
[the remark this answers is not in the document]
sarahkoop: I think that sounds good!

jaxdesmarais (approved):

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoLauncher.kt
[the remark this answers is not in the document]
tdchow: Added the additional events here: f27b8ee75455e2de34ca9eb703b55183187da9bf

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoAnalytics.kt
```diff
@@ -8,7 +8,14 @@ internal object VenmoAnalytics {
...
+    // Launching App Switch events
+    const val APP_SWITCH_STARTED = "venmo:tokenize:app-switch:started"
     const val APP_SWITCH_SUCCEEDED = "venmo:tokenize:app-switch:succeeded"
     const val APP_SWITCH_FAILED = "venmo:tokenize:app-switch:failed"
+
+    // Handle return events
+    const val HANDLE_RETURN_STARTED = "venmo:tokenize:handle-return:started"
+    const val HANDLE_RETURN_SUCCEEDED = "venmo:tokenize:handle-return:succeeded"
+    const val HANDLE_RETURN_FAILED = "venmo:tokenize:handle-return:failed"
+    const val HANDLE_RETURN_NO_RESULT = "venmo:tokenize:handle-return:no-result"
```
tdchow, line 20: Thoughts on adding this `no-result` event?

It would be an Android specific event, but it'd give us some insight into when buyers are canceling via the CCT X or entering CCT picture-in-picture mode.

sarahkoop: I think it would be helpful

PR #1223, opened 2024-11-22
Add deep link support as a fallback to app links

tdchow:
### Summary of changes

 - Add optional `returnUrlScheme` property to the `PayPalClient` constructor, which is used for deep linking back to the merchant app if the user has unchecked "Open supported links".
 - Add `GetReturnLinkTypeUseCase` that returns which link type should be sent to the PayPal request and payment URL

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage


BraintreeCore/src/main/java/com/braintreepayments/api/core/GetReturnLinkTypeUseCase.kt
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Use case that returns which link type should be used for navigating from App Switch / CCT back into the merchant app.
+ *
+ * If a user unchecks the "Open supported links" checkbox in the Android OS settings for the merchant's app. If this
+ * setting is unchecked, this use case will return [ReturnLinkType.DEEP_LINK], otherwise [ReturnLinkType.APP_LINK]
+ * will be returned.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class GetReturnLinkTypeUseCase(private val merchantRepository: MerchantRepository) {
```
tdchow, line 16: Thoughts on adding a domain layer / use case pattern to the SDK? I wasn't able to find an existing pattern to share business logic outside of creating additional Client classes.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -152,10 +155,20 @@ class PayPalClient internal constructor(
...
             .launchAsNewTask(braintreeClient.launchesBrowserSwitchAsNewTask())
             .metadata(metadata)
+            .apply {
+                when (getReturnLinkTypeUseCase()) {
+                    GetReturnLinkTypeUseCase.ReturnLinkType.APP_LINK -> {
+                        appLinkUri(merchantRepository.appLinkReturnUri)
+                    }
+
+                    GetReturnLinkTypeUseCase.ReturnLinkType.DEEP_LINK -> {
+                        returnUrlScheme(merchantRepository.returnUrlScheme)
```
sarahkoop, line 168: Do we need to check that returnUrlScheme isn't null?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -50,12 +50,28 @@ internal class PayPalInternalClient(
                     payPalRequest.enablePayPalAppSwitch = isPayPalInstalled(context)
                 }
 
+                val returnLinkType = getReturnLinkTypeUseCase()
+                val navigationLink = when (returnLinkType) {
+                    GetReturnLinkTypeUseCase.ReturnLinkType.APP_LINK -> merchantRepository.appLinkReturnUri
+                    GetReturnLinkTypeUseCase.ReturnLinkType.DEEP_LINK -> merchantRepository.returnUrlScheme
```
sarahkoop, line 56: Same comment here - should we be checking that the merchant has included a returnUrlScheme? If a merchant only wants to support app links and app links are not enabled on the device we should have an error scenario I assume

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetReturnLinkTypeUseCase.kt
sarahkoop: Can you expand on what you envision for adding a domain layer?

tdchow: Yep! I see it as a layer that encapsulates business logic that can be reused across the SDK. It's also the layer between the UI and the data layer, where it reads data from the repository and applies some sort of business logic.

https://developer.android.com/topic/architecture/domain-layer

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -54,8 +56,9 @@ class PayPalClient internal constructor(
     constructor(
         context: Context,
         authorization: String,
-        appLinkReturnUrl: Uri
-    ) : this(BraintreeClient(context, authorization, null, appLinkReturnUrl))
+        appLinkReturnUrl: Uri,
+        returnUrlScheme: String? = null
```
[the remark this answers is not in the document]
tdchow: Good call, let me update the kdoc.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
tdchow: That's a good question. I think we should check if returnUrlScheme is non-null. However, how should we handle the null case?

If the buyer has the `Open supported links` unchecked, and the merchant passes a null `returnUrlScheme` the deep link would fail. Maybe we should throw an error before the app switch/CCT is launched and fail fast?

sarahkoop: Yes I think fail fast and let the merchant decide if they want to tell the user to turn on that setting or how they want to handle it if they don't want to support deep links

saperi22:
QQ: would it be possible to use the already sent in app link url for deeplink return url?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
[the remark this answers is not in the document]
jaxdesmarais: Do we need to consider any type of okay from security for adding this back? I know we were moving towards removing url schemes across the SDKs for security reasons but wasn't sure what applied with it as an opt in.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
tdchow: We're actually defaulting the returnUrlScheme to the merchant's package name appended with `.braintree`, if a null value is passed in.

https://github.com/braintree/braintree_android/blob/bafeda710b4c65780395ef2f1ca872254d648e9c/BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt#L54

So the `returnUrlScheme` in `MerchantRepository` is non null. I believe this was the existing deep link behavior.

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetReturnLinkTypeUseCase.kt
jaxdesmarais: I'm okay with this pattern. At some point we should consider documenting the Android and iOS patterns internally since they have diverged quite a bit in recent major versions. Unless this is more of a best practice in Android, in which case probably fine to leave undocumented and just slowly migrate to each languages best practices!

tdchow: I'd say MVVM is the architecture pattern that Google is pushing since they have entire doc sections on the different layers of MVVM. That said, it wouldn't hurt to have a callout somewhere in the repo of the architecture patterns the SDK is using.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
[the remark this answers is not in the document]
tdchow: I don't think we need to run this by security as our other modules are still relying on deep links for navigating back to the merchant app. I also don't think security originally required the switch to app links. @sarahkoop - do you have context on if security was originally involved with the switch to app links?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
tdchow: Similar to the comment above, we're defaulting the returnUrlScheme to the merchant's package name. Did we want to change this behavior?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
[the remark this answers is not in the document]
sarahkoop: Yes I think since we are still supporting deep links across multiple flows it's fine to add back in - this originated as a merchant request not an internal security mandate

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
tdchow: Yep, it would fail if they do not define the intent filter. Hmm I guess having it more explicit would be better for the deep link fallback. I'm thinking we may need to have a new field that stores the deep link fallback URL scheme if the other modules are still relying on the existing URL scheme with this default value.

tdchow: I just verified that we're using the existing `returnUrlScheme` with the default value for the other modules. I'll add a new param to the MerchantRepository for holding the fallback URL scheme.

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetReturnLinkTypeUseCase.kt
saperi22: Pretty widely used pattern in android. AFAIK, folks on NXO iOS also made use of this.

sarahkoop (approved):
Needs CHANGELOG but otherwise LGTM 👍

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
tdchow: I also like the explicit opt-in to use a deep link fallback. I can see some merchants only wanting to support deep links for the increased security.

Another option could be to reuse the app link and have the merchant pass in a boolean value if they want to enable the deep link fallback, but they'd still need to add the correct intent filter in their manifest. So it might be clearer to have them pass in that URL scheme they define in their manifest.

tdchow:
> Needs CHANGELOG but otherwise LGTM 👍

Good catch. I just pushed 52e9adeda61dc835b42265fe6127e277ccf3c1b3

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
saperi22: go ahead :)

PR #1226, opened 2024-12-03
Add URL to PayPal App Switch Events

jaxdesmarais:
### Summary of changes

- Add `url` to `APP_SWITCH_SUCCEEDED` and `APP_SWITCH_FAILED` events
- Add unit tests

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors

@jaxdesmarais

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
@@ -28,5 +28,5 @@ data class AnalyticsEventParams @JvmOverloads constructor(
     var endpoint: String? = null,
     val experiment: String? = null,
     val paymentMethodsDisplayed: List<String> = emptyList(),
-    var appSwitchUrl: String? = null
+    val appSwitchUrl: String? = null
 )
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
@@ -28,5 +28,5 @@ data class AnalyticsEventParams @JvmOverloads constructor(
     var endpoint: String? = null,
     val experiment: String? = null,
     val paymentMethodsDisplayed: List<String> = emptyList(),
-    val appSwitchUrl: String? = null
+    var appSwitchUrl: String? = null
 )
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
@@ -43,6 +43,11 @@ class PayPalClient internal constructor(
     private var isVaultRequest = false
 
     /**
+     * The final URL string used to open the app switch flow
+     */
+    private var appSwitchUrlString: String? = null
+
+    /**
      * Initializes a new [PayPalClient] instance
      *
      * @param context          an Android Context
@@ -187,6 +192,7 @@ class PayPalClient internal constructor(
         val isAppSwitchFlow = !switchInitiatedTime.isNullOrEmpty()
 
         if (isAppSwitchFlow) {
+            appSwitchUrlString = approvalUrl
             braintreeClient.sendAnalyticsEvent(
                 PayPalAnalytics.HANDLE_RETURN_STARTED,
                 analyticsParams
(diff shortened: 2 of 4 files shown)


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -304,7 +310,8 @@ class PayPalClient internal constructor(
         braintreeClient.sendAnalyticsEvent(PayPalAnalytics.TOKENIZATION_FAILED, analyticsParams)
 
         if (isAppSwitchFlow) {
-            braintreeClient.sendAnalyticsEvent(PayPalAnalytics.APP_SWITCH_FAILED, analyticsParams)
+            val appSwitchAnalyticsParams = analyticsParams.apply { appSwitchUrl = appSwitchUrlString }
```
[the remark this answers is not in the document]
jaxdesmarais: 🤔 it looks like that doesn't set the property since `analyticsParams` is a `val`. Changing it to a `var` gives some compiler warnings and updating to the suggestion doesn't seem to be setting the other properties. Likely just my misunderstanding on the best way to do this since I agree it feels a bit ugly as it is in the PR.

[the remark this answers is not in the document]
sarahkoop: Ah missed that - this way is good then 👍

sarahkoop (approved):

richherrera (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
```diff
@@ -28,5 +28,5 @@ data class AnalyticsEventParams @JvmOverloads constructor(
     var endpoint: String? = null,
     val experiment: String? = null,
     val paymentMethodsDisplayed: List<String> = emptyList(),
-    val appSwitchUrl: String? = null
+    var appSwitchUrl: String? = null
```
tdchow, line 31: Is it possible to keep this property a `val`? We may need to move where this field is getting set before sending the analytic event, but I see some benefits in keeping these fields immutable. It prevents state from being held in multiple places and eliminates the chance of fields getting out of date.

agedd (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
jaxdesmarais: Wouldn't we want to make that change then across the board? The properties `payPalContextId`, `linkType`, `isVaultRequest`, `startTime`, `endTime`, and `endpoint` are all `var`s in `AnalyticsEventParams`. I'm not opposed to changing the pattern but was following what looked to be an already established pattern. If we'd like to move away from the pattern it seems like we should tackle all of `AnalyticsEventParams` as a whole vs this one property. Thoughts?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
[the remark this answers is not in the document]
jaxdesmarais: Oh nice! Updated to use copy here: https://github.com/braintree/braintree_android/pull/1226/commits/45c98f720d4a0fc58c23e5f142cf12046c4f1d53

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
tdchow: Yep, all of the other properties in the events param class should ideally be `val`s, but I think this is existing tech debt. A good rule of thumb is that we should try to keep properties immutable if the parent data class is just used to pass data into another function.

tdchow: Yep, all of the other properties in the events param class should ideally be `val`s, but I think this is existing tech debt. A good rule of thumb is that we should try to keep properties immutable if the parent data class is just used to pass data into another function.

jaxdesmarais: Gotcha, yeah it is pretty confusing since that isn't the pattern we established here. In any case, update in commit https://github.com/braintree/braintree_android/pull/1226/commits/736fee1cb65e29498ee5937f03355e50ceea847a and created ticket DTMOBILES-1298 to address this tech debt.

tdchow (approved):

PR #1232, opened 2024-12-05
Add `ThreeDSecureRequest.requestorAppUrl`

jaxdesmarais:
### Summary of changes

 - Add `ThreeDSecureRequest.requestorAppUrl` - this is a new property required by Mastercard for 3DS for out of band transactions. There was an internal request from a large merchant for exposing this field.

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors

@jaxdesmarais

### DIFF
--- a/ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/CardinalClient.kt
+++ b/ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/CardinalClient.kt
@@ -109,8 +109,8 @@ internal class CardinalClient {
             cardinalConfigurationParameters.uiCustomization = it.cardinalUiCustomization
         }
 
-        request.threeDSRequestorAppUrl?.let {
-            cardinalConfigurationParameters.threeDSRequestorAppURL = request.threeDSRequestorAppUrl
+        request.requestorAppUrl?.let {
+            cardinalConfigurationParameters.threeDSRequestorAppURL = request.requestorAppUrl
         }
         try {
             Cardinal.getInstance().configure(context, cardinalConfigurationParameters)
--- a/ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/CardinalClient.kt
+++ b/ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/CardinalClient.kt
@@ -109,6 +109,9 @@ internal class CardinalClient {
             cardinalConfigurationParameters.uiCustomization = it.cardinalUiCustomization
         }
 
+        request.threeDSRequestorAppUrl?.let {
+            cardinalConfigurationParameters.threeDSRequestorAppURL = request.threeDSRequestorAppUrl
+        }
         try {
             Cardinal.getInstance().configure(context, cardinalConfigurationParameters)
         } catch (e: RuntimeException) {
--- a/ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureRequest.kt
+++ b/ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureRequest.kt
@@ -49,7 +49,7 @@ import org.json.JSONObject
  * @property customFields Optional. Object where each key is the name of a custom field which has
  * been configured in the Control Panel. In the Control Panel you can configure 3D Secure Rules
  * which trigger on certain values.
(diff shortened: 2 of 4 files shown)


ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureRequest.kt
```diff
@@ -68,7 +71,8 @@ data class ThreeDSecureRequest @JvmOverloads constructor(
     var v2UiCustomization: ThreeDSecureV2UiCustomization? = null,
     var uiType: ThreeDSecureUiType = ThreeDSecureUiType.BOTH,
     var renderTypes: List<ThreeDSecureRenderType>? = null,
-    var customFields: Map<String, String>? = null
+    var customFields: Map<String, String>? = null,
+    var threeDSRequestorAppUrl: String? = null
```
sarahkoop, line 75: Would `requestorAppUrl` be simpler since all the params are on the `ThreeDSecureRequest` or do we prefer to keep the params aligned with Cardinal's naming? Either is fine with me

jaxdesmarais: I think leaving 3DS off of it makes sense! I will update this and the iOS PR. Agree reading it with the request name is a mouthful. 😂

jaxdesmarais: https://github.com/braintree/braintree_android/pull/1232/commits/1fcc38d4f258005a96db094b199ac37ba2286a24

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/CardinalClient.kt
```diff
@@ -109,6 +109,9 @@ internal class CardinalClient {
             cardinalConfigurationParameters.uiCustomization = it.cardinalUiCustomization
         }
 
+        request.requestorAppUrl?.let {
+            cardinalConfigurationParameters.threeDSRequestorAppURL = request.requestorAppUrl
```
tdchow, line 113:
```suggestion
            cardinalConfigurationParameters.threeDSRequestorAppURL = it
```

We can reference `it` here within the `let` scoping function, which will have a non-null type. `threeDSRequestorAppURL` probably handles the null case, but using `it` here will prevent us from ever passing a null value, since `requestorAppUrl` could change between the null check and referencing the value.

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/CardinalClient.kt
```diff
@@ -109,6 +109,9 @@ internal class CardinalClient {
             cardinalConfigurationParameters.uiCustomization = it.cardinalUiCustomization
         }
 
+        request.requestorAppUrl?.let {
```
tdchow, line 112: Could we add a unit tests for this new condition?

sarahkoop (approved):

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/CardinalClient.kt
tdchow: Scope functions can get really confusing in Kotlin 🥴 Here's a good doc page on them that I'll frequently reference for the Function Selection table: https://kotlinlang.org/docs/scope-functions.html

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/CardinalClient.kt
jaxdesmarais: Ah, I had assumed this wasn't testable on Android since it isn't on iOS. But it looks like I was wrong, test added here: https://github.com/braintree/braintree_android/pull/1232/commits/d9ce940f69b7fa30d894e882aa4076b2dd7b1644

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/CardinalClient.kt
jaxdesmarais: This doc is super helpful! Thanks for sharing, I'll get there on Android/Kotlin eventually. 😅

tdchow (approved):

PR #1235, opened 2024-12-06
Shopper insights rp1 feature include session

warmkesselj:
Summary of changes

Add public property shopperSessionId to BTPaypalRequest. Value should be passed to both the create_payment_resource and setup_billing_agreement endpoints. Demo app should be updated enable/disable passing this value.

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### Authors
@stechiu 
@warmkesselj 
@jwarmkessel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
   * Add `isPayPalAppInstalled` and `isVenmoAppInstalled` methods
   * Add `shopperSessionId` parameter to `ShopperInsightsClient`
 * BraintreePayPal
-  * Add `shopperSessionInsight`
+  * Add `shopperSessionId` to `PayPalCheckoutRequest` and `PayPalVaultRequest`
 
 ## 5.2.0 (2024-10-30)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 * ShopperInsights (BETA)
   * Add `isPayPalAppInstalled` and `isVenmoAppInstalled` methods
   * Add `shopperSessionId` parameter to `ShopperInsightsClient`
+* BraintreePayPal
+  * Add `shopperSessionInsight`
 
 ## 5.2.0 (2024-10-30)
 
--- a/Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
@@ -148,7 +148,8 @@ public class PayPalFragment extends BaseFragment {
                 activity,
                 buyerEmailAddress,
                 buyerPhoneCountryCode,
-                buyerPhoneNationalNumber
+                buyerPhoneNationalNumber,
+                    null
             );
         } else {
             payPalRequest = createPayPalCheckoutRequest(
@@ -156,7 +157,8 @@ public class PayPalFragment extends BaseFragment {
                 amount,
                 buyerEmailAddress,
                 buyerPhoneCountryCode,
-                buyerPhoneNationalNumber
+                buyerPhoneNationalNumber,
+                    null
             );
         }
         payPalClient.createPaymentAuthRequest(requireContext(), payPalRequest,
--- a/Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
@@ -42,7 +42,6 @@ public class PayPalRequestFactory {
 
         if (!shopperInsightsSessionId.toString().isEmpty()) {
             request.setShopperSessionId(shopperInsightsSessionId.toString());
-            request.setUserAuthenticationEmail(buyerEmailAddress);
         }
 
         if (Settings.isPayPalAppSwithEnabled(context)) {
--- a/Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
@@ -118,21 +118,20 @@ public class PayPalRequestFactory {
         String buyerEmailAddress,
         String buyerPhoneCountryCode,
         String buyerPhoneNationalNumber,
-        Optional<String> shopperInsightsSessionId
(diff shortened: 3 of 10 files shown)


Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
```diff
@@ -18,26 +18,32 @@
...
         String buyerEmailAddress,
         String buyerPhoneCountryCode,
-        String buyerPhoneNationalNumber
+        String buyerPhoneNationalNumber,
+        String shopperInsightsSessionId
     ) {
 
         PayPalVaultRequest request = new PayPalVaultRequest(true);
 
-        if (!buyerEmailAddress.isEmpty()) {
```
sarahkoop, line 33: Do we need to change all these other params? It seems like it should be unaffected by this PR

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalRequest.kt
```diff
@@ -72,6 +72,8 @@ import org.json.JSONException
  * where the user has a PayPal Account with the same email.
  * @property userPhoneNumber User phone number used to initiate a quicker authentication flow in
  * cases where the user has a PayPal Account with the phone number.
+ * @property shopperSessionId value should be the shopper session ID returned from your server SDK
```
sarahkoop, line 75: I think this doc string was included in the ticket but I'm wondering if it's clear enough for when we are outside of the shopper insights module. Should we include something like `the shopper session ID returned from your shopper insights server SDK integration?`. Also should this param be beta/experimental API @jaxdesmarais?

Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
warmkesselj: My understanding is that the original code is potentially a run time crash if the value is null. In the case of the demo app and how its setup, setting the toggles to null didn't have an impact on this flow because the values would always be true.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalRequest.kt
warmkesselj: I like it. 
Current: 
shopperSessionId value should be the shopper session ID returned from your server SDK request.
To:
shopperSessionId  the shopper session ID returned from your shopper insights server SDK integration?

warmkesselj: I'll make the changes in iOS as well.

Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
sarahkoop: Ah ok in that case do we need to check that it's not null _and_ it's not empty?

warmkesselj: Yes, great catch!

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalRequest.kt
jaxdesmarais: Yep we can for sure give this the beta/experimental API tag

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt
```diff
@@ -126,6 +126,10 @@ class PayPalCheckoutRequest @JvmOverloads constructor(
 
         userPhoneNumber?.let { parameters.put(PHONE_NUMBER_KEY, it.toJson()) }
 
+        shopperSessionId?.let {
+            if (it.isNotEmpty()) parameters.put(SHOPPER_SESSION_ID, it)
+        }
```
sarahkoop, line 131: We should be able to use the same simplified putOpt logic from the `PayPalVaultRequest` here

sarahkoop:
Did we manually test and confirm passing this parameter is working with the checkout and vault flows?

warmkesselj:
> Did we manually test and confirm passing this parameter is working with the checkout and vault flows?

Yes, but will need to test again with the updates.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 * ShopperInsights (BETA)
   * Add `isPayPalAppInstalled` and `isVenmoAppInstalled` methods
   * Add `shopperSessionId` parameter to `ShopperInsightsClient`
+* BraintreePayPal
+  * Add `shopperSessionInsight`
```
jaxdesmarais, line 11:
```suggestion
  * Add `shopperSessionId` to `PayPalCheckoutRequest` and `PayPalVaultRequest`
```

sarahkoop (approved):

Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
```diff
@@ -148,15 +148,17 @@ private void launchPayPal(
                 activity,
                 buyerEmailAddress,
                 buyerPhoneCountryCode,
-                buyerPhoneNationalNumber
+                buyerPhoneNationalNumber,
+                    null
```
jaxdesmarais, line 152: nit: it looks like there are some extra spaces for the 2 null properties here

Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragment.kt
```diff
@@ -229,7 +231,9 @@ class ShopperInsightsFragment : BaseFragment() {
                 activity,
                 emailInput.editText?.text.toString(),
                 countryCodeInput.editText?.text.toString(),
-                nationalNumberInput.editText?.text.toString()
+                nationalNumberInput.editText?.text.toString(),
+                shopperSessionId
+
```
jaxdesmarais, line 236:
```suggestion
```

jaxdesmarais (approved):
Couple of very minor cleanup comments but otherwise looks great!

PR #1241, opened 2024-12-16
Add Shopper Session ID param

sarahkoop:
### Summary of changes

 - Remove previous updates to pass existing `sessionId` for shopper insights
 - Add new `shopper_session_id` FPTI tag to be used specifically for shopper insights and related payment flows
 - Tested manually and verified `shopper_session_id` is getting to FPTI as expected
 - Aligns with iOS PR: https://github.com/braintree/braintree_ios/pull/1485

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors
@sarahkoop

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
@@ -17,7 +17,8 @@ import androidx.annotation.RestrictTo
  * the experiment, as a JSON string, that the merchant sent to the us.
  * @property paymentMethodsDisplayed A ShopperInsights module specific event that indicates the
  * order of payment methods displayed to the shopper by the merchant.
- * @property shopperSessionId The Shopper Insights customer session ID created by a merchant's server SDK or graphQL integration.
+ * @property shopperSessionId The Shopper Insights customer session ID created by a merchant's
+ * server SDK or graphQL integration.
  */
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 data class AnalyticsEventParams @JvmOverloads constructor(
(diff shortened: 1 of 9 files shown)


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -44,6 +42,11 @@ class PayPalClient internal constructor(
      */
     private var isVaultRequest = false
 
+    /**
+     * Used for sending Shopper Insights session ID provided by merchant to FPTI
+     */
+    private var shopperSessionId: String? = null
```
tdchow, line 48: Should we save this in a repository singleton outside of the Client class? Maybe a PayPalRepository that holds all PayPal specific values.

sarahkoop: I considered adding it to the `AnalyticsParamRepository` but since it is passed in by the merchant at each step there didn't seem much value there and it seemed a little confusing that we don't do that for any other general analytics event params. Open to a new param repository or adding it to the existing one if folks think that would be more helpful!

tdchow: Ah gotcha. That makes sense 👍

tdchow (approved):

warmkesselj (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
```diff
@@ -17,6 +17,7 @@ import androidx.annotation.RestrictTo
  * the experiment, as a JSON string, that the merchant sent to the us.
  * @property paymentMethodsDisplayed A ShopperInsights module specific event that indicates the
  * order of payment methods displayed to the shopper by the merchant.
+ * @property shopperSessionId A ShopperInsights session ID provided by the merchant
```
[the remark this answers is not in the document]
sarahkoop: Updated 👍

jaxdesmarais (approved):

PR #1249, opened 2025-01-07
Remove `paymentMethodsDisplayed`

jaxdesmarais:
### Summary of changes

 - Remove `paymentMethodsDisplayed` - no longer needed for v1.5 of Shopper Insights

### Checklist

 - ~[ ] Tested and confirmed payment flows affected by this change are functioning as expected~

### Authors

@jaxdesmarais

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
@@ -15,7 +15,6 @@ import androidx.annotation.RestrictTo
  * @property endpoint The endpoint being called.
  * @property experiment Currently a ShopperInsights module specific event that indicates
  * the experiment, as a JSON string, that the merchant sent to the us.
- * order of payment methods displayed to the shopper by the merchant.
  * @property shopperSessionId The Shopper Insights customer session ID created by a merchant's
  * server SDK or graphQL integration.
  * @property buttonType buttonType Represents the tapped button type.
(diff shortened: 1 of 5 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
```diff
@@ -15,7 +15,6 @@ import androidx.annotation.RestrictTo
  * @property endpoint The endpoint being called.
  * @property experiment Currently a ShopperInsights module specific event that indicates
  * the experiment, as a JSON string, that the merchant sent to the us.
- * @property paymentMethodsDisplayed A ShopperInsights module specific event that indicates the
  * order of payment methods displayed to the shopper by the merchant.
```
sarahkoop, line 18:
```suggestion
```

sarahkoop (approved):

richherrera (approved):

saperi22:
@jaxdesmarais do we want to update the docs for the `sendPayPalPresentedEvent` to say that it is no longer required? I don't think we'd want to disturb the signature for backwards compatibility.

jaxdesmarais:
> @jaxdesmarais do we want to update the docs for the `sendPayPalPresentedEvent` to say that it is no longer required? I don't think we'd want to disturb the signature to preserve backwards compatibility.

For shopper insights we have updated all of the signatures to remove this property, but never cleaned up the FPTI tag. We have agreed with product to make these breaking changes per our docstrings and tags, so will not maintain backwards compatibility once this is released: https://github.com/braintree/braintree_android/blob/934d2ba4dd420937cd5ea043c0c41994e7378612/ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/ShopperInsightsClient.kt#L24-L26

PR #1255, opened 2025-01-14
Bug Fix: Handle Missing 3DS `dfReferenceId`

jaxdesmarais:
### Summary of changes

 - Return error if no `dfReferenceId` is returned in the 3D Secure flow
    - Confirmed with Cardinal that we should not continue the flow if we do not get a `dfReferenceId` back, there is also no fallback mechanism on mobile like exists on web

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors

@jaxdesmarais

### DIFF
--- a/ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureClient.kt
+++ b/ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureClient.kt
@@ -204,11 +204,17 @@ class ThreeDSecureClient internal constructor(
                     configuration,
                     request
                 ) { consumerSessionId: String?, _ ->
-                    if (consumerSessionId != null) {
-                        try {
+                    if (consumerSessionId != null && !consumerSessionId.isEmpty()) {
                             lookupJSON.put("dfReferenceId", consumerSessionId)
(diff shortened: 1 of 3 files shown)


ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureClient.kt
```diff
@@ -204,11 +204,17 @@ class ThreeDSecureClient internal constructor(
                     configuration,
                     request
                 ) { consumerSessionId: String?, _ ->
-                    if (consumerSessionId != null) {
-                        try {
+                    if (consumerSessionId != null && !consumerSessionId.isEmpty()) {
```
richherrera, line 207:
```suggestion
                    if (!consumerSessionId.isNullOrEmpty()) {
```

richherrera (approved):

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureClient.kt
jaxdesmarais: Oh nice! I didn't even know that existed. 🚀

PR #1258, opened 2025-01-17
Fix order intent use-case

saperi22:
### Summary of changes

 - Update helper function to ignore case when checking for string equality.
 - Our demo app isn't setup for creating a customer when doing a checkout -- so the transaction itself fails, but this indicates that we're sending the intent=order as expected.

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+* PayPal
+  * Fix bug where `intent=order` was not working
```
jaxdesmarais, line 5: Very minor nit just to clarify:
```suggestion
  * Fix bug where `intent=order` was not being set as expected
```

jaxdesmarais (approved):

sarahkoop (approved):

PR #1267, opened 2025-02-07
Fix issue that's causing BA flows to fail on stage

saperi22:
### Summary of changes

 - Changes to fix for BA flow stage testing.

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
@@ -156,7 +156,7 @@ public class PayPalFragment extends BaseFragment {
                 buyerEmailAddress,
                 buyerPhoneCountryCode,
                 buyerPhoneNationalNumber,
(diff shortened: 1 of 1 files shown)


jaxdesmarais (approved):

Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
```diff
@@ -156,7 +156,7 @@ private void launchPayPal(
                 buyerEmailAddress,
                 buyerPhoneCountryCode,
                 buyerPhoneNationalNumber,
-                    "fake-session-id"
```
sarahkoop, line 159: It's just a stage problem that this isn't working? I would expect this field to be passable. We should make sure to put this back and test on prod before release.

saperi22:
Closing this as this is expected to work on prod. I'll just keep this in mind while testing on stage.

PR #1269, opened 2025-02-13
Fix 3ds npe

saperi22:
### Summary of changes

 - Demo app fix for crash on 3DS card screen.

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/BaseFragment.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/BaseFragment.java
@@ -15,18 +15,19 @@ public abstract class BaseFragment extends Fragment {
 
     @CallSuper
     public void onPaymentMethodNonceCreated(PaymentMethodNonce paymentMethodNonce) {
-        getActivity().setProgressBarIndeterminateVisibility(true);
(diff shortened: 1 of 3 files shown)


warmkesselj:
What does NPE stand for?

jaxdesmarais:
> What does NPE stand for?

@warmkesselj null pointer exception

sarahkoop:
So is this a change we would suggest our merchants do? Or is this change to fix the NPE in our demo to show that the 3DS issue is _not_ and issue on v5?

saperi22:
> So is this a change we would suggest our merchants do? Or is this change to fix the NPE in our demo to show that the 3DS issue is _not_ and issue on v5?

It's the latter. Our demo app is crashing on v5. That's a fix that the merchants have to do for their own UI. I haven't found a crash on v5 otherwise. If there's a concrete stacktrace that v5 is an issue, we can take a look at it then.

sarahkoop (approved):

warmkesselj:
Hey @saperi22 just curious if this is ready for review?

jaxdesmarais (approved):

PR #1271, opened 2025-02-20
Fix app switch vault analytics

saperi22:
### Summary of changes

 - We were logging the app-switch analytics when a couple static checks were satisfied even if the launchPayPalApp was false from the BT gateway.
 - In this PR, I'm making changes to take into consideration if launchPayPalApp is being returned true or not.

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage


jaxdesmarais (approved):
Changes look good. Can we add a changelog entry for this so future us doesn't forget when this change was made? Something like "Fix bug to correctly log isAppSwitchFlow" or whatever the FPTI key is. Approving ahead of time.

PR #1273, opened 2025-03-03
Configure RoomDB to fallback to destructive migration when a migratio…

saperi22:
…n is not provided.

We do not provide a migration path from a higher number to a lower number and in those cases, the DB is cleared out instead of the host app crashing. Since the DB we use today only handles the analytics events and they are uploaded every 30 secs, the risk of dropping data should be minimal.

### Summary of changes

- Enable fallbackToDestructiveMigration() on RoomDB.

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* PayPal
```
sarahkoop, line 5:
```suggestion
* All Modules
```

tdchow (approved):

sarahkoop (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsDatabase.kt
```diff
@@ -42,7 +42,8 @@ internal abstract class AnalyticsDatabase : RoomDatabase() {
                     context.applicationContext,
                     AnalyticsDatabase::class.java,
                     "analytics_database"
-                ).build()
+                ).fallbackToDestructiveMigration()
```
richherrera, line 45: 💯

richherrera (approved):

PR #1283, opened 2025-03-25
Update Event Batching Logic and AnalyticsClient Refactor

tdchow:
### Summary of changes

 - Update analytic event batching logic to only batch non conversion events. Conversion related events will be sent immediately.
 - Added `AnalyticsEventRepository` to hold events that are batched in memory
 - Separated out the FPTI API call from `AnalyticsClient` into  a new `AnalyticsApi` class
 - Removed Room and WorkManager classes

### Happy Path API Calls
Screenshot 2025-03-25 at 4 07 13 PM

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage


BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
```diff
@@ -0,0 +1,156 @@
+package com.braintreepayments.api.core
+
+import org.json.JSONArray
+import org.json.JSONException
+import org.json.JSONObject
+
+/**
+ * This API call sends analytic events to FPTI.
+ */
+internal class AnalyticsApi(
```
tdchow, line 10: All of this logic was moved from AnalyticsClient.

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
```diff
@@ -78,6 +78,8 @@ class BraintreeClient internal constructor(
                 it.deepLinkFallbackUrlScheme = deepLinkFallbackUrlScheme
             }
         }
+
+        applicationContext.deleteDatabase("analytics_database")
```
tdchow, line 82: Thoughts on removing the old analytics database on SDK launch? I think it's good to clean users device of stale data, but might be wasteful to do this every launch. 🤔

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventRepository.kt
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Flushes the current event buffer and returns the accumulated events.
+     *
+     * This function atomically clears the internal `events` list and returns a new list containing all the events that
+     * were previously in the buffer.
+     *
+     * @return A new list containing all events that were previously in the buffer.
+     */
+    fun flushAndReturnEvents(): List<AnalyticsEvent> {
+        synchronized(events) {
```
sarahkoop, line 26: Since this is synchronized should we be calling the analytics send method on a new thread each time?

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
sarahkoop: Hmmm I can't decide on this one. If a user deletes the merchant app this would be handled for us, but unsure how often that is likely to happen. It seems like a lot to try and delete the database on every `BraintreeClient` init. Is there somewhere else we could invoke that is called less often? @saperi22 curious what you think on this too

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventRepository.kt
tdchow: I believe we're calling `AnalyticsClient.sendEvent()` on the main thread in every instance, so having this synchronized block is probably overkill. But just wanted to add some safety if we end up sending events on different threads.

tdchow: We should be okay! This was all happening on the main thread previously. I was originally planning on introducing Coroutines for sending events, but saw we have a `BraintreeHttpClient.post()` that takes a callback and executes the call on a background thread. So I just went down that route to reduce the scope of the PR. I still think we can make future enhancements migrating to Coroutines.

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
tdchow: I couldn't find another spot that guaranteed execution 🙁  The lingering database should also be extremely small in size, so it wouldn't be the end of the world if we didn't remove it.

sarahkoop (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
saperi22: >I think it's good to clean users device of stale data, but might be wasteful to do this every launch.
>The lingering database should also be extremely small in size, so it wouldn't be the end of the world if we didn't remove it.

I agree with Tim on both the points. 😅 

>It seems like a lot to try and delete the database on every BraintreeClient init.

Unless this is an expensive operation, I don't see why this would be an issue. My _feeling_ on this is that it'll just be a maintenance thing forever.

One concern I have is if this API handles non-existent database deletion gracefully or not.

I don't have a strong opinion on this one other than the concern above.

tdchow: My assumption is that the `deleteDatabase()` call is lightweight. The API returns true if the database was visible and removed, or false otherwise. So it should handle when a database is not found.

```java 
  public boolean deleteDatabase(@NonNull String name) {
    checkArgument(!TextUtils.isEmpty(name), "Database name cannot be empty or null");
    if (databases.contains(name)) {
      if (context.deleteDatabase(getPrefixName(name))) {
        databases.remove(name);
        return true;
      }
    }
    return false;
  }
```

saperi22 (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
tdchow: Hm after thinking about this again. I think we should remove this line and just have the lingering database. I feel it's small enough to not cause concern.

@saperi22 @sarahkoop - does that sound good to you both?

PR #1285, opened 2025-03-27
link_type FPTI Parameter Updates

tdchow:
### Summary of changes

 - Update the logic of the `link_type` FPTI parameter - it now represents when are enabled/disabled by the user on the merchant app
 - Update PayPalClient and VenmoClient to set the `linkType` property on initialization
 - Separated out the app link vs deep link logic from `GetReturnLinkUseCase` into `GetReturnLinkTypeUseCase`
 
`GetReturnLinkTypeUseCase` - returns the type of link to be used for navigating back to the merchant app
`GetReturnLinkUseCase` - returns the actual link value (deep link or app link string value)

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsParamRepository.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsParamRepository.kt
@@ -10,6 +10,8 @@ class AnalyticsParamRepository(
     private val uuidHelper: UUIDHelper = UUIDHelper()
 ) {
 
+    var linkType: LinkType? = null
+
     private lateinit var _sessionId: String
 
     /**
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/GetReturnLinkTypeUseCase.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/GetReturnLinkTypeUseCase.kt
@@ -1,0 +1,34 @@
+package com.braintreepayments.api.core
+
+import android.content.Intent
+import android.content.pm.PackageManager
+import androidx.annotation.RestrictTo
+
+/**
+ * Use case that returns a return link type that will be used for navigating from App Switch or browser back into the
+ * merchant app.
+ *
+ * If a user unchecks the "Open supported links" checkbox in the Android OS settings for the merchant's app. If this
+ * setting is unchecked, this use case will return [ReturnLinkTypeResult.DEEP_LINK], otherwise
+ * [ReturnLinkTypeResult.APP_LINK] will be returned.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class GetReturnLinkTypeUseCase(private val merchantRepository: MerchantRepository) {
+
+    enum class ReturnLinkTypeResult {
+        APP_LINK, DEEP_LINK
+    }
+
+    operator fun invoke(): ReturnLinkTypeResult {
+        val context = merchantRepository.applicationContext
+        val intent = Intent(Intent.ACTION_VIEW, merchantRepository.appLinkReturnUri).apply {
+            addCategory(Intent.CATEGORY_BROWSABLE)
+        }
+        val resolvedActivity = context.packageManager.resolveActivity(intent, PackageManager.MATCH_DEFAULT_ONLY)
+        return if (resolvedActivity?.activityInfo?.packageName == context.packageName) {
+            ReturnLinkTypeResult.APP_LINK
+        } else {
+            ReturnLinkTypeResult.DEEP_LINK
+        }
+    }
+}
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
@@ -5,6 +5,7 @@ import android.net.Uri
 import android.text.TextUtils
 import com.braintreepayments.api.BrowserSwitchOptions
 import com.braintreepayments.api.core.AnalyticsEventParams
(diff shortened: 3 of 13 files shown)


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -28,20 +31,17 @@ class PayPalClient internal constructor(
     private val braintreeClient: BraintreeClient,
     private val internalPayPalClient: PayPalInternalClient = PayPalInternalClient(braintreeClient),
     private val merchantRepository: MerchantRepository = MerchantRepository.instance,
+    getReturnLinkTypeUseCase: GetReturnLinkTypeUseCase = GetReturnLinkTypeUseCase(merchantRepository),
     private val getReturnLinkUseCase: GetReturnLinkUseCase = GetReturnLinkUseCase(merchantRepository),
-    private val getAppSwitchUseCase: GetAppSwitchUseCase = GetAppSwitchUseCase(AppSwitchRepository.instance)
+    private val getAppSwitchUseCase: GetAppSwitchUseCase = GetAppSwitchUseCase(AppSwitchRepository.instance),
+    analyticsParamRepository: AnalyticsParamRepository = AnalyticsParamRepository.instance
```
sarahkoop, line 37: Can these all be private?

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetReturnLinkTypeUseCase.kt
```diff
@@ -0,0 +1,34 @@
...
+ * If a user unchecks the "Open supported links" checkbox in the Android OS settings for the merchant's app. If this
+ * setting is unchecked, this use case will return [ReturnLinkTypeResult.DEEP_LINK], otherwise
+ * [ReturnLinkTypeResult.APP_LINK] will be returned.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class GetReturnLinkTypeUseCase(private val merchantRepository: MerchantRepository) {
+
+    enum class ReturnLinkTypeResult {
+        APP_LINK, DEEP_LINK
+    }
```
sarahkoop, line 20: Since this enum is exactly the same as the existing `LinkType` enum, would it make sense to just return a `LinkType` from from use case?

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetReturnLinkTypeUseCase.kt
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Use case that returns a return link type that will be used for navigating from App Switch or browser back into the
+ * merchant app.
+ *
+ * If a user unchecks the "Open supported links" checkbox in the Android OS settings for the merchant's app. If this
+ * setting is unchecked, this use case will return [ReturnLinkTypeResult.DEEP_LINK], otherwise
+ * [ReturnLinkTypeResult.APP_LINK] will be returned.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class GetReturnLinkTypeUseCase(private val merchantRepository: MerchantRepository) {
```
saperi22, line 16: mark this internal, since this is kt file?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
tdchow: `getReturnLinkTypeUseCase` and `analyticsParamRepository` are not properties on the class and are just params pass into the constructor. Since they're only used in the `init` functions, they don't need to be defined as class properties.

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetReturnLinkTypeUseCase.kt
tdchow: Unfortunately, this use case is use across modules - paypal and venmo. So it needs to stay public with the `@RestrictTo` annotation.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsParamRepository.kt
```diff
@@ -10,6 +10,8 @@ class AnalyticsParamRepository(
     private val uuidHelper: UUIDHelper = UUIDHelper()
 ) {
 
+    var linkType: LinkType? = null
```
saperi22, line 13: Do we move the other parameters here too? How do you differentiate between what goes here and what goes elsewhere?

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetReturnLinkTypeUseCase.kt
tdchow: I think there's some benefit in keeping separation between the analytics specific `LinkType` and the result type of this use case. For example the `GetReturnLinkUseCase` relies on the `GetReturnLinkTypeUseCase` but does not need the analytic specific `LinkType.stringValue` String.

I'm also open to collapsing them down to a single enum, but the above was my thought process in keeping them separate.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsParamRepository.kt
tdchow: I think long term we should move all analytic params to this repository. I feel it's much easier to manage than having individual properties to manage in each of the Client classes.

sarahkoop (approved):

saperi22 (approved):

PR #1286, opened 2025-03-31
Add App Switch Analytic Params

tdchow:
### Summary of changes

 - Add `merchant_enabled_app_switch` and `attempted_app_switch` analytics params
 - Add properties to `AnalyticsParamRepository` to hold the state of these new params
 - Rename `AnalyticsParamRepository.resetSessionId()` to `AnalyticsParamRepository.reset()` as it now clears all stored properties
 - Fix an issue where `PayPalClient` was not invoking `AnalyticsParamRepository.reset()`
 - Update LinkType.APP_LINK value to "applink"

Example request:
``` json
{
  "events": [
    {
      "batch_params": {
        "app_id": "com.braintreepayments.demo",
        "app_name": "Braintree Demo",
        "c_sdk_ver": "5.8.1-SNAPSHOT",
        "client_os": "Android API 35",
        "comp": "braintreeclientsdk",
        "device_manufacturer": "Google",
        "mobile_device_model": "sdk_gphone64_arm64",
        "event_source": "mobile-native",
        "merchant_sdk_env": "sandbox",
        "api_integration_type": "custom",
        "is_simulator": true,
        "mapv": "5.8.1-SNAPSHOT",
        "merchant_id": "dcpspy2brwdjr3qn",
        "platform": "Android",
        "session_id": "54fe650375a14f8cb36488a03f90d656",
        "venmo_installed": false,
        "paypal_installed": false,
        "authorization_fingerprint": "..."
      },
      "event_params": [
        {
          "event_name": "paypal:tokenize:handle-return:succeeded",
          "t": 1743524457002,
          "is_vault": false,
          "tenant_name": "Braintree",
          "paypal_context_id": "3JN584945N4561731",
          "link_type": "deeplink",
          "url": "com.braintreepayments.demo.braintree",
          "merchant_enabled_app_switch": true,
          "attempted_app_switch": false
        }
      ]
    }
  ],
  "authorizationFingerprint": "..."
}
```

### Checklist

 - [ ] Added a changelog entry
 - [X] Relevant test coverage

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
@@ -81,6 +81,8 @@ internal class AnalyticsApi(
             .putOpt(FPTI_KEY_BUTTON_TYPE, event.buttonType)
             .putOpt(FPTI_KEY_BUTTON_POSITION, event.buttonOrder)
             .putOpt(FPTI_KEY_PAGE_TYPE, event.pageType)
+            .putOpt(FPTI_KEY_MERCHANT_ENABLED_APP_SWITCH, event.merchantEnabledAppSwitch)
+            .putOpt(FPTI_KEY_PAYPAL_ATTEMPTED_APP_SWITCH, event.payPalServerSideAttemptedAppSwitch)
     }
 
     @Throws(JSONException::class)
@@ -133,6 +135,8 @@ internal class AnalyticsApi(
         private const val FPTI_KEY_BUTTON_TYPE = "button_type"
         private const val FPTI_KEY_BUTTON_POSITION = "button_position"
         private const val FPTI_KEY_PAGE_TYPE = "page_type"
+        private const val FPTI_KEY_MERCHANT_ENABLED_APP_SWITCH = "merchant_enabled_app_switch"
+        private const val FPTI_KEY_PAYPAL_ATTEMPTED_APP_SWITCH = "attempted_app_switch"
 
         private const val FPTI_BATCH_KEY_VENMO_INSTALLED = "venmo_installed"
         private const val FPTI_BATCH_KEY_PAYPAL_INSTALLED = "paypal_installed"
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
@@ -81,8 +81,8 @@ internal class AnalyticsApi(
             .putOpt(FPTI_KEY_BUTTON_TYPE, event.buttonType)
             .putOpt(FPTI_KEY_BUTTON_POSITION, event.buttonOrder)
             .putOpt(FPTI_KEY_PAGE_TYPE, event.pageType)
-            .putOpt(FPTI_KEY_MERCHANT_ENABLED_APP_SWITCH, event.merchantEnabledAppSwitch)
-            .putOpt(FPTI_KEY_PAYPAL_ATTEMPTED_APP_SWITCH, event.payPalServerSideAttemptedAppSwitch)
+            .putOpt(FPTI_KEY_MERCHANT_ENABLED_APP_SWITCH, event.didEnablePayPalAppSwitch)
+            .putOpt(FPTI_KEY_PAYPAL_ATTEMPTED_APP_SWITCH, event.didPayPalServerAttemptAppSwitch)
     }
 
     @Throws(JSONException::class)
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsParamRepository.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsParamRepository.kt
@@ -39,7 +39,11 @@ class AnalyticsParamRepository(
         }
 
     /**
-     * Clears the session ID value from the repository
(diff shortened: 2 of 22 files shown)


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -103,8 +103,10 @@ class PayPalClient internal constructor(
         payPalRequest: PayPalRequest,
         callback: PayPalPaymentAuthCallback
     ) {
+        analyticsParamRepository.reset()
```
sarahkoop, line 106: Will this cause a problem with ShopperInsights now because the session ID will be different across between the shopper insights analytics and the start of the PayPal flow?

tdchow: Ah good catch. Yeah this will cause issues with ShopperInsights. I think we may need to reset the analyticsParamRepository at the end of every flow. I'll do some investigating and see if this is possible.

The issue i ran into is that the param repository was never cleared, so subsequent launches of the SDK had lingering values for `merchantEnabledAppSwitch` and `payPalServerSideAttemptedAppSwitch`.

tdchow: This was existing, but are we okay with session_id remaining consistent for subsequent launches of the SDK?

I did some testing with resetting the repository at the end of the flow and things seem to be working as expected. Although we need to invoke `reset()` in a number of places (cancel, failure, success cases).

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsParamRepository.kt
```diff
@@ -26,10 +39,17 @@ class AnalyticsParamRepository(
         }
 
     /**
-     * Clears the session ID value from the repository
+     * Resets the [sessionId] and clears all other repository values.
+     *
+     * Note that this function is called in different spots of the SDK lifecycle for different payment modules. Some
+     * modules call reset during launch of the SDK. The PayPal module calls reset at the end of the payment flow to
+     * persist the [sessionId] value set from the Shopper Insights module.
```
sarahkoop, line 46: Can be handled in a separate PR but should we also call reset at the end of the payment flow for Venmo for shopper insights?

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
```diff
@@ -133,6 +135,8 @@ internal class AnalyticsApi(
         private const val FPTI_KEY_BUTTON_TYPE = "button_type"
         private const val FPTI_KEY_BUTTON_POSITION = "button_position"
         private const val FPTI_KEY_PAGE_TYPE = "page_type"
+        private const val FPTI_KEY_MERCHANT_ENABLED_APP_SWITCH = "merchant_enabled_app_switch"
+        private const val FPTI_KEY_PAYPAL_ATTEMPTED_APP_SWITCH = "attempted_app_switch"
```
sarahkoop, line 139: Are we holding on merge/release until these tags are added FPTI? Not sure if there were existing ones that could be used or we want these specific names

sarahkoop (approved):
A few questions but LGTM

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
tdchow: Unfortunately, there weren't any tags that we could reuse. I think we can hold the release until the new tags are approved. I do already see them showing up in FPTI explorer.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsParamRepository.kt
tdchow: Good catch! Yep, I'll create a follow up PR to add this to the Venmo module.

jaxdesmarais:
Take it or leave it but I opted to use the internal names `didEnablePayPalAppSwitch` and `didPayPalServerAttemptAppSwitch` in the iOS PR: https://github.com/braintree/braintree_ios/pull/1551. Not sure if we want to align those across platform - also open to using different names if we don't think those are clear.

tdchow:
> Take it or leave it but I opted to use the internal names `didEnablePayPalAppSwitch` and `didPayPalServerAttemptAppSwitch` in the iOS PR: [braintree/braintree_ios#1551](https://github.com/braintree/braintree_ios/pull/1551). Not sure if we want to align those across platform - also open to using different names if we don't think those are clear.

Those names make sense. I'll align to what you have on iOS.

warmkesselj (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
jaxdesmarais: Both of these are now listed as approved in FPTI so we should be good to merge/release whenever. 🚀

jaxdesmarais (approved):
Not sure if we want a CHANGELOG for this on Android. I added one on iOS as it makes it a bit more clear to external teams (and us) when new tags are released.

tdchow:
> Not sure if we want a CHANGELOG for this on Android. I added one on iOS as it makes it a bit more clear to external teams (and us) when new tags are released.

Good call. Let me push an update for the changelog.

PR #1287, opened 2025-04-02
Reset AnalyticsParamRepository at the end of the Venmo flow

tdchow:
### Summary of changes

 - Reset AnalyticsParamRepository at the end of the Venmo flow

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### DIFF
--- a/Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.kt
+++ b/Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.kt
@@ -379,21 +379,25 @@ class VenmoClient internal constructor(
     ) {
         braintreeClient.sendAnalyticsEvent(VenmoAnalytics.TOKENIZE_FAILED, analyticsParams)
         callback.onVenmoPaymentAuthRequest(request)
+        analyticsParamRepository.reset()
     }
 
     private fun callbackSuccess(callback: VenmoTokenizeCallback, venmoResult: VenmoResult) {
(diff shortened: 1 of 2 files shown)


warmkesselj (approved):

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.kt
```diff
@@ -379,21 +379,25 @@ class VenmoClient internal constructor(
     ) {
         braintreeClient.sendAnalyticsEvent(VenmoAnalytics.TOKENIZE_FAILED, analyticsParams)
         callback.onVenmoPaymentAuthRequest(request)
+        analyticsParamRepository.reset()
```
saperi22, line 382: consider moving the logic to a single method and reusing the method everywhere

saperi22 (approved):

Venmo/src/main/java/com/braintreepayments/api/venmo/VenmoClient.kt
tdchow: Each of the callback functions are slightly different in the event and callback result/request it returns. So I didn't find a good way to extract to a separate function.

saperi22: @tdchow I just meant, you could add a `cleanup()` method and have this one line in there. :)

PR #1290, opened 2025-04-05
Convert test files to kotlin

saperi22:
### Summary of changes

 - Converting a test file from java to kotlin
 - Moved from using Mockito for java to mockk for kotlin test file

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage


PayPal/src/test/java/com/braintreepayments/api/paypal/MockPayPalInternalClientBuilder.java
```diff
@@ -9,6 +9,10 @@
 import org.mockito.invocation.InvocationOnMock;
 import org.mockito.stubbing.Answer;
 
+/**
+ * This outputs a Mockito mock of PayPalInternalClient. For newer kotlin tests, please use MockkPayPalInternalClientBuilder that outputs a mockk mock.
+ */
```
tdchow, line 14:
```suggestion
 * @deprecated Use {@link com.braintreepayments.api.testutils.MockkBraintreeClientBuilder} instead.
 */
```

Might be helpful to have a link to the class's replacement.

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalClientUnitTest.kt
```diff
@@ -0,0 +1,1022 @@
...
+    private var mockkPayPalTokenizeCallback = mockk<PayPalTokenizeCallback>(relaxed = true)
+    private var mockkPaymentAuthCallback: PayPalPaymentAuthCallback = mockk(relaxed = true)
+
+    private var mockkMerchantRepository: MerchantRepository = mockk(relaxed = true)
+    private var mockkGetReturnLinkTypeUseCase: GetReturnLinkTypeUseCase =
+        mockk<GetReturnLinkTypeUseCase>(relaxed = true)
+    private var mockkGetReturnLinkUseCase: GetReturnLinkUseCase = mockk(relaxed = true)
+    private val appSwitchRepository: AppSwitchRepository? = null
+    private var mockkGetAppSwitchUseCase: GetAppSwitchUseCase = mockk(relaxed = true)
+    private var mockkAnalyticsParamRepository: AnalyticsParamRepository = mockk(relaxed = true)
```
tdchow, line 55: Do you have strong opinions on the `mockk` prefix for these? I think it gets a little hard to scan the variables names. What if we added a `mock` postfix instead? Also in other files, I think we just omit the `mock` naming.

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalClientUnitTest.kt
```diff
@@ -0,0 +1,1022 @@
...
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.skyscreamer.jsonassert.JSONAssert
+
+@RunWith(RobolectricTestRunner::class)
+class PayPalClientUnitTest {
+    private var activity = mockk<FragmentActivity>(relaxed = true)
```
tdchow, line 42: Can we make all of the mocks `val`s?

TestUtils/src/main/java/com/braintreepayments/api/testutils/MockBraintreeClientBuilder.java
```diff
@@ -19,6 +19,10 @@
 
 import org.mockito.stubbing.Answer;
 
+/**
+ * This outputs a Mockito mock of BraintreeClient. For newer kotlin tests, please use MockkBraintreeClientBuilder that outputs a mockk mock.
```
tdchow, line 23: Same here - could we link the replacement class here?

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalClientUnitTest.kt
```diff
@@ -0,0 +1,1022 @@
...
+
+        val sut = PayPalClient(
+            braintreeClient,
+            payPalInternalClient,
+            mockkMerchantRepository,
+            mockkGetReturnLinkTypeUseCase,
+            mockkGetReturnLinkUseCase,
+            mockkGetAppSwitchUseCase,
+            mockkAnalyticsParamRepository
+        )
```
tdchow, line 233: What are your thoughts on creating a function to create the sut instance? Every time there's a change in dependencies, every test needs to be updated. I recently ran into this a couple times.

PayPal/src/test/java/com/braintreepayments/api/paypal/MockPayPalInternalClientBuilder.java
saperi22: `MockkBraintreeClientBuilder` is not a replacement for MockBraintreeClientBuilder, though. We'd want to use `MockkBraintreeClientBuilder` in case the implementation requires mockk object and not a Mockito object.

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalClientUnitTest.kt
saperi22: Nah, I can get rid of the prefix. The prefix was an in between step to make me differentiate an older Mockito object and the mockk object while I was refactoring.

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalClientUnitTest.kt
saperi22: Probably as a different refactor PR. don't want to increase scope and sit on this PR.
I'll make the change if it is a small refactor.

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalClientUnitTest.kt
```diff
@@ -0,0 +1,1022 @@
...
+    @Throws(JSONException::class)
+    fun beforeEach() {
+        every { mockkMerchantRepository.returnUrlScheme } returns "com.braintreepayments.demo"
+        every { mockkGetReturnLinkUseCase.invoke() } returns AppLink(Uri.parse("www.example.com"))
+        every { mockkGetAppSwitchUseCase.invoke() } returns true
+        every { mockkGetReturnLinkTypeUseCase.invoke() } returns ReturnLinkTypeResult.APP_LINK
+    }
+
+    @Test
+    fun initialization_sets_app_link_in_analyticsParamRepository() {
```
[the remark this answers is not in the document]
saperi22: Honestly, if the naming is easy to read, anything is fine. If there's a tool that can enforce backtick one, that'd be good to use, otherwise, I don't want to think about if I'm using the correct format of naming or not.
In this case, the names come from java to kt conversion.

PayPal/src/test/java/com/braintreepayments/api/paypal/MockPayPalInternalClientBuilder.java
tdchow: That's fair. Maybe we can just add the link annotation to where we call out `MockkPayPalInternalClientBuilder` to make navigating files easier.

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalClientUnitTest.kt
saperi22: made the change 👍

PayPal/src/test/java/com/braintreepayments/api/paypal/MockPayPalInternalClientBuilder.java
saperi22: @tdchow your thoughts?

tdchow: I agree, we don't want to call it a direct replacement. All I'm calling out is possibly adding the @link annotation to make files easier to navigate.

saperi22: > @tdchow your thoughts?

I should refresh more often :/

tdchow (approved):

jaxdesmarais (approved):

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalClientUnitTest.kt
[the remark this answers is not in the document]
saperi22: @sarahkoop do you have a strong preference on me using the backticks in this PR?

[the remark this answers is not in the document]
sarahkoop: No all good - can be for the future style guide 😄

PR #1294, opened 2025-04-11
Convert GooglePayLauncherUnitTest to Kotlin

saralvasquez:
### Summary of changes

 - Converts `GooglePayLauncherUnitTest` file from Java to Kotlin

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage


GooglePay/src/test/java/com/braintreepayments/api/googlepay/GooglePayLauncherUnitTest.kt
```diff
@@ -0,0 +1,69 @@
...
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+class GooglePayLauncherUnitTest {
+
+    private val activityResultLauncher: ActivityResultLauncher<GooglePayPaymentAuthRequestParams?> =
+        mockk<ActivityResultLauncher<GooglePayPaymentAuthRequestParams?>>(relaxed = true)
+    private val callback: GooglePayLauncherCallback? = mockk<GooglePayLauncherCallback>(relaxed = true)
+    private val activityResultRegistry: ActivityResultRegistry? =
+        mockk<ActivityResultRegistry>(relaxed = true)
```
tdchow, line 23:
```suggestion
    private val activityResultLauncher = mockk<ActivityResultLauncher<GooglePayPaymentAuthRequestParams?>>(relaxed = true)
    private val callback = mockk<GooglePayLauncherCallback>(relaxed = true)
    private val activityResultRegistry = mockk<ActivityResultRegistry>(relaxed = true)
```

In Kotlin the type can actually be inferred. So you can either declare the type and not include the type in the `mockk()` constructor, or remove the declared type and specify the type in the mockk constructor.

richherrera (approved):
🚀

GooglePay/src/test/java/com/braintreepayments/api/googlepay/GooglePayLauncherUnitTest.kt
```diff
@@ -0,0 +1,69 @@
...
+        } returns activityResultLauncher
+    }
+
+    @Test
+    fun constructor_createsActivityLauncher() {
+        val expectedKey = "com.braintreepayments.api.GooglePay.RESULT"
+        val lifecycleOwner = FragmentActivity()
+
+        val registry = mockk<ActivityResultRegistry>(relaxed = true)
+        GooglePayLauncher(registry, lifecycleOwner, callback!!)
```
tdchow, line 43: The GooglePayLauncher constructor takes in all non null parameters:

```kotlin
class GooglePayLauncher internal constructor(
    registry: ActivityResultRegistry,
    lifecycleOwner: LifecycleOwner,
    callback: GooglePayLauncherCallback
)
```

So we should be able to make all of the test dependencies non null - without the `?`.  This way we won't have to use the `!!` to convert to a non null type. It's a little less important for a unit test, but you usually want to avoid the non null assertion if possible.

Here's a good read on null safety from the Kotlin docs: https://kotlinlang.org/docs/null-safety.html

GooglePay/src/test/java/com/braintreepayments/api/googlepay/GooglePayLauncherUnitTest.kt
saralvasquez: Ah yes! That looks much better. Thanks for the info!

tdchow (approved):

PR #1303, opened 2025-04-23
Add new page types for Payment Ready v2

tdchow:
### Summary of changes

 - Add `PAYMENT_METHOD_SELECTION` and `PAYMENT_METHOD_ADDITION` for Payment Ready v2

### Checklist

 - [x] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/PageType.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/PageType.kt
@@ -55,6 +55,16 @@ enum class PageType(internal val stringValue: String) {
     MINI_CART("mini_cart"),
 
     /**
+     * The payment method selection page
+     */
+    PAYMENT_METHOD_SELECTION("payment_method_selection"),
+
+    /**
+     * The payment method addition page
+     */
+    PAYMENT_METHOD_ADDITION("payment_method_addition"),
+
+    /**
      * Some other page
      */
     OTHER("other");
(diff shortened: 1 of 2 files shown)


ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/PageType.kt
```diff
@@ -54,6 +54,16 @@ enum class PageType(internal val stringValue: String) {
      */
     MINI_CART("mini_cart"),
 
+    /**
+     * The payment method selection page
+     */
+    PAYMENT_METHOD_SELECTION("payment_method_selection"),
```
jaxdesmarais, line 60: Are we able to add any sort of tests for these? Not sure if we have those currently.

tdchow: It doesn't look like we have tests for this enum.

I feel that testing hardcoded values of enum properties is not worth it as there is no logic. However, if there are functions defined within enums that change based on external input, it'd be worth testing.

jaxdesmarais: Gotcha, wasn't sure if we were testing the `page_type` being passed correctly anywhere. It seems like no, but was still worth asking.

sarahkoop (approved):

jaxdesmarais (approved):

PR #1304, opened 2025-04-23
Move existing logic to ShopperInsightsClientV2

tdchow:
### Summary of changes

 - Add analytics calls and app installed checks to ShopperInsightsClientV2
 - Clean up a few shopper insights enums

### Checklist

 - [ ] Added a changelog entry
 - [X] Relevant test coverage

### DIFF
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/ShopperInsightsClientV2.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/ShopperInsightsClientV2.kt
@@ -2,8 +2,14 @@ package com.braintreepayments.api.shopperinsights.v2
 
 import android.content.Context
 import com.braintreepayments.api.core.AnalyticsClient
+import com.braintreepayments.api.core.AnalyticsEventParams
 import com.braintreepayments.api.core.BraintreeClient
+import com.braintreepayments.api.core.DeviceInspector
 import com.braintreepayments.api.core.ExperimentalBetaApi
+import com.braintreepayments.api.shopperinsights.ButtonType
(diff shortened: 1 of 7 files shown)


warmkesselj (approved):

ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/ShopperInsightsClientV2.kt
```diff
@@ -31,9 +38,62 @@ class ShopperInsightsClientV2 internal constructor(
...
+
+    /**
+     * Call this method when the PayPal, Venmo or Other button has been successfully displayed to the buyer.
+     * This method sends analytics to help improve the Shopper Insights feature experience.
+     * @param buttonType Type of button presented - PayPal, Venmo, or Other.
+     * @param presentmentDetails Detailed information, including button order, experiment type, and page type about the
+     * payment button that is sent to analytics to help improve the Shopper Insights feature experience.
+     * @param sessionId The shopper session ID
+     */
+    fun sendPresentedEvent(
```
jaxdesmarais, line 54: Do we want to consider having these all defined in 1 place that can be shared? Thinking of an instance where we have 3-4 clients all with the same methods and want to add something to all of them, we would need to manually make the change in 3-4 places. Not sure if that's feasible or not though, so take it or leave it!

tdchow: I felt that there wasn't much going on in the function - mapping `PresentmentDetails` to an `AnalyticsEventParams` type and sending the event. I think if the mapping became more complicated we could create a use case that could be used in all Clients.

jaxdesmarais (approved):

PR #1308, opened 2025-04-25
Checkout rba metadata feature amount breakdown and unit amount

warmkesselj:
### Summary of changes

- Add unitAmount of type String to PayPalRecurringBillingDetails
- Add new RecurringBillingAmountBreakdown class
- Include the following from HLD: itemTotal, shipping, handling, taxTotal, insurance, shippingDiscount, discount
- Add amountBreakdown of type RecurringBillingAmountBreakdown to PayPalRecurringBillingDetails

### Checklist

 - [X] Added a changelog entry
 - [X] Relevant test coverage

### Authors
@jwarmkessel @warmkesselj


PayPal/src/main/java/com/braintreepayments/api/paypal/RecurringBillingAmountBreakdown.kt
```diff
@@ -0,0 +1,26 @@
...
+ * @property shipping - Cost of shipping the items
+ * @property handling - Cost associated with handling the items (e.g., packaging or processing)
+ * @property taxTotal - Total tax amount applied to the transaction
+ * @property insurance - Cost of insurance applied to the shipment or items
+ * @property shippingDiscount - Discount amount applied specifically to shipping
+ * @property discount - General discount applied to the total transaction
+ */
+@Parcelize
+data class RecurringBillingAmountBreakdown(
+    val itemTotal: Double,
```
sarahkoop, line 19: I think all the other "amount" type values in this feature are strings - should these all be strings instead?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalRecurringBillingDetails.kt
```diff
@@ -22,6 +22,8 @@ import org.json.JSONObject
  * @property productQuantity Quantity associated with the product.
  * @property shippingAmount The shipping amount for the billing cycle at the time of checkout.
  * @property taxAmount The taxes for the billing cycle at the time of checkout.
+ * @property unitAmount This property is for checkout with vault requests only.
+ * @property amountBreakdown Breakdown of items associated to the total cost
  */
```
sarahkoop, line 27: Both of these properties should specify that they are for checkout with vault requests only, but they should also include a description of what the property is first.

Ex: `@property amountBreakdown Breakdown of items associated to the total cost - this property is for checkout with vault requests only`

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Braintree Android SDK Release Notes
 
+## unreleased 
+* PayPal
```
sarahkoop, line 4: I think both `recurringBillingDetails` and `recurringBillingPlanType` are supposed to move to the `PayPalRequest` so they can be used on both the `PayPalCheckoutRequest` and `PayPalVaultRequest` (and update CHANGELOG to reflect that) @jaxdesmarais is my understanding of that correct?

PayPal/src/main/java/com/braintreepayments/api/paypal/RecurringBillingAmountBreakdown.kt
warmkesselj: Updating!

CHANGELOG.md
jaxdesmarais: That's correct. Not sure if I missed those in the ticket but should be covered in the HLD linked there. We will want to move both of those to the `PayPalRequest` in addition to adding the new properties.

warmkesselj: I see this in the ticket:
This work should be done off a new checkout-rba-metadata-feature branch
**Add unitAmount of type String to PayPalRecurringBillingDetails**
In docstrings include that this property is for checkout with vault requests only
Add **new RecurringBillingAmountBreakdown** class
Include the following from HLD: itemTotal, shipping, handling, taxTotal, insurance, shippingDiscount, discount
Add docstrings for all new properties
Add **amountBreakdown**  of type **RecurringBillingAmountBreakdown to PayPalRecurringBillingDetails**

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalRecurringBillingDetails.kt
```diff
@@ -35,6 +38,8 @@ data class PayPalRecurringBillingDetails @JvmOverloads constructor(
     var productQuantity: Int? = null,
     var shippingAmount: String? = null,
     var taxAmount: String? = null,
+    val unitAmount: String? = null,
+    val amountBreakdown: RecurringBillingAmountBreakdown? = null,
```
tdchow, line 42: Should we make these `var`s since they're optional properties? This would create the Java setter methods.

PayPal/src/main/java/com/braintreepayments/api/paypal/RecurringBillingAmountBreakdown.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+@Parcelize
+data class RecurringBillingAmountBreakdown(
+    val itemTotal: String,
+    val shipping: String,
+    val handling: String,
+    val taxTotal: String,
+    val insurance: String,
+    val shippingDiscount: String,
+    val discount: String
```
tdchow, line 25: Are all of these required fields? We might want to double check the API spec.

CHANGELOG.md
jaxdesmarais: We synced offline on this and clarified that we should add both `recurringBillingDetails` and `recurringBillingPlanType` to the Checkout request

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalVaultRequest.kt
```diff
@@ -33,8 +33,8 @@ class PayPalVaultRequest
 @JvmOverloads constructor(
     override val hasUserLocationConsent: Boolean,
     var shouldOfferCredit: Boolean = false,
-    var recurringBillingDetails: PayPalRecurringBillingDetails? = null,
-    var recurringBillingPlanType: PayPalRecurringBillingPlanType? = null,
+    override var recurringBillingDetails: PayPalRecurringBillingDetails? = null,
+    override var recurringBillingPlanType: PayPalRecurringBillingPlanType? = null,
```
sarahkoop, line 37: We will also need to add these to the `PayPalCheckoutRequest` to enable this for the required flow

PayPal/src/main/java/com/braintreepayments/api/paypal/RecurringBillingAmountBreakdown.kt
```diff
@@ -0,0 +1,26 @@
...
+import kotlinx.android.parcel.Parcelize
+
+/**
+ * A recurring billing amount breakdown.
+ *
+ * @property itemTotal - Total amount of the items before any taxes or discounts
+ * @property shipping - Cost of shipping the items
+ * @property handling - Cost associated with handling the items (e.g., packaging or processing)
+ * @property taxTotal - Total tax amount applied to the transaction
+ * @property insurance - Cost of insurance applied to the shipment or items
```
jaxdesmarais, line 13: What do folks think of adding `total` to the end of `shipping`, `handling`, and `insurance`? Reading it now and realizing it may make it more clear to merchants.

PayPal/src/main/java/com/braintreepayments/api/paypal/RecurringBillingAmountBreakdown.kt
warmkesselj: We've removed some due to updated spec but they are required. There is a validation check on V2 orders.

PR #1310, opened 2025-05-01
Convert ThreeDSecureClientUnitTest to Kotlin

saralvasquez:
### Summary of changes

 - Converts `ThreeDSecureClientUnitTest` from Java to Kotlin
 - Converts `MockCardinalClientBuilder` from Java to Kotlin
 - Adds necessary functionality to `MockkBraintreeClientBuilder`

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage


ThreeDSecure/src/test/java/com/braintreepayments/api/threedsecure/MockCardinalClientBuilder.java
```diff
@@ -13,6 +13,7 @@
 
 import org.mockito.stubbing.Answer;
 
+@Deprecated
```
jaxdesmarais, line 16: General question, are there other tests still using this one or can it be removed now that `ThreeDSecureClientUnitTest` has been converted?

jaxdesmarais (approved):

ThreeDSecure/src/test/java/com/braintreepayments/api/threedsecure/ThreeDSecureClientUnitTest.kt
```diff
@@ -0,0 +1,819 @@
...
+        )
+
+        val paymentAuthResult = ThreeDSecurePaymentAuthResult("jwt", validateResponse, threeDSecureParams, null)
+        sut.tokenize(paymentAuthResult, threeDSecureTokenizeCallback)
+
+        val captorList = mutableListOf<ThreeDSecureResult>()
+        verify { threeDSecureTokenizeCallback.onThreeDSecureResult(capture(captorList)) }
+        val result = captorList.first()
+        assertTrue(result is ThreeDSecureResult.Failure)
+        assertEquals(result.nonce, paymentAuthResult.threeDSecureParams!!.threeDSecureNonce)
```
tdchow, line 760:
```suggestion
        assertEquals(result.nonce, paymentAuthResult.threeDSecureParams?.threeDSecureNonce)
```

Instead of throwing a NPE, we can use the [safe call operator](https://kotlinlang.org/docs/null-safety.html#safe-call-operator) here. If `paymentAuthResult.threeDSecureParams` is null, null will be passed into the second parameter of `assertEquals` which would result in a clearer log if the assertion failed.

tdchow (approved):
Just had a non-blocking suggestion for adding a null check.

ThreeDSecure/src/test/java/com/braintreepayments/api/threedsecure/ThreeDSecureClientUnitTest.kt
saralvasquez: Ah nice! Great to know!

PR #1311, opened 2025-05-02
Convert GooglePayClientUnitTest to Kotlin

saralvasquez:
### Summary of changes

 - Converts `MockGooglePayInternalClientBuilder` to kotlin and deprecates java version
 - Adds `activityInfo` to `MockkBraintreeClientBuilder`
 - Converts `GooglePayClientUnitTest` to kotlin

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### Authors
@saralvasquez


tdchow (approved):

GooglePay/src/test/java/com/braintreepayments/api/googlepay/MockGooglePayInternalClientBuilder.java
```diff
@@ -11,6 +11,7 @@
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
 
+@Deprecated
```
jaxdesmarais, line 14: Not a blocker but do we still need this one now that the tests are converted? This may be a super basic question but here I am. :)

saralvasquez: That's a great question. I saw that this was done with a similar class which is why I've been deprecating rather than completely removing the converted builder classes, but never asked why. @tdchow Is there any reason not to just delete the old file after conversion? Do we suspect that merchants use these or something?

tdchow: I believe the other test support classes which were converted to Kotlin were not an exact replacement, so there needed to be updates to the actual tests that relied on those classes. To keep the scope small for a given PR, I think the Java variant was marked as deprecated.

If these builder classes are no longer referenced, we should be good to remove them completely!

saralvasquez: Ok great! I'll go ahead and make that change to both PRs I have open rn

PR #1315, opened 2025-05-05
Checkout rba metadata feature gateway

warmkesselj:
### Summary of changes

Add AmountBreakdown, update the post body to communicate with Gateway to make sure we're getting a successful response.

### Checklist

 - [x] Relevant test coverage

### Authors
@jwarmkessel @warmkesselj

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
@@ -125,31 +125,6 @@ public class PayPalRequestFactory {
         return request;
     }
 
-    private static List<PayPalLineItem> buildLineItems(Float unitItemPrice,
-                                                       Float setupFee,
-                                                       Float immediateBillingAmount) {
-        Float totalAmount = unitItemPrice
-                + (setupFee != null ? setupFee : 0.00f)
-                + (immediateBillingAmount != null ? immediateBillingAmount : 0.00f);
-
-        PayPalLineItem item = new PayPalLineItem(
-                PayPalLineItemKind.CREDIT,
-                "Subscription Setup + First Cycle",
-                "1",
-                totalAmount.toString()
-        );
-
-        item.setDescription("Includes setup and first cycle");
-        item.setImageUrl("http://example.com/image.jpg");
-        item.setProductCode("sub-setup-001");
-        item.setUnitTaxAmount("0.50"); // Only include if you use taxTotal in AmountBreakdown
-        item.setUpcType(PayPalLineItemUpcType.UPC_TYPE_2);
-        item.setUpcCode("upc-001");
-        item.setUrl("http://example.com");
-
-        return Collections.singletonList(item);
-    }
-
     public static PayPalCheckoutRequest createPayPalCheckoutRequest(
         Context context,
         String amount,
@@ -160,30 +135,6 @@ public class PayPalRequestFactory {
         String shopperInsightsSessionId
     ) {
         PayPalCheckoutRequest request = new PayPalCheckoutRequest(amount, true);
-
-        List<PayPalLineItem> lineItems = buildLineItems(
-                10.00f,
-                5.00f,
-                3.00f
-        );
-
-        request.setLineItems(lineItems);
-
-        var amountF = Float.parseFloat(amount) - 1;
-
-        // Should equal item_total + tax_total + shipping + handling + insurance
-        // - shipping_discount - discount.
-        AmountBreakdown breakdown = new AmountBreakdown(
-                Float.toString(amountF), // item_total (required)
-                "0.50",  // tax_total (add) (optional, include only if line_items.tax_amount exists)
-                "0.50",  // shipping_total (add)
-                "0.50",  // handling (add)
-                "0.50",  // insurance (add)
-                "0.50",  // shipping_discount (subtract)
-                "0.50"   // discount (subtract)
-        );
-
-        request.setAmountBreakdown(breakdown);
 
         if (buyerEmailAddress != null && !buyerEmailAddress.isEmpty()) {
             request.setUserAuthenticationEmail(buyerEmailAddress);
--- a/Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
@@ -3,7 +3,6 @@ package com.braintreepayments.demo;
 import android.content.Context;
 
 import com.braintreepayments.api.core.PostalAddress;
-import com.braintreepayments.api.paypal.AmountBreakdown;
 import com.braintreepayments.api.paypal.PayPalBillingCycle;
 import com.braintreepayments.api.paypal.PayPalBillingInterval;
 import com.braintreepayments.api.paypal.PayPalBillingPricing;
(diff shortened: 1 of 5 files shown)


sarahkoop:
Can we update base to the feature branch?

PayPal/src/main/java/com/braintreepayments/api/paypal/AmountBreakdown.kt
```diff
@@ -41,4 +42,17 @@ data class AmountBreakdown(
...
+
+    fun toJson(): JSONObject {
+        return JSONObject().apply {
+            put("item_total", itemTotal)
+            put("shipping", shippingTotal)
+            put("handling", handlingTotal)
+            put("tax_total", taxTotal)
+            put("insurance", insuranceTotal)
+            put("shipping_discount", shippingDiscount)
+            put("discount", discountTotal)
```
sarahkoop, line 55: Can we pull the string values of these out into a companion object to match other request param objects?

PayPal/src/main/java/com/braintreepayments/api/paypal/AmountBreakdown.kt
```diff
@@ -41,4 +42,17 @@ data class AmountBreakdown(
     val insuranceTotal: String?,
     val shippingDiscount: String?,
     val discountTotal: String?
-) : Parcelable
+) : Parcelable {
+
+    fun toJson(): JSONObject {
+        return JSONObject().apply {
+            put("item_total", itemTotal)
+            put("shipping", shippingTotal)
```
sarahkoop, line 50: We should use `putOpt` for any optional fields

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt
```diff
@@ -239,6 +239,11 @@ class PayPalCheckoutRequest @JvmOverloads constructor(
         }
 
         parameters.put(EXPERIENCE_PROFILE_KEY, experienceProfile)
+
+        amountBreakdown?.let {
+            parameters.put(AMOUNT_BREAKDOWN_KEY, it.toJson())
+        }
```
sarahkoop, line 245: We should maybe consider moving all of these to `let`, but nitpick for now let's be consistent with the null check logic we use for all other params

Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
```diff
@@ -119,6 +125,31 @@ public static PayPalVaultRequest createPayPalVaultRequest(
...
+                totalAmount.toString()
+        );
+
+        item.setDescription("Includes setup and first cycle");
+        item.setImageUrl("http://example.com/image.jpg");
+        item.setProductCode("sub-setup-001");
+        item.setUnitTaxAmount("0.50"); // Only include if you use taxTotal in AmountBreakdown
+        item.setUpcType(PayPalLineItemUpcType.UPC_TYPE_2);
+        item.setUpcCode("upc-001");
+        item.setUrl("http://example.com");
```
sarahkoop, line 148: Is this PR being used for the demo app updates ticket as well? I don't think these changes are needed for passing the parameters to the GW.

If we are updating the demo app in this PR too, I think we should probably have settings for passing RBA metadata and line items or not with the checkout request. Currently this is going to pass line items on every PayPal request. Maybe we can also limit the line item params set to just the required ones and not pass the fake image/URL etc?

warmkesselj: This PR is only being used for communicating with Gateway. There's a separate ticket for a demo app update.

sarahkoop: Ok let's remove these unrelated changes then

PayPal/src/main/java/com/braintreepayments/api/paypal/AmountBreakdown.kt
```diff
@@ -41,4 +42,27 @@ data class AmountBreakdown(
...
+    }
+
+    companion object {
+        private const val KEY_ITEM_TOTAL = "item_total"
+        private const val KEY_TAX_TOTAL = "tax_total"
+        private const val KEY_SHIPPING = "shipping"
+        private const val KEY_HANDLING = "handling"
+        private const val KEY_INSURANCE = "insurance"
+        private const val KEY_SHIPPING_DISCOUNT = "shipping_discount"
+        private const val KEY_DISCOUNT = "discount"
```
sarahkoop, line 66: These keys should be used in the above `toJson` method

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalCheckoutRequestUnitTest.java
```diff
@@ -399,4 +399,39 @@ public void newPayPalCheckoutRequest_setsAmountBreakdown_AllFields() {
...
+        String requestBody = request.createRequestBody(
+                mock(Configuration.class),
+                mock(Authorization.class),
+                "success_url",
+                "cancel_url",
+                null
+        );
+
+        // Amount breakdown assertions
+        assertTrue(requestBody.contains("\"item_total\":\"20.00\""));
```
saperi22, line 429: nit: you could use https://github.com/skyscreamer/JSONassert which we already pull in to test this in a bit more readable format.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt
```diff
@@ -239,6 +239,11 @@ class PayPalCheckoutRequest @JvmOverloads constructor(
         }
 
         parameters.put(EXPERIENCE_PROFILE_KEY, experienceProfile)
+
+        if (amountBreakdown != null) {
```
saperi22, line 243: is this check required? putOpt takes care of the nullable value, correct?

Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
```diff
@@ -18,7 +18,9 @@
 import com.braintreepayments.api.paypal.PayPalPhoneNumber;
 import com.braintreepayments.api.paypal.PayPalVaultRequest;
 
+import java.math.BigDecimal;
```
jaxdesmarais, line 21:
```suggestion
```

Demo/src/main/java/com/braintreepayments/demo/PayPalRequestFactory.java
```diff
@@ -18,7 +18,9 @@
...
 import java.util.ArrayList;
+import java.util.Collections;
```
jaxdesmarais, line 23:
```suggestion
```

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt
```diff
@@ -239,6 +239,12 @@ class PayPalCheckoutRequest @JvmOverloads constructor(
         }
 
         parameters.put(EXPERIENCE_PROFILE_KEY, experienceProfile)
+
+        parameters.putOpt(AMOUNT_BREAKDOWN_KEY, amountBreakdown?.toJson())
```
jaxdesmarais, line 243:
```suggestion
        if (amountBreakdown.isNotEmpty()) {
            parameters.put(AMOUNT_BREAKDOWN_KEY, it.toJson())
        }
```

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt
warmkesselj: Matching the pattern of the rest of the function.

jaxdesmarais (approved):

PayPal/src/main/java/com/braintreepayments/api/paypal/AmountBreakdown.kt
```diff
@@ -41,4 +42,27 @@ data class AmountBreakdown(
     val insuranceTotal: String?,
     val shippingDiscount: String?,
     val discountTotal: String?
-) : Parcelable
+) : Parcelable {
+
+    fun toJson(): JSONObject {
```
tdchow, line 47: Sorry I missed this in the initial review - can we mark this function as internal? The merchant shouldn't need access to this function.

tdchow (approved):

PR #1317, opened 2025-05-09
Add CustomerSessionRequest and PurchaseUnit request objects

tdchow:
### Summary of changes

 - Add CustomerSessionRequest and PurchaseUnit request objects

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/CustomerSessionRequest.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/CustomerSessionRequest.kt
@@ -1,0 +1,21 @@
+package com.braintreepayments.api.shopperinsights.v2
+
+import com.braintreepayments.api.core.ExperimentalBetaApi
+
+/**
+ * Request for creating a customer session.
+ *
+ * @property hashedEmail Customer email address hashed via SHA256.
+ * @property hashedPhoneNumber Customer phone number hashed via SHA256.
+ * @property payPalAppInstalled If the PayPal app is installed on the device.
+ * @property venmoAppInstalled If the Venmo app is installed on the device.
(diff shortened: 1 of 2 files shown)


sarahkoop (approved):

ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/CustomerSessionRequest.kt
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @property hashedEmail Customer email address hashed via SHA256.
+ * @property hashedPhoneNumber Customer phone number hashed via SHA256.
+ * @property payPalAppInstalled If the PayPal app is installed on the device.
+ * @property venmoAppInstalled If the Venmo app is installed on the device.
+ * @property purchaseUnits List of purchase units containing the amount and currency code.
+ */
+@ExperimentalBetaApi
+data class CustomerSessionRequest(
+    var hashedEmail: String? = null,
```
saperi22, line 16: an reason to have these as `var`s?
would it be better to make them `val`s and force use copy constructor?

saperi22 (approved):

ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/CustomerSessionRequest.kt
tdchow: I believe we've been following the pattern where nullable properties are defined as `var`s so that java integrations can use setter functions. We shouldn't be modifying the variables internally.

saperi22: good to know!

PR #1319, opened 2025-05-12
Convert CardClientUnitTest to Kotlin

saralvasquez:
### Summary of changes

 - Converts `CardClientUnitTest` from Java to Kotlin

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### Authors
@saralvasquez


Card/src/test/java/com/braintreepayments/api/card/CardClientUnitTest.kt
```diff
@@ -0,0 +1,269 @@
...
+import org.robolectric.RobolectricTestRunner
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+@RunWith(RobolectricTestRunner::class)
+class CardClientUnitTest {
+    private val card: Card = Card()
+    private val cardTokenizeCallback: CardTokenizeCallback = mockk(relaxed = true)
+
+    private var apiClient: ApiClient = mockk(relaxed = true)
```
tdchow, line 28: It looks like we're setting the `apiClient` in every test. We could remove this class variable.

tdchow (approved):
Looks good! Just had a small suggestion.

Card/src/test/java/com/braintreepayments/api/card/CardClientUnitTest.kt
saralvasquez: It's getting set in most tests but not all. I think there are like 4 or 5 tests where it's using the class variable mockk

tdchow: Apologies! I missed those tests.

PR #1320, opened 2025-05-13
Add UpdateCustomerSessionApi

tdchow:
extract request and response logic to CustomerSessionRequestBuilder and ShopperInsightsResponseParser

### Summary of changes

 - Add UpdateCustomerSessionApi for updating a customer session via the GraphQL mutation
 - Extract request json building to CustomerSessionRequestBuilder
 - Extract response parsing to ShopperInsightsResponseParser
 - Move API related classes to an `internal` package

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### DIFF
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/internal/CustomerSessionRequestBuilder.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/internal/CustomerSessionRequestBuilder.kt
@@ -1,0 +1,57 @@
+package com.braintreepayments.api.shopperinsights.v2.internal
+
+import com.braintreepayments.api.core.ExperimentalBetaApi
+import com.braintreepayments.api.shopperinsights.v2.CustomerSessionRequest
+import org.json.JSONArray
+import org.json.JSONObject
+
+/**
+ * Builder for creating the JSON request objects for the Shopper Insights v2 APIs.
+ */
+@ExperimentalBetaApi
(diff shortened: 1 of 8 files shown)


ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/internal/CustomerSessionRequestBuilder.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.json.JSONObject
+
+/**
+ * Builder for creating the JSON request objects for the Shopper Insights v2 APIs.
+ */
+@ExperimentalBetaApi
+class CustomerSessionRequestBuilder {
+
+    internal data class JsonRequestObjects(
+        val customer: JSONObject,
```
jaxdesmarais, line 15: Not at all a blocker, just curious. Is there a reason why we use `Json` when Kotlin seems to prefer `JSON`? Not sure if that is a Java holdover or if the language follows a bit of an anti-pattern. 🙈

tdchow: That's a great question! In general, we should be following pascal case and camel case, as they're the preferred, modern naming conventions.

The Json classes that Android includes is from the `org.json` library, which is an extremely old Java Json library. It was created when Java naming conventions were less formalized.

warmkesselj (approved):

PR #1322, opened 2025-05-14
Add CustomerRecommendationsResult and PaymentOptions

tdchow:
### Summary of changes

 - Add `CustomerRecommendationsResult` and `PaymentOptions`
 - Add warning copy to `CustomerSessionRequest`'s doc string

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/CustomerRecommendationsResult.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/CustomerRecommendationsResult.kt
@@ -1,0 +1,19 @@
+package com.braintreepayments.api.shopperinsights.v2
+
+import com.braintreepayments.api.core.ExperimentalBetaApi
+
+/**
+ * Customer recommendations for what payment options to show.
+ *
+ * @property sessionId The session ID for the customer session.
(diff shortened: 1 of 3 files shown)


saperi22 (approved):

ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/CustomerRecommendationsResult.kt
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @property sessionId The session ID for the customer session.
+ * @property isInPayPalNetwork Whether the customer is in the PayPal network.
+ * @property paymentRecommendations The payment recommendations for the shopper.
+ *
+ * Warning: This feature is in beta. It's public API may change or be removed in future releases.
+ */
+@ExperimentalBetaApi
+data class CustomerRecommendationsResult(
+    val sessionId: String?,
```
jaxdesmarais, line 16: Is the session ID ever not returned or do we always get one back?

tdchow: I believe we should always get a session ID back from the API, but in the GQL spec, it's defined as optional. So to match the spec, I defined `sessionId` as nullable.

Issue #1321, opened 2025-05-14
Unexpected double callback in 3DSecure flow after OTP Retry


caruspi:
### Braintree SDK Version

5.10.0

### Environment

Sandbox

### Android Version & Device

Pixel 6a, API 33

### Braintree dependencies

I tried the demo app on the main's HEAD, that at the time of writing is [this commit](https://github.com/braintree/braintree_android/commit/d3a15aa71e2c1703ef86d13220edc9bb4f47734c).

### Describe the bug

We're encountering an issue when testing 3D Secure flows using the Braintree Android SDK. The problem occurs only with a specific test card (`status: challenge_required` use case `4000000000001109`) and not with all 3D Secure test scenarios.

Specifically, when the user enters an incorrect OTP on the first attempt and then provides the correct OTP on the second attempt, the flow does not behave as expected.

The [ThreeDSecureClient](https://github.com/braintree/braintree_android/blob/main/ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureTokenizeCallback.kt)'s callback is invoked twice:

- First with a `ThreeDSecureResult.Failure`
- Then again with a `ThreeDSecureResult.Success`

This leads to inconsistent state or duplicate processing that expects a single result.

### To reproduce

1. Clone and run the [Braintree Android Demo App](https://github.com/braintree/braintree-android-drop-in).

2. Select "_Credit or Debit Cards_".

3. Enter the following test card details (`status: challenge_required` use case ([link](https://developer.paypal.com/braintree/docs/guides/3d-secure/testing/ruby/))):
  - Card number: `4000 0000 0000 1109`
  - Expiration date: 01/2028

4. Proceed with the 3D Secure challenge:
  - Enter an **incorrect OTP** on the first attempt.
  - Enter the **correct OTP** on the second attempt.

Note:
This issue only occurs with the test card `4000000000001109`. Other cards that trigger a 3D Secure challenge do not exhibit this behavior under the same conditions.

### Expected behavior

The callback should be invoked only once, returning a `ThreeDSecureResult.Success`.

### Screenshots

| Credit Card Info  | First Attempt (incorrect OTP) | Second Attempt (correct OTP) | Result (Success + error) |
| ------------- | ------------- | ------------- | ------------- |
[...]

PR #1323, opened 2025-05-14
Surround the 3DS success callback when an error exists

tdchow:
### Summary of changes

 - Resolves https://github.com/braintree/braintree_android/issues/1321

There was a missing else block when a 3DS error is returned. The error callback was invoked along with the success callback.

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* ThreeDSecure
+    * Fix an issue where the `ThreeDSecureTokenizeCallback` would return a `Failure` and `Success` result (fixes #1321)
+
 ## 5.10.0 (2025-04-29)
 
 * PayPal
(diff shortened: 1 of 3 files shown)


caruspi:
@tdchow Thanks for the quick response and the PR!

I had a look at the fix, and I just wanted to confirm one detail to be sure: 

If the user initially enters an incorrect OTP and then retries with the correct one, the expected callback result should be a `Failure` or `Success`?

From the current logic, it seems that the callback will be invoked with a `Failure`, because debugging the use case mentioned in the [issue](https://github.com/braintree/braintree_android/issues/1321) I opened, the first check:

` if (threeDSecureResult.hasError())`

will be true.

tdchow:
@caruspi - No worries!

The test card you provided in the original issue should result in a single error:

> Failed Challenge Authentication
> Cardholder enrolled, authentication unsuccessful. Merchants should prompt customers for another form of payment.
> 
> status: challenge_required

In my testing, I was seeing both success and error results triggering due to the missing else block. I believe that entering an incorrect OTP code is unrelated to reproducing the issue.

Feel free to pull down this branch and do some testing if you'd like!

jaxdesmarais:
Can we add a CHANGELOG for this? If you add `(fixes #1321)` in the entry it should link it to the issue

PR #1328, opened 2025-05-19
Add createCustomerSession function to ShopperInsightsClientV2

tdchow:
### Summary of changes

 - Add `ShopperInsightsClientV2. createCustomerSession()` function

### Checklist
 - [x] Relevant test coverage
~- [ ] Tested and confirmed payment flows affected by this change are functioning as expected~


ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/ShopperInsightsClientV2.kt
```diff
@@ -43,6 +45,31 @@ class ShopperInsightsClientV2 internal constructor(
...
 
+    /**
+     * Creates a new customer session.
+     *
+     * @param customerSessionRequest: a [CustomerSessionRequest] object containing the request parameters
+     * @param customerSessionCallback: a callback that returns the result of the customer session creation
+     *
+     * Note: **This feature is in beta. It's public API may change in future releases.**
+     */
+    fun createCustomerSession(
```
warmkesselj, line 56: Does this function needs its own function level @ExperimentalBetaApi type directive since the beta is being called out in the comment above?

tdchow: We don't need to mark each individual function with the `@ExperimentalBetaApi` annotation. Since the entire `ShopperInsightsClientV2` class has the annotation, it applies to all of the public functions/properties.

tdchow:
> Approving ahead of time. Just a minor comment and a note that there are some unchecked boxes in the PR description.

Thanks! I don't think we need a changelog entry for this since the entire V2 flow of Shopper Insights is net new. Also the function is not yet called in the demo app, so there are no payment flows affected by this change.

richherrera (approved):

PR #1332, opened 2025-06-04
Move app switch analytics to the PayPalLauncher class

tdchow:
### Summary of changes

 - Move app switch analytics (started and error events) to PayPalLauncher to better reflect the stage the user is in
 - Inject `Context` into `DeviceInspector`
 - Created a `DeviceInspectorProvider` for injecting `DeviceInspector`
 - Update `SetAppSwitchUseCase` logic to store the state of all 3 app switch checks
 - Add Library scope to `GetAppSwitchUseCase` and `SetAppSwitchUseCase`

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/DeviceInspector.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/DeviceInspector.kt
@@ -14,6 +14,7 @@ import com.braintreepayments.api.sharedutils.SignatureVerifier
  */
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 class DeviceInspector(
+    private val context: Context,
     private val appHelper: AppHelper = AppHelper(),
     private val signatureVerifier: SignatureVerifier = SignatureVerifier(),
 ) {
@@ -58,7 +59,7 @@ class DeviceInspector(
         return isVenmoIntentAvailable && isVenmoSignatureValid
     }
 
-    fun isPayPalInstalled(context: Context?): Boolean {
+    fun isPayPalInstalled(): Boolean {
         return appHelper.isAppInstalled(context, PAYPAL_APP_PACKAGE)
     }
 
@@ -135,6 +136,7 @@ class DeviceInspector(
                     "$VENMO_APP_PACKAGE.$VENMO_APP_SWITCH_ACTIVITY"
                 )
             )
+
         internal fun getDropInVersion(): String? {
             try {
                 val dropInBuildConfigClass = Class.forName("com.braintreepayments.api.dropin.BuildConfig")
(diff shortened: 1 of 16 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/core/SetAppSwitchUseCase.kt
```diff
@@ -1,8 +1,30 @@
...
+    /**
+     * Sets the status of the app switch flow. This should be called once the PayPal response is received. Since it is
+     * the final check to see if the app switch flow should be shown.
+     *
+     * App Switch Logic:
+     * 1. Merchant enabled app switch
+     * 2. PayPal app installed
+     * 3. PayPal response indicates an app switch flow
+     *
+     * @param appSwitchFlowFromPayPalResponse whether the PayPal response indicates an app switch flow.
```
jaxdesmarais, line 21: Do we need a param for `merchantEnabledAppSwitch` too?

tdchow: Good catch! I just added the missing param.

BraintreeCore/src/test/java/com/braintreepayments/api/core/SetAppSwitchUseCaseUnitTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    fun `invoke sets isAppSwitchFlow true when all conditions are true`() {
+        every { deviceInspector.isPayPalInstalled() } returns true
+
+        setAppSwitchUseCase.invoke(
+            merchantEnabledAppSwitch = true,
+            appSwitchFlowFromPayPalResponse = true
+        )
+
+        verify { appSwitchRepository.isAppSwitchFlow = true }
+        confirmVerified(appSwitchRepository)
```
saperi22, line 33: Is `confirmVerified` basically checking that no other interactions are happening on the mock? `appSwitchRepository` in this case.

saperi22 (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/DeviceInspector.kt
```diff
@@ -14,6 +14,7 @@ import com.braintreepayments.api.sharedutils.SignatureVerifier
  */
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 class DeviceInspector(
+    private val context: Context,
```
saperi22, line 17: Since we are adding this as a class variable, would you consider cleaning up the individual method calls that have `context` to keep the using the same pattern?! (In a follow up PR, of course.)

jaxdesmarais (approved):

BraintreeCore/src/test/java/com/braintreepayments/api/core/SetAppSwitchUseCaseUnitTest.kt
tdchow: Yep, that's correct! Credit goes to Copilot for adding it to the tests 😄

BraintreeCore/src/main/java/com/braintreepayments/api/core/DeviceInspector.kt
tdchow: Yep! We should update the other functions that take in `context`. I wanted to keep the scope of this PR small.

PR #1337, opened 2025-06-16
Latency optimizations

saperi22:
### Summary of changes

 - Move `braintreeClient.sendAnalyticsEvent()` inside `braintreeClient.getConfiguration()` to reduce a network call to get configuration.
 -  Prefetch configuration on init of BraintreeClient to reduce user perceived latency.

Do these need to be called out in the CHANGELOG? 🤔 

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage


BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
```diff
@@ -78,6 +78,18 @@ class BraintreeClient internal constructor(
                 it.deepLinkFallbackUrlScheme = deepLinkFallbackUrlScheme
             }
         }
+
+        prefetchConfiguration()
```
tdchow, line 82: Do we need to do some testing of the other modules to make sure that nothing breaks by moving the config call here?

saperi22: Forgot to comment here.
I tested PayPal, ShopperInsights flows and they were both working.

jaxdesmarais: General question, is there a reason we created a method for this vs just calling directly like we do on iOS: https://github.com/braintree/braintree_ios/blob/6582191edb667a33a5e5900421b27edf3db2f88e/Sources/BraintreeCore/BTAPIClient.swift#L64-L67

BraintreeCore/src/test/java/com/braintreepayments/api/core/BraintreeClientUnitTest.kt
```diff
@@ -377,19 +377,24 @@ class BraintreeClientUnitTest {
...
 
-        val callbackSlot = slot<ConfigurationLoaderCallback>()
+        val callbackSlots = mutableListOf<ConfigurationLoaderCallback>()
         verify {
-            configurationLoader.loadConfiguration(capture(callbackSlot))
+            configurationLoader.loadConfiguration(capture(callbackSlots))
         }
 
-        callbackSlot.captured.onResult(ConfigurationLoaderResult.Success(configuration))
+        callbackSlots[0].onResult(ConfigurationLoaderResult.Success(configuration))
```
jaxdesmarais, line 385: Do we want to consider only ever having 1 configuration? This would likely be a seperate PR, but we could consider something similar to [this logic on iOS](https://github.com/braintree/braintree_ios/blob/6582191edb667a33a5e5900421b27edf3db2f88e/Sources/BraintreeCore/Configuration/ConfigurationLoader.swift#L47-L50) where we only ever return 1 config result even if there are multiple in flight at the same time on slower networks.

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
tdchow: I tested all the payment methods. For the card integration, I'm now seeing 4 config calls when the demo fragment is loaded. On the main branch I am seeing 2 config calls. Should we look into this to see if we can address the card module? I should have time tomorrow to take a look at this.

tdchow: I spent a little time digging into the card integration in the demo app. We're creating 4 client classes in the fragment's onCreate:

```kotlin
        americanExpressClient = new AmericanExpressClient(requireContext(), super.getAuthStringArg());
        cardClient = new CardClient(requireContext(), super.getAuthStringArg());
        threeDSecureClient = new ThreeDSecureClient(requireContext(), super.getAuthStringArg());

        dataCollector = new DataCollector(requireContext(), super.getAuthStringArg());
```

So I feel we should be okay since this doesn't reflect an actual merchant integration.

BraintreeCore/src/test/java/com/braintreepayments/api/core/BraintreeClientUnitTest.kt
tdchow: Yep! I think that's what we'll want to move towards. Sai and I talked about replacing the configuration loader callbacks with Coroutine Flow. This way all parallel request to configuration will rely on a single API call or configuration stored in cache.

This would be a large change to the SDK so we decided to tackle it after as a tech debt item.

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
saperi22: > General question, is there a reason we created a method for this vs just calling directly like we do on iOS: https://github.com/braintree/braintree_ios/blob/6582191edb667a33a5e5900421b27edf3db2f88e/Sources/BraintreeCore/BTAPIClient.swift#L64-L67

Wanted it to be more explicit since we aren't doing anything with the result.

BraintreeCore/src/test/java/com/braintreepayments/api/core/BraintreeClientUnitTest.kt
jaxdesmarais: Awesome! Yeah definitely outside of the scope for this PR.

PR #1341, opened 2025-06-25
added updateCustomerSession function to ShopperInsightsClientV2

sailaya99:

### Summary of changes

 - 

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


jaxdesmarais:
I am guessing this is showing a large diff because the base branch should be switched to `payment-ready-v2` and your branch should be off of that base branch with these changes.

PR #1343, opened 2025-07-01
Add generateCustomerRecommendations to ShopperInsightsClientV2

saralvasquez:
### Summary of changes

 - Added `generateCustomerRecommendations` to `ShopperInsightsClientV2`
 - Moved `CustomerRecommendations` data class to new separate class from the results class
 - Added a failure and success case to `CustomerRecommendationsResult`
 - Unit test support for `generateCustomerRecommendations`

**Note: used Github Copilot AI tool support**

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors
@saralvasquez

### DIFF
--- a/ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/CustomerRecommendations.kt
+++ b/ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/CustomerRecommendations.kt
@@ -1,0 +1,19 @@
+package com.braintreepayments.api.shopperinsights.v2
+
+import com.braintreepayments.api.core.ExperimentalBetaApi
+
+/**
+ * Customer recommendations for what payment options to show.
+ *
+ * @property sessionId The session ID for the customer session.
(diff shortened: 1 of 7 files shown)


ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/CustomerRecommendations.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Customer recommendations for what payment options to show.
+ *
+ * @property sessionId The session ID for the customer session.
+ * @property isInPayPalNetwork Whether the customer is in the PayPal network.
+ * @property paymentRecommendations The payment recommendations for the shopper.
+ *
+ * Warning: This feature is in beta. It's public API may change or be removed in future releases.
+ */
+@ExperimentalBetaApi
+data class CustomerRecommendations(
```
tdchow: I know this is an existing class, but I don't think merchants would be creating new instances of this class. Should we define the constructor as internal?


```suggestion
data class CustomerRecommendations internal constructor(
```

PR #1349, opened 2025-07-10
Update paypalContextId handling for analytics

tdchow:
### Summary of changes

 - Remove in memory repository for storing paypalContextId - this lead to issues where the token that was returned in the intent was different than the one that was persisted in memory.
 - Update each step of the flow to mange paypalContextId in a variable scoped to the current function. This will ensure that the value sent to analytics is always correct.

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage


PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -106,22 +102,40 @@ class PayPalClient internal constructor(
...
-                    callback,
-                    PayPalPaymentAuthRequest.Failure(createPayPalError())
-                )
-            } else if (configuration != null) {
-                sendPayPalRequest(context, payPalRequest, configuration, callback)
-            } else {
-                callbackCreatePaymentAuthFailure(
-                    callback,
-                    PayPalPaymentAuthRequest.Failure(BraintreeException("No configuration or error returned"))
-                )
```
tdchow, line 124: Refactored this if/else chain into a when statement.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -106,22 +102,40 @@ class PayPalClient internal constructor(
...
+            val analyticsEventParams = AnalyticsEventParams(
+                payPalContextId = null,
```
tdchow, line 106: `payPalContextId` defaults to null, but thought it'd be more straight forward to explicitly set it null.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -358,33 +389,18 @@ class PayPalClient internal constructor(
...
-    private val analyticsParams: AnalyticsEventParams
-        get() {
-            return AnalyticsEventParams(
-                payPalContextId = payPalContextId,
-                isVaultRequest = isVaultRequest,
-                shopperSessionId = shopperSessionId
-            )
-        }
-
-    private fun payPalConfigInvalid(configuration: Configuration?): Boolean {
```
tdchow, line 376: Removed this to do the `!configuration.isPayPalEnabled` inline.

warmkesselj (approved):

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -106,22 +102,40 @@ class PayPalClient internal constructor(
...
+
+                configuration == null -> {
+                    callbackCreatePaymentAuthFailure(
+                        callback,
+                        PayPalPaymentAuthRequest.Failure(BraintreeException("No configuration or error returned")),
+                        analyticsEventParams
+                    )
+                }
+
+                !configuration.isPayPalEnabled -> {
```
jaxdesmarais, line 130: This is pretty cool logic to me. Seems like what Kotlin does in a when statement is check that `configuration` is not null and if not you can access it as an unwrapped property (may not be totally accurate for Kotlin but what we'd call it in Swift) safely below. Does it keep that property in scope of only this when/else statement? The syntax is pretty different than Swift so want to make sure I am understanding it correctly.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -318,11 +346,12 @@ class PayPalClient internal constructor(
...
         callback: PayPalPaymentAuthCallback,
-        failure: PayPalPaymentAuthRequest.Failure
+        failure: PayPalPaymentAuthRequest.Failure,
+        analyticsEventParams: AnalyticsEventParams
     ) {
         braintreeClient.sendAnalyticsEvent(
-            PayPalAnalytics.TOKENIZATION_FAILED,
-            analyticsParams.copy(errorDescription = failure.error.message)
+            eventName = PayPalAnalytics.TOKENIZATION_FAILED,
+            params = analyticsEventParams.copy(errorDescription = failure.error.message)
```
jaxdesmarais, line 354: Is the reason we are switching to `eventName =/params =` mainly for readability?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
tdchow: Yep, that's exactly right! For all the other cases past the `configuration == null` check, `configuration` will  have a non null type. The compiler is smart enough to know which code paths this would apply for.

In the previous implementation, we had duplicate null checks for `configuration`. The original else if case that was checking for null would never have executed. I believe it was added to pass in a non null `configuration` to `sendPayPalRequest(context, payPalRequest, configuration, callback)`.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
tdchow: Yep, I added names to the call arguments for better readability. IDEs can show this, but I've found it useful having argument names in code reviews, or viewing code in GitHub.

saperi22 (approved):

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
jaxdesmarais: Nice! That's super cool, appreciate the explanation and glad I wasn't too far off in my understanding.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
jaxdesmarais: Cool, yeah agree it helps with readability when you aren't pulling down the branch to look through alongside.

PR #1352, opened 2025-07-15
migrated PayPalInternalClientUnitTest to Kotlin

sailaya99:

### Summary of changes

 - 

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


saperi22 (approved):

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalInternalClientUnitTest.kt
```diff
@@ -0,0 +1,1229 @@
...
+
+    @Test
+    fun sendRequest_withPayPalVaultRequest_sendsAllParameters() {
+        val braintreeClient = MockkBraintreeClientBuilder().build()
+
+        every { clientToken.bearer } returns "client-token-bearer"
+        every { merchantRepository.authorization } returns clientToken
+        every { merchantRepository.appLinkReturnUri } returns Uri.parse("https://example.com")
+
+        val sut = PayPalInternalClient(
```
tdchow: Are we using the same mocks that are set in `createSutWithMocks()`? Should we use that create function for all of the unit tests?

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalInternalClientUnitTest.kt
```diff
@@ -0,0 +1,1229 @@
...
+        payPalRequest.riskCorrelationId = "risk-correlation-id"
+
+        sut.sendRequest(context, payPalRequest, configuration, payPalInternalClientCallback)
+
+        val slot = slot<PayPalPaymentAuthRequestParams>()
+        verify {
+            payPalInternalClientCallback.onResult(capture(slot), null)
+        }
+        val payPalPaymentAuthRequestParams = slot.captured
+        Assert.assertEquals("risk-correlation-id", payPalPaymentAuthRequestParams.clientMetadataId)
```
tdchow:
```suggestion
        assertEquals("risk-correlation-id", payPalPaymentAuthRequestParams.clientMetadataId)
```

What are your thoughts on importing `Assert.assertEquals` and not having to reference `Assert` in each call?

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalInternalClientUnitTest.kt
```diff
@@ -0,0 +1,1229 @@
...
+        val payPalRequest = PayPalCheckoutRequest("1.00", true)
+
+        sut.sendRequest(context, payPalRequest, configuration, payPalInternalClientCallback)
+
+        val slot = slot<PayPalPaymentAuthRequestParams>()
+        verify {
+            payPalInternalClientCallback.onResult(capture(slot), null)
+        }
+        val payPalPaymentAuthRequestParams = slot.captured
+        TestCase.assertNull(payPalRequest.riskCorrelationId)
```
tdchow: Instead of using `TestCase.assertNull` can we stay consistent and use `Assert.assertNull`?

PR #1353, opened 2025-07-16
Demo app updates

saperi22:
### Summary of changes

 - Demo app changes for Payment Ready v2

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragmentV2.kt
+++ b/Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragmentV2.kt
@@ -122,9 +122,10 @@ class ShopperInsightsFragmentV2 : BaseFragment() {
                     val sessionId = viewModel.sessionId.collectAsState().value
                     Text(if (sessionId.isNotEmpty()) "Session Id = $sessionId" else "")
                     val recommendations = viewModel.recommendations.collectAsState().value
-                    Text(if (recommendations.isNotEmpty()) "Recommendations = " else "")
+                    Text(if (recommendations.isNotEmpty()) "Recommendations = $recommendations" else "")
+                    val isInPayPalNetwork = viewModel.isInPayPalNetwork.collectAsState().value
                     Button(
-                        enabled = true, //viewModel.isInPayPalNetwork.collectAsState().value,
+                        enabled = isInPayPalNetwork,
                         onClick = { launchPayPalVault(emailText, countryCodeText, nationalNumberText, sessionId) }
                     ) {
                         Text(text = "PayPal")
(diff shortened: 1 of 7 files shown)


Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragmentV2.kt
```diff
@@ -0,0 +1,298 @@
...
+                        Text(text = "PayPal")
+                    }
+                }
+            }
+        }
+    }
+
+    override fun onResume() {
+        super.onResume()
+//        handleVenmoReturnToApp()
```
tdchow: Do we need to handle Venmo's handle return to app function?

Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragmentV2.kt
```diff
@@ -0,0 +1,298 @@
...
+                    // Handle error, e.g., show error message
+                }
+            }
+        }
+    }
+
+    private fun handleCreateCustomerSession(emailText: String, countryCodeText: String, nationalNumberText: String) {
+        val customerSessionRequest = CustomerSessionRequest(
+            hashedEmail = emailText.sha256(),
+//            hashedPhoneNumber = nationalNumberText.sha256()
```
tdchow: Should we pass in the hashedPhoneNumber here?

Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragmentV2.kt
```diff
@@ -0,0 +1,298 @@
...
+                }
+                is CustomerSessionResult.Failure -> {
+                    // Handle failure
+                }
+            }
+        }
+    }
+
+    private fun handleGetRecommendations(sessionId: String) {
+        val sessionId = "94f0b2db-5323-4d86-add3-paypal000000"
```
tdchow: Should we make this hardcoded sessionId the default value of the `sessionId` function param?

Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragmentV2.kt
saperi22: I wanted to, but I wasn't sure if we just do (countryCodeText+nationalNumberText).sha256() as a phoneNumber or if it is just the nationalNumber.sha256().

tdchow: Did this end up working?

tdchow (approved):

Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsFragmentV2.kt
saperi22: I am just sending the nationalNumber as the phone number and it seems to work. That's what was shown as an example in the mocked data.

detekt/detekt-config.yml
```diff
@@ -139,7 +139,8 @@ complexity:
     ignoreStringsRegex: '$^'
   TooManyFunctions:
     active: true
-    excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**', '**/iosTest/**']
+    excludes: ['**/test/**', '**/androidTest/**', '**/commonTest/**', '**/jvmTest/**', '**/jsTest/**',
+               '**/iosTest/**', '**/Demo/**']
```
saperi22: skipping `TooManyFunctions` check on the demo module. Let me know if this is a bad idea.

tdchow: It might be worth leaving in. I feel ideally our demo app modules should follow the same ruleset as the other modules.

richherrera (approved):

PR #1358, opened 2025-07-24
Convert AmexClientTest from Java to Kotlin

noguier:
### Summary of changes

 - Converted `AmericanExpressClientUnitTest` from Java to Kotlin
 - Added functionality to `MockkBraintreeClientBuilder`

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage


TestUtils/src/main/java/com/braintreepayments/api/testutils/MockkBraintreeClientBuilder.kt
```diff
@@ -76,6 +79,16 @@ class MockkBraintreeClientBuilder {
         return this
     }
 
+    fun sendGETSuccessfulResponse(sendGETSuccess: String): MockkBraintreeClientBuilder {
+        this.sendGETSuccess = sendGETSuccess
```
saperi22, line 83: nit: consider renaming to `sendGetSuccess`
see : https://kotlinlang.org/docs/coding-conventions.html#choose-good-names

> When using an acronym as part of a declaration name, follow these rules:

> For two-letter acronyms, use uppercase for both letters. For example, IOStream.

> For acronyms longer than two letters, capitalize only the first letter. For example, XmlFormatter or HttpInputStream.

noguier: I was following the naming conventions, that were established in this class before. If I am to change sendGETSuccess, I would have to change 6 more variables and 6 functions definitions as well. I will make a follow up ticket to change the naming in this class by your suggestion.

PR #1360, opened 2025-07-25
Lint Warning Cleanup

tdchow:
### Summary of changes

 - Resolve various lint warnings

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


sailaya99 (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeHttpClient.kt
```diff
@@ -136,48 +136,6 @@ internal class BraintreeHttpClient(
...
-     * Makes a synchronous HTTP POST request to Braintree.
-     *
-     * @param path the path or url to request from the server via HTTP POST
-     * @param data the body of the post request
-     * @param configuration configuration for the Braintree Android SDK.
-     * @param authorization
-     * @return the HTTP response body
-     */
-    @Throws(Exception::class)
-    fun post(
```
jaxdesmarais, line 149: I'm assuming this is removed because we aren't using it? Just funny to see a post request removed but makes sense if we aren't using it anywhere.

jaxdesmarais (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeHttpClient.kt
tdchow: Yep, it was unused! The post function that's used across the SDK is async - callback is passed in. While this post function seems to be synchronous.

PR #1361, opened 2025-07-25
Change var/function declarations in MockkBuilder and its references, to follow naming conventions

noguier:
### Summary of changes

 - Changed `MockkBraintreeClientBuilder` variables and functions to follow naming conventions 
 - Changed function calls in Unit tests to match the changes
### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/TestUtils/src/main/java/com/braintreepayments/api/testutils/MockkBraintreeClientBuilder.kt
+++ b/TestUtils/src/main/java/com/braintreepayments/api/testutils/MockkBraintreeClientBuilder.kt
@@ -13,13 +13,13 @@ import io.mockk.mockk
 class MockkBraintreeClientBuilder {
 
     private var sendGraphQLPostSuccess: String? = null
-    private var sendGraphQLPOSTError: Exception? = null
+    private var sendGraphQLPostError: Exception? = null
 
-    private var sendGETSuccess: String? = null
-    private var sendGETError: Exception? = null
+    private var sendGetSuccess: String? = null
+    private var sendGetError: Exception? = null
 
-    private var sendPOSTSuccess: String? = null
-    private var sendPOSTError: Exception? = null
+    private var sendPostSuccess: String? = null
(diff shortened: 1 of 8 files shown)


saperi22 (approved):

TestUtils/src/main/java/com/braintreepayments/api/testutils/MockkBraintreeClientBuilder.kt
```diff
@@ -13,13 +13,13 @@ import io.mockk.mockk
 class MockkBraintreeClientBuilder {
 
     private var sendGraphQLPostSuccess: String? = null
```
jaxdesmarais, line 15: Should `GraphQL` be `GraphQl`? Tagging @saperi22 and @tdchow here since I am not sure on Android best practices for this. 😄

noguier: I think 2-letter acronyms are to be capitalized :) https://kotlinlang.org/docs/coding-conventions.html#choose-good-names

jaxdesmarais: Ah, nice find. Interesting though since we specifically don't capitalize those in the rest of the SDK. 🙈 For example in the android SDK today we have id's written as `Id` but seems like those should be `ID`. Not in scope for this PR though.

tdchow: I don't think there were any recommended acronym naming conventions for Java, so some of the naming patterns in the SDK may not make sense in Kotlin. But I think `id` is a strange one 🙂 

https://stackoverflow.com/questions/1699944/what-is-correct-java-naming-convention-for-id

PR #1366, opened 2025-07-29
Convert `AuthenticationInsightUnitTest` from Java to Kotlin

noguier:
### Summary of changes

 - Converted `AuthenticationInsightUnitTest` from Java to Kotlin

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage


Card/src/test/java/com/braintreepayments/api/card/AuthenticationInsightUnitTest.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    @Test
+    @Throws(JSONException::class)
+    fun `parses unknown regulation environment from JSON and returns unknown`() {
+        val response = JSONObject().apply {
+            put("customerAuthenticationRegulationEnvironment", "FaKeVaLuE")
+        }
+        val authenticationInsight = AuthenticationInsight.fromJson(response)
+
+        assertEquals("fakevalue", authenticationInsight?.regulationEnvironment)
```
warmkesselj, line 34: Suggestion: I'm guessing "FaKeVaLuE" is lower cased as a side effect. Would it be helpful to leave a comment pointing that out?

warmkesselj: Also the name of the function and the behavior seems strange to me. `JSON and returns unknown`. Seems to return exactly the value passed in instead of "Unknown".

noguier: Hi Justin, thank you for your suggestion, I think it is a little more clear to rename the function name. Let me know if it works: [[2c134c7](https://github.com/braintree/braintree_android/pull/1366/commits/2c134c708194c06035174add35ee9bd411d53308)](url)

warmkesselj: Seems to make more sense. Thanks! I would have either expected the function name to be updated or that the expected value returned would be "unknown". I'll approve for now but tagging @tdchow @saperi22 for some visibility!

warmkesselj (approved):

Card/src/test/java/com/braintreepayments/api/card/AuthenticationInsightUnitTest.kt
```diff
@@ -36,7 +36,7 @@ class AuthenticationInsightUnitTest {
 
     @Test
     @Throws(JSONException::class)
-    fun `parses regulation environment key from JSON and returns its value`() {
+    fun `parses regulation environment key from JSON and returns its lowercase value`() {
```
warmkesselj, line 39: Now I'm wondering if this second unit test is redundant.

noguier: Yes, I can see what you mean. I was following the tests that were already defined in the Java Unit tests, but, of course, open for any improvements or refactoring.

warmkesselj: I would just check and make sure there aren't expected values based on setting them. And if not then remove one or the other.

saperi22: ~the first parameter is different for the test cases, I'd leave it be.~

noguier: Hi all, I have tried to parse the code of the `fromJson` function , and I believe it is worth keeping both tests, because it tests two different conditional branches `fromJson` for GraphQL and REST.

PR #1369, opened 2025-07-30
Convert shared utils exception classes and AppHelper to Kotlin

tdchow:
### Summary of changes

 - Convert shared utils exception classes and AppHelper to Kotlin

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


saperi22 (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/DeviceInspector.kt
```diff
@@ -63,7 +63,7 @@ class DeviceInspector(
         return appHelper.isAppInstalled(context, PAYPAL_APP_PACKAGE)
     }
 
-    fun isVenmoInstalled(context: Context?): Boolean {
+    fun isVenmoInstalled(context: Context): Boolean {
```
jaxdesmarais, line 66: This is probably a super basic question but is there a reason Venmo requires a context and PayPal does not for the app install checks?

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/UpgradeRequiredException.kt
```diff
@@ -0,0 +1,7 @@
+package com.braintreepayments.api.sharedutils
+
+/**
+ * Exception thrown when a 426 HTTP_UPGRADE_REQUIRED response is encountered. Indicates that the
+ * API used or current SDK version is deprecated and must be updated.
+ */
+class UpgradeRequiredException(message: String) : Exception(message)
```
jaxdesmarais, line 7: 😅 I've never seen this one before, interesting we have a custom exception for it.

jaxdesmarais (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/DeviceInspector.kt
tdchow: This is a good question! The PayPal app check uses the context that's injected into DeviceInspector:

```kotlin
class DeviceInspector(
    private val context: Context,
    private val appHelper: AppHelper = AppHelper(),
    private val signatureVerifier: SignatureVerifier = SignatureVerifier(),
)
```

I'll open a follow up PR updating the Venmo app check function to use the same.

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/UpgradeRequiredException.kt
tdchow: I haven't seen this one either! It appears to be thrown when the API returns a 426 status code.

PR #1373, opened 2025-07-31
Convert various SharedUtils classes to Kotlin

tdchow:

### Summary of changes

 - Convert `CertificateHelper`, `HttpResponseParser`, `ManifestValidator`, `Scheduler`, `SignatureVerifier`, and `ThreadScheduler` classes to Kotlin

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


saperi22 (approved):

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/ManifestValidator.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import android.content.Context
+import android.content.pm.ActivityInfo
+import android.content.pm.PackageManager
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class ManifestValidator {
+
+    fun <T> getActivityInfo(context: Context, klass: Class<T>): ActivityInfo? {
```
jaxdesmarais, line 11: Do we need the space here or can it be removed:
```suggestion
    fun <T>getActivityInfo(context: Context, klass: Class<T>): ActivityInfo? {
```

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/SignatureVerifier.kt
```diff
@@ -0,0 +1,57 @@
...
+     * Check if an app has the correct, matching, signature. Used to prevent malicious apps from
+     * impersonating other apps.
+     *
+     * @param context                Android Context
+     * @param packageName            the package name of the app to verify.
+     * @param base64EncodedSignature the base64 encoded signature to verify.
+     * @return true is signature is valid or signature verification has been disabled.
+     */
+    @Suppress("SwallowedException", "ReturnCount")
+    fun isSignatureValid(context: Context, packageName: String, base64EncodedSignature: String): Boolean {
```
jaxdesmarais: nit: do we want some spaces to break up some of this method? It's a little bunched up which sometimes makes things a bit harder to grock.

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/SignatureVerifier.kt
```diff
@@ -0,0 +1,57 @@
...
+        val signatures: Array<Signature>?
+        try {
+            signatures = packageManager.getPackageInfo(packageName, PackageManager.GET_SIGNATURES).signatures
+        } catch (e: NameNotFoundException) {
+            return false
+        }
+        if (signatures == null || signatures.isEmpty()) {
+            return false
+        }
+        for (signature in signatures) {
```
jaxdesmarais, line 40: Do we have a preference for using `for` vs `forEach`?

jaxdesmarais: Ah, found the [Kotlin convention](https://kotlinlang.org/docs/coding-conventions.html#loops) but it's weirdly opaque. 😅

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/SignatureVerifier.kt
```diff
@@ -0,0 +1,57 @@
...
+            val currentSignature: String
+            try {
+                val md = MessageDigest.getInstance("SHA-256")
+                val encodedCert = certificateHelper.getEncodedCertificate(signature.toByteArray())
+                currentSignature = Base64.encodeToString(md.digest(encodedCert), Base64.DEFAULT)
+            } catch (e: NoSuchAlgorithmException) {
+                return false
+            } catch (e: CertificateException) {
```
jaxdesmarais: Is there a way to combine these?

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/ManifestValidator.kt
```diff
@@ -0,0 +1,20 @@
...
+        return try {
+            val packageInfo = context.packageManager.getPackageInfo(context.packageName, PackageManager.GET_ACTIVITIES)
+            val activities = packageInfo.activities
+            activities?.firstOrNull { it.name == klass.name }
+        } catch (_: PackageManager.NameNotFoundException) {
+            null
```
warmkesselj, line 17: Suggestion: Would it be worthwhile to Log when the package is not found here?

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/Scheduler.kt
```diff
@@ -0,0 +1,6 @@
+package com.braintreepayments.api.sharedutils
+
```
warmkesselj, line 2: Do we need another `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)` here?

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/ThreadScheduler.kt
```diff
@@ -0,0 +1,20 @@
+package com.braintreepayments.api.sharedutils
+
+import android.os.Handler
+import android.os.Looper
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+
```
warmkesselj, line 7: Do we need another `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)` here?

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/ManifestValidator.kt
tdchow: In Kotlin the correct formatting for generic functions is to include a space.

https://kotlinlang.org/docs/generics.html#generic-functions

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/SignatureVerifier.kt
tdchow: I wasn't even aware for that preference 😅 but that's good to know.

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/SignatureVerifier.kt
tdchow: Unfortunately there isn't a clean way to catch multiple exception types in the same block in Kotlin like you can in Java.

```kotlin
try {
    ...
} catch (e: Exception) {
    if (e is NoSuchAlgorithmException || e is CertificateException) {
        return false
    }
    throw e
}
```

This is one way to do it, but I feel it doesn't read as clear as catching the individual exception types.

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/SignatureVerifier.kt
tdchow: Yep, let me see what I can do to make this function easier to read.

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/ManifestValidator.kt
tdchow: Great suggestion. Since this is a PR scoped to Java -> Kotlin conversion, I don't think we should modify any of the actual logic. I think we can follow up in a separate PR.

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/Scheduler.kt
tdchow: Since this interface has the `internal` visibility modifier, we don't need to use the `RestrictTo` annotation. 

`internal` will scope the class/interface to the surrounding module. We're using the `RestrictTo` annotation for classes/interfaces that need to be public to be shared across modules, but should be hidden from merchants.

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/ThreadScheduler.kt
tdchow: We don't. See thread for Scheduler.kt.

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/SignatureVerifier.kt
tdchow: Just pushed an update, let me know if you think it needs any more tweaks.

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/SignatureVerifier.kt
jaxdesmarais: True true. 🤔

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/ManifestValidator.kt
jaxdesmarais: 🤯 interesting!

PR #1381, opened 2025-08-06
[Analytics] Add Context Type

jaxdesmarais:
**Note:** this work is going into a feature branch, once all changes are completed this will be verified and merged in with the appropriate teams. This is part of an alignment on the tags we pass to FPTI as well as some new tags and the same work is being done on iOS.

### Summary of changes

- Add `context_type` -this is a new tag passed by all services to determine if a request is BA vs EC token. The values were determined by the FPTI team for this new tag.
- Verified this tag is now sent in FPTI
- Add unit test
- iOS PR: https://github.com/braintree/braintree_ios/pull/1625

### Checklist

- [x] Added a changelog entry
- [x] Relevant test coverage

### Authors

- @jaxdesmarais


BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
```diff
@@ -86,6 +86,7 @@ internal class AnalyticsApi(
             .putOpt(FPTI_KEY_MERCHANT_ENABLED_APP_SWITCH, event.didEnablePayPalAppSwitch)
             .putOpt(FPTI_KEY_PAYPAL_ATTEMPTED_APP_SWITCH, event.didPayPalServerAttemptAppSwitch)
             .putOpt(FPTI_KEY_ERROR_DESC, event.errorDescription)
+            .putOpt(FPTI_KEY_CONTEXT_TYPE, event.contextType)
```
tdchow: It looks like this new tag's logic aligns with the existing `isVaultRequest`. Instead of adding additional logic to PayPalClient, what are your thoughts on if we use `event.isVaultRequest` to determine the value of `FPTI_KEY_CONTEXT_TYPE`?

I feel that this would clean up the duplicate logic added along with scope the values of contextType to the AnalyticsApi class.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -46,6 +46,11 @@ class PayPalClient internal constructor(
      */
     private var shopperSessionId: String? = null
 
+    /**
+     * Used for analytics purpose to determine if the context type is `BA_TOKEN` or `EC_TOKEN`
+     */
+    private var contextType: String? = null
```
tdchow: We may not need this new variable if we contain the contextType logic to the AnalyticsApi class. But just wanted to call out that we should try to avoid adding additional state to the Client classes. We're in the process of transitioning to a Repository pattern where SDK state is scoped to Repository classes.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
```diff
@@ -20,6 +20,7 @@ import androidx.annotation.RestrictTo
  * @property buttonOrder The order or ranking in which payment buttons appear.
  * @property pageType The page or view that a button is displayed on.
  * @property errorDescription use this to provide more detail on a failure event.
+ * @property contextType Indicates whether the request was a Billing Agreement (BA-TOKEN) or Checkout (EC-TOKEN) request.
```
jaxdesmarais:
```suggestion
 * @property contextType Indicates whether the request was a Billing Agreement (BA-TOKEN) 
 * or Checkout (EC-TOKEN) request.
```

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
jaxdesmarais: That's a great question, I had the same thought as well since they essentially represent the same value to us but different values to FPTI. I opted to keep them separate in hope that we can remove `isVaultRequest` in a future version when it's no longer being used. But if it makes it easier short term open to containing the logic in 1 place until that comes up.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
jaxdesmarais: Ah, yeah I was curious why some stuff was on the class and other in `analyticsParamRepository`. It sounds like that is where this should live instead if we decide to keep this separated?

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
tdchow: Ah I see. If we end up removing `FPTI_KEY_IS_VAULT` from our FPTI call, we could still pass in `isVaultRequest` as a part of `AnalyticsEventParams`. This way the implementation details of what we send to FPTI will be scoped to the single AnalyticsApi class instead of spanning across multiple classes in the SDK.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
tdchow: If we decide to keep contextType separate, we'd probably need to find/create a repository class within the PayPal module as `AnalyticsEventRepository` is in BraintreeCore. `contextType` would only apply to the PayPal integration.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsEventParams.kt
```diff
@@ -20,6 +20,8 @@ import androidx.annotation.RestrictTo
  * @property buttonOrder The order or ranking in which payment buttons appear.
  * @property pageType The page or view that a button is displayed on.
  * @property errorDescription use this to provide more detail on a failure event.
+ * @property contextType Indicates whether the request was a Billing Agreement (BA-TOKEN)
```
jaxdesmarais:
```suggestion
 * @property contextType Indicates whether the request was a Billing Agreement (BA-TOKEN)
```

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
jaxdesmarais: 🤔 My understanding of solid principles is that we should only be creating abstractions once necessary. I do worry that if everything goes into a separate abstraction layer before needed we end up going against solid principles and clean architecture by adding complexity before it's necessary. I think it's a balance we will need to find as we move to this architecture, but not necessarily in scope for this PR.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
jaxdesmarais: Sure, I can move it there if we'd rather for this PR, though we will diverge on iOS vs Android for this implementation. What do you think?

tdchow: I feel that it may be worth doing in this PR since it would be close to the architecture that we're trying to move towards. It may reduce some tech debt later down the line.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
tdchow: I feel this wouldn't necessarily be an abstraction but would follow single responsibility in SOLID principles. In clean architecture the app/sdk would be split into layers that have dependencies pointing inward to the entities (or data) layer.

This layer would be responsible for holding and maintaining the business objects of the application - in our case it'd be all of the variables or state in the SDK. So ideally all other layers of the SDK would technically be stateless, and we would inject repository classes wherever that data is needed.



Happy to continue chatting about this 😄

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
jaxdesmarais: Can do! What's our general style for conditional logic like this? Would we do something like:
```kotlin
.putOpt(FPTI_KEY_CONTEXT_TYPE, if (isVaultRequest) "BA-TOKEN" else "EC-TOKEN")
```
OR
```kotlin
if (isVaultRequest) {
    .putOpt(FPTI_KEY_CONTEXT_TYPE, "BA-TOKEN")
} else {
    .putOpt(FPTI_KEY_CONTEXT_TYPE, "EC-TOKEN")
}
```
both of these are somewhat sudo code so probably don't actually compile but hopefully gets the idea across and will fix the compiler errors obviously.

tdchow: I feel the first option is a bit clearer as the second option would prevent us from being able to chain to the other `.putOpt()` calls. We'd need to store the created JSONObject to a temp variable and then call put opt for the context type:

```kotlin
val json = JSONObject()...

if (isVaultRequest) {
    json.putOpt(FPTI_KEY_CONTEXT_TYPE, "BA-TOKEN")
} else {
    json.putOpt(FPTI_KEY_CONTEXT_TYPE, "EC-TOKEN")
}

return json
```

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
jaxdesmarais: For sure down to continue chatting architecture. 😄  It _does_ seem like `AnalyticsEventRepository` has some PayPal only state on it already in the form of `didEnablePayPalAppSwitch` and `didPayPalServerAttemptAppSwitch`. I'm fine moving this logic to be contained only within `AnalyticsApi` but it will be worth addressing that in the future since we are now holding and maintaining state for different things in different places, which I think breaks the above.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -94,14 +94,15 @@ class PayPalClient internal constructor(
     ) {
         shopperSessionId = payPalRequest.shopperSessionId
         isVaultRequest = payPalRequest is PayPalVaultRequest
+
```
jaxdesmarais:
```suggestion
```

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -94,14 +94,15 @@ class PayPalClient internal constructor(
...
         analyticsParamRepository.didEnablePayPalAppSwitch = payPalRequest.enablePayPalAppSwitch
 
         braintreeClient.getConfiguration { configuration: Configuration?, error: Exception? ->
             val analyticsEventParams = AnalyticsEventParams(
                 payPalContextId = null,
                 isVaultRequest = isVaultRequest,
                 shopperSessionId = shopperSessionId
-            )
+                )
```
jaxdesmarais:
```suggestion
            )
```

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -165,7 +166,7 @@ class PayPalClient internal constructor(
                                     payPalContextId = payPalContextId,
                                     isVaultRequest = isVaultRequest,
                                     shopperSessionId = shopperSessionId
-                                )
+                                    )
```
jaxdesmarais:
```suggestion
                                )
```

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt
```diff
@@ -260,7 +261,7 @@ class PayPalClient internal constructor(
             isVaultRequest = isVaultRequest,
             shopperSessionId = shopperSessionId,
             appSwitchUrl = paymentAuthResult.browserSwitchSuccess.returnUrl.toString()
-        )
+            )
```
jaxdesmarais:
```suggestion
        )
```

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
jaxdesmarais: Updated: https://github.com/braintree/braintree_android/pull/1381/commits/9b686dae57ab537080bb76a7aaff15782bd2d56e

Though I couldn't find a particularly easy way to test this in the paypal client tests, but seems like some other similar params aren't tested there. May be missing something though, and happy to pair on it if so.

tdchow: It doesn't look like there are any existing conditions on how the JSONObject is constructed. We'd need separate unit tests that pass in `isVaultRequest` as true and another test as false. Then we can verify the `analyticsRequest` that will be posted.

jaxdesmarais: It looks like it's tested via that setup in AnalyticsApiUnitTest

tdchow: It looks like we can have both cases in the different authorization types returned in the `getExpectedJson()` function.

jaxdesmarais: Yep, just pushed an update to AnalyticsApiUnitTest to separate out the event construction and test both.

tdchow: 👍 thanks for making the updates!

PR #1388, opened 2025-08-08
Payment ready v2 update demo

warmkesselj:
### Summary of changes
Updates to payment ready v2 demo app so it's showing outcomes of the network calls.
### Checklist

### Authors
@jwarmkessel @warmkesselj

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsV2ViewModel.kt
+++ b/Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsV2ViewModel.kt
@@ -21,11 +21,12 @@ import kotlinx.coroutines.flow.update
 class ShopperInsightsV2ViewModel : ViewModel() {
     private lateinit var shopperInsightsClient: ShopperInsightsClientV2
 
-    var sessionId = MutableStateFlow<String>("")
+    var sessionId = MutableStateFlow<String>("94f0b2db-5323-4d86-add3-paypal000000")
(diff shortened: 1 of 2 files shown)


Demo/src/main/java/com/braintreepayments/demo/ShopperInsightsV2ViewModel.kt
```diff
@@ -15,17 +15,20 @@ import com.braintreepayments.api.shopperinsights.v2.PaymentOptions
...
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.update
 
 @OptIn(ExperimentalBetaApi::class)
 class ShopperInsightsV2ViewModel : ViewModel() {
     private lateinit var shopperInsightsClient: ShopperInsightsClientV2
 
-    var sessionId = MutableStateFlow<String>("")
+    private val _sessionId = MutableStateFlow<String>("94f0b2db-5323-4d86-add3-paypal000000")
```
agedd, line 25: jw where do we get this sessionId value from 👀

warmkesselj: This is from the confluence doc provide by the rp8 team. They had a series of scenarios that they shared with us for testing. This is one of the sessionId's that returned an in paypal network condition as true with recommended options.

PR #1392, opened 2025-08-14
Convert PayPalLineItemUnitTest to kotlin

saralvasquez:
### Summary of changes

 - converted `PayPalLineItemUnitTest` to kotlin

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalLineItemUnitTest.kt
```diff
@@ -0,0 +1,42 @@
...
+class PayPalLineItemUnitTest {
+
+    @Test
+    fun `toJson sets keys and values`() {
+        val item = PayPalLineItem(
+            PayPalLineItemKind.DEBIT,
+            "An Item",
+            "1",
+            "2"
+        )
```
tdchow:
```suggestion
        val item = PayPalLineItem(
            kind = PayPalLineItemKind.DEBIT,
            name = "An Item",
            quantity = "1",
            unitAmount = "2"
        )
```

Adding names to call arguments can help with readability.

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalLineItemUnitTest.kt
```diff
@@ -0,0 +1,42 @@
...
+        item.description = "A new item"
+        item.imageUrl = "http://example.com/image.jpg"
+        item.productCode = "abc-123"
+        item.unitTaxAmount = "1.50"
+        item.upcType = PayPalLineItemUpcType.UPC_TYPE_2
+        item.upcCode = "upc-code"
+        item.url = "http://example.com"
```
tdchow: Instead of calling the setter functions, we can set these in the PayPalLineItem constructor above.

```kotlin
data class PayPalLineItem
@JvmOverloads constructor(
    val kind: PayPalLineItemKind,
    val name: String,
    val quantity: String,
    val unitAmount: String,
    var description: String? = null,
    var imageUrl: String? = null,
    var productCode: String? = null,
    var unitTaxAmount: String? = null,
    var upcCode: String? = null,
    var upcType: PayPalLineItemUpcType? = null,
    var url: String? = null
)
```

saperi22 (approved):

PR #1394, opened 2025-08-15
Convert Json and JsonTest to Kotlin

tdchow:
### Summary of changes

 - Convert Json and JsonTest to Kotlin

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage


SharedUtils/src/androidTest/java/com/braintreepayments/api/sharedutils/JsonTest.java
```diff
@@ -1,48 +0,0 @@
...
-
-    @Before
-    public void setup() throws JSONException {
-        json = new JSONObject("{\"key\":null}");
-    }
-
-    @Test
-    public void android_optString_returnsIncorrectNullValue() {
-        assertEquals("null", json.optString("key"));
-        assertEquals("null", json.optString("key", null));
```
tdchow, line 29: This line was removed in the Kotlin file as the second parameter of `JSONObject.optString()` is annotated as non-null.

```java
@NonNull public String optString(@Nullable String name, @NonNull String fallback) {
    Object object = opt(name);
    String result = JSON.toString(object);
    return result != null ? result : fallback;
}
```

PR #1399, opened 2025-08-19
DTNFSPEED-521 - Add Required Params to App Switch Universal Link

anibalb2500:

### Summary of changes

 - Added new query parameters that will be required from the PayPal app switch universal link. 
 - Updated the Demo App so that we'd be able to test the Pay Later functionality. 

### Checklist

 - [x] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
@@ -38,8 +38,7 @@ internal class PayPalInternalClient(
         callback: PayPalInternalClientCallback
     ) {
         try {
-            val isBillingAgreement = payPalRequest is PayPalVaultRequest
-            val endpoint = if (isBillingAgreement) {
+            val endpoint = if (payPalRequest.isBillingAgreement()) {
                 SETUP_BILLING_AGREEMENT_ENDPOINT
             } else {
                 CREATE_SINGLE_PAYMENT_ENDPOINT
@@ -158,7 +157,14 @@ internal class PayPalInternalClient(
                 )
                 if (getAppSwitchUseCase()) {
                     if (!contextId.isNullOrEmpty()) {
-                        paymentAuthRequest.approvalUrl = createAppSwitchUri(parsedRedirectUri).toString()
+                        val flowType = if (payPalRequest.isBillingAgreement()) VAULT_FLOW_TYPE else ECS_FLOW_TYPE
+                        val merchantId = configuration.merchantId
+
+                        paymentAuthRequest.approvalUrl = createAppSwitchUri(
+                            parsedRedirectUri,
+                            flowType,
+                            merchantId
+                        ).toString()
                     } else {
                         callback.onResult(null, BraintreeException("Missing Token for PayPal App Switch."))
                     }
@@ -172,10 +178,12 @@ internal class PayPalInternalClient(
         }
     }
 
-    private fun createAppSwitchUri(uri: Uri): Uri {
+    private fun createAppSwitchUri(uri: Uri, flowType: String, merchantId: String): Uri {
         return uri.buildUpon()
             .appendQueryParameter("source", "braintree_sdk")
             .appendQueryParameter("switch_initiated_time", System.currentTimeMillis().toString())
+            .appendQueryParameter("merchant", merchantId)
+            .appendQueryParameter("flow_type", flowType)
             .build()
     }
 
@@ -184,8 +192,12 @@ internal class PayPalInternalClient(
             ?: redirectUri.getQueryParameter("token")
     }
 
+    private fun PayPalRequest.isBillingAgreement(): Boolean = this is PayPalVaultRequest
+
     companion object {
         private const val CREATE_SINGLE_PAYMENT_ENDPOINT = "paypal_hermes/create_payment_resource"
         private const val SETUP_BILLING_AGREEMENT_ENDPOINT = "paypal_hermes/setup_billing_agreement"
+        private const val VAULT_FLOW_TYPE = "va"
+        private const val ECS_FLOW_TYPE = "ecs"
     }
 }


Demo/src/main/res/values/strings.xml
```diff
@@ -90,6 +90,7 @@
     <string name="paypal_display_name_summary">Name to be displayed in the PayPal flow</string>
     <string name="paypal_landing_page_type_billing">Billing</string>
     <string name="paypal_landing_page_type_login">Login</string>
+    <string name="paypal_offer_pay_later">Offer Pay Later</string>
```
tdchow: Would it be worth calling out that this setting would only apply for the `PayPalCheckoutRequest`, one time checkout flow?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -158,7 +157,14 @@ internal class PayPalInternalClient(
...
                     if (!contextId.isNullOrEmpty()) {
-                        paymentAuthRequest.approvalUrl = createAppSwitchUri(parsedRedirectUri).toString()
+                        val flowType = if (payPalRequest.isBillingAgreement()) VAULT_FLOW_TYPE else ECS_FLOW_TYPE
+                        val merchantId = configuration.merchantId
+
+                        paymentAuthRequest.approvalUrl = createAppSwitchUri(
+                            parsedRedirectUri,
+                            flowType,
+                            merchantId
+                        ).toString()
```
tdchow:
```suggestion
                        paymentAuthRequest.approvalUrl = createAppSwitchUri(
                            uri = parsedRedirectUri,
                            flowType = if (payPalRequest.isBillingAgreement()) VAULT_FLOW_TYPE else ECS_FLOW_TYPE,
                            merchantId = configuration.merchantId
                        ).toString()
```

Take it or leave it suggestion - not sure if it reads a little better without storing flowType and merchantId into variables.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -184,8 +192,12 @@ internal class PayPalInternalClient(
             ?: redirectUri.getQueryParameter("token")
     }
 
+    private fun PayPalRequest.isBillingAgreement(): Boolean = this is PayPalVaultRequest
+
     companion object {
         private const val CREATE_SINGLE_PAYMENT_ENDPOINT = "paypal_hermes/create_payment_resource"
         private const val SETUP_BILLING_AGREEMENT_ENDPOINT = "paypal_hermes/setup_billing_agreement"
+        private const val VAULT_FLOW_TYPE = "va"
+        private const val ECS_FLOW_TYPE = "ecs"
```
tdchow:
```suggestion
        private const val SINGLE_PAYMENT_FLOW_TYPE = "ecs"
```

`ECS` might be a little confusing since the acronym doesn't currently exist in the codebase.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -172,10 +178,12 @@ internal class PayPalInternalClient(
         }
     }
 
-    private fun createAppSwitchUri(uri: Uri): Uri {
+    private fun createAppSwitchUri(uri: Uri, flowType: String, merchantId: String): Uri {
```
tdchow: It might be good to add a Kdoc and explain why `flowType` and `merchantId` are added to the app switch Uri.

tdchow:
Would you also be able to add an entry to CHANGELOG.md for the new query params added to the app switch url?

Demo/src/main/res/values/strings.xml
anibalb2500: Yes, and I just noticed that these strings are for settings. Should I move this to the PayPalActivity section?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
anibalb2500: Taking this suggestion. Thanks.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
anibalb2500: An astute observation.

Demo/src/main/res/values/strings.xml
tdchow: Ah good catch, I didn't even realize. Yep, I think we can move this new string to where the other `PayPalActivity` strings live.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
anibalb2500: @tdchow  The lint will likely fail due to line length still, if want to avoid storing flowType in a variable. 

I see other lines exceeding the line length limit, so I'm assuming exceptions were made in the past.

tdchow: I think in that case we can store flowType to a variable.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
   * Only send `paypal:tokenize:browser-login:canceled` for non app switch flows
   * Fix an issue where non app switch flows were sending app switch events
   * Update `url` FPTI tag to represent the approval URL prior to checkout and the success URL after checkout
+  * Added `merchant` and `flow_type` as query parameters to the app switch URL.
```
tdchow: We'll want to move this entry under the `## unreleased` heading (after line 6). This current section is for a release that has already been released.

Basically when we release a new version of the SDK, or deploy script renames `## unreleased` to the version number and all new entries will go in the `## unreleased` heading.

PR #1401, opened 2025-08-19
Add required query params

anibalb2500:

### Summary of changes

 - Added new query parameters that will be required from the PayPal app switch universal link. 
 - Updated the Demo App so that we'd be able to test the Pay Later functionality. 

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Video
Showcases running with the new toggled and then untoggled. 
https://github.com/user-attachments/assets/0fd9a310-8b9c-4865-be99-672be264fe43

### DIFF
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
@@ -160,11 +160,7 @@ internal class PayPalInternalClient(
                         val flowType = if (payPalRequest.isBillingAgreement()) VAULT_FLOW_TYPE else ECS_FLOW_TYPE
                         val merchantId = configuration.merchantId
 
-                        paymentAuthRequest.approvalUrl = createAppSwitchUri(
-                            parsedRedirectUri,
-                            flowType,
-                            merchantId
-                        ).toString()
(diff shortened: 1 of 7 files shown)


tdchow (approved):

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -172,10 +175,23 @@ internal class PayPalInternalClient(
...
+     *
+     * @param uri The base [Uri] to build upon.
+     * @param flowType The checkout flow type.
+     * @param merchantId The merchant ID for the integration.
+     */
+    private fun createAppSwitchUri(uri: Uri, merchantId: String, flowType: String): Uri {
         return uri.buildUpon()
             .appendQueryParameter("source", "braintree_sdk")
             .appendQueryParameter("switch_initiated_time", System.currentTimeMillis().toString())
+            .appendQueryParameter("merchant", merchantId)
```
jaxdesmarais, line 191: Is this okay to pass from a security standpoint?

anibalb2500: Yes, this is fine.

PR #1403, opened 2025-08-20
Set TLSCertificateFactory in OkHttpClient Builder & decompress gzip-encoded HTTP responses

tdchow:
### Summary of changes

 - Set `TLSCertificateFactory` on `OkHttpClient` Builder (The existing `TLSCertificateFactory` was mostly unchanged)
 - Add `GzipRequestInterceptor` for decompressing gzip-encoded HTTP responses
 - Move `TLSCertificatePinning` from the core module to the sharedutils module

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage


BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeGraphQLClient.kt
```diff
@@ -28,18 +30,11 @@ internal class BraintreeGraphQLClient(
             .data(data)
             .baseUrl(configuration.graphQLUrl)
             .addHeader("User-Agent", "braintree/android/" + BuildConfig.VERSION_NAME)
-            .addHeader("Authorization",
-                String.format(Locale.US, "Bearer %s", authorization.bearer))
+            .addHeader(
+                "Authorization",
+                String.format(Locale.US, "Bearer %s", authorization.bearer)
```
warmkesselj, line 35: Should the Locale be pulled from the device?

tdchow: Good question. This was existing code that was just reformatted. But I don't think we'd want to used device locale as this is formatting the authorization header. We'd want the header to be consistent no matter the device locale.

SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/GzipRequestInterceptor.kt
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * This interceptor checks for responses with "Content-Encoding: gzip" header and
+ * transparently decompresses the response body. After decompression, it removes
+ * the Content-Encoding and Content-Length headers since they no longer apply
+ * to the decompressed content.
+ *
+ * The interceptor handles case-insensitive matching for the gzip encoding
+ * (e.g., "gzip", "GZIP", "Gzip" are all supported).
+ */
+internal class GzipRequestInterceptor : Interceptor {
```
jaxdesmarais: Did a bit of reading on this since we don't do this on iOS. Seems like the iOS URLSession handles it automatically as of iOS 7+. Pretty interesting that okhttp doesn't do the same but finding a lot that indicates it may require a more manual implementation on Android.

tdchow: OkHttp should handle Gzip decoding by default, but I don't think this is the case when setting an SSLSocketFactory on the OkHttpClient. So unfortunately we need to manually decode responses.

jaxdesmarais: 🤔 is it worth testing? I am finding some stuff online from a few years ago that indicates they do. But of course could be outdated.

jaxdesmarais: Looking here: https://square.github.io/okhttp/features/calls/#rewriting-requests

tdchow: I originally tested these new classes before I added the Gzip interceptor. The responses returned were gzip encoded 🙁

tdchow: Let me dig into the OkHttp source to see why the response was not getting decoded.

jaxdesmarais: For sure, not a blocker for this PR at all just went down a rabbit hole, so I trust whatever you ultimately find. I did find this online:
> OkHttp automatically decompresses responses that include the Content-Encoding: gzip header, provided that the Accept-Encoding: gzip header was not manually set in the request.
>
> If the Accept-Encoding header is manually added to the request, OkHttp will not automatically decompress the response, even if the Content-Encoding: gzip header is present in the response. In such cases, manual decompression of the response body is required. 

It looks like we do set that in our current code here: https://github.com/braintree/braintree_android/blob/4c84ba98e1247acb235e44147ac1d72a42cd3e07/SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/HttpRequest.kt#L72 So I wonder if that is causing issues.

tdchow: I just went through OkHttp source and you're exactly right! When i was testing, I was including all of the headers from HttpRequest. I was just mapping that existing request to the new OkHttpRequest object - which included the "Accept-Encoding" header.

I found the exact logic that validates the quote you shared above. With that accept encoding header, OkHttp was not doing transparent Gzip.

https://github.com/square/okhttp/blob/parent-4.12.0/okhttp/src/main/kotlin/okhttp3/internal/http/BridgeInterceptor.kt#L69

I'll go ahead and remove the custom gzip implementation. I tested removing the interceptor on my current branch with removing the original accept encoding header and everything is working as expected. Thanks for your investigation and follow up on this!

jaxdesmarais: 🎉 it seems like okhttp will bring us a lot of parity in the networking layers, which will be really nice! Thanks for digging into it with me, I think we are all learning a lot still about okhttp so it's been nice to dig into it a bit.

jaxdesmarais: Plus less code is always a win 😄

tdchow: Always happy about less code to maintain 😂

PR #1407, opened 2025-08-22
Update HttpClient to call OkHttpSynchronusHttpClient and update BraintreeGraphQLClient andBraintreeHttpClient

tdchow:
### Summary of changes

 - Update HttpClient to call OkHttpSynchronusHttpClient
 - Update BraintreeGraphQLClient and BraintreeHttpClient to assemble `OkHttpRequest`s

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeHttpClient.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeHttpClient.kt
@@ -121,11 +121,9 @@ internal class BraintreeHttpClient(
         val headers = mutableMapOf(
             USER_AGENT_HEADER to "braintree/android/" + BuildConfig.VERSION_NAME
         )
-        if (authorization is TokenizationKey) {
-            headers["Client-Key"] = authorization.bearer
-        }
-        authorization?.bearer?.let { token ->
-            headers["Authorization"] = "Bearer $token"
+        when (authorization) {
+            is TokenizationKey -> headers["Client-Key"] = authorization.bearer
+            is ClientToken -> headers["Authorization"] = "Bearer ${authorization.bearer}"
         }
         headers.putAll(additionalHeaders)
         return headers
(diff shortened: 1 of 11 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeHttpClient.kt
```diff
@@ -99,48 +62,77 @@ internal class BraintreeHttpClient(
...
         if (authorization is TokenizationKey) {
-            request.addHeader(CLIENT_KEY_HEADER, authorization.bearer)
+            headers["Client-Key"] = authorization.bearer
         }
-        authorization?.bearer?.let { token -> request.addHeader("Authorization", "Bearer $token") }
-        additionalHeaders.forEach { (name, value) -> request.addHeader(name, value) }
-        httpClient.sendRequest(request, callback)
+        authorization?.bearer?.let { token ->
+            headers["Authorization"] = "Bearer $token"
+        }
```
tdchow: Existing behavior - Does this header need to be added when authorization is TokenizationKey?

Currently we're sending both `authorization` and `client-key` headers for every call when the authorization type is tokenization key.

jaxdesmarais: It looks like for the REST APIs we have logic to add one or the other based on client token vs tokenization key on iOS: https://github.com/braintree/braintree_ios/blob/ec471c75f4b05f8ccf2225f95099cf4fc8f28db4/Sources/BraintreeCore/BTHTTP.swift#L235-L240

For GraphQL we always add it: https://github.com/braintree/braintree_ios/blob/ec471c75f4b05f8ccf2225f95099cf4fc8f28db4/Sources/BraintreeCore/BTGraphQLHTTP.swift#L63

tdchow: Thanks for the iOS context! The Android GraphQL Client has matching logic.

But for Rest APIs it sounds like we can match iOS's logic in only setting one or the other based on the Authorization type.

tdchow: Just pushed https://github.com/braintree/braintree_android/pull/1407/commits/2007e9a757869871cc3ab14e81be9b033cbce623 updating the auth header based on the Authorization type

warmkesselj (approved):

PR #1410, opened 2025-08-26
Change configuration cache length

saralvasquez:
### Summary of changes

 - changed `ConfigurationCache` `TIME_TO_LIVE` from 5 minutes to 1 day

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors
@saralvasquez


BraintreeCore/src/main/java/com/braintreepayments/api/core/ConfigurationCache.kt
```diff
@@ -42,7 +42,7 @@ internal class ConfigurationCache(
     }
 
     companion object {
-        private val TIME_TO_LIVE = TimeUnit.MINUTES.toMillis(5)
+        private val TIME_TO_LIVE = TimeUnit.DAYS.toMillis(1)
```
warmkesselj, line 45: Just curious about the reasoning for this update. Like, what kind of side-effects to expect from our side with testing?

BraintreeCore/src/main/java/com/braintreepayments/api/core/ConfigurationLoader.kt
```diff
@@ -18,6 +17,7 @@ internal class ConfigurationLoader(
     },
 ) {
     private val analyticsClient: AnalyticsClient by lazyAnalyticsClient
+    internal var merchantId: String? = null
```
warmkesselj: This name confuses me a bit. Would it make more sense to use `configurationCacheKey` instead of merchantId?

warmkesselj: Maybe some commenting could help?

BraintreeCore/src/main/java/com/braintreepayments/api/core/ConfigurationCache.kt
saralvasquez: This length of cache was decided in the discussions for the story that lead to these changes

BraintreeCore/src/main/java/com/braintreepayments/api/core/ConfigurationLoader.kt
saralvasquez: That's fair. I think I decided on `merchantId` as that's what was most descriptive of what was being saved to the variable, but I could also see an argument for `cacheKey` which is what was used before

BraintreeCore/src/main/java/com/braintreepayments/api/core/ConfigurationCache.kt
warmkesselj: I wonder if it would make sense to have different duration for a debug versus release scheme?

BraintreeCore/src/main/java/com/braintreepayments/api/core/ConfigurationLoader.kt
tdchow: Hm now that I'm thinking about it, I'm not sure if it makes sense to base the cache key on the merchant ID. Since the merchant ID value is returned in the configuration, the merchant ID / cache key would never change which would never invalidate the cache. The cache would only be cleared based on the time constraint.

I'm not sure if this would be an issue, but if a single merchant integration can handle multiple merchant IDs, the cache would never be cleared for subsequent launches of the SDK for a different merchant ID.

saralvasquez: Yeah Justin and I were just discussing that. I think the fact that the merchant ID will never change makes the 1 day cache length tricky. I think if we wanted to have the longer cache length and the merchant ID as the cache key we would need to add extra functionality to invalidate the cache. I think it probably makes more sense to keep is as the bearer + url combo

saralvasquez: Additionally, the merchant ID as cache key was really just to keep consistent with the gateway right? Like there isn't any real benefit on our end to changing it

saralvasquez: Also why 1 day for cache length? I think from the thread in slack it was to avoid invalidating the cache during a checkout flow right? Would a couple hours be a better time frame? 5 minutes to 1 day feels like a big jump

tdchow: I'm guessing GW is managing merchant ID in a different way. But since the SDK gets the merchant ID in the configuration response, i think keeping the existing cache key makes sense.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
   * Add `paymentId` to `PayPalAccountNonce`
 * GooglePay
     * Add `softwareInfo` details to the `merchantInfo` field in the `loadPaymentData` request.
+* BraintreeCore
+  * Change configuration cache time to 1 day
```
jaxdesmarais, line 12: I think we will need to pull latest from main since we did a release yesterday

BraintreeCore/src/main/java/com/braintreepayments/api/core/ConfigurationCache.kt
saralvasquez: Yeah that's a good question and now I'm sort of wondering if having it set to such a long time frame is wise. @tdchow is there a reason behind such a big jump? Would having a shorter window still fix the issue of cache invalidation during a payment flow?

saralvasquez:
I think changes to the configuration cache need a bit more discussion. Closing this for now

PR #1422, opened 2025-09-03
Legacy HTTP Clean up and Result type

tdchow:
### Summary of changes

 - Remove legacy HTTP classes
 - Add a `Result` sealed class to the `NetworkResponseCallback`

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/ConfigurationLoader.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/ConfigurationLoader.kt
@@ -40,40 +40,55 @@ internal class ConfigurationLoader(
         cachedConfig?.let {
             callback.onResult(ConfigurationLoaderResult.Success(it))
         } ?: run {
-            httpClient.get(
-                path = configUrl,
-                configuration = null,
-                authorization = authorization,
-            ) { result ->
-                when (result) {
-                    is NetworkResponseCallback.Result.Success -> {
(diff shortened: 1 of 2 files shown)


saperi22 (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/ConfigurationLoader.kt
```diff
@@ -43,32 +44,34 @@ internal class ConfigurationLoader(
...
-                        val configuration = Configuration.fromJson(responseBody)
-                        saveConfigurationToCache(configuration, authorization, configUrl)
-                        callback.onResult(ConfigurationLoaderResult.Success(configuration, timing))
+            ) { result ->
+                when (result) {
+                    is NetworkResponseCallback.Result.Success -> {
+                        val responseBody = result.response.body
+                        val timing = result.response.timing
+                        try {
+                            val configuration = Configuration.fromJson(responseBody ?: "")
```
saralvasquez: Just for my knowledge, are we expecting there to be cases where the response body is null or is this just to be safe?

tdchow: That's a good question. We should be able to assume that if the call was successful, there should be a non null body. I'll add a null check and add an error case for that.

PR #1432, opened 2025-09-05
Convert ThreeDSecureNonceUnitTest to kotlin

saralvasquez:
### Summary of changes

 - converts `ThreeDSecureNonceUnitTest` to kotlin

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### Authors
@saralvasquez


saperi22 (approved):

ThreeDSecure/src/test/java/com/braintreepayments/api/threedsecure/ThreeDSecureNonceUnitTest.kt
```diff
@@ -0,0 +1,143 @@
...
+    @Test
+    fun `fromJSON parses ThreeDSecureNonce`() {
+        val sut = ThreeDSecureNonce.fromJSON(JSONObject(Fixtures.PAYMENT_METHODS_RESPONSE_VISA_CREDIT_CARD))
+
+        assertEquals("Visa", sut.cardType)
+        assertEquals("123456-12345-12345-a-adfa", sut.string)
+        assertEquals("11", sut.lastTwo)
+        assertEquals("1111", sut.lastFour)
+        assertNotNull(sut.threeDSecureInfo)
+        assertFalse { sut.threeDSecureInfo.liabilityShifted }
```
noguier: minor suggestion: I noticed that some assertFalse, assertTrue pass in a boolean value, and some a lambda block. Maybe we can use one or the other in this test, just for the unity.

noguier (approved):

ThreeDSecure/src/test/java/com/braintreepayments/api/threedsecure/ThreeDSecureNonceUnitTest.kt
saralvasquez: Interesting! Do you have an example on hand? I've primarily been doing it this way but only because I haven't seen otherwise

saralvasquez: Never mind! I did a quick search and found both. It seems like there's a pretty even split across all the test files. We could go through an unify them all but I'm not sure how useful that would be. Maybe that would be a good open dev story

saralvasquez: I'd say as long as it's consistent within the same file it should be ok so I'll go ahead and change this file to stick to one.

PR #1434, opened 2025-09-08
PayPal app link resolution for improved app switch success rate & instrumentation

Harish2207:

### Behaviour changes

Existing behaviour: App-switch will be enabled by the SDK if there's merchant-opt-in & if the Paypal app is installed

#### Enhancement
- App-switch will be enabled by the SDK if the Paypal app can resolve the universal link in addition to the existing checks before sending it as enabled in the checkout requests. This is a preemptive measure to ensure more reliable Paypal app invocations.
- A new `ResolvePayPalUseCase` that checks if the PayPal app will resolve the app-switch URL.

### Instrumentation changes

- A new FPTI tag `paypal_app_switch_url_received` will be sent to more accurately reflect that the SDK has received an app switch URL from the Paypal APIs.
- The existing FPTI tag `attempted_app_switch` will now be sent to more accurately reflect app-switch attempts after verifying the Paypal app can resolve the URL.

Tested with the Paypal app's open supported links toggled on and off with multiple browsers.

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage


BraintreeCore/src/main/java/com/braintreepayments/api/core/ShouldFallbackToBrowserUseCase.kt
```diff
@@ -0,0 +1,46 @@
...
+    enum class Result {
+        FALLBACK, APP_SWITCH
+    }
+
+    operator fun invoke(
+        uri: Uri?,
+        appPackage: String = PAYPAL_APP_PACKAGE
+    ): Result {
+        if (uri == null) {
+            return Result.APP_SWITCH
```
saperi22: Should this be app switch?
Both app switch and browser switch both require a uri to work. Would it be better if we handle the uri being null before it gets to this class?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalLauncher.kt
```diff
@@ -72,12 +76,42 @@ class PayPalLauncher internal constructor(
             return sendLaunchFailureEventAndReturn(error, isAppSwitch, analyticsEventParams)
         }
 
+        val fallbackToBrowser = when {
+            !isAppSwitch -> false
```
saperi22: I'm having a hard time reading this.
when `isAppSwitch` is false (i.e. it is not app switch), `!isAppSwtich` evaluates to true and `fallbackToBrowser` is false. Is this the intended behavior?
If it isn't app switch and it isn't browser switch, what is it?

BraintreeCore/src/main/java/com/braintreepayments/api/core/ShouldFallbackToBrowserUseCase.kt
```diff
@@ -0,0 +1,42 @@
...
+import android.content.pm.PackageManager
+import android.net.Uri
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class ShouldFallbackToBrowserUseCase(
+    private val merchantRepository: MerchantRepository
+) {
+
+    companion object {
```
saperi22: nit: `companion object` typically goes at the bottom of the file so the important stuff is up top.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalLauncher.kt
```diff
@@ -72,12 +76,45 @@ class PayPalLauncher internal constructor(
...
+        }
+
+        val opensInBrowser = when {
+            !isAppSwitch -> true
+            else -> {
+                val result = shouldFallbackToBrowserUseCase(uri)
+
+                when (result) {
+                    ShouldFallbackToBrowserUseCase.Result.FALLBACK -> {
+                        analyticsClient.sendEvent(PayPalAnalytics.APP_SWITCH_FAILED,
```
saperi22: Should this be app switch failed? I don't think so.
As I understand, today, `APP_SWITCH_FAILED` refers to an attempt and something went wrong.
This could be a new state -- unable to app switch. Thoughts?

saperi22: I see that you've introduced app switch attempted, this could potentially be kept as failed. Let me check.

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetPaypalLaunchTypeUseCase.kt
```diff
@@ -0,0 +1,42 @@
...
+
+        return if (opensInTargetApp) {
+            Result.APP
+        } else {
+            Result.BROWSER
+        }
+    }
+
+    companion object {
+        private const val PAYPAL_APP_PACKAGE = "com.paypal.android.p2pmobile"
```
jaxdesmarais: This is already defined here: https://github.com/braintree/braintree_android/blob/007ef5ffe350e1ba99cc90f06fc97c5e9e805838/BraintreeCore/src/main/java/com/braintreepayments/api/core/DeviceInspector.kt#L127

It'd be great to not maintain this in 2 places if it ever needs to change.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalAnalytics.kt
```diff
@@ -19,6 +19,7 @@ internal object PayPalAnalytics {
     const val HANDLE_RETURN_NO_RESULT = "paypal:tokenize:handle-return:no-result"
 
     // App Switch events
+    const val APP_SWITCH_ATTEMPTED = "paypal:tokenize:app-switch:attempted"
```
jaxdesmarais: The event names should be the same across iOS and Android and are not in the 2 PRs open. Please decide and align on names to use across the 2 platforms.

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetPaypalLaunchTypeUseCase.kt
ram-gh: Addressed in https://github.com/braintree/braintree_android/commit/5734531ac3de26ba454f2d93bdcc10d42804edd1

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalAnalytics.kt
ram-gh: This extra event for Android is intentional. This is due to an Android specific app-setting that users can use to turn off "Open supported links" for the PayPal app. The funnel between `paypal:tokenize:app-switch:attempted` and `paypal:tokenize:app-switch:started` can be used to track how many users fail to app-switch due to turning this setting off and more accurately instrument actual app-switch attempts vs starts. 

For iOS with changes in [1648](https://github.com/braintree/braintree_ios/pull/1648),  we have a completion handler which will be invoked with a clear indication if the universal link opened the Paypal app.

ram-gh:
> Hi @Harish2207 - I just took a look at the code changes. It looks like we may be duplicating some logic. There is already a check to see if the user has disabled "Open supported links" checkbox in the merchant app settings. Take a look at: https://github.com/braintree/braintree_android/blob/main/BraintreeCore/src/main/java/com/braintreepayments/api/core/GetReturnLinkTypeUseCase.kt
> 
> Currently, app switch should only be enabled when app links are available. If the "Open supported links" checkbox has been disabled, we should be attempting to use an opt in deep link parameter set by the merchant, which would open a CCT. Is this logic not working as expected?

Hi @tdchow The logic we have in `GetReturnLinkTypeUseCase.kt` is for tracking the merchant app's "Open supported links" setting. This PR introduces `GetPaypalLaunchTypeUseCase.kt` for tracking Paypal app's "Open supported links". 

While this PR only improves instrumentation, ideally, we should also use `GetPaypalLaunchTypeUseCase.kt`'s result as a marker for opt-in in addition to `GetReturnLinkTypeUseCase.kt`

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetPaypalLaunchTypeUseCase.kt
```diff
@@ -0,0 +1,38 @@
+package com.braintreepayments.api.core
```
tdchow: Does this class need to be defined in the core module? Can it live in `com.braintreepayments.api.paypal`?

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetPaypalLaunchTypeUseCase.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.net.Uri
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class GetPaypalLaunchTypeUseCase(
```
tdchow: Can we add a Kdoc for this class - it'll help give context on when and why it's used.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalAnalytics.kt
tdchow: Would it be better to add the requirement of having the PayPal app having "Open supported links" enabled to the actual app switch logic instead of the instrumentation only change? I feel that adding another event in the funnel only for a single platform may introduce some confusion.

If we add this check to app switch eligibility on Android, we could add a new FPTI tag that represents if the "Open supported links" checkbox is enabled.

ram-gh: Yeah, this was the original plan. We had to settle for an instrumentation only change to avoid scope creep. Now that we mutually agree / align on the value prop, we will change the existing behaviour as to when actual app-switch is enabled by adding in this condition. 

We will have changes ready on Sep 17. 
Please help with a re-review & we can aim to merge this PR by the end of this week. @tdchow

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetPaypalLaunchTypeUseCase.kt
ram-gh: Addressed in [13ff38c](https://github.com/braintree/braintree_android/pull/1434/commits/13ff38c6fe8cd32f9c236e6354a757a1e1af653c)

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetPaypalLaunchTypeUseCase.kt
ram-gh: Please review the latest changes.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalAnalytics.kt
ram-gh: @tdchow Please re-review the latest changes.

ram-gh:
> Tested with the Paypal app's open supported links toggled on and off with multiple browsers here: https://paypal.atlassian.net/wiki/x/Gq7Io

Tested with the Paypal app's open supported links toggled on and off with multiple browsers here: https://paypal.atlassian.net/wiki/x/Gq7Io

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalAppSwitchResolver.kt
```diff
@@ -0,0 +1,25 @@
+package com.braintreepayments.api.paypal
+
+import android.net.Uri
+import com.braintreepayments.api.core.MerchantRepository
+
+/**
+ * Determines if the PayPal app can resolve app-switch URLs.
+ */
+internal object PayPalAppSwitchResolver {
```
tdchow: Is there a benefit in splitting out the logic in this class? We're trying to follow clean/mvvm architecture, so this looks like a use case as it does not contain any state.

Can we either rename this class to append `UseCase` or can we consolidate the logic with the `GetPaypalLaunchTypeUseCase`?

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
```diff
@@ -141,9 +142,10 @@ internal class AnalyticsApi(
         private const val FPTI_KEY_BUTTON_POSITION = "button_position"
         private const val FPTI_KEY_PAGE_TYPE = "page_type"
         private const val FPTI_KEY_MERCHANT_ENABLED_APP_SWITCH = "merchant_enabled_app_switch"
-        private const val FPTI_KEY_PAYPAL_ATTEMPTED_APP_SWITCH = "attempted_app_switch"
```
tdchow, line 144: Should we be removing this existing FPTI tag? I feel we'd still want to report when the PP APIs tell the SDK if we should app switch or launch the CCT.

Can we make the new tag an addition?

PayPal/src/main/java/com/braintreepayments/api/paypal/GetPaypalLaunchTypeUseCase.kt
```diff
@@ -0,0 +1,58 @@
...
+    enum class Result {
+        /** The payment flow will be launched in a web browser */
+        BROWSER,
+
+        /** The payment flow will be launched in the PayPal app */
+        APP
+    }
+
+    /** Detects if users disable the "Open supported links" setting in the Paypal app.
+     *
```
tdchow:
```suggestion
    /** 
     * Detects if users disable the "Open supported links" setting in the Paypal app.
```

PayPal/src/main/java/com/braintreepayments/api/paypal/GetPaypalLaunchTypeUseCase.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.braintreepayments.api.core.MerchantRepository
+
+/**
+ * Use case that determines whether the PayPal app-switch link will open in the PayPal app
+ * or in a browser.
+ *
+ * Detects if a user unchecks the "Open supported links" checkbox in the Android OS settings for the PayPal app.
+ * If this setting is unchecked, this use case will return [Result.APP], otherwise [Result.BROWSER] will be returned.
+ *
+ */
```
tdchow:
```suggestion
 * If this setting is unchecked, this use case will return [Result.APP], otherwise [Result.BROWSER] will be returned.
 */
```

PayPal/src/main/java/com/braintreepayments/api/paypal/GetPaypalLaunchTypeUseCase.kt
```diff
@@ -0,0 +1,58 @@
...
+) {
+
+    /**
+     * Represents the possible launch types for PayPal payment flows.
+     */
+    enum class Result {
+        /** The payment flow will be launched in a web browser */
+        BROWSER,
+
+        /** The payment flow will be launched in the PayPal app */
```
tdchow:
```suggestion
        /** 
         * The payment flow will be launched in the PayPal app 
         */
```

PayPal/src/main/java/com/braintreepayments/api/paypal/GetPaypalLaunchTypeUseCase.kt
```diff
@@ -0,0 +1,58 @@
...
+ */
+internal class GetPaypalLaunchTypeUseCase(
+    private val merchantRepository: MerchantRepository
+) {
+
+    /**
+     * Represents the possible launch types for PayPal payment flows.
+     */
+    enum class Result {
+        /** The payment flow will be launched in a web browser */
```
tdchow:
```suggestion
        /** 
         * The payment flow will be launched in a web browser
         */
```

PayPal/src/main/java/com/braintreepayments/api/paypal/GetPaypalLaunchTypeUseCase.kt
```diff
@@ -0,0 +1,58 @@
...
+ *
+ */
+internal class GetPaypalLaunchTypeUseCase(
+    private val merchantRepository: MerchantRepository
+) {
+
+    /**
+     * Represents the possible launch types for PayPal payment flows.
+     */
+    enum class Result {
```
tdchow: I'm wondering if we want to restructure this result type. The result of this use case seems to represent if the PayPal app has "Open supported links" enabled, not if the SDK will actually launch the payment flow in a Chrome Custom Tab or the PayPal app. There is additional eligibility logic in the SDK to determine if app switch should happen. 

Maybe we can restructure the result to `PAYPAL_SUPPORTED_LINKS_ENABLED` and `PAYPAL_SUPPORTED_LINKS_DISABLED`. There's probably better naming for this, but just wanted to provide an example.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
ram-gh: We are not removing it. We have two tags in the same file.

`FPTI_KEY_PAYPAL_RECEIVED_APP_SWITCH_URL` : To denote if the PP APIs returned an app-switch url.
`FPTI_KEY_PAYPAL_ATTEMPTED_APP_SWITCH` : To denote if the SDK really attempted an app-switch.

PayPal/src/main/java/com/braintreepayments/api/paypal/GetPaypalLaunchTypeUseCase.kt
ram-gh: The SDK always launches the payment flow in the CCT which then opens the Paypal app if it is able to resolve the app-link or loads the page in the same CCT context if the app-link is not resolved.

This use-case is invoked in two places: 
1. The vital invocation: `processAppSwitchAttempt` in `PaypalLauncher`: This is after we receive an app-switch eligible app-link from Paypal APIs. I feel it is more appropriate to call the returned result as App & Browser. This feeds our updated instrumentation tag `attempted_app_switch`
2. The functional invocation: Last validation to enable app-switch after other eligibility conditions are met (merchant-opt-in, app installation check). 

Also, this naming is consistent with the existing use-case which checks for open supported links of the merchant app: `GetReturnLinkTypeUseCase`

tdchow: I was more thinking of the use case when the PayPal app has "Open supported links" enabled, but the merchant's app has "Open supported links" disabled. So the actual payment flow will not app switch. I think the result types of this Use Case are a little misleading. I could see an engineer getting confused when this UseCase returns `App` but the actual flow happens in the CCT.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
tdchow: Apologies, I'm now seeing this constant moved to the bottom of the list. Any reason why it was moved? Might be good to grouop all of the App Switch tags together.

ram-gh: Addressed in [058b0c5](https://github.com/braintree/braintree_android/pull/1434/commits/058b0c51bf1e2538ecc0cbcc36f5da529e1d2488)

PayPal/src/main/java/com/braintreepayments/api/paypal/GetPaypalLaunchTypeUseCase.kt
ram-gh: `I could see an engineer getting confused when this UseCase returns App but the actual flow happens in the CCT.` --> The code flow ensures this doesn't happen already. After we receive a response from PayPal APIs, the merchant app's "Open supported links" is checked before the PayPal app's "Open supported links" is checked. If the merchant app's "Open supported links" were to be disabled, we ensure the latter is not executed and is returned promptly. 

Agree that we can refactor the return types. 
We have refactored it to be more simple and concise in the latest commit.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalAppSwitchResolver.kt
ram-gh: Addressed in [20a894d](https://github.com/braintree/braintree_android/pull/1434/commits/20a894da0b8dad3efc8fd2b1b29dd5d888e588b6)

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsParamRepository.kt
```diff
@@ -21,10 +21,16 @@ class AnalyticsParamRepository(
...
-     * App switch attempted based on the server side response
+     * Whether or not app-switch URL was received from the PayPal server response
      */
     var didPayPalServerAttemptAppSwitch: Boolean? = null
 
+    /**
+     * Whether or not the SDK attempted to perform an app switch.
+     */
+
+    var didSdkAttemptAppSwitch: Boolean? = null
```
tdchow, line 32:
```suggestion
     */
    var didSdkAttemptAppSwitch: Boolean? = null
```

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalLauncher.kt
```diff
@@ -158,7 +166,7 @@ class PayPalLauncher internal constructor(
         }
     }
 
-    private fun sendLaunchFailureEventAndReturn(
+private fun sendLaunchFailureEventAndReturn(
```
tdchow:
```suggestion
    private fun sendLaunchFailureEventAndReturn(
```

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalLauncher.kt
```diff
@@ -67,10 +67,18 @@ class PayPalLauncher internal constructor(
...
+                isAppSwitch,
+                analyticsEventParams
+            )
+
+        options.url
+            ?: return buildLaunchFailureAndReturn(
+                "BrowserSwitchOptions URL is null",
+                isAppSwitch,
+                analyticsEventParams
+            )
```
tdchow, line 81: Suggestion - named arguments may help with readability here, especially for the first argument.

```suggestion
            ?: return buildLaunchFailureAndReturn(
                errorMessage = "BrowserSwitchOptions is null",
                isAppSwitch = isAppSwitch,
                analyticsEventParams = analyticsEventParams
            )

        options.url
            ?: return buildLaunchFailureAndReturn(
                errorMessage = "BrowserSwitchOptions URL is null",
                isAppSwitch = isAppSwitch,
                analyticsEventParams = analyticsEventParams
            )
```

tdchow (approved):
Changes look good. Just had a few minor formatting suggestions.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
```diff
@@ -141,6 +142,7 @@ internal class AnalyticsApi(
         private const val FPTI_KEY_BUTTON_POSITION = "button_position"
         private const val FPTI_KEY_PAGE_TYPE = "page_type"
         private const val FPTI_KEY_MERCHANT_ENABLED_APP_SWITCH = "merchant_enabled_app_switch"
+        private const val FPTI_KEY_PAYPAL_RECEIVED_APP_SWITCH_URL = "paypal_app_switch_url_received"
```
jaxdesmarais, line 145: It does not look like this has been approved as an FPTI tag yet. Please work with the FPTI team to get the tag approved to ensure it doesn't get removed down the road.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsParamRepository.kt
```diff
@@ -21,10 +21,15 @@ class AnalyticsParamRepository(
     var didEnablePayPalAppSwitch: Boolean? = null
 
     /**
-     * App switch attempted based on the server side response
+     * Whether or not app-switch URL was received from the PayPal server response
      */
     var didPayPalServerAttemptAppSwitch: Boolean? = null
 
+    /**
+     * Whether or not the SDK attempted to perform an app switch.
```
jaxdesmarais: Can we add more detail here on what we are using to determine if the SDK attempted to perform app switch similar to the docstring above?

BraintreeCore/src/main/java/com/braintreepayments/api/core/DeviceInspector.kt
```diff
@@ -124,7 +124,7 @@ class DeviceInspector(
         get() = getDropInVersion()
 
     companion object {
-        private const val PAYPAL_APP_PACKAGE = "com.paypal.android.p2pmobile"
+        const val PAYPAL_APP_PACKAGE = "com.paypal.android.p2pmobile"
```
jaxdesmarais, line 127: I expect we don't want merchants to use this. Should we add docstrings around that or are we covered here? cc: @tdchow

tdchow: We should be okay here since the entire class is annotated with `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)`. It would apply this scope to the entire class and it's public properties.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
ram-gh: https://internal.paypalinc.com/fptigov/tags/dictionary?tagStatus=PENDING&tagId=98324
Will follow-up and get it approved. This should not be a blocker for merge.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsParamRepository.kt
ram-gh: Addressed in [669c698](https://github.com/braintree/braintree_android/pull/1434/commits/669c69857fe5dcfbeb73ec31fb6ebe0e17db8eeb)

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
ram-gh: Tag is approved 

Screenshot 2025-09-25 at 12 21 56

PR #1436, opened 2025-09-11
Centralize Java version configuration using Gradle variables

Jake-Jiang:
Centralize Java version configuration using Gradle variables

In all modules' build.gradle files, replace hard-coded `jvmTarget` and `kotlinOptions` with `version variables defined in the root project.

This change centralizes Java compatibility version configuration in one place, improving project maintainability. Future Java version updates require only one modification.

### DIFF
--- a/Demo/build.gradle
+++ b/Demo/build.gradle
@@ -56,8 +56,18 @@ android {
     }
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_17
-        targetCompatibility JavaVersion.VERSION_17
+        sourceCompatibility versions.javaSourceCompatibility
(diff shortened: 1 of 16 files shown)


Demo/build.gradle
```diff
@@ -56,8 +56,18 @@ android {
     }
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_17
-        targetCompatibility JavaVersion.VERSION_17
```
tdchow, line 60: I think we still may want to set source and target versions to 17 for the demo app.

Jake-Jiang: OK，I have changed the commit to keep source and target versions to 17 for the demo app.

noguier (approved):

PR #1447, opened 2025-10-03
Fix Card instrumentation tests and properly handle json errors

tdchow:
### Summary of changes

 - Add network_security_config.xml to fix SSL errors
 - Update logic in card client to properly handle errors in json
 - Fix Instrumentation tests
 
Screenshot 2025-10-03 at 3 09 44 PM

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
 # Braintree Android SDK Release Notes
+
+## unreleased
+
+* Card
+  * Fix an issue where a tokenize card errors are masked with a JSON parsing error
   
 ## 5.18.0 (2025-10-01)
 


tdchow:
@tdchow to update changelog for updated error behavior.

richherrera (approved):

Card/src/androidTest/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <domain-config>
+        <domain includeSubdomains="true">braintree-demo-merchant-63b7a2204f6e.herokuapp.com</domain>
```
saperi22, line 4: Should we be including this at the top level or is it sufficient if we include for the Card module?

saperi22 (approved):

Card/src/androidTest/res/xml/network_security_config.xml
tdchow: I think we'll just want this config set for our instrumentation tests and not for the source SDK.

PR #1454, opened 2025-10-17
Remove Assertions class and use assertEquals instead

tdchow:
### Summary of changes

 - Remove Assertions class that was used for comparing `BinData` objects
 - Replace with `assertEquals()` as `BinData` is a `data` class
 - Remove unused `BraintreeActivityTestRule` class

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


TestUtils/src/main/java/com/braintreepayments/api/testutils/BraintreeActivityTestRule.java
noguier, line 1: Just out of curiosity, what is this file, and why do we no longer need it?

noguier (approved):

TestUtils/src/main/java/com/braintreepayments/api/testutils/BraintreeActivityTestRule.java
tdchow: The class seems to clear all shared prefs from the device/emulator and enable keyguard for UI tests. I think the UI tests that relied on this class were removed. Not exactly sure why the tests were removed though 🤔

PR #1455, opened 2025-10-21
Add Pay Later Button

6c6z7ymqmk-sudo:

### Summary of changes

 - Add a paylater button on Demo 
 - Add background color for Buttons 

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### Screenshot

Screenshot_20251028_095301

Screenshot 2025-10-30 at 4 31 30 PM

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
@@ -52,7 +52,6 @@ public class PayPalFragment extends BaseFragment {
         Button singlePaymentButton = view.findViewById(R.id.paypal_single_payment_button);
         Button singlePaymentPayLaterButton = view.findViewById(R.id.paypal_single_payment_pay_later_button);
         Button paymentWithCreditButton = view.findViewById(R.id.paypal_pay_with_credit_button);
-        Switch offerPayLater = view.findViewById(R.id.offer_pay_later_switch);
         Switch contactInformationSwitch = view.findViewById(R.id.contact_info_switch);
         Switch amountBreakdownSwitch = view.findViewById(R.id.amount_breakdown_switch);
 
@@ -68,6 +67,7 @@ public class PayPalFragment extends BaseFragment {
                 amountBreakdownSwitch.isChecked()
             );
         });
+
         paymentWithCreditButton.setOnClickListener(v -> {
             launchPayPal(
                 false,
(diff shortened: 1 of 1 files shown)


Demo/src/main/res/layout/fragment_paypal.xml
```diff
@@ -59,6 +59,7 @@
         android:id="@+id/offer_pay_later_switch"
         android:layout_width="match_parent"
         android:layout_height="@dimen/margin_40"
+        android:visibility="gone"
         android:text="@string/paypal_offer_pay_later" />
```
tdchow: Would it be better to remove this `Switch` entirely? It doesn't look like it's ever set to visible.

6c6z7ymqmk-sudo: Good point. I suggest keeping this feature, as it’s still used in other places. We can add a feature switch to control the visibility of the second button I implemented, as well as this switch.

tdchow: I think with your current changes we're not able to test all of the combinations of one time checkout, billing agreement, pay later enabled, pay later disabled.

Since we're adding a dedicated button for pay later, can we add one for one time checkout and separate one for launching the billing agreement flow?

6c6z7ymqmk-sudo: After reviewing the codebase, Billing Agreements don’t currently support Pay Later. The first two buttons handle (1) single payments and (2) billing agreements, both with payLater = false. The newly added button handles single payments with payLater = true. This setup fulfills our current business needs.

tdchow: Got it. So it sounds like the this `Switch` is still not used. Can we remove it as a part of this PR?

6c6z7ymqmk-sudo: Removed

saperi22 (approved):

Demo/src/main/java/com/braintreepayments/demo/PayPalFragment.java
```diff
@@ -71,9 +73,23 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
+        });
+
+        singlePaymentPayLaterButton.setOnClickListener(v -> {
+            launchPayPal(
+                    false,
+                    buyerEmailEditText.getText().toString(),
+                    buyerPhoneCountryCodeEditText.getText().toString(),
+                    buyerPhoneNationalNumberEditText.getText().toString(),
+                    contactInformationSwitch.isChecked(),
+                    true,
```
jaxdesmarais, line 106: Why are we not using the value of `isPayLaterEnabled` here? It seems odd that we are setting this value in 2 different ways across this PR in addition to providing a default.

6c6z7ymqmk-sudo: I renamed `isPayLaterEnabled` to `isPayLaterSelected`, since the variable is meant to track whether the customer has chosen the Pay Later option and pass that state to `ShowNonceFragment`.
I also darkened the background to improve text contrast and make the content stand out more. The original demo app had several visible design issues, so I made these adjustments to bring it more in line with the design standards of our other apps.

saperi22:
@jaxdesmarais do you have enough info. and do you agree on merging this PR in?

PR #1459, opened 2025-10-30
Revert "Add Scanning Pattern for Forks"

saperi22:
Reverts braintree/braintree_android#1457

Sounds like this was a test and wasn't supposed to be merged in yet.
@JLESUS please correct me if I'm wrong.


JLESUS (approved):
LGTM, both files should be removed. We will need to make rulesets on the repo level for now as a short term fix as we examine better ways to address the issue on an org-level.

PR #1460, opened 2025-11-03
Unsupported browser usecase

saperi22:
### Summary of changes

 - Add use case to fetch the default browser for the device
 - Add use case to check if app links are supported for the default browser

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/GetAppLinksCompatibleBrowserUseCase.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/GetAppLinksCompatibleBrowserUseCase.kt
@@ -12,8 +12,8 @@ class GetAppLinksCompatibleBrowserUseCase(
 ) {
 
     operator fun invoke(): Boolean {
-        getDefaultBrowserUseCase()?.let {
-            return appLinkCompatibleBrowsers.contains(it)
+        getDefaultBrowserUseCase()?.let { defaultBrowser ->
+            return appLinkCompatibleBrowsers.any { defaultBrowser.contains(it) }
         }
         return false
     }
--- a/BraintreeCore/src/test/java/com/braintreepayments/api/core/GetAppLinksCompatibleBrowserUseCaseUnitTest.kt
+++ b/BraintreeCore/src/test/java/com/braintreepayments/api/core/GetAppLinksCompatibleBrowserUseCaseUnitTest.kt
@@ -30,6 +30,24 @@ class GetAppLinksCompatibleBrowserUseCaseUnitTest {
     }
 
     @Test
+    fun `when invoke is called with Chrome canary as default browser, returns true`() {
+        every { getDefaultBrowserUseCase() } returns "com.android.chrome.canary"
+
+        val result = sut()
+
+        assertEquals(true, result)
+    }
+
+    @Test
+    fun `when invoke is called with incomplete appId of Chrome as default browser, returns false`() {
+        every { getDefaultBrowserUseCase() } returns "com.android"
+
+        val result = sut()
+
+        assertEquals(false, result)
+    }
+
+    @Test
     fun `when invoke is called with Brave as default browser, returns true`() {
         every { getDefaultBrowserUseCase() } returns "com.brave.browser"
 


BraintreeCore/src/main/java/com/braintreepayments/api/core/AppLinkCompatibleBrowserUseCase.kt
```diff
@@ -0,0 +1,36 @@
+package com.braintreepayments.api.core
+
+import android.content.pm.PackageManager
+
+class AppLinkCompatibleBrowserUseCase(
```
tdchow: Can we make mark these use cases as internal?

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetAppLinksCompatibleBrowserUseCase.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    operator fun invoke(): Boolean {
+        getDefaultBrowserUseCase(packageManager)?.let {
+            return appLinkCompatibleBrowsers.contains(it)
+        }
+        return false
+    }
+
+    companion object {
+        private val appLinkIncompatibleBrowsers = listOf(
```
tdchow: Do we need to include the list of incompatible browsers if we're only using the list of compatible browsers as an allow list?

BraintreeCore/src/test/java/com/braintreepayments/api/core/AppLinkCompatibleBrowserUseCaseUnitTest.kt
```diff
@@ -0,0 +1,150 @@
...
+class AppLinkCompatibleBrowserUseCaseUnitTest {
+
+    private val getDefaultBrowserUseCase: GetDefaultBrowserUseCase = mockk(relaxed = true)
+    private val packageManager: PackageManager = mockk(relaxed = true)
+
+    lateinit var subject: AppLinkCompatibleBrowserUseCase
+
+    @Before
+    fun setUp() {
+        subject = AppLinkCompatibleBrowserUseCase(getDefaultBrowserUseCase, packageManager)
```
saralvasquez: Nit: is this the same as sut? If so I think consistency in the unit test files would be good

BraintreeCore/src/test/java/com/braintreepayments/api/core/GetDefaultBrowserUseCaseUnitTest.kt
```diff
@@ -0,0 +1,97 @@
...
+    private val merchantRepository: MerchantRepository = mockk(relaxed = true)
+    private val packageManager: PackageManager = mockk(relaxed = true)
+    private val appLinkReturnUri = Uri.parse("https://example.com")
+
+    lateinit var subject: GetDefaultBrowserUseCase
+
+    @Before
+    fun setUp() {
+        every { merchantRepository.appLinkReturnUri } returns appLinkReturnUri
+        subject = GetDefaultBrowserUseCase(merchantRepository)
```
saralvasquez: Same nit as the other unit test file

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetAppLinksCompatibleBrowserUseCase.kt
```diff
@@ -0,0 +1,36 @@
+package com.braintreepayments.api.core
+
+import android.content.pm.PackageManager
+
+class AppLinkCompatibleBrowserUseCase(
+    private val getDefaultBrowserUseCase: GetDefaultBrowserUseCase,
+    private val packageManager: PackageManager
+) {
+
+    operator fun invoke(): Boolean {
```
saralvasquez, line 14: Given this approach to calling this class I wonder if a different class name would make it more intuitive to use. Like a verb based name similar to the `GetDeafultBrowserUseCase`

saralvasquez (approved):
Couple non-blocking nits but otherwise looks good to me!

Also a question. `UseCase` feels like an unintuitive naming convention in this context. Where did that come from? I ask because the only context I've heard "use case" has been in high level planning but not inside the code itself

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetAppLinksCompatibleBrowserUseCase.kt
saralvasquez: Oh also if we do keep the incompatible list could we name it something more distinguishable from the allow list? It's a bit hard to see the difference at a quick glance

saperi22:
> Couple non-blocking nits but otherwise looks good to me!
> 
> Also a question. `UseCase` feels like an unintuitive naming convention in this context. Where did that come from? I ask because the only context I've heard "use case" has been in high level planning but not inside the code itself

Great question!
Google defines it to represent the logic in the domain layer. See [this](https://developer.android.com/topic/architecture/domain-layer#conventions)

BraintreeCore/src/test/java/com/braintreepayments/api/core/AppLinkCompatibleBrowserUseCaseUnitTest.kt
saperi22: claude came up with subject ;)
i'll change it

BraintreeCore/src/test/java/com/braintreepayments/api/core/GetDefaultBrowserUseCaseUnitTest.kt
saperi22: claude came up with subject ;)
i'll change it

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetAppLinksCompatibleBrowserUseCase.kt
saperi22: I don't think I understand. Are you referring to the class name itself or the method name?

saralvasquez: Ah! Sorry yeah that comment location is confusing. I'm talking about the method name

saperi22: `invoke()` is a standard function in Kotlin. It serves a specific purpose. See [this](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/invoke.html). Also [invoke operator](https://kotlinlang.org/docs/operator-overloading.html#invoke-operator)

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetAppLinksCompatibleBrowserUseCase.kt
saperi22: I could remove it.
I wanted to get some comments on it though.
I am using the supported browsers instead of the unsupported because we've only tested for the supported ones. If will try to use app links for the browsers we know will support, and everything else would be deep links. Does that sound good?

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetAppLinksCompatibleBrowserUseCase.kt
saralvasquez: Oh jeeze sorry I'm multitasking and said the entirely wrong thing. I mean the class name not the `invoke` method name 🙃

saperi22: Is the new change in line with what you expect?

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetAppLinksCompatibleBrowserUseCase.kt
```diff
@@ -0,0 +1,30 @@
...
+        getDefaultBrowserUseCase()?.let {
+            return appLinkCompatibleBrowsers.contains(it)
+        }
+        return false
+    }
+
+    companion object {
+        private val appLinkCompatibleBrowsers = listOf(
```
jaxdesmarais, line 22: General question, are these names pretty stable or do they change occasionally? I know apple changes user agents all the time, but not sure if that's a problem on Android. Asking because I am thinking through a situation where a name changes and suddenly a valid browser stops working and we'd need to make an update.

saperi22: AFAIK, that's not a thing in the Android world.

BraintreeCore/src/main/java/com/braintreepayments/api/core/AppLinkCompatibleBrowserUseCase.kt
saperi22: Used `@RestrictTo` annotation instead of internal marker as we're to use this across modules.

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetAppLinksCompatibleBrowserUseCase.kt
tdchow: Yep, I think the best option is to use an allow list instead of a blocklist for the reason you mentioned.

BraintreeCore/src/main/java/com/braintreepayments/api/core/GetAppLinksCompatibleBrowserUseCase.kt
tdchow: It's not possible for apps to change their package names after they're published or installed on the end user's device. However if a new version of a browser is added to the PlayStore as a new app listing, it'd have a different package name. I don't think this would ever happen for mainstream browsers though.

jaxdesmarais: @saperi22 this wasn't blocking just a general question for my own knowledge. 😄

PR #1461, opened 2025-11-04
Unit test conversion cleanup

saralvasquez:
### Summary of changes

 - removes old `MockitoConfiguration` files
 - converts the java `MockLocalPaymentApiBuilder` file to kotlin `MockkLocalPaymentApiBuilder`
 - removes the deprecated `MockPayPalInternalClientBuilder`
 - removes unused `MockApiClientBuilder`
 - removes deprecated`MockBraintreeClientBuilder`
 - removes unused file `MockThreadScheduler`
 - converts java `FixturesHelper` to kotlin
 - converts java `JSONBuilder` to kotlin
 - converts java `TestActivity` to kotlin
 - converts java `TestConfigurationBuilder` to kotlin

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


TestUtils/src/main/java/com/braintreepayments/api/testutils/JSONBuilder.kt
```diff
@@ -0,0 +1,30 @@
+package com.braintreepayments.api.testutils
+
+import org.json.JSONException
+import org.json.JSONObject
+
+abstract class JSONBuilder protected constructor(json: JSONObject) {
```
tdchow: We could remove the empty constructor on line 10 if we set a default value for this constructor.

```suggestion
abstract class JSONBuilder protected constructor(json: JSONObject = JSONObject()) {
```

TestUtils/src/main/java/com/braintreepayments/api/testutils/TestConfigurationBuilder.kt
```diff
@@ -0,0 +1,207 @@
...
+
+    fun paypal(): TestPayPalConfigurationBuilder {
+        return try {
+            TestPayPalConfigurationBuilder(jsonBody.getJSONObject("paypal"))
+        } catch (ignored: JSONException) {
+            TestPayPalConfigurationBuilder(true)
+        }
+    }
+
+    class TestPayPalConfigurationBuilder : JSONBuilder {
```
tdchow, line 103: What are your thoughts on moving these classes into separate files?

saralvasquez: I guess we could but I'm not sure there would be that much benefit. They are only used in a single unit test file each and most are also utilizing other fields from the base builder class. My thinking is that context-wise it's all the same functionality and so breaking into separate files would be breaking up the context behind the builders. It would narrow the scope of the file but clutter up testUtils with files that are basically the same thing. I don't feel super strongly about it though

PR #1468, opened 2025-11-10
Add remaining modules to instrumentation test CI

tdchow:
### Summary of changes

 - Add LocalPayment and DataCollector modules instrumentation tests to CI

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage


noguier (commented):
seems like some modules are failing, sorry I didnt catch that earlier. It is curious that only some of the modules result in an error.

jaxdesmarais (approved):

PR #1469, opened 2025-11-12
import new version Browser Switch and add its integration

noguier:
### Summary of changes

 - Added the new version of Browser-Switch that supports AuthTab to core SDK
 - Added integration of the new AuthTab flow to demo app
 - **IMPORTANT** data binding needed to be disabled to compile with Gradle 8.9.1
 - Added a boolean to force Chrome Custom Tabs if we need to perform app switch, to fix applink support for Authtab

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors
@noguier

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@
     * Add `restorePendingRequest` method to restore browser state after app switch
 * BraintreeCore
     * Update Browser Switch version to 3.4.0
-    * Update Gradle version to 8.9.1
+    * Update Android Gradle Plugin version to 8.9.1
     * Update compileSdkVersion and targetSdkVersion to 36
 
 ## 5.19.0 (2025-11-18)


saralvasquez:
Looks like there's a unit test failure. Does that happen locally for you as well?

Demo/build.gradle
```diff
@@ -47,7 +47,6 @@ android {
     }
 
     buildFeatures {
-        dataBinding = true
```
jaxdesmarais, line 50: Is this a change that merchants may need to make as well to pull in this new feature? If so we should make sure all of the potential changes are documented somewhere to ensure merchants can pick up this new change easily.

gradle/libs.versions.toml
```diff
@@ -1,5 +1,5 @@
 [versions]
-gradle = "8.6.1"
+gradle = "8.9.1"
```
jaxdesmarais, line 2: Same question here on if this is a merchant facing change that they would need to make when updating the SDK version.

build.gradle
```diff
@@ -50,11 +50,11 @@ allprojects {
 version '5.18.1-SNAPSHOT'
 group 'com.braintreepayments'
 ext {
-    compileSdkVersion = 35
+    compileSdkVersion = 36
```
jaxdesmarais, line 53: Same question here on if this is a merchant facing change

noguier:
great questions @jaxdesmarais and @saralvasquez. 
In browser-switch we updated `androidx.browser:browser` to 1.9.0 to take advantage of the AuthTab, this introduced the need to update `com.android.tools.build:gradle` to at least 8.9.1, as well as compile/targetSdk to at least 36. 

When importing the library into this repository we also needed to make the same updates, however, we also needed to disable data binding in _only_  the demo module. This is because 8.8.0 was the first version of the android gradle plugin that requires kotlin 2.x if a project is using the `databinding ktx` library.
So that we do not introduce even more version changes, me and @tdchow decided that it is fair to assume that if the merchant wants to use `databinding ktx` they will have to update their kotlin versions. 

Please let me know if it makes sense, I am happy to add `CHANGELOG` entry as well. However, I do think that @saralvasquez raises a fair question of whether or not it should be a minor version change for our SDK. 
Please let me know what would be the best way to proceed?

saralvasquez:
Ok yeah the data binding changes being isolated to the demo module sounds reasonable to me. I think the only thing that might be a little wonky then is the gradle version for a minor update

saralvasquez:
Ok yeah looking back at the CHANGELOG, gradle version updates didn't require a major version bump. Cool! I think this is ready once a CHANGELOG entry gets added 🥳

saperi22 (approved):

CHANGELOG.md
```diff
@@ -6,14 +6,30 @@
...
+    * Update compileSdkVersion and targetSdkVersion to 36
 
 ## 5.19.0 (2025-11-18)
 
 * Card
   * Fix an issue where a tokenize card errors are masked with a JSON parsing error
 * PayPal
-    * Add `shouldOfferCredit`property to `PayPalCheckoutRequest`. Enables merchants to offer PayPal Credit for Checkout requests.
-  
+  * Add `shouldOfferCreditproperty` to `PayPalCheckoutRequest`. Enables merchants to offer PayPal Credit for Checkout requests.
```
tdchow:
```suggestion
  * Add `shouldOfferCredit` property to `PayPalCheckoutRequest`. Enables merchants to offer PayPal Credit for Checkout requests.
```

I think we were originally missing a space between the property name and "property".

https://github.com/braintree/braintree_android/blob/f5c3f3d7a5d43a0361c16d3e3b47b7359dff2ee3/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalRequest.kt#L107

tdchow (approved):

PR #1471, opened 2025-11-13
Create PayPal button class

saralvasquez:
### Summary of changes

 - Adds PayPal button class
 - sets height, minimum width, and corner radius according to figma
 - Centers logo in the middle of the button according to figma
 - Adds custom attrs to allow user to set color

Screenshot_20251114_121628

Screenshot_20251114_121510

Screenshot_20251114_121558

### Checklist

 - [x] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
+++ b/UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
@@ -9,6 +9,9 @@ import androidx.appcompat.widget.AppCompatButton
 import androidx.core.content.ContextCompat
 import androidx.core.graphics.toColorInt
 
+/**
+    * A customizable PayPal branded button to initiate the PayPal flow
+ */
 class PayPalButton @JvmOverloads constructor(
     context: Context,
     attrs: AttributeSet? = null,


saralvasquez:
I made the button the right blue and added an example to the bottom of the PayPal fragment just to make it easier to test with and view. Let me know if this looks like the right idea and I'll open a real PR once I get a first pass sign off

saralvasquez:
That's a great idea! One sec

UIComponents/src/main/res/drawable/paypal_logo_black.png
tdchow, line 1: Can we use the .svg asset instead of the .png version? We should try to use vector assets for any drawables. This makes it much easier to support multiple screen densities.

https://developer.android.com/develop/ui/views/graphics/vector-drawable-resources

saralvasquez: For sure! I was wondering what the preference was and just went with what we had in the demo

saralvasquez:
I'm also going to go ahead and just add the logic for color switching now since the story for color options later is more about the demo app than the button itself

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
```diff
@@ -0,0 +1,89 @@
...
+import android.graphics.drawable.GradientDrawable
+import android.util.AttributeSet
+import androidx.appcompat.widget.AppCompatButton
+import androidx.core.content.ContextCompat
+import androidx.core.graphics.toColorInt
+import kotlin.text.toInt
+import kotlin.times
+
+@Suppress("MagicNumber")
+class PayPalButton @JvmOverloads constructor(
```
jaxdesmarais, line 15: Should we have docstrings here or will that be a follow up story?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
```diff
@@ -0,0 +1,89 @@
...
+import android.graphics.drawable.Drawable
+import android.graphics.drawable.GradientDrawable
+import android.util.AttributeSet
+import androidx.appcompat.widget.AppCompatButton
+import androidx.core.content.ContextCompat
+import androidx.core.graphics.toColorInt
+import kotlin.text.toInt
+import kotlin.times
+
+@Suppress("MagicNumber")
```
tdchow: Instead of suppressing the warning, what are your thoughts on creating constants for them?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
```diff
@@ -0,0 +1,89 @@
...
+class PayPalButton @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
```
jaxdesmarais, line 17: Wondering if this is the best merchant facing name or if that will be included in a follow up.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
   * Fix an issue where a tokenize card errors are masked with a JSON parsing error
 * PayPal
     * Add `shouldOfferCredit`property to `PayPalCheckoutRequest`. Enables merchants to offer PayPal Credit for Checkout requests.
+* UIComponents
+  * Add `PayPalButton` class to generate a PayPal-branded button for launching PayPal Checkout flow
```
jaxdesmarais: small clarifying point so it's clear this is for either checkout or vault:
```suggestion
  * Add `PayPalButton` class to generate a PayPal-branded button for launching PayPal flow
```

jaxdesmarais (commented):
My comments are non-blocking, just general questions

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
saralvasquez: Actually yeah this is a great reason to start a dimens file to add to values. I'll do that

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
tdchow: Unfortunately, this is the standard naming convention for custom views in Android. However, it's very rare for the actual usage of the custom view to reference the constructor programmatically. This constructor will be internally called when merchants add the custom view to their XML layout files.

jaxdesmarais: Ah - that makes sense, yeah I know almost nothing about Android UI best practices so for sure defer to you all. 😄

tdchow: It's a good question to raise! The old way of doing custom views is not very intuitive 😅

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
saralvasquez: Great call out. I recall a docs story, but I don't think docstrings was included in it. I'll add em

saralvasquez: I wasn't sure how much needed to be added here so let me know if it needs more detail or context. Other class file docstring descriptions seemed pretty minimal until callable methods were added

saperi22 (approved):
Since Jax had comments, maybe wait for his PR approval.

PR #1479, opened 2025-11-24
Add PayPal Button to demo app

saralvasquez:
### Summary of changes

 - Adds new `Payment Buttons` page and home screen button
 - Adds the `PayPalButton` and style change buttons to the `Payment Buttons` page
 - Adds ability to change style of `PayPalButton` via the style change buttons
 - Adds new separate style class for `PayPalButton`
 - Refactor of `PayPalButton` to clean up button color change functionality

[Screen_recording_20251124_171427.webm](https://github.com/user-attachments/assets/aa0581f3-4830-4a0f-85ce-b5f4a46b3ed9)

### Checklist

 - [x] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/PaymentButtonsFragment.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/PaymentButtonsFragment.java
@@ -12,7 +12,6 @@ import com.braintreepayments.api.uicomponents.PayPalButton;
 import com.braintreepayments.api.uicomponents.PayPalButtonColor;
 import com.google.android.material.button.MaterialButtonToggleGroup;
 
-
 public class PaymentButtonsFragment extends BaseFragment {
 
     private PayPalButton payPalButton;


Demo/src/main/java/com/braintreepayments/demo/PaymentButtonsFragment.java
```diff
@@ -0,0 +1,57 @@
...
+import androidx.annotation.Nullable;
+
+import com.braintreepayments.api.uicomponents.PayPalButton;
+import com.google.android.material.button.MaterialButtonToggleGroup;
+
+
+public class PaymentButtonsFragment extends BaseFragment {
+
+    private PayPalButton payPalButton;
+    private MaterialButtonToggleGroup toggleGroup;
```
jaxdesmarais, line 18: 😍  this looks similar to segmented controls on iOS so I think we will be well aligned! Plus now I know what this is actually called for next time.

Demo/src/main/res/layout/fragment_payment_buttons.xml
```diff
@@ -0,0 +1,87 @@
...
+            android:orientation="vertical">
+
+            <LinearLayout
+                android:layout_width="match_parent"
+                android:layout_height="match_parent"
+                android:gravity="center"
+                android:padding="15dp"
+                android:orientation="horizontal">
+
+                <com.google.android.material.button.MaterialButtonToggleGroup
```
jaxdesmarais, line 31: Are we able to have the width of this match the button width? Not sure if that's possible.

Demo/src/main/res/layout/fragment_payment_buttons.xml
```diff
@@ -0,0 +1,87 @@
...
+        android:layout_weight="1">
+
+        <LinearLayout
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_gravity="center_horizontal"
+            android:orientation="vertical">
+
+
+
```
jaxdesmarais: Not sure if we have extra spaces here or if the 3 are intentional.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
```diff
@@ -77,13 +59,24 @@ class PayPalButton @JvmOverloads constructor(
...
+        d.setBounds(left, top, left + w, top + h)
+        d.draw(canvas)
+    }
+
+    /**
+     * Sets the color style of the PayPal button
+     *
+     * @property color A string representing the button color. Valid values are "white", "black", and "blue"
+     */
+    fun setButtonColor(color: String) {
```
jaxdesmarais: Do we want to make the color an enum vs the merchant passing us a string? On iOS I think we used `primary`, `white`, `black` as the values and default to `primary`.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButtonStyle.kt
```diff
@@ -0,0 +1,22 @@
+package com.braintreepayments.api.uicomponents
+
+import androidx.annotation.ColorInt
+import androidx.core.graphics.toColorInt
+
+enum class PayPalButtonStyle(
```
jaxdesmarais: If this is public we should have a docstring. We should also align iOS and Android on what we call this.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 * UIComponents
   * Add `PayPalButton` class to generate a PayPal-branded button for launching PayPal flow
   * Add `VenmoButton` class to generate a Venmo-branded button for launching Venmo flow
+  * Add Payment Buttons page to the demo app
```
jaxdesmarais: We can remove this since it's not merchant facing.

Demo/src/main/java/com/braintreepayments/demo/PaymentButtonsFragment.java
saralvasquez: Yeah! Sai also found something called `Chips` that's very common in Android that looks similar as well

Demo/src/main/res/layout/fragment_payment_buttons.xml
saralvasquez: I can try to finagle it a bit, but I'm not very knowledgeable about UI so it might be more effort than it's worth

Demo/src/main/res/layout/fragment_payment_buttons.xml
saralvasquez: Oh this was where the venmo button section was gonna go so I just left a bit of space there to denote that, but I can remove it

CHANGELOG.md
saralvasquez: Ah yeah good point

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
saralvasquez: Ah yeah I added a `PayPalButtonStyle` class that would make so much more sense to use here rather than a string. I think I'd prefer `blue` to `primary` since it's more descriptive, but I'll defer to the iOS pattern for consistency

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButtonStyle.kt
saralvasquez: I see a `VenmoButtonColor` file in iOS so I'll rename this to `PayPalButtonColor`

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
jaxdesmarais: I think on iOS we call the enum `VenmoButtonColor` or `PayPalButtonColor` vs using Style. I don't have a strong preference on blue vs primary. @buzzamus unless you feel strongly about `primary` can we update Venmo to use `blue`. @stechiu for when you're back to do the same in the PayPal PR.

buzzamus: I am fine either way, so I can update our references to blue instead of primary

saralvasquez: Op! I didn't see this until now and updated it to primary instead of blue 🙃

saralvasquez: I can change it back to blue real quick!

Demo/src/main/res/layout/fragment_payment_buttons.xml
saralvasquez: Got it figured out without too much effort

Demo/src/main/java/com/braintreepayments/demo/PaymentButtonsFragment.java
```diff
@@ -0,0 +1,58 @@
...
+import android.view.ViewGroup;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.braintreepayments.api.uicomponents.PayPalButton;
+import com.braintreepayments.api.uicomponents.PayPalButtonColor;
+import com.google.android.material.button.MaterialButtonToggleGroup;
+
+
```
jaxdesmarais: Small nit on double space here:

```suggestion
```

PR #1484, opened 2025-12-01
Venmo Button - Invoke Venmo flow

noguier:
### Summary of changes

 - invoke Venmo flow by the press of the Venmo Button
 - Add a way to display nonce in the `PaymentButtonsFragment`

### Checklist

 - [ ] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/PaymentButtonsFragment.java
+++ b/Demo/src/main/java/com/braintreepayments/demo/PaymentButtonsFragment.java
@@ -1,25 +1,47 @@
 package com.braintreepayments.demo;
 
+import android.net.Uri;
 import android.os.Bundle;
+import android.text.TextUtils;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.fragment.app.FragmentActivity;
+import androidx.navigation.NavDirections;
+import androidx.navigation.fragment.NavHostFragment;
 
+import com.braintreepayments.api.core.UserCanceledException;
 import com.braintreepayments.api.uicomponents.PayPalButton;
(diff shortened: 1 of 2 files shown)


Demo/src/main/java/com/braintreepayments/demo/PaymentButtonsFragment.java
```diff
@@ -51,7 +73,7 @@ public void onButtonChecked(MaterialButtonToggleGroup group, int checkedId, bool
 
         venmoButton = view.findViewById(R.id.venmo_payment_button);
         venmoToggleGroup = view.findViewById(R.id.venmo_button_toggle_group);
-
+        venmoButton.setOnClickListener(this::launchVenmo);
```
tdchow, line 76: Is there a way we can have the click listener set internally in the payment buttons (both PayPalButton and VenmoButtons)? This way a merchant doesn't need to worry about connecting the click listener manually.

Demo/src/main/java/com/braintreepayments/demo/PaymentButtonsFragment.java
```diff
@@ -1,26 +1,48 @@
...
 
+import java.util.ArrayList;
+
 public class PaymentButtonsFragment extends BaseFragment {
 
     private PayPalButton payPalButton;
     private MaterialButtonToggleGroup payPalToggleGroup;
     private VenmoButton venmoButton;
+    private VenmoClient venmoClient;
+    private VenmoLauncher venmoLauncher;
```
tdchow, line 42: Are we able to internalize the VenmoClient and VenmoLauncher classes within the VenmoButton? This would reduce the number of integration steps a merchant would have to do to use the payment buttons.

noguier: Yes, that would be great, let me try to make it work!

PR #1485, opened 2025-12-03
Complete PayPal flow that is initiated by the click of the branded button

noguier:
### Summary of changes

 - Abstracted initializing PayPalClient into the button class
 - Abstracted calling PayPalLauncher.launch() into the button class
 - All of the flow, except processing the final PayPalResult and storing the pending request is abstracted from the merchant

### Checklist

 - [x] Added a changelog entry
 - [ ] Relevant test coverage

### DIFF
--- a/UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
+++ b/UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
@@ -45,9 +45,16 @@ class PayPalButton @JvmOverloads constructor(
      * The PayPal client used to create payment auth requests and tokenize results.
      * Must be initialized via [initialize] before use.
      */
-    lateinit var payPalClient: PayPalClient
+    private lateinit var payPalClient: PayPalClient
     private val payPalLauncher: PayPalLauncher
 
+    /**
+     * Callback invoked when the PayPal payment authentication request is launched.
+     *
+     * This callback is used to receive notifications about the launch status, such as
+     * [PayPalPendingRequest.Started] or [PayPalPendingRequest.Failure]. The pending request
+     * returned on success should be stored and passed to [handleReturnToApp] to complete the payment flow.
+     */
     var payPalLaunchCallback: PayPalLaunchCallback? = null
 
     init {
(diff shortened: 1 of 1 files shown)


saperi22 (approved):

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
```diff
@@ -26,6 +39,17 @@ class PayPalButton @JvmOverloads constructor(
     private val desiredHeight = resources.getDimension(R.dimen.pay_button_height).toInt()
     private val minDesiredWidth = resources.getDimension(R.dimen.pay_button_min_width).toInt()
 
+    private var payPalRequest: PayPalRequest? = null
+
+    /**
+     * The PayPal client used to create payment auth requests and tokenize results.
+     * Must be initialized via [initialize] before use.
+     */
+    lateinit var payPalClient: PayPalClient
```
tdchow: I think we can make the `payPalClient` private to the PayPalButton class. The merchant shouldn't need to set/inspect the variable.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
```diff
@@ -26,6 +39,17 @@ class PayPalButton @JvmOverloads constructor(
...
+    private val payPalLauncher: PayPalLauncher
+
+    var payPalLaunchCallback: PayPalLaunchCallback? = null
```
tdchow, line 58: Can we add a kdoc for this variable? It would be good to give merchants some context on what this callback is for.

There might be a case where the merchant doesn't set this callback but the button is clicked by the buyer. What are your thoughts on adding a check to see if the `payPalLaunchCallback` is null before executing anything in the onClickListener? I feel that failing faster is better to be able to notify the merchant that they have not integrated the button correctly.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
```diff
@@ -36,6 +60,128 @@ class PayPalButton @JvmOverloads constructor(
...
+    fun setPayPalRequest(payPalRequest: PayPalRequest) {
+        setOnClickListener {
+            payPalClient.createPaymentAuthRequest(
+                context = context,
+                payPalRequest = payPalRequest
+            ) { paymentAuthRequest: PayPalPaymentAuthRequest ->
+                when (paymentAuthRequest) {
+                    is PayPalPaymentAuthRequest.ReadyToLaunch -> {
+                        getActivity()?.let { activity ->
+                            val payPalPendingRequest = payPalLauncher.launch(
```
tdchow: The `setPayPalRequest()` function is looking a bit long and is doing quite a bit. What are your thoughts on creating a new private function for moving out all of the `payPalLauncher.launch()` code?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
noguier: yes, great idea, I will add that !

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
noguier: yep, i think it makes total sense, I will address it right now

noguier: addressed in 79fa6d7

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
noguier: addressed in 79fa6d7

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
noguier: addressed in 79fa6d7

PR #1521, opened 2026-01-22
Update Magnes to 5.6.0 and `getClientMetadataId` to receive a callback

noguier:
### Summary of changes

 - added the ability to receive a callback to `getClientMetadataId` when the submit completes
 - updated PayPalInternalClient and LocalPaymentClient call to `getClientMetadataId`
 - updated test coverage 

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage


DataCollector/src/main/java/com/braintreepayments/api/datacollector/MagnesInternalClient.kt
```diff
@@ -44,20 +46,22 @@ class MagnesInternalClient(
...
-            val result = magnesSDK.collectAndSubmit(
+            magnesSDK.collectAndSubmit(
                 context.applicationContext,
                 request.clientMetadataId,
                 request.additionalData
-            )
-            return result.paypalClientMetaDataId
+            ) { status, clientMetadataId ->
+                // Callback is invoked when device data collection and submit API completes
+                callback(clientMetadataId)
```
jaxdesmarais: Should we be checking that the status is a success before returning the callback here? For the timeout and error status we can instead return an empty string or an error.

CHANGELOG.md
```diff
@@ -2,9 +2,13 @@
 
 ## unreleased
 
+
+* DataCollector
+    * Bump Magnes version to `5.6.0`
 * PayPal
     * Pass `funding_source` to the app switch url link and to analytics events.
 
+
```
jaxdesmarais: Looks like we've got a double space here now

DataCollector/src/test/java/com/braintreepayments/api/datacollector/MagnesInternalClientUnitTest.kt
```diff
@@ -74,20 +79,25 @@ class MagnesInternalClientUnitTest {
...
+        sut.getClientMetadataId(
             null, sandboxConfiguration,
             dataCollectorInternalRequest
-        )
+        ) { clientMetadataId ->
+            receivedClientMetadataId = clientMetadataId
+        }
 
-        Assert.assertTrue(result.isEmpty())
+        Assert.assertNotNull(receivedClientMetadataId)
```
saralvasquez: Do we need the extra NotNull assert here since you've got the `!!` below?

DataCollector/src/test/java/com/braintreepayments/api/datacollector/MagnesInternalClientUnitTest.kt
```diff
@@ -177,41 +187,56 @@ class MagnesInternalClientUnitTest {
...
+        sut.getClientMetadataId(
             context, prodConfiguration,
             dataCollectorInternalRequest
-        )
+        ) { clientMetadataId ->
+            receivedClientMetadataId = clientMetadataId
+        }
 
-        Assert.assertTrue(result.isEmpty())
+        Assert.assertNotNull(receivedClientMetadataId)
```
saralvasquez: Same question as above

saralvasquez (approved):
Couple small nits, but looks good otherwise!

DataCollector/src/test/java/com/braintreepayments/api/datacollector/MagnesInternalClientUnitTest.kt
```diff
@@ -177,41 +187,56 @@ class MagnesInternalClientUnitTest {
         requestWithInvalidGUID.applicationGuid = "invalid guid"
 
         val sut = MagnesInternalClient(magnesSDK)
-        val result = sut.getClientMetadataId(context, prodConfiguration, requestWithInvalidGUID)
+        var receivedClientMetadataId: String? = null
```
saperi22:
```suggestion
        var receivedClientMetadataId: String = "non-empty"
```
would this work for your use-case? you could potentially remove the `!!` in the assert.

DataCollector/src/main/java/com/braintreepayments/api/datacollector/MagnesInternalClient.kt
noguier: I added the check for the status in b52b032, let me know if it is sufficient!

saperi22 (approved):

DataCollector/src/test/java/com/braintreepayments/api/datacollector/MagnesInternalClientUnitTest.kt
noguier: thank you for this suggestion, I have updated the test not to use `!!` in 123ebec!

DataCollector/src/main/java/com/braintreepayments/api/datacollector/DataCollector.kt
```diff
@@ -47,12 +47,13 @@ class DataCollector @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP) constructor(
     fun getClientMetadataId(
         context: Context,
         configuration: Configuration?,
-        hasUserLocationConsent: Boolean
-    ): String {
+        hasUserLocationConsent: Boolean,
+        callback: (String) -> Unit
```
jaxdesmarais: Is either this or the below changing the public interface for merchants that use the data collector directly? I am not sure how the `RestrictTo` logic works.

noguier: As far as I understand `@RestrictTo` means this function is internal to our library group - merchants shouldn't be calling it directly, the only function that merchants should call is  `collectDeviceData`.

DataCollector/src/main/java/com/braintreepayments/api/datacollector/MagnesInternalClient.kt
```diff
@@ -44,20 +47,26 @@ class MagnesInternalClient(
...
                 request.clientMetadataId,
                 request.additionalData
-            )
-            return result.paypalClientMetaDataId
+            ) { status, clientMetadataId ->
+                // Callback is invoked when device data collection and submit API completes
+                if (status == MagnesSubmitStatus.SUCCESS) {
+                    callback(clientMetadataId)
+                } else {
+                    callback("")
```
jaxdesmarais: On iOS it looks like we return an error if we get a timeout or failed result. Not sure if we have the ability to do the same on Android. If not, we may want to leave ourselves a next major version note so we can add it in a future major version. We typically add those like `// NEXT_MAJOR_VERSION: return an error if we do not get a success status`

noguier: I made some changes to handle timeout and error in a77b191, please let me know if it is what you were looking for.

DataCollector/src/main/java/com/braintreepayments/api/datacollector/MagnesInternalClient.kt
```diff
@@ -20,10 +21,12 @@ class MagnesInternalClient(
...
         context: Context?,
         configuration: Configuration?,
-        request: DataCollectorInternalRequest?
-    ): String {
+        request: DataCollectorInternalRequest?,
+        callback: (String?, Exception?) -> Unit
+    ) {
         if (context == null || configuration == null || request == null) {
-            return ""
+            callback("", null)
```
jaxdesmarais: Should we consider an exception if all of these are null?

DataCollector/src/main/java/com/braintreepayments/api/datacollector/MagnesInternalClient.kt
```diff
@@ -44,20 +47,27 @@ class MagnesInternalClient(
...
+            magnesSDK.collectAndSubmit(
                 context.applicationContext,
                 request.clientMetadataId,
                 request.additionalData
-            )
-            return result.paypalClientMetaDataId
+            ) { status, clientMetadataId ->
+                // Callback is invoked when device data collection and submit API completes
+                when (status) {
+                    MagnesSubmitStatus.SUCCESS -> callback(clientMetadataId, null)
```
jaxdesmarais: iirc this callback returns a debug ID instead CMID. I think on success we want `result.paypalClientMetaDataId`

DataCollector/src/main/java/com/braintreepayments/api/datacollector/MagnesInternalClient.kt
```diff
@@ -44,20 +47,27 @@ class MagnesInternalClient(
...
+            }
         } catch (e: InvalidInputException) {
             // Either clientMetadataId or appGuid exceeds their character limit
             Log.e(
                 "Exception",
                 "Error fetching client metadata ID. Contact Braintree Support for assistance.",
                 e
             )
-            return ""
+            callback("", null)
```
jaxdesmarais: We could consider returning the above exception here?

DataCollector/src/main/java/com/braintreepayments/api/datacollector/MagnesInternalClient.kt
noguier: I was just following the pattern that was already established before, I can throw an `IllegalArgumentException` here as well?

jaxdesmarais: I think it makes sense to bubble up that either context, configuration, or request was null is helpful for merchants and for us to debug.

DataCollector/src/main/java/com/braintreepayments/api/datacollector/MagnesInternalClient.kt
noguier: thank you very much for catching this, you are correct

DataCollector/src/main/java/com/braintreepayments/api/datacollector/MagnesInternalClient.kt
noguier: great point, I fixed it in 00cb8ee

DataCollector/src/main/java/com/braintreepayments/api/datacollector/MagnesInternalClient.kt
noguier: should be updated in 00cb8ee

DataCollector/src/main/java/com/braintreepayments/api/datacollector/MagnesInternalClient.kt
noguier: got it, should be updated in 00cb8ee

jaxdesmarais (approved):
🚀

PR #1564, opened 2026-04-07
LI- Fix crashes caused by uncaught exception

saralvasquez:
### Summary of changes

 Fix uncaught exceptions in analytics and config prefetch coroutines   
AnalyticsApi.execute() launched a fire-and-forget coroutine with no
   exception handling, causing UnknownHostException (and any other network
   error) to propagate to the thread's uncaught exception handler and
   terminate the host app. Wrap the httpClient.post() call in try-catch
   since analytics delivery is best-effort.

   BraintreeClient.prefetchConfiguration() only caught IOException, but
   ConfigurationLoader can also return a JSONException (malformed config
   response) which is not an IOException subtype and would similarly crash
   the app. Broaden the catch to Exception.

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage

### Authors
@saralvasquez

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
@@ -7,6 +7,7 @@ import kotlinx.coroutines.launch
 import org.json.JSONArray
 import org.json.JSONException
 import org.json.JSONObject
+import kotlin.coroutines.cancellation.CancellationException
 
 /**
  * This API call sends analytic events to FPTI.
@@ -43,7 +44,7 @@ internal class AnalyticsApi(
                     authorization = merchantRepository.authorization,
                 )
             } catch (e: Exception) {
-                // no op - failure to send analytics should not impact the user experience
+                if (e is CancellationException) throw e
             }
         }
     }
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
@@ -11,6 +11,7 @@ import org.json.JSONObject
 /**
  * This API call sends analytic events to FPTI.
  */
+@Suppress("SwallowedException", "TooGenericExceptionCaught")
 internal class AnalyticsApi(
     private val httpClient: BraintreeHttpClient = BraintreeHttpClient(),
     private val deviceInspector: DeviceInspector = DeviceInspectorProvider().deviceInspector,
@@ -34,12 +35,16 @@ internal class AnalyticsApi(
         val analyticsRequest =
             createFPTIPayload(merchantRepository.authorization, jsonEvents, metadata)
         coroutineScope.launch {
-            httpClient.post(
-                path = FPTI_ANALYTICS_URL,
-                data = analyticsRequest.toString(),
-                configuration = null,
-                authorization = merchantRepository.authorization,
-            )
+            try {
+                httpClient.post(
+                    path = FPTI_ANALYTICS_URL,
+                    data = analyticsRequest.toString(),
+                    configuration = null,
+                    authorization = merchantRepository.authorization,
+                )
+            } catch (e: Exception) {
+                // no op - failure to send analytics should not impact the user experience
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
```diff
@@ -34,12 +35,16 @@ internal class AnalyticsApi(
...
-                authorization = merchantRepository.authorization,
-            )
+            try {
+                httpClient.post(
+                    path = FPTI_ANALYTICS_URL,
+                    data = analyticsRequest.toString(),
+                    configuration = null,
+                    authorization = merchantRepository.authorization,
+                )
+            } catch (e: Exception) {
```
noguier, line 46: Just so I am clear, why do we need to surround the post() call here with try...catch? I thought we are catching any exceptions on the caller site?

noguier (approved):
I just one curiosity question, otherwise looks great, thank you for fixing it!

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
saralvasquez: Yeah good question. In the caller sites we often only catch JSONExceptions. In the issue that was opened a different type of exception was returned which wasn't recognized so wasn't caught. This ultimately was what caused the crash. The try catch here just ensures all exceptions when trying to make analytics calls will be caught and not cause a crash no matter what is being caught when it's called further up-stream

saperi22 (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
jaxdesmarais, line 46: Since this is in a `coroutineScope.launch` I am finding some info that indicates we may want to handle `CancellationException` to prevent potential memory leaks. I think it would be something like this here and below in `BraintreeClient`:
```kotlin
} catch (e: CancellationException) {
    throw e  // let coroutine machinery handle this error if thrown
} catch (e: Exception) {
    // no op - failure to send analytics should not impact the user experience
}
```

Def want to def to y'all on the Android side for this though if my understanding isn't quite right and we don't need to handle this in both places.

noguier: Yeah, I think it is generally a good practice to always catch a `CancelationException` with coroutine flows.

noguier: Can we just generalize the JSONExceptions to Exceptions at call sites?

BraintreeCore/src/main/java/com/braintreepayments/api/core/BraintreeClient.kt
```diff
@@ -93,8 +93,8 @@ class BraintreeClient internal constructor(
         // It ensures that the configuration is loaded and ready for use in subsequent requests.
         coroutineScope.launch {
             try {
-                val config = getConfiguration()
-            } catch (e: IOException) {
+                getConfiguration()
+            } catch (e: Exception) {
                 // no op
```
noguier: I think it is useful to add `CancellationException` here as well

BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt
saralvasquez: Ah after some learning it does sound quite important I also don't think we've been adhering to it very strictly along the conversion so we might want to make a ticket to go in and add any missing cases

saralvasquez: Again a great question and I don't think I answered well the first time. There are some important pieces to clarify. Firstly, there isn't always a try/catch around this call. Also the places where there is atry/catch is to catch exceptions of operations that are initiated inside that function. In the case of the analytics `execute` function, the operation that might throw an exception (`httpClient.post`) is several steps away from where the call is initiated.

What this means is that were are hundreds of places where we would need to catch all exceptions to make sure none were uncaught which is a huge lift and not really best practice. The simpler and more concrete way to ensure that all exceptions are caught from the `httpClient.post` related to any analytics call is to do it where the post is called. This also prevents us from having to remember to add a try/catch in any place we add new analytics calls.

tldr- it's best practice to catch exceptions as close to where they could be thrown as possible

PR #1570, opened 2026-04-09
Create BaseTextInputView for Card Fields

noguier:
### Summary of changes

 - Add `BaseTextInputView` — a custom `LinearLayout` that serves as the base view for card number, CVV, and expiration input fields in the CardFields feature
 - Custom outlined border with gray/blue-focus/red-error states and animated floating hint label
 - Accessibility support via `AccessibilityDelegateCompat` for error announcements and `contentDescription` on the EditText
 - Error display with icon, border color change, and label visibility toggling

https://github.com/user-attachments/assets/7ac36f01-95ac-4282-a25e-d0313ff33ab5

### Checklist

 - [x] Added a changelog entry
 - [x] Relevant test coverage


UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/BaseTextInputView.kt
```diff
@@ -0,0 +1,205 @@
...
+import android.widget.TextView
+import androidx.annotation.RestrictTo
+import androidx.core.content.ContextCompat
+import androidx.core.view.AccessibilityDelegateCompat
+import androidx.core.view.ViewCompat
+import androidx.core.view.accessibility.AccessibilityNodeInfoCompat
+import androidx.core.widget.doAfterTextChanged
+import com.braintreepayments.api.uicomponents.R
+
+// TODO change class to internal before releasing
```
saralvasquez, line 26: Just out of curiosity why can't it be internal now?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/BaseTextInputView.kt
```diff
@@ -0,0 +1,205 @@
...
+        updateHintPosition(animate = false)
+    }
+
+    override fun onDetachedFromWindow() {
+        super.onDetachedFromWindow()
+        hintAnimator?.cancel()
+        hintAnimator = null
+    }
+
+    private fun updateBorderState() {
```
saralvasquez, line 116: Just since it wasn't in the video, do we have confirmation that the boarder color changes when an error goes away?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/BaseTextInputView.kt
```diff
@@ -0,0 +1,205 @@
...
+
+        when {
+            hasError -> borderDrawable.setStroke(borderFocusedWidth, errorColor)
+            hasFocus -> borderDrawable.setStroke(borderFocusedWidth, focusedBorderColor)
+            else -> borderDrawable.setStroke(borderWidth, defaultBorderColor)
+        }
+        inputContainer.invalidate()
+    }
+
+    private fun updateHintPosition(animate: Boolean) {
```
saralvasquez, line 128: Similar question. Do we have confirmation that the hint goes back when text is completely removed?

saralvasquez (commented):
This looks great! Just a couple questions. Also how did you get that video? Would it make sense to add changes to the demo app so it's easier to test the changes as they come in?

noguier:
> This looks great! Just a couple questions. Also how did you get that video? Would it make sense to add changes to the demo app so it's easier to test the changes as they come in?

Hey, yes, so I had to essentially add a very simplified implementation to the demo app locally, I can add it to this PR, I just didn't want to make this PR too big.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/BaseTextInputView.kt
noguier: since i didn't add error validation , right now this error is just hard-coded , I can see if i can create a very simplified error handling to see the changes

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/BaseTextInputView.kt
noguier: Because I use it in demo app to test my changes, once we will have the child views and Card fields container with all three views,  we will not be calling this class directly, hence we can make it internal!

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/BaseTextInputView.kt
saralvasquez: ah ok! that can come later then. cool

noguier:
I did add the implementation to the demo app actually @saralvasquez!

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/BaseTextInputView.kt
noguier: https://github.com/user-attachments/assets/3728c4a5-79f6-4705-bde6-baac96aae9ff

Great question, here is a quick little video to confirm the behavior!

saralvasquez: Awesome! Thanks for doing that

saralvasquez (approved):
🎉

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/BaseTextInputView.kt
```diff
@@ -0,0 +1,205 @@
...
+                duration = ANIMATION_DURATION
+                start()
+            }
+        } else {
+            hintLabel.translationY = targetTranslationY
+            hintLabel.setTextSize(TypedValue.COMPLEX_UNIT_PX, targetTextSize)
+        }
+    }
+
+    fun setHint(hint: CharSequence?) {
```
saperi22, line 165: Consider adding a default hint in the init.
That way, there's something out of the box and can be updated if the merchant chooses to.

saperi22 (approved):

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/BaseTextInputView.kt
noguier: I think it will make more sense to actually set the default hint on the child views, since the baseTextInoutView must be general, and i can't think of a meaningful default hint at this level, but at child views level I will add it. Let me know what do you think?

PR #1571, opened 2026-04-14
Add `CardNumberTextInputView` for CardFields feature

noguier:
### Summary of changes

- Extend `BaseTextInputView` to support subclasses
- Add `setCardBrandIcon()` method to `BaseTextInputView` for displaying icons in the textInput objects
- Create `CardNumberTextInputView` with leading card brand icon and `CardNumberTextInputViewUnitTest` to test new View
- Add card brand vector drawables (unknown, Visa) and accessibility strings for all supported card brands
- Update Demo app to use `CardNumberTextInputView`

cardUnknown
cardVisa

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage


saperi22 (approved):

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/BaseTextInputView.kt
```diff
@@ -199,6 +216,32 @@ class BaseTextInputView @JvmOverloads constructor(
         updateBorderState()
     }
 
+    internal fun setLeadingDrawable(@DrawableRes iconRes: Int, contentDescription: String) {
```
saralvasquez: Is there a reason this is named `setLeadingDrawable`? Is it a standard naming convention? I only ask because it doesn't feel particularly descriptive

saralvasquez (approved):
Looks good! Just some basic questions. Also is there a reason you removed the other fields from the demo app?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/BaseTextInputView.kt
noguier: Great question, I was trying to be descriptive, but still general enough because it is a Base class. Additionally, I think later I will have to add setTrailingDrawable because some input fields designed to have an icon at the end(right-side) of the text input field. 
Looking at the figma design though it looks like we will only be using the leading Drawable for the card text input field. Do you think we should rename it to `setCardBrandIcon()`?

noguier:
> Also is there a reason you removed the other fields from the demo app?

Before the inputs in the Demo were using BaseTextInput directly, since in the final product we will not be using the Base class directly, I thought it would be smart to remove them now.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/BaseTextInputView.kt
saralvasquez: Yeah I'd say make it as descriptive and specific as possible now. And we can rename it to something broader later since it's an internal function

noguier: Sounds good, should be addressed now @ bb7a434

saralvasquez: great thank you!

PR #1574, opened 2026-04-16
Add missing cancellation exceptions

saralvasquez:
### Summary of changes

 - Added cancellation exception "catch and re-throw" in all areas where it was missing
 - Added unit tests for all these cases
 - Did a bit of unit test package cleanup from junit -> kotlin

This is a follow-up to the coroutines conversion that has been ongoing. Since this is still internal a changelog entry is not required

### Checklist

 - [ ] Added a changelog entry
 - [x] Relevant test coverage

### Authors
@saralvasquez


ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/internal/GenerateCustomerRecommendationsApi.kt
```diff
@@ -54,6 +55,7 @@ internal class GenerateCustomerRecommendationsApi(
                 val recommendationsResult = parseRecommendationsResponse(responseBody)
                 GenerateCustomerRecommendationsResult.Success(recommendationsResult)
             } catch (e: Exception) {
+                if (e is CancellationException) throw e
                 GenerateCustomerRecommendationsResult.Error(e)
             }
         } catch (e: JSONException) {
```
noguier, line 61: Just to avoid the same LI you have already fixed, are we 100% sure only JSONException can be thrown here?

noguier: same question for CreateCustomerSessionApi.kt and UpdateCustomerSessionApi.kt

noguier (approved):
I have just a "sanity check" question, otherwise it looks very good, thank you very much for adding tests for all of them.

ShopperInsights/src/main/java/com/braintreepayments/api/shopperinsights/v2/internal/GenerateCustomerRecommendationsApi.kt
saralvasquez: good question! JSONException is the only thing that can be thrown there since it's just params.put that can throw anything and the only exception is JSONException

saperi22 (approved):

PR #1577, opened 2026-04-21
Add Pre-Commit Hook

agedd:

> Before submitting this PR, note we cannot accept language translation PRs. We support the same languages that are supported by PayPal, and have a dedicated localization team to provide the translations. If there is an error in a specific translation, you may open an issue and we will escalate it to the localization team.

### Summary of changes
- Add pre-commit hooks to standardize commit format 
- `.pre-commit-config.yaml` declares the commit-msg hook
- `scripts/commit-msg` is the validation script

**note**: once the PR is merged, every contributor must run the following once from the repo root after cloning:

```
mkdir -p .git/hooks
cp scripts/commit-msg .git/hooks/commit-msg
chmod +x .git/hooks/commit-msg
```
#### install (via `pre-commit` framework)
If you have `pre-commit` installed (or install it via `pip3 install pre-commit`):
```
pre-commit install --hook-type commit-msg
```

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- ~[ ] Tested and confirmed payment flows affected by this change are functioning as expected~

- @agedd

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -72,7 +72,7 @@ The individual components may be of interest for advanced integrations and are e
 
 ## Git Hooks
  
-We use a `commit-msg` hook to enforce consistent commit message formatting.
+We use a `commit-msg` hook to enforce consistent commit message formatting. Follow the guidelines below to prevent issues with improperly formatted messages.
  
 ### Install (Manual — recommended)
  
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -70,6 +70,51 @@ The individual components may be of interest for advanced integrations and are e
 * Commits should be small but atomic. Tests should always be passing; the product should always function appropriately.
 * Commit messages should be concise and descriptive.
 
+## Git Hooks
+ 
+We use a `commit-msg` hook to enforce consistent commit message formatting.
+ 
+### Install (Manual — recommended)
+ 
+Run the following once from the repo root after cloning:
+ 
+```
+mkdir -p .git/hooks
+cp scripts/commit-msg .git/hooks/commit-msg
+chmod +x .git/hooks/commit-msg
+```
+ 
+### Install (via `pre-commit` framework)
+ 
+If you have `pre-commit` installed (or install it via `pip3 install pre-commit`):
+ 
+```
+pre-commit install --hook-type commit-msg
+```
+ 
+### Commit format
+ 
+Commits must follow `<type>: <description>` (or `<type>(scope): <description>`), where `<description>` starts. Allowed types:
+ 
+| Type | When to use |
+|------|-------------|
+| `fix` | Patches a bug |
+| `feature` | Introduces a new feature |
+| `docs` | Documentation changes only |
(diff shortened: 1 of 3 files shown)


scripts/commit-msg
```diff
@@ -0,0 +1,46 @@
...
+  echo ""
+  echo "   Your message: \"$COMMIT_MSG\""
+  echo ""
+  exit 1
+fi
+
+# Warn about vague descriptions
+VAGUE_TERMS="cleanup|fix stuff|update|changes|misc|wip|temp"
+if echo "$COMMIT_MSG" | grep -qiE ": ($VAGUE_TERMS)$"; then
+  echo "⚠️  Warning: commit description looks vague. Consider being more specific."
```
saralvasquez: Can we add something in the warning that explains what terms are considered vague?

DEVELOPMENT.md
```diff
@@ -70,6 +70,51 @@ The individual components may be of interest for advanced integrations and are e
 * Commits should be small but atomic. Tests should always be passing; the product should always function appropriately.
 * Commit messages should be concise and descriptive.
 
+## Git Hooks
+ 
+We use a `commit-msg` hook to enforce consistent commit message formatting.
```
saralvasquez: I feel like we should call out that devs need to install something in this description or they might just read this and move on without doing anything below

saralvasquez (commented):
Looks good! Just a couple small suggestions

saperi22 (approved):

saperi22:
@agedd git hooks have an update from Git.
Take a look at [this](https://github.blog/open-source/git/highlights-from-git-2-54/) update from yesterday.

Instead of a third party tool, we could use new config that works across all repos. Might be a good config to have for each developer locally. Might also be a much simpler thing to do for setup. Requires Git 2.54, which might be the latest.

saperi22 (approved):

DEVELOPMENT.md
agedd: yep i mention it on line 89 [here](https://github.com/braintree/braintree_android/blob/23c4196e3c3c12f538de5a698926af34baf6236b/DEVELOPMENT.md#install-via-pre-commit-framework)

saralvasquez: I more mean that since this is the first section that folks are going to read we should clarify that they need to continue reading or they will miss something. Something like "We use a `commit-msg` hook to enforce consistent commit message formatting. Follow the directions below to avoid issues with improperly formatted commit messages."

I know that sometimes I gloss over stuff if it's not explicit on telling me that something won't work if I don't do the following commands.

agedd: ahh gotcha! - i've updated [here](2e00492282b6e1923ac251c4677ad07880832781)

saralvasquez (approved):
Looks great! Thanks

PR #1580, opened 2026-04-22
test: add more instrumentation tests to BTCore

noguier:
### Summary of changes
- Added more instrumentation tests to BraintreeCore
- Achieved 76% of test coverage

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Test generation

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @noguier


BraintreeCore/src/androidTest/java/com/braintreepayments/api/core/usecase/GetReturnLinkTypeUseCaseTest.kt
```diff
@@ -0,0 +1,52 @@
...
+        val result = sut()
+        assertEquals(ReturnLinkTypeResult.DEEP_LINK, result)
+    }
+
+    @Test(timeout = 1000)
+    fun invoke_returnsDeepLink_whenAppLinkReturnUriIsNull() {
+        merchantRepository.appLinkReturnUri = null
+        val result = sut()
+        assertEquals(ReturnLinkTypeResult.DEEP_LINK, result)
+    }
```
saperi22, line 44: did we want to test app_link somewhere?

BraintreeCore/src/androidTest/java/com/braintreepayments/api/core/usecase/GetReturnLinkUseCaseTest.kt
```diff
@@ -0,0 +1,62 @@
...
+        merchantRepository.deepLinkFallbackUrlScheme = null
+        val sut = GetReturnLinkUseCase(merchantRepository)
+
+        val result = sut()
+
+        assertTrue(result is ReturnLinkResult.Failure)
+    }
+
+    @Test(timeout = 1000)
+    fun invoke_returnsFailure_whenDeepLinkFallbackIsNull() {
```
saperi22: this and the above one are the same in content

BraintreeCore/src/androidTest/java/com/braintreepayments/api/core/usecase/GetReturnLinkTypeUseCaseTest.kt
noguier: so i tried, but app-links are very hard to test without mocking, and i thought we were not supposed to use mocks in instrumentation tests?

BraintreeCore/src/androidTest/java/com/braintreepayments/api/core/usecase/GetReturnLinkUseCaseTest.kt
noguier: Thank you!

AmericanExpress/build.gradle
```diff
@@ -25,6 +25,12 @@ android {
         }
     }
 
+    buildTypes {
+        debug {
+            testCoverageEnabled true
```
saralvasquez, line 30: If I'm reading your bullet list correctly, this is to report on test coverage right? Where would we find the results of that test coverage?

BraintreeCore/src/androidTest/java/com/braintreepayments/api/core/ConfigurationCacheTest.kt
```diff
@@ -0,0 +1,118 @@
...
+        sut.saveConfiguration(configuration, cacheKey, saveTimestamp)
+        val result = sut.getConfiguration(cacheKey, justBeforeExpiry)
+
+        assertNotNull(result)
+    }
+
+    @Test(timeout = 1000)
+    fun saveConfiguration_overwritesPreviousValue() {
+        val configA = Configuration.fromJson(Fixtures.CONFIGURATION_WITH_ENVIRONMENT)
+        val configB = Configuration.fromJson(Fixtures.CONFIGURATION_WITH_MERCHANT_ID)
```
saralvasquez: The merchant ids for these two configurations is the same so the assertion below doesn't mean much. We should find two configurations with the same field but different values for that field

BraintreeCore/src/androidTest/java/com/braintreepayments/api/core/DeviceInspectorTest.kt
```diff
@@ -0,0 +1,140 @@
...
+    fun getDeviceMetadata_returnsNonEmptyDeviceModel() {
+        val metadata = sut.getDeviceMetadata(context, configuration, "test-session", IntegrationType.CUSTOM)
+        requireNotNull(metadata.deviceModel)
+        assertTrue(metadata.deviceModel.isNotEmpty())
+    }
+
+    @Test(timeout = 1000)
+    fun getDeviceMetadata_detectsEmulatorOnCIEmulator() {
+        val metadata = sut.getDeviceMetadata(context, configuration, "test-session", IntegrationType.CUSTOM)
+        assertTrue("Expected emulator detection to return true on CI emulator", metadata.isSimulator)
```
saralvasquez, line 98: The value `isSimulator` is supposed to be a boolean. I think AI didn't catch that the string response here might be due to how it's testing and isn't actually testing the value we would expect

saralvasquez (commented):
Couple of wonky tests and a qq but otherwise looks great!

AmericanExpress/build.gradle
noguier: Hi, yes this is to report the coverage. We would call it like
 `./gradlew :[module]:createDebugAndroidTestCoverageReport` and the report will be generated locally at 
`[full_path_to_repo]/braintree_android/[module]/build/reports/coverage/androidTest/debug/connected/index.html`

BraintreeCore/src/androidTest/java/com/braintreepayments/api/core/ConfigurationCacheTest.kt
noguier: good catch, thank you very much

BraintreeCore/src/androidTest/java/com/braintreepayments/api/core/DeviceInspectorTest.kt
noguier: Hey Sara, i am a bit confused for your comment , are you talking about this string ""Expected emulator detection to return true on CI emulator"? if so, it is just a message for the assertion. assertTrue still checks that the metadata.IsSimulator is true.

BraintreeCore/src/androidTest/java/com/braintreepayments/api/core/ConfigurationCacheTest.kt
noguier: addressed in f52aa24

BraintreeCore/src/androidTest/java/com/braintreepayments/api/core/DeviceInspectorTest.kt
saralvasquez: Oh jeeze lol. I didn't read the actual assertion and thought it was `assertEquals`. This makes sense. My bad!

saralvasquez (approved):
🎉

saralvasquez:
I'd recommend changing the PR title before merging though. This is more of a `test:` type PR than a `feat:` type PR

saperi22 (approved):

PR #1591, opened 2026-04-28
feature: adds new github action workflow for PR digest

saralvasquez:
### Summary of changes
- Adds a new GH actions workflow to pull all PRs across all our mobile repos that need some attention and posts it to our core slack channel twice a day

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
I used claude desktop to iterate on ideas for how to accomplish the custom digest as I wanted it. It generated the yml file and I made small tweaks to get it closer to what I wanted in terms of output message and specific criteria. I also used claude code to give the code a first pass review

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @saralvasquez

### DIFF
--- a/.github/workflows/pr-review-digest.yml
+++ b/.github/workflows/pr-review-digest.yml
@@ -1,0 +1,169 @@
+name: PR Review Digest
+
+on:
+  schedule:
+    # 9:00 AM Central (CDT = UTC-5). Update to '0 15 * * 1-5' when clocks fall back to CST (UTC-6) in November.
+    - cron: '0 14 * * 1-5'
+    # 3:00 PM Central (CDT = UTC-5). Update to '0 21 * * 1-5' when clocks fall back to CST (UTC-6) in November.
+    - cron: '0 20 * * 1-5'
+  workflow_dispatch:
+
+jobs:
+  digest:
+    name: Post PR review digest to Slack
+    runs-on: ubuntu-latest
+    steps:
+      # ── Step 1: Authenticate as the GitHub App ─────────────────────────────
+      # This action handles JWT generation and token exchange automatically.
+      - name: Generate GitHub App token
+        id: app-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.GH_APP_ID }}
+          private-key: ${{ secrets.GH_APP_PRIVATE_KEY }}
+          # Scopes the token to the whole org so it can read any repo with our GH app installed.
+          owner: ${{ github.repository_owner }}
+
+      # ── Step 2: Find PRs needing review and post digest to Slack ───────────
+      - name: Find PRs needing review and post to Slack
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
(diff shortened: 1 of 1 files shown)


jaxdesmarais (commented):
This is cool! Is there a way to have it live somewhere else vs needing to duplicate the same logic across all of the repos? That way we could have 1 workflow that pulls in everything we need instead of 9 of the same file. I also wonder if there is a way to have it only pull in the inner source PRs on the days we have agreed to review those PRs per the inner source guidelines?

saralvasquez:
> Is there a way to have it live somewhere else vs needing to duplicate the same logic across all of the repos?

This does pull all the repos. I only added it here to get a proof of concept, but ideally we would have a separate tooling repo where we keep stuff that spans all our code bases. Oh and I only added the active repos since the others are inactive and have been marked for deprecation

saralvasquez:
As for the inner source PRs, we do have the data to do that but I'd prefer to keep the logic here simple and iterate later after we have proof of concept. If we really want to distinguish which PRs are inner sourced we can have the devs add a tag to the PR and it will show up in the digest

jaxdesmarais:
> This does pull all the repos. I only added it here to get a proof of concept, but ideally we would have a separate tooling repo where we keep stuff that spans all our code bases. Oh and I only added the active repos since the others are inactive and have been marked for deprecation

Oops, I missed this part. 🤦 thanks for clarifying.

> As for the inner source PRs, we do have the data to do that but I'd prefer to keep the logic here simple and iterate later after we have proof of concept. If we really want to distinguish which PRs are inner sourced we can have the devs add a tag to the PR and it will show up in the digest

I am fine to iterate on this! I really like the idea of tags here as a way of easily filtering inner sourced PRs in the future.

.github/workflows/pr-review-digest.yml
```diff
@@ -0,0 +1,169 @@
...
+        run: |
+          set -euo pipefail
+          
+          THRESHOLD="${OPEN_THRESHOLD_HOURS:-12}"
+          NOW=$(date +%s)
+          TOTAL=0
+          SLACK_BODY=""
+          
+          if [[ -z "$REPOS" ]]; then
+            echo "REPOS variable is empty — Please set it to a comma-separated list of repositories (e.g. 'org/repo1, org/repo2') in Actions Variables."
```
jaxdesmarais, line 44: Ah, so the repos are set in actions or somewhere else? Sorry probably a basic question but just trying to understand. Not a blocker though.

jaxdesmarais (approved):

.github/workflows/pr-review-digest.yml
saralvasquez: That's a good question. It's set in the settings for the repo under `Secrets and Variables` -> `Actions`-> `Variables`

agedd (approved):
this is so cool! ty for adding this 🚀

PR #1595, opened 2026-05-06
feat: Add card brand drawables, credit card formatting and brand detection

noguier:
Screenshot_20260506_104008

[Screen_recording_20260506_134452.webm](https://github.com/user-attachments/assets/724c56d0-efd4-4b47-a96d-776c655d27da)

### Summary of changes
- Create vector drawables for each card brand
- Add `CardNumberFormatter`, that strips non-digits, detects the card brand via `CardBrand.detect()`, inserts spaces, enforces max digit length, and preserves cursor position across reformats
- Wire `CardNumberFormatter` into `CardNumberTextInputView` so the card brand icon updates dynamically as the user types                                  
- Extend `CardBrand` enum with ELO, MIR, and VERVE as supported card brands
- Unit test all new logic

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Vector tweaking to align exactly with the design, cursor positioning logic, unit tests generation

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry

@noguier


UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardNumberFormatter.kt
```diff
@@ -0,0 +1,136 @@
...
+            return when {
+                matches.size == 1 -> matches[0]
+                else -> CardBrand.UNKNOWN
+            }
+        }
+
+        fun formatCardNumber(digits: String, formatGaps: IntArray): String {
+            if (digits.isEmpty()) return ""
+            val sb = StringBuilder(digits.length + formatGaps.size)
+            for (i in digits.indices) {
```
jaxdesmarais: nit: are we able to use fully typed out names for these vs shorthand?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardNumberFormatter.kt
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        fun countDigitsBeforeIndex(text: String, index: Int): Int {
+            return text.take(index).count { it.isDigit() }
+        }
+
+        fun findIndexForDigitPosition(text: String, digitPosition: Int): Int {
+            if (digitPosition == 0) return 0
+            var count = 0
+            for (i in text.indices) {
```
jaxdesmarais: same here, can we use `index` here instead of `i`?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardBrand.kt
```diff
@@ -88,6 +125,44 @@ internal enum class CardBrand(
 
     val maxLength: Int get() = validLengths.max()
 
+    @get:DrawableRes
+    val iconRes: Int
```
jaxdesmarais, line 129: what does "res" stand for here?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardBrand.kt
```diff
@@ -88,6 +125,44 @@ internal enum class CardBrand(
...
+            HIPER -> R.drawable.card_fields_cc_unknown
+            HIPERCARD -> R.drawable.card_fields_cc_unknown
+            ELO -> R.drawable.card_fields_cc_unknown
+            MIR -> R.drawable.card_fields_cc_unknown
+            VERVE -> R.drawable.card_fields_cc_unknown
+            UNKNOWN -> R.drawable.card_fields_cc_unknown
+        }
+
+    @get:StringRes
+    val iconContentDescriptionRes: Int
```
jaxdesmarais, line 148: same q for res?

CHANGELOG.md
```diff
@@ -5,7 +5,8 @@
 * GooglePay
    * Add optional `DisplayItems` in GooglePay request
 * UIComponents
-   * Add `BaseTextInputView` for CardFields input components
+   * Add `BaseTextInputView` for CardFields input components 
+   * Add `CardNumberTextInputView`, a specialized card number input with dynamic card brand icon display
```
jaxdesmarais: Are these two exposed to merchants? If not we should omit them.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardBrand.kt
noguier: res = resource 
It usually comes from android naming conventions, here is the example of some 
Screenshot 2026-05-06 at 4 02 30 PM

for our variables I can use it as `val iconResource: Int` if it is more readable?
The reason it is an int because it means the value is an Android resource ID (an R.* reference)

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardBrand.kt
noguier: responded above

CHANGELOG.md
noguier: Makes sense, will delete this entry because it will not be merchant-facing!

noguier: updated in b50e484

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardNumberFormatter.kt
noguier: yes, great suggestion, it is easier to follow with explicit names, especially because two functions look very similar, but a have different role. 
Updated in b50e484

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardNumberFormatter.kt
noguier: Updated in b50e484

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardBrand.kt
jaxdesmarais: Ah - will leave that up to y'all what pattern we want to use. If it's the standard in Android that is fine by me. :)

jaxdesmarais (approved):

UIComponents/src/main/res/drawable-xxhdpi/card_fields_cc_discover.png
saralvasquez, line 1: Is there a reason this is the only drawable that's a png instead of a vector drawable?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardBrand.kt
```diff
@@ -78,6 +82,39 @@ internal enum class CardBrand(
         cvvLength = 3
     ),
 
+    ELO(
+        prefixPatterns = listOf(
```
saralvasquez, line 86: Just cuz I didn't ask on the last PR, we got these prefixes from somewhere right? Like iOS or the HLD or something? Just worth double checking

saralvasquez (commented):
Ok I'm still getting through the logic for the number formatter but the rest looks good and I have a couple small questions so far I figured I'd put out while I'm still looking through the formatting file

UIComponents/src/main/res/drawable-xxhdpi/card_fields_cc_discover.png
noguier: yes, it is due to discover logo being too complicated to convert to vector drawable (subtle lines on the logo, color gradient). I could not find vector drawable for it either. So I had to go with png, on a most common android screen resolution.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardBrand.kt
noguier: yes, for every card except elo, mir and verve, the regex comes directly from drop-in, otherwise I found [this](https://github.com/braintree/credit-card-type/blob/main/src/lib/card-types.ts) BT repo that has the regex for elo, mir, and verve.

saralvasquez (commented):
I _think_ this is ready to go but I did notice that the Union pay matcher didn't work for cards that start with 622 but work for all other 62_ digits. Is that expected?

noguier:
> I think this is ready to go but I did notice that the Union pay matcher didn't work for cards that start with 622 but work for all other 62_ digits. Is that expected?

Thank you for this comment, 622_ must resolve to Discover, but it is not the case in my implementation, I will fix it!

saralvasquez (approved):
🎉

PR #1600, opened 2026-05-14
feature(PayPal): collect and send device_info to determine app switch eligibility

santugowda:
## Summary of changes

Collects and sends device information as `app_switch_context.device_info` in PayPal Hermes requests when `enablePayPalAppSwitch` is enabled. This data allows the PayPal server to determine app switch eligibility, primarily for low-RAM devices.

### Background

The PayPal App Switch flow launches the PayPal native app directly, bypassing browser checkout. On low-RAM devices, running two apps simultaneously (the merchant app + PayPal app) can cause the OS to kill the merchant app in the background due to memory pressure — leading to a broken checkout experience when the user tries to return.

The PayPal server uses `memory_available_mb` and `memory_total_mb` to gate app switch eligibility — if the device doesn't have enough RAM, the server skips app switch and falls back to browser checkout (Chrome Custom Tab), which is lighter on memory.

This PR adds collection and injection of:
- `model` — device model (e.g. `SM-A166U`) for any device-specific overrides the server may have
- `memory_available_mb` — free RAM at request time
- `memory_total_mb` — total device RAM

These are injected into the request body under `app_switch_context.device_info`, giving the server everything it needs to decide whether to serve an app switch URL or a browser URL in the response.

### JSON structure sent to BTGW

> **Note:** The flat top-level `os_type`, `os_version`, and `merchant_app_return_url` keys are **intentionally retained**. BTGW reads these flat fields to determine app switch eligibility — sending only the nested structure broke the billing/vault app-switch flow (see commit 434accfd). The nested `app_switch_context` now carries **only** `device_info`; the previously-nested `native_app` block (`os_type`/`os_version`/`app_url`) was **removed** because it simply duplicated the flat fields and is not consumed by BTGW. Retiring the flat fields depends on a future BTGW migration, tracked separately.

```json
{
  "launch_paypal_app": true,
  "os_type": "Android",
  "os_version": "36",
  "merchant_app_return_url": "https://...",
  "app_switch_context": {
    "device_info": {
      "model": "SM-A166U",
      "memory_available_mb": 350,
      "memory_total_mb": 4096
    }
  }
}
```

**Changes:**
- `PayPalRequest`: removed the now-unused `NATIVE_APP_KEY`, `APP_URL_KEY`, and `OS_TYPE_VALUE` constants
- `PayPalCheckoutRequest` / `PayPalVaultRequest`: removed the nested `app_switch_context.native_app` block (redundant with the retained flat `os_type`/`os_version`/`merchant_app_return_url` keys)
- `PayPalInternalClient`: `injectDeviceInfo()` adds `device_info` (model, memory_available_mb, memory_total_mb) into `app_switch_context`, creating the object when absent
- Updated unit tests to verify the new request shape

### Steps to Test

1. Set `enablePayPalAppSwitch = true` on `PayPalCheckoutRequest` or `PayPalVaultRequest`
2. Ensure the merchant app is configured with an App Link return URL
3. Trigger a PayPal checkout or vault flow
4. Intercept the request to `paypal_hermes/create_payment_resource` or `paypal_hermes/setup_billing_agreement`
5. Confirm `app_switch_context.device_info` is present in the request body with `model`, `memory_available_mb`, and `memory_total_mb`
6. On a device with sufficient RAM, confirm the server returns an app switch URL (PayPal app launches)
7. On a low-RAM device, confirm the server falls back to browser checkout (Chrome Custom Tab launches)

### AI Usage

**Which AI Agent Was Used?**
- [x] Claude

**How was AI used?**
Code generation, cross-codebase analysis (Android SDK, BTGW, backend), refactoring the request structure, on-device request/response verification, and writing tests.

- [ ] less than 30%
- [x] 30 - 60%
- [ ] 60 - 100%

### Checklist
- [x] Added a changelog entry

### Authors
- @santugowda

----
### Inner Source Process
Internal to PayPal contributors should fill out this section. All others can delete.

PR should follow these steps before codeowners review will begin:
1. Comment `/inner source` on this PR — this will automatically add the `inner source` and `tech lead review required` labels. Open the PR in a draft state.
2. PR should be reviewed by and approved by your team's technical lead, we do not allow LGTM reviews, there should be comments and feedback provided on all PR reviews
3. Once the above steps are completed, comment `/ready` on this PR — this will automatically remove the `tech lead review required` label. Move the PR to ready to review.
4. PR comments must be addressed within 24 hours, if you are unable to address within this timeframe, move the PR back to a draft state so our team knows not to review

### Demo

Request body sent (key fields):
```json
{
  "launch_paypal_app": true,
  "os_type": "Android",
  "os_version": "36",
  "merchant_app_return_url": "https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com/braintree-payments",
  "app_switch_context": {
    "device_info": {
      "model": "SM-S936U",
      "memory_available_mb": 350,
      "memory_total_mb": 11113
    }
  }
}
```

**token=232345522J779231L**

EC flow

https://github.com/user-attachments/assets/28b3af60-5024-467e-8f04-8b3710c514a4

Billing Successful app-switch

https://github.com/user-attachments/assets/11197a51-823e-4ccc-9949-d0ed03878f3c

Billing on Low Ram device not app-switching

https://github.com/user-attachments/assets/1cc24822-d0a8-4be3-867b-a2b5ba66f72d

### Inner Source Checklist
- [x] Added all labels to the PR
- [x] Provide steps to test the flows changed, if applicable in the summary
- [x] Demo video of the functionality, if applicable
- [x] All upstream dependencies are merged in and this PR can be released at any time; PRs should not be opened until this is true
- [x] Unit tests and builds have been run locally and pass/compile as expected


jaxdesmarais:
@santugowda please follow the `Inner Source Process` in the PR description.

santugowda:
@anbojorquez  @ayer-ribeiro could you please review this PR? Thanks!

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt
```diff
@@ -172,9 +172,13 @@ class PayPalCheckoutRequest @JvmOverloads constructor(
...
         if (enablePayPalAppSwitch && !appLink.isNullOrEmpty()) {
             parameters.put(ENABLE_APP_SWITCH_KEY, enablePayPalAppSwitch)
-            parameters.put(OS_VERSION_KEY, Build.VERSION.SDK_INT.toString())
-            parameters.put(OS_TYPE_KEY, "Android")
-            parameters.put(MERCHANT_APP_RETURN_URL_KEY, appLink)
+            parameters.put(APP_SWITCH_CONTEXT_KEY, JSONObject().apply {
+                put(NATIVE_APP_KEY, JSONObject().apply {
+                    put(APP_URL_KEY, appLink)
+                    put(OS_TYPE_KEY, OS_TYPE_VALUE)
+                    put(OS_VERSION_KEY, Build.VERSION.SDK_INT.toString())
```
ayer-ribeiro: Since we will have multiple versions of Braintree in merchant apps, is it okay to send multiple JSON contracts in the checkout request body?

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -187,6 +198,19 @@ internal class PayPalInternalClient(
...
 
+    private fun injectDeviceInfo(context: Context, requestBody: String): String {
+        val activityManager = context.getSystemService(Context.ACTIVITY_SERVICE) as ActivityManager
+        val memInfo = ActivityManager.MemoryInfo().also { activityManager.getMemoryInfo(it) }
+        return JSONObject(requestBody).apply {
+            val appSwitchContext = optJSONObject(PayPalRequest.APP_SWITCH_CONTEXT_KEY) ?: return@apply
+            appSwitchContext.put(PayPalRequest.DEVICE_INFO_KEY, JSONObject().apply {
+                put(PayPalRequest.DEVICE_MODEL_KEY, Build.MODEL)
+                put(PayPalRequest.MEMORY_AVAILABLE_MB_KEY, (memInfo.availMem / BYTES_PER_MB).toInt())
+                put(PayPalRequest.MEMORY_TOTAL_MB_KEY, (memInfo.totalMem / BYTES_PER_MB).toInt())
```
ayer-ribeiro, line 209: Does it make sense to move this to `createRequestBody`?

We could pass `deviceModel`, `totalAvailableMemoryMB`, and `totalMemoryMB` as parameters of the `createRequestBody` function.

Then we would:

- avoid instantiating multiple JSONObject objects
- ensure createRequestBody contains the device info even if it’s called in other places

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -187,6 +198,19 @@ internal class PayPalInternalClient(
             .build()
     }
 
+    private fun injectDeviceInfo(context: Context, requestBody: String): String {
+        val activityManager = context.getSystemService(Context.ACTIVITY_SERVICE) as ActivityManager
+        val memInfo = ActivityManager.MemoryInfo().also { activityManager.getMemoryInfo(it) }
+        return JSONObject(requestBody).apply {
+            val appSwitchContext = optJSONObject(PayPalRequest.APP_SWITCH_CONTEXT_KEY) ?: return@apply
```
ayer-ribeiro: To ensure this function is robust and doesn’t depend on previous conditions, should we create the `APP_SWITCH_CONTEXT_KEY` field when it doesn’t exist and ensure the device info is added?

That would prevent us from skipping the device info handling in scenarios where there is no app switch context.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -69,14 +74,20 @@ internal class PayPalInternalClient(
...
+        val baseRequestBody = payPalRequest.createRequestBody(
             configuration = configuration,
             authorization = merchantRepository.authorization,
             successUrl = successUrl,
             cancelUrl = cancelUrl,
             appLink = appLinkParam
         ) ?: throw JSONException("Error creating requestBody")
 
+        val requestBody = if (payPalRequest.enablePayPalAppSwitch && appLinkParam != null) {
+            withContext(Dispatchers.IO) { injectDeviceInfo(context, baseRequestBody) }
```
ayer-ribeiro: do we really need using Dispatchers.IO here? I'd change to Default dispatcher,
AI analysis attached below:

> Dispatchers.IO is intended for operations that may block on external I/O, such as disk access, network calls, or blocking socket operations. The work performed inside injectDeviceInfo does not appear to be I/O-bound:
> 
> ActivityManager.getMemoryInfo() — This is a Binder IPC call to ActivityManagerService, not direct file or disk I/O from the app process. It retrieves system memory information through the Android framework and is not documented as requiring a worker thread or as an I/O-bound operation.
> Build.MODEL — This is a static String value exposed by the Android framework. Accessing it is effectively a simple field read.
> JSONObject operations — These are in-memory operations involving JSON construction or mutation. They are CPU work, not blocking I/O.
> 
> Because this block is a mix of a lightweight framework call and small in-memory work, routing it through Dispatchers.IO is semantically misleading. If a dispatcher switch is needed for defensive reasons, Dispatchers.Default is a better fit than Dispatchers.IO; otherwise, the withContext may not be necessary at all depending on the caller context.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt
santugowda: This was discussed and agreed upon with the BTGW team. BTGW normalizes the request body before forwarding downstream to PayPal backend services. 
The flat top-level fields `(os_type, os_version, merchant_app_return_url) `coexist with the nested `app_switch_context` in both the old and new contract — BTGW handles the cleanup on their end regardless of SDK version

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
santugowda: The device_info fields require Android system calls (ActivityManager, Build.MODEL) which are platform-specific and don't belong in the request model layer. The BTGW team reviewed and agreed to this contract structure. That said, open to moving it if the team prefers — but it would mean threading Context into createRequestBody.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
santugowda: `injectDeviceInfo` is only called when `enablePayPalAppSwitch && appLink != null`. At that point, `app_switch_context` is always present since createRequestBody builds it. The null case is not reachable, no change needed.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
santugowda: Agreed, removed the `withContext(Dispatchers.IO) `wrapper - none of these operations are I/O-bound. 
 Will address this.

ayer-ribeiro (commented):
LGTM

saralvasquez:
In testing the PayPal checkout flows I noticed that single payment with app switch works as expected with the new fields. However, when I enable PP app switch and try to checkout with vault I get an error `PayPal is currently unable to handle the request`. Were you able to get the SDK demo app vault flow to work as expected?

santugowda:
> In testing the PayPal checkout flows I noticed that single payment with app switch works as expected with the new fields. However, when I enable PP app switch and try to checkout with vault I get an error `PayPal is currently unable to handle the request`. Were you able to get the SDK demo app vault flow to work as expected?

@saralvasquez  This was existing issue in BT Demo App(I see this issue on Main branch as well). 
So to test vault flow, I created a new BT SDK - integrated in XOSphere(PayPal - internal demo app) please check` Billing Successful app-switch` video in the PR description.

saralvasquez:
I get different behavior on the main branch vs this branch. On the main branch when the vault option is selected app switch happens and then inside the PP app the payment can't complete. On this branch when the vault option is selected there is no app switch at all and we simply get the error inside the demo app

santugowda:
> I get different behavior on the main branch vs this branch. On the main branch when the vault option is selected app switch happens and then inside the PP app the payment can't complete. On this branch when the vault option is selected there is no app switch at all and we simply get the error inside the demo app

Ok, I do see the error now on this branch alone. (had to use a physical device)
But on using BT-SDK in XOSphere app - AppSwitch is working for Billing/Vault flows. 

Below is the error response I get - 
`java.io.IOException: Unexpected code Response{protocol=h2, code=503, message=, url=https://api.braintreegateway.com/merchants/dfy45jdj3dxkmz5m/client_api/v1/paypal_hermes/setup_billing_agreement} with body {"error":"PayPal is currently unable to handle the request"}`

I don't see any token or correlationID to investigate on this error on BE side - any pointers on this.
FYI via Postman the /setup_billing_agreement is working as expected. Let me know, we can have a working session if needed.

santugowda:
> > I get different behavior on the main branch vs this branch. On the main branch when the vault option is selected app switch happens and then inside the PP app the payment can't complete. On this branch when the vault option is selected there is no app switch at all and we simply get the error inside the demo app
> 
> Ok, I do see the error now on this branch alone. (had to use a physical device) But on using BT-SDK in XOSphere app - AppSwitch is working for Billing/Vault flows.
> 
> Below is the error response I get - `java.io.IOException: Unexpected code Response{protocol=h2, code=503, message=, url=https://api.braintreegateway.com/merchants/dfy45jdj3dxkmz5m/client_api/v1/paypal_hermes/setup_billing_agreement} with body {"error":"PayPal is currently unable to handle the request"}`
> 
> I don't see any token or correlationID to investigate on this error on BE side - any pointers on this. FYI via Postman the /setup_billing_agreement is working as expected. Let me know, we can have a working session if needed.

@saralvasquez  Found the issue - it was missing fields in the request body. Will push this fix in a bit.

santugowda:
> > > I get different behavior on the main branch vs this branch. On the main branch when the vault option is selected app switch happens and then inside the PP app the payment can't complete. On this branch when the vault option is selected there is no app switch at all and we simply get the error inside the demo app
> > 
> > 
> > Ok, I do see the error now on this branch alone. (had to use a physical device) But on using BT-SDK in XOSphere app - AppSwitch is working for Billing/Vault flows.
> > Below is the error response I get - `java.io.IOException: Unexpected code Response{protocol=h2, code=503, message=, url=https://api.braintreegateway.com/merchants/dfy45jdj3dxkmz5m/client_api/v1/paypal_hermes/setup_billing_agreement} with body {"error":"PayPal is currently unable to handle the request"}`
> > I don't see any token or correlationID to investigate on this error on BE side - any pointers on this. FYI via Postman the /setup_billing_agreement is working as expected. Let me know, we can have a working session if needed.
> 
> @saralvasquez Found the issue - it was missing fields in the request body. Will push this fix in a bit.

Fixed this in the lastest commit i.e. #1600 434accfde9f4aed2b070bce74867d457da5bfe57
Attaching video for reference as well, please review when you get a chance. 

https://github.com/user-attachments/assets/bdb23f63-87c9-4359-ab5e-53372d887a11

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalVaultRequest.kt
```diff
@@ -103,6 +103,13 @@ class PayPalVaultRequest
             parameters.put(OS_VERSION_KEY, Build.VERSION.SDK_INT.toString())
             parameters.put(OS_TYPE_KEY, "Android")
             parameters.put(MERCHANT_APP_RETURN_URL_KEY, appLink)
```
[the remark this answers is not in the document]
santugowda: Updated the description.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt
```diff
@@ -175,6 +175,13 @@ class PayPalCheckoutRequest @JvmOverloads constructor(
             parameters.put(OS_VERSION_KEY, Build.VERSION.SDK_INT.toString())
             parameters.put(OS_TYPE_KEY, "Android")
             parameters.put(MERCHANT_APP_RETURN_URL_KEY, appLink)
```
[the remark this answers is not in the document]
santugowda: Updated the description.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -187,6 +196,19 @@ internal class PayPalInternalClient(
...
+        val memInfo = ActivityManager.MemoryInfo().also { activityManager.getMemoryInfo(it) }
+        return JSONObject(requestBody).apply {
+            val appSwitchContext = optJSONObject(PayPalRequest.APP_SWITCH_CONTEXT_KEY) ?: return@apply
+            appSwitchContext.put(PayPalRequest.DEVICE_INFO_KEY, JSONObject().apply {
+                put(PayPalRequest.DEVICE_MODEL_KEY, Build.MODEL)
+                put(PayPalRequest.MEMORY_AVAILABLE_MB_KEY, (memInfo.availMem / BYTES_PER_MB).toInt())
+                put(PayPalRequest.MEMORY_TOTAL_MB_KEY, (memInfo.totalMem / BYTES_PER_MB).toInt())
+            })
+        }.toString()
+    }
```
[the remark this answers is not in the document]
santugowda: Sure, this makes sense added safe cast `(as? ActivityManager)` that returns the original request body if the service is unavailable

noguier:
As a follow up, i tested the flows and inspected request that are sent to `paypal_hermes/create_payment_resource` or `paypal_hermes/setup_billing_agreement`. Confirmed that the changes are working as expected.

CHANGELOG.md
noguier, line 1: i think it is a bit too wordy for the changelog entry, additionally I am not sure if the merchants will know what PayPal Hermes is.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -187,6 +196,20 @@ internal class PayPalInternalClient(
             .build()
     }
 
+    private fun injectDeviceInfo(context: Context, requestBody: String): String {
+        val activityManager =
+            context.getSystemService(Context.ACTIVITY_SERVICE) as? ActivityManager ?: return requestBody
+        val memInfo = ActivityManager.MemoryInfo().also { activityManager.getMemoryInfo(it) }
+        return JSONObject(requestBody).apply {
```
noguier, line 203: would you help me understand why we need activityManager here?

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalInternalClientUnitTest.kt
```diff
@@ -1076,4 +1085,143 @@ class PayPalInternalClientUnitTest {
...
+            assertEquals(4096, deviceInfo.getInt("memory_total_mb"))
+
+            val nativeApp = appSwitchContext.getJSONObject("native_app")
+            assertEquals("ANDROID", nativeApp.getString("os_type"))
+            assertNotNull(nativeApp.getString("os_version"))
+            assertNotNull(nativeApp.getString("app_url"))
+        }
+
+    @Test
+    fun `sendRequest with app switch enabled and unavailable ActivityManager omits device_info`() =
```
noguier, line 1122: in what condition ActivityManager might be unavailable?

CHANGELOG.md
santugowda: ack - let me update this.  Below is the content : 

Collect device information to improve PayPal app switch eligibility when it's enabled

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalInternalClientUnitTest.kt
santugowda: This was added in response to the earlier Copilot feedback to avoid a hard ClassCastException on the request path.In practice on a real device it's effectively never null — ACTIVITY_SERVICE is a core system service. The safe cast (as?) and this test are defensive rather than for a common runtime path. 


But `getSystemService()` can return null (or a non-ActivityManager type) in edge cases like a Context that isn't fully initialized/associated with a system context, or a wrapped/mocked Context in unusual embedded or instrumented environments.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
santugowda: @noguier  -- ActivityManager is the only Android API that exposes the device's runtime memory. 
We use it to read `MemoryInfo.availMem` and `MemoryInfo.totalMem`, which populate device_info.memory_available_mb and memory_total_mb in the request body. PayPal's backend uses these values to decide app-switch eligibility on low-RAM devices.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt
```diff
@@ -175,6 +175,13 @@ class PayPalCheckoutRequest @JvmOverloads constructor(
...
+            parameters.put(APP_SWITCH_CONTEXT_KEY, JSONObject().apply {
+                put(NATIVE_APP_KEY, JSONObject().apply {
+                    put(APP_URL_KEY, appLink)
+                    put(OS_TYPE_KEY, OS_TYPE_VALUE)
+                    put(OS_VERSION_KEY, Build.VERSION.SDK_INT.toString())
```
jaxdesmarais: Sending some of these in 2 places seems redundant. I thought the idea was that we wanted to move to the new formatting and remove the old? This would also allow the GW to eventually remove their old code too. Are you able to clarify with them as sending them the same value in 2 places seems like tech debt we will need to clean up eventually.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalRequest.kt
```diff
@@ -155,5 +159,9 @@ abstract class PayPalRequest internal constructor(
         internal const val CONTACT_PREFERENCE_KEY: String = "contact_preference"
         internal const val SHOPPER_SESSION_ID_KEY: String = "shopper_session_id"
         internal const val AMOUNT_BREAKDOWN_KEY: String = "amount_breakdown"
+        internal const val DEVICE_INFO_KEY: String = "device_info"
+        internal const val DEVICE_MODEL_KEY: String = "model"
```
jaxdesmarais, line 160: We use `mobile_device_model` on iOS. We should send the same tag for the same things across both platforms.

santugowda: @jaxdesmarais  : `mobile_device_model` is mainly used for FPTI analytics and on Android we are using the same in [AnalyticApi.kt](https://github.com/braintree/braintree_android/blob/706c842008117fece8c82c3dc7e2eb4438be315c/BraintreeCore/src/main/java/com/braintreepayments/api/core/AnalyticsApi.kt#L181)

`model` in this PR is a different object, though — it's not an FPTI tag but a field in the app_switch_context request body to the GW/PayPal backend for the (Android-only) low-RAM eligibility feature. That device_info shape (model, memory_available_mb, memory_total_mb) is the contract we agreed with the BTGW and PayPal backend teams, so model is what they're keyed to parse. 
FPTI stays consistent with iOS; this field just follows the backend contract.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt
santugowda: Agreed, this is redundant — `os_type, os_version, and merchant_app_return_url `(as `native_app.app_url`) are sent both flat at the top level and nested under `app_switch_context`. The  nested block is the new format we agreed with BTGW.

I actually did try nested-only first, but had to revert (434accfd) and retain the flat params because removing them broke the billing agreement / vault flow — BTGW still reads  the flat fields(Please see my comments above with @saralvasquez). So sending both is what we agreed with BTGW for now.

saralvasquez: Is there a confirmed plan for GW to change how they are managing the vault flow so that the top level values can be removed? I  would hate for this to get lost in the shuffle and then we have duplicate values hanging around that we can never clean up

santugowda: @saralvasquez  - the nested `app_switch_context` is the agreed new format, but removing the flat top-level fields depends on BTGW migrating the vault flow off them.                    
                                                                                                                                                                                     
Also, the nesting is currently Android-only,  it was introduced as part of this feature. iOS still sends the flat param request, so ideally this needs a tech-backlog item to clean it up across both native platforms and BTGW.
cc: @jaxdesmarais

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalRequest.kt
```diff
@@ -143,9 +143,13 @@ abstract class PayPalRequest internal constructor(
...
         internal const val USER_ACTION_KEY: String = "user_action"
         internal const val ENABLE_APP_SWITCH_KEY: String = "launch_paypal_app"
+        internal const val APP_SWITCH_CONTEXT_KEY: String = "app_switch_context"
+        internal const val NATIVE_APP_KEY: String = "native_app"
+        internal const val APP_URL_KEY: String = "app_url"
+        internal const val MERCHANT_APP_RETURN_URL_KEY: String = "merchant_app_return_url"
         internal const val OS_VERSION_KEY: String = "os_version"
         internal const val OS_TYPE_KEY: String = "os_type"
-        internal const val MERCHANT_APP_RETURN_URL_KEY: String = "merchant_app_return_url"
+        internal const val OS_TYPE_VALUE: String = "ANDROID"
```
saralvasquez: It looks like this value in the top level above is `Android` why to switch to all caps? Will that cause issues with GW in the future when we eventually remove the top level? I feel like we should keep it consistent across the two values if possible

saralvasquez (commented):
A couple open questions but the logic makes sense to me

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt
jaxdesmarais: > Also, the nesting is currently Android-only, it was introduced as part of this feature. iOS still sends the flat param request, so ideally this needs a tech-backlog item to clean it up across both native platforms and BTGW.

We were not involved in these conversations nor told this was an item needed for our backlog/prioritized. If these changes are not needed for this code change, lets revert. If they are needed we would expect XO to handle this cleanup in both repos as part of this workstream.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalRequest.kt
santugowda: > why to switch to all caps?

No specific reason, just the payload we agreed with BTGW.

> Will that cause issues with GW in the future when we eventually remove the top level

Removing the top-level fields isn't something we've committed to - so far we haven't discussed removing any request objects with BTGW. 
For now BTGW reads the flat and nested fields via separate paths, so the casing difference isn't an issue.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt
saralvasquez: > @saralvasquez - the nested `app_switch_context` is the agreed new format, but removing the flat top-level fields depends on BTGW migrating the vault flow off them.


Yeah I understand that it requires GW migrating off of them. I'm more asking if they have prioritized or formalized that work in any way or if we're going to need to follow up with them in the future

santugowda: To clarify: the nested `app_switch_context` isn't optional - it's the container for `device_info` (model + memory), which is the core of this PR's low-RAM feature, so we can't revert it without dropping the feature.

@saralvasquez  - no we need to discuss this with BTGW and prioritize this.

noguier:
Hey there again, I tested the flows one more time, and the Checkout flow does not app switch anymore. The vault completes with no problems though. Please note we will want both flows working exactly as expected before merging. Attaching the video: 

https://github.com/user-attachments/assets/56057b96-19ef-49b7-b5e2-cf68a4914cd2

saralvasquez:
I noticed the same behavior when re-testing just now

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt
saralvasquez: Ok yeah how it is now makes sense to me and I'm cool with it as long as we have confirmation from GW that they intend to change stuff so we can clean it up later

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalRequest.kt
jaxdesmarais: Ah - gotcha. Sorry didn't realize we were in the request file. :)

santugowda:
ok, should be a error on BE. It's the same experience on `main` branch as well. 
Let me debug this flow at my end.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt
jaxdesmarais: @santugowda which `OS_TYPE_KEY` and `OS_VERSION_KEY` takes precedent when both are passed? Can we remove 1 set in the PR?

santugowda: @jaxdesmarais — BTGW uses the flat top-level os_type (e.g. "Android") / os_version (e.g. "36"). These existed before this PR and app switch has always relied on them. The nested copies inside `app_switch_context` are the new format BTGW is moving to, so they're extra for now.

I also tested removing the flat top-level `os_type/os_version`: Single Payment (EC) is unaffected, but Billing (Vault) fails on app switch. So we can't remove them yet.

I'll start a Slack thread with the BTGW team to help clean up these extra flat top-level fields.

santugowda:
Single Payment is working now.

https://github.com/user-attachments/assets/295fc547-3f1d-4039-866b-09265bbf711f



Slack thread has the info about yesterday's error, yet to close out on the root cause.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalCheckoutRequest.kt
santugowda: @jaxdesmarais  : ignore my above comment.  Please finds the latest below : 

Reverted back to the flat os_type/os_version request. app_switch_context now carries only device_info (the low-RAM data), which is Android-specific and not needed on iOS. No   tech-backlog/cleanup follow-up needed. Details shared in the Slack thread.

Resolving this conversation.

PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalInternalClient.kt
```diff
@@ -187,6 +196,21 @@ internal class PayPalInternalClient(
             .build()
     }
 
+    private fun injectDeviceInfo(context: Context, requestBody: String): String {
```
jaxdesmarais: nit: will defer to android folks but `append` may be a better name vs `inject`

saralvasquez: Yeah I agree here. Append is preferable

jaxdesmarais:
@santugowda instrumentation tests are failing here but passing on other branches. Is this something you want to take a look at?

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalCheckoutRequestUnitTest.kt
```diff
@@ -225,9 +225,15 @@ class PayPalCheckoutRequestUnitTest {
...
         assertTrue(jsonObject.getBoolean("launch_paypal_app"))
+
+        // flat app switch parameters
         assertEquals("Android", jsonObject.getString("os_type"))
-        assertEquals(appLink, jsonObject.getString("merchant_app_return_url"))
         assertNotNull(jsonObject.getString("os_version"))
+        assertEquals(appLink, jsonObject.getString("merchant_app_return_url"))
+
+        // app_switch_context is no longer built here; device_info is injected later by
+        // PayPalInternalClient, so the request body itself carries no app_switch_context.
```
saralvasquez: nit: I don't think the comments in this file are necessary

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalInternalClientUnitTest.kt
```diff
@@ -1076,4 +1085,133 @@ class PayPalInternalClientUnitTest {
...
+
+            val payPalRequest = PayPalCheckoutRequest("1.00", true).apply {
+                enablePayPalAppSwitch = true
+            }
+
+            sut.sendRequest(context, payPalRequest, configuration)
+
+            val actual = JSONObject(slot.captured)
+            // With native_app removed, app_switch_context is only created when device_info is
+            // injected; an unavailable ActivityManager means it is never added.
```
saralvasquez: nit: same as above

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalVaultRequestUnitTest.kt
```diff
@@ -263,10 +263,17 @@ class PayPalVaultRequestUnitTest {
...
+        val jsonObject = JSONObject(requestBody)
+        assertTrue(jsonObject.getBoolean("launch_paypal_app"))
+
+        // flat app switch parameters
+        assertEquals("Android", jsonObject.getString("os_type"))
+        assertEquals(versionSDK.toString(), jsonObject.getString("os_version"))
+        assertEquals("universal_url", jsonObject.getString("merchant_app_return_url"))
+
+        // app_switch_context is no longer built here; device_info is injected later by
+        // PayPalInternalClient, so the request body itself carries no app_switch_context.
```
saralvasquez: nit: same as above

PayPal/src/test/java/com/braintreepayments/api/paypal/PayPalVaultRequestUnitTest.kt
```diff
@@ -375,9 +382,15 @@ class PayPalVaultRequestUnitTest {
...
         assertTrue(jsonObject.getBoolean("launch_paypal_app"))
+
+        // flat app switch parameters
         assertEquals("Android", jsonObject.getString("os_type"))
-        assertEquals(appLink, jsonObject.getString("merchant_app_return_url"))
         assertNotNull(jsonObject.getString("os_version"))
+        assertEquals(appLink, jsonObject.getString("merchant_app_return_url"))
+
+        // app_switch_context is no longer built here; device_info is injected later by
+        // PayPalInternalClient, so the request body itself carries no app_switch_context.
```
saralvasquez: nit: same as above. Also applies to the comment above

saralvasquez (commented):
I've tested both checkout and vault and both work as expected. I just have a couple nits about comments and I'd like to see the change Jax mentioned about `inject` vs `append`. I'll approve once we have a clean build besides the known security check failure

santugowda:
Pushed the changes addressing all the above nit comments, `append` refactor and instrumentation changes.

saralvasquez (approved):
Once we get a clean build I'd say this is ready to go in

santugowda:
CI is 🟢 except for the known Dependency Review failure.

agedd (approved):

jaxdesmarais:
@santugowda can you please update with the latest from main and move the changelog entry under a new unreleased section?

santugowda:
> @santugowda can you please update with the latest from main and move the changelog entry under a new unreleased section?

@jaxdesmarais  - rebased and updated Changelog.md. Hoping to get this merged today.

mzlangreder (approved):

PR #1601, opened 2026-05-14
feature: add expiration date input field

saralvasquez:
### Summary of changes
- Adds new text view for expiration date
- Adds formatter class for the expiration input
- Adds tests for the new functionality
- Adds the expiration view to the demo app  

[Screen_recording_20260514_151517.webm](https://github.com/user-attachments/assets/9229af58-0052-45a8-9211-e2fe146c878a)

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
I experimented with letting Claude write 100% of this code since there was a lot of context from existing classes. I've thoroughly tested and examined the output, but this is just a note to use a close eye when reviewing this. For what it's worth I had to do very little tweaking on what Claude code put together

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @saralvasquez


UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/ExpirationDateFormatter.kt
```diff
@@ -0,0 +1,90 @@
...
+
+        fun formatExpiration(digits: String): String = when {
+            digits.length <= 2 -> digits
+            else -> "${digits.take(2)}/${digits.drop(2)}"
+        }
+
+        fun findCursorPosition(formatted: String, digitIndex: Int): Int {
+            if (digitIndex == 0) return 0
+            var count = 0
+            for (i in formatted.indices) {
```
noguier: nit: can we rename `i` to a more meaningful name in the for loop?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/ExpirationTextInputView.kt
```diff
@@ -0,0 +1,42 @@
...
+    internal val formatter = ExpirationDateFormatter()
+
+    init {
+        formatter.onMonthRejected = {
+            setError(context.getString(R.string.expiration_invalid_month_error))
+        }
+        setInputType(InputType.TYPE_CLASS_NUMBER)
+        setHint(context.getString(R.string.expiration_hint))
+        editText.contentDescription = context.getString(R.string.expiration_hint_accessibility)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
noguier: what is the purpose of this if statement?

saralvasquez: Good question! That `setAutofillHints` only works on some versions of Android that we allow with the SDK so that checks to make sure it's available before setting the accessibility hint. Although now that I think about it, it would be better to try and find a way that works universally. I'll give it a look

Demo/src/main/res/layout/fragment_card_fields.xml
```diff
@@ -11,4 +11,10 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content" />
 
+    <com.braintreepayments.api.uicomponents.cardfields.ExpirationTextInputView
+        android:id="@+id/expiration_input"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="12dp" />
```
jaxdesmarais, line 18: Probably a very basic android q, but is the the padding between this field and the one above or something else? Wasn't sure why `CardNumberTextInputView` didn't have one, so am assuming it sets the spacing in between those?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/ExpirationDateFormatter.kt
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    companion object {
+        const val MAX_DIGITS = 4
+        const val MAX_LENGTH = 5 // MM/YY
+
+        fun applyLeadingZero(digits: String): String =
+            if (digits.length == 1 && digits[0] > '1') "0$digits" else digits
+
+        fun isValidMonth(digits: String): Boolean {
```
jaxdesmarais: Do we need one of these for year? Or to make sure the year wasn't in the past? Wasn't sure how/if we handed this on iOS. Maybe @buzzamus can help clarify?

buzzamus: I largely yanked the logic from the Drop-In. What iOS does for year is since a card is valid for the entire month listed as the expiration date we move the month forward by 1 from current month and if it is less than card expiration date we return it as invalid. 

We also do some logic that if the exp month is in December we move the year up 1 and set month as 1 to compare

Demo/src/main/res/layout/fragment_card_fields.xml
saralvasquez: Yeah that's exactly what it is. This does bring up an interesting point though. I wonder if there is like a universal padding in the figma that we should add to a theme for all the card field objects

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/ExpirationDateFormatter.kt
saralvasquez: Oh that method only makes sure that the month number isn't over 12 since there's only 12 months. So like the only thing it would reject is if someone tried to put in like "18/20" or something like that

saralvasquez: We don't have to keep it. This was something I added after testing since it felt weird for a user to be able to put in a month number that doesn't exist

noguier: I agree with Sara that it is a nice to have error notification, especially because we do not automatically prepend 0 to 1, because 11 and 12 are valid months.

saralvasquez: Would it help if I renamed the method to something like `isValidMonthNumber`?

Demo/src/main/res/layout/fragment_card_fields.xml
saralvasquez: I took a look at the backlog and I think creating universal spacing will come up in a later story. We've also got a later story for demo app updates that we can use to clean up how the demo app looks so we should treat these changes as temporary for testing purposes only

noguier: Yes, because we will need to create a Card field component that encompasses all the text views, then we can tweak the spacing so that everything looks good together

noguier (approved):
This looks good, error handling and padding is addressed in the next tickets, so I think it is ready to be merged in! 
Great job ❤️

Demo/src/main/res/layout/fragment_card_fields.xml
jaxdesmarais: Thank you both! My XML knowledge is fuzzy so just wanted to make sure I was understanding how things were constructed correctly. :)

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/ExpirationDateFormatter.kt
jaxdesmarais: Ah, your explanation makes sense to me! I think up to y'all what is more clear. Mainly wanted to make sure we didn't need a year counterpart to make sure the year entered wasn't in the past. Good point on prepending a 0... is that something we want to do on both in a future ticket? I know as a customer I rarely remember the 0 unless it's a dropdown.

gradle/libs.versions.toml
```diff
@@ -10,6 +10,7 @@ androidxRoom = "2.6.1"
 androidxTest = "1.5.0"
 androidxAppcompat = "1.6.0"
 androidxAnnotation = "1.7.0"
+androidxAutofill = "1.3.0"
```
jaxdesmarais, line 13: Do we need to call this out in the new dependencies.md?

jaxdesmarais (approved):

gradle/libs.versions.toml
saralvasquez: Ah yes! Thank you for the reminder

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/ExpirationDateFormatter.kt
saralvasquez: Since it's MM/YY there isn't really a way to tell if it's in the past or far future lol but that's good instinct. As for prepending a 0, this PR already has that for month. I don't think year would benefit from the prepend since there's no good way to tell if a 0 is needed before the digit that is typed. I guess we could add a check after the focus changes to make sure there is the right number of digits

saralvasquez: I think I'm going to reserve the changes I suggested at the bottom for when we are making the full card field view and hooking it up to the back end because that's where a lot of the error handling is going to be managed

jaxdesmarais: Ah that's true - and love the optimism of this being around during the next set of 0 starting years. 🙈 You're way ahead of me on the moth prepending! That all sounds great to me.

PR #1602, opened 2026-05-15
fix(paypal): fix malformed return_url in AppLink mode and add bidirectional intent resilience

pedrofsn:
## Problem

In AppLink mode, `PayPalInternalClient` constructed the OneTouch return URL by using the full HTTPS App Link URI as a scheme prefix:

```
https://example.com/paypal-app-switch://onetouch/v1/success  ← malformed
```

This URL has no registered intent handler on Android, so `BrowserSwitchClient` returns `NoResult` instead of `Success`. The billing agreement stays in `PENDING` state.

A secondary issue: `BrowserSwitchOptions` was missing `returnUrlScheme` in the AppLink branch and `appLinkUri` in the DeepLink branch, meaning a mismatch between the intent type chosen at request time and the intent type delivered at return time would always fail both `matchesDeepLinkUrlScheme()` and `matchesAppLinkUri()`.

## Changes

**`GetReturnLinkUseCase`** — extend `ReturnLinkResult.AppLink` to carry `deepLinkFallbackUrlScheme` alongside `appLinkReturnUri`. Add `resolvedDeepLinkFallbackUrlScheme` to `MerchantRepository` as a non-null accessor that falls back to `returnUrlScheme` when the merchant has not explicitly set one.

**`PayPalInternalClient`** — use `deepLinkFallbackUrlScheme` (a valid custom scheme) instead of `appLinkReturnUri.toString()` for the `return_url` and `cancel_url` in AppLink mode. `merchant_app_return_url` continues to use the HTTPS App Link URI.

**`PayPalClient`** — bidirectional resilience in `buildBrowserSwitchOptions()`:
- AppLink branch: add `returnUrlScheme` (only when `deepLinkFallbackUrlScheme` is explicitly configured by the merchant — skipped for App Link–only integrations that have no deep link scheme in their manifest)
- DeepLink branch: add `appLinkUri` when `appLinkReturnUri` is available

## Test coverage

- `GetReturnLinkUseCaseUnitTest`: assert both fields on `AppLink` result, including the fallback-to-`returnUrlScheme` case
- `PayPalInternalClientUnitTest`: assert `return_url` uses the custom scheme and `merchant_app_return_url` uses the HTTPS URI
- `PayPalClientUnitTest`: assert both `returnUrlScheme` + `appLinkUri` are set in AppLink mode; assert both are set in DeepLink mode when `appLinkReturnUri` is configured

## Pre-PR checklist
- [x] `./gradlew --continue clean testRelease` — all modules pass
- [x] `./gradlew clean lint` — no issues
- [x] `./gradlew detekt --auto-correct` — no issues

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/MerchantRepository.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/MerchantRepository.kt
@@ -18,6 +18,9 @@ class MerchantRepository {
 
     var deepLinkFallbackUrlScheme: String? = null
 
+    val resolvedDeepLinkFallbackUrlScheme: String
+        get() = deepLinkFallbackUrlScheme ?: returnUrlScheme
+
     companion object {
 
         /**
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/usecase/GetReturnLinkUseCase.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/usecase/GetReturnLinkUseCase.kt
@@ -30,7 +30,10 @@ class GetReturnLinkUseCase(
 ) {
 
     sealed class ReturnLinkResult {
-        data class AppLink(val appLinkReturnUri: Uri) : ReturnLinkResult()
+        data class AppLink(
+            val appLinkReturnUri: Uri,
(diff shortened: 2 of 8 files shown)


jaxdesmarais:
Closing for now - an update solution will be put up shortly

PR #1604, opened 2026-05-18
feat:  Expose supported integrations and card types via Configuration.fetch()

noguier:
### Summary of changes
- Add a public `Configuration.fetch(context, authorization, callback)` that allows merchants to query which payment integrations are enabled for their account
- Make `ConfigurationCallback` public (remove `@RestrictTo(LIBRARY_GROUP)`)
- Two new `loadConfiguration` overloads were added: 
   - `loadConfiguration()` (no-arg), an existing internal path, uses MerchantRepository authorization and sends analytics for payment flows. Behavior is unchanged. 
   - `loadConfiguration(authorization)`, a new overload for queries via `Configuration.fetch()`. Skips analytics because the full SDK lifecycle (session ID, event queue) is not initialized in this path, and orphaned analytics events would be malformed. Both delegate to a shared private implementation with a `sendAnalytics` flag.

### AI Usage

**Which AI Agent Was Used?**
- [x] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Brainstorm the way to expose the config to the merchants . Copilot was used to give a first review and highlighting a big issue that helped with further developing of a solution.

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

@noguier


BraintreeCore/src/main/java/com/braintreepayments/api/core/Configuration.kt
```diff
@@ -63,6 +68,47 @@ class Configuration internal constructor(configurationString: String) {
...
+            context: Context,
+            authorization: String,
+            callback: ConfigurationCallback
+        ) {
+            fetch(
+                BraintreeClient(context, authorization),
+                CoroutineScope(Dispatchers.Main),
+                callback
+            )
+        }
```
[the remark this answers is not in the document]
noguier: refactored not to spin off a new BT client, but use ConfigurationLoader instead

BraintreeCore/src/main/java/com/braintreepayments/api/core/Configuration.kt
```diff
@@ -63,6 +68,59 @@ class Configuration internal constructor(configurationString: String) {
...
+            callback: ConfigurationCallback
+        ) {
+            SdkComponent.create(context.applicationContext)
+            fetch(
+                ConfigurationLoader(),
+                Authorization.fromString(authorization),
+                CoroutineScope(Dispatchers.Main),
+                callback
+            )
+        }
```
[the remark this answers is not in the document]
noguier: I was not sure if we want to add a CHANGELOG here, since we do not want to encourage the merchants to use new fetch() function?

BraintreeCore/src/test/java/com/braintreepayments/api/core/ConfigurationFetchUnitTest.kt
```diff
@@ -0,0 +1,88 @@
...
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import kotlin.coroutines.cancellation.CancellationException
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertNull
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@RunWith(RobolectricTestRunner::class)
+class ConfigurationFetchUnitTest {
```
saralvasquez: Do we want to have a separate unit test class for this or just include it in the ConfigurationUnitTest class?

BraintreeCore/src/test/java/com/braintreepayments/api/core/ConfigurationFetchUnitTest.kt
```diff
@@ -0,0 +1,88 @@
...
+            configurationLoader.loadConfiguration(authorization)
+        } returns ConfigurationLoaderResult.Success(expectedConfiguration)
+
+        var capturedConfiguration: Configuration? = null
+        var capturedError: Exception? = null
+
+        Configuration.fetch(
+            configurationLoader,
+            authorization,
+            TestScope(UnconfinedTestDispatcher(testScheduler)),
```
saralvasquez: What does UnconfinedTestDispatcher do? I don't think I've seen that before

saralvasquez (commented):
This looks pretty good to me. I had a couple questions and I also wanna make sure we've manually confirmed that only the values we want to expose to merchants get returned in the fetch call

BraintreeCore/src/test/java/com/braintreepayments/api/core/ConfigurationFetchUnitTest.kt
noguier: Totally valid

BraintreeCore/src/test/java/com/braintreepayments/api/core/ConfigurationFetchUnitTest.kt
noguier: thank you for this comment, the UnconfinedTestDispatcher just doesn't need `advanceUntilIdle` call. But you are right we are using the regular TestDispatcher throughout the repo, so I will refactor it

noguier:
> we've manually confirmed that only the values we want to expose to merchants get returned in the fetch call

Merchants will only receive the values that were exposed to them in v4!

noguier:
Open questions: should we add a CHANGELOG? and should we add a kdoc similar to iOS: https://github.com/braintree/braintree_ios/blob/bb35d2ec3059737ea4f6262698275f7b1d80a79b/Sources/BraintreeCore/Configuration/BTConfiguration.swift#L8?

saralvasquez:
> Open questions: should we add a CHANGELOG? and should we add a kdoc similar to iOS: https://github.com/braintree/braintree_ios/blob/bb35d2ec3059737ea4f6262698275f7b1d80a79b/Sources/BraintreeCore/Configuration/BTConfiguration.swift#L8?

I think a definite yes to a CHANGELOG so that folks who have been asking about it know it's available. For the kdoc I feel like it couldn't hurt to add a suggestion not to use it unless absolutely necessary. I'm not sure about the internal use part though

jaxdesmarais:
> Open questions: should we add a CHANGELOG? and should we add a kdoc similar to iOS: https://github.com/braintree/braintree_ios/blob/bb35d2ec3059737ea4f6262698275f7b1d80a79b/Sources/BraintreeCore/Configuration/BTConfiguration.swift#L8?

Yes to kdoc, we should align since there may be a reason to remove this at some point as this API technically is not merchant facing. Yes to changelog too, you can do something like `Allow fetching of configuration (fixes #XXXX)` and it'll autolink to those GH issues.

jaxdesmarais (approved):

agedd (approved):
this looks great! 🚀

PR #1608, opened 2026-06-01
feat: Create validation use cases

noguier:
### Summary of changes
- add `CardNumberValidationUseCase` that validates card number length per detected brand & Luhn checksum
- add `ExpirationValidationUseCase` that validates MMYY format, month 01-12, and not expired. Takes currentMonth/currentYear for testability
- add `CvvValidationUseCase` that validates CVV length matches expected length for the card brand 
- move `resolveBrand()` to CardBrand companion to eliminate duplication across formatter and validation layers

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Brainstorming, and tests generation

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

@noguier


UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardNumberValidationUseCase.kt
```diff
@@ -0,0 +1,25 @@
...
+        val brand = CardBrand.resolveBrand(cardNumber)
+
+        val validLength = cardNumber.length in brand.validLengths
+        val luhnValid = if (brand == CardBrand.UNIONPAY) true else LuhnValidator.isLuhnValid(cardNumber)
+
+        return Result(
+            isValid = validLength && luhnValid,
+            detectedBrand = brand
+        )
+    }
```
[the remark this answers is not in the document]
noguier: The input views already strip non-digit characters at the input layer before this use case is ever called. Adding a second filter here would be redundant.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CvvValidationUseCase.kt
```diff
@@ -0,0 +1,8 @@
+package com.braintreepayments.api.uicomponents.cardfields
+
+internal class CvvValidationUseCase {
+
+    operator fun invoke(cvv: String, cardBrand: CardBrand): Boolean {
+        return cvv.length == cardBrand.cvvLength
+    }
```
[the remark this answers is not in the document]
noguier: Same as above, `CvvFormatter` already restricts input to digits only, and `CvvTextInputView.getRawCvv()` returns a digit-only string.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardNumberValidationUseCase.kt
```diff
@@ -0,0 +1,25 @@
...
+        }
+
+        val brand = CardBrand.resolveBrand(cardNumber)
+
+        val validLength = cardNumber.length in brand.validLengths
+        val luhnValid = if (brand == CardBrand.UNIONPAY) true else LuhnValidator.isLuhnValid(cardNumber)
+
+        return Result(
+            isValid = validLength && luhnValid,
+            detectedBrand = brand
```
buzzamus, line 22: How does this treat unknown card brands? are they considered invalid?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CvvValidationUseCase.kt
buzzamus, line 7: In a weird situation where a customer enters a card number for say, a Mastercard and then enters their cvv, but then changes to Amex how does this logic handle the change to invalid, or is that being taken care of in a later ticket?

noguier: Hi, yes, this is handled in a later ticket. Validation use cases are stateless, they will be invoked by ViewModel, so will be the re-triggering CVV validation in the scenario you described! Thank you for asking!

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardNumberValidationUseCase.kt
noguier: No, as long as `UNKNOWN` cards lengths are satisfied (for unknowns valid lengths are from 12 to 19, so very permissive) and the Luhn check returns valid `UNKNOWN` cards will also be valid

buzzamus: got it! Thank you. That is how we handle on iOS as well since we use `.unknown` for brands that we don't have the image for and want it to still pass as long as it is luhn valid and meets all the other criteria

buzzamus (approved):
Thanks for answering my questions!

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardNumberValidationUseCase.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    operator fun invoke(cardNumber: String): Result {
+        if (cardNumber.isEmpty()) {
+            return Result(isValid = false, detectedBrand = CardBrand.UNKNOWN)
+        }
+
+        val brand = CardBrand.resolveBrand(cardNumber)
+
+        val validLength = cardNumber.length in brand.validLengths
+        val luhnValid = if (brand == CardBrand.UNIONPAY) true else LuhnValidator.isLuhnValid(cardNumber)
```
saralvasquez, line 18: Just out of curiosity why is UnionPay different here?

saralvasquez (approved):
One small question but otherwise looks good!

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardNumberValidationUseCase.kt
noguier: UnionPay is the only card brand that does not specifically follow the Luhn Algorithm when constructing the card numbers, so we have to exclude it from validating against Luhn.

PR #1610, opened 2026-06-03
Create card fields view model

noguier:
### Summary of changes
Introduce the internal ViewModel that orchestrates card number, expiration, and CVV validation and exposes observable state for the container view to consume
- new ViewModel observes field changes (`onCardNumberChanged`,`onExpiryChanged`, `onCvvChanged`) and focus changes
- exposes StateFlow<ValidationResult> per field, plus `detectedCardBrand` and `isFormValid`
- Implements `suppressErrorDuringTyping()` so errors never appear while editing, only Valid state propagates immediately; everything else downgrades to Validating (matches iOS)
- Implements `resolveOnBlur() `so incomplete fields resolve to Invalid(required) when empty or Invalid(error) when partially filled
- Revalidates CVV when the detected brand changes

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Code organization brainstorming, test coverage

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

@noguier


UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsViewModel.kt
```diff
@@ -0,0 +1,164 @@
...
+            CardField.CARD_NUMBER -> {
+                val result = cardNumberValidationUseCase(currentCardNumber)
+                _detectedCardBrand.value = result.detectedBrand
+                _cardNumberValidation.value = resolveOnBlur(
+                    useCaseResult = result.validation,
+                    isEmpty = currentCardNumber.isEmpty(),
+                    requiredRes = R.string.card_number_required,
+                    errorRes = R.string.card_number_error
+                )
+            }
```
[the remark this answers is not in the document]
noguier: I don't think the stale-CVV scenario is reachable in the current design, so I'll leave the blur path as is. `currentCardNumber` is only ever written in `onCardNumberChanged`, which runs on every keystroke and already updates `detectedCardBrand` and calls `revalidateCvv()` when the brand changes.

saralvasquez:
Was this tested at all in the demo app?

noguier:
> Was this tested at all in the demo app?

No, because it is just logic, without a UI. UI will tested when CardField class will be created!

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsViewModel.kt
```diff
@@ -0,0 +1,164 @@
...
+        @StringRes errorRes: Int
+    ): ValidationResult = when {
+        useCaseResult is ValidationResult.Valid -> useCaseResult
+        useCaseResult is ValidationResult.Invalid -> useCaseResult
+        isEmpty -> ValidationResult.Invalid(requiredRes)
+        else -> ValidationResult.Invalid(errorRes)
+    }
+
+    /** Sets [isFormValid] to true when each of the input fields are valid*/
+    private fun updateIsValid() {
```
saralvasquez, line 159: This makes sense to me, but what will this value be used for? I assume this will become important during a later PR?

saralvasquez (commented):
Looks good just a couple questions!

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsViewModel.kt
noguier: Thank you for this question, the idea is that this value will be broadcasted to merchants. So that they can enable/disable their pay button, so that we do not attempt to "tokenize" on invalid card info :)

saralvasquez: Ah makes sense!

saralvasquez (approved):
🎉

UIComponents/src/test/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsViewModelUnitTest.kt
```diff
@@ -0,0 +1,324 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+class CardFieldsViewModelUnitTest {
+
+    private fun createViewModel() = CardFieldsViewModel(
+        currentDateProvider = { Pair(6, 2026) }
+    )
```
buzzamus, line 13: Is this date actually being used anywhere for testing, or does it not matter and is strictly for creating the view model?

The reason I ask is I don't know enough about Android testing libraries to know if this will be a problem or not, but on iOS I initially did this, but then realized there are certain situations where setting the date to an actual date will cause issues in the future vs. using the actual current date (like `Date.now()` or whatever it would be in Kotlin.

noguier: so this is the date only used in tests, in the production code we are using 
`private val currentDateProvider: () -> Pair<Int, Int> = {
        val cal = Calendar.getInstance()
        Pair(cal.get(Calendar.MONTH) + 1, cal.get(Calendar.YEAR))
    }`
 Which will correctly pick the current date :)

jaxdesmarais: I think Brent's comment is around, say next month July(7) 2026, these will start failing since the date should be invalid because June will be in the past. So we want to build it to always iterate on the current date in our tests.

buzzamus: Yes, I worded it poorly, but that was the issue I was trying to relay

noguier: ah, I see, well these tests only test June 2026 agains 03/25 (which always be expired), and against 12/28 which will always be in the future for fixed June 2026. I can switch to real date, though, and have the future date like 12/99?

buzzamus (approved):

PR #1615, opened 2026-06-09
feat: Create CardFields class

noguier:
### Summary of changes
- Implements `CardFields` merchant facing class
- Inflates three views ( cardNumber, expiration, cvv)
- Wires the three views to the `CardFieldsViewModel` for state management
- Adds the full single component to demo
- Exposes `CardFields.setOnValidationChangeListener` for merchant to enable/disable pay button when the card becomes valid

https://github.com/user-attachments/assets/4803b352-8f39-4e5e-bae3-1ed0909cc01a

### AI Usage

**Which AI Agent Was Used?**
- [x] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Brainstorming, line completion, test generation

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

@noguier


UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardFields.kt
```diff
@@ -0,0 +1,142 @@
...
+        scope.launch {
+            viewModel.detectedCardBrand.collect { brand ->
+                cvvView.updateCardBrand(brand)
+            }
+        }
+        scope.launch {
+            viewModel.isFormValid.collect { isValid ->
+                validationChangedListener?.onValidationChanged(isValid)
+            }
+        }
```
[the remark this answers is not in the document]
noguier: fix it by dropping initial StateFlow emission with .drop(1) at 9d9cba7

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardFields.kt
```diff
@@ -0,0 +1,142 @@
...
+
+    fun interface OnValidationChangedListener {
+        fun onValidationChanged(isFormValid: Boolean)
+    }
+
+    fun setOnValidationChangedListener(listener: OnValidationChangedListener?) {
+        validationChangedListener = listener
+        listener?.onValidationChanged(viewModel.isFormValid.value)
+    }
+
```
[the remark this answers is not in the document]
noguier: fix it by dropping initial StateFlow emission with .drop(1) at 9d9cba7

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardFields.kt
```diff
@@ -0,0 +1,142 @@
...
+        attrs: AttributeSet? = null,
+        defStyleAttr: Int = 0
+    ) : this(context, attrs, defStyleAttr, CardFieldsViewModel())
+
+    private val cardNumberView: CardNumberTextInputView
+    private val expirationView: ExpirationTextInputView
+    private val cvvView: CvvTextInputView
+    private var observerScope: CoroutineScope? = null
+    private var validationChangedListener: OnValidationChangedListener? = null
+
```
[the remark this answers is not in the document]
noguier: fix it by dropping initial StateFlow emission with .drop(1) at 9d9cba7

saralvasquez:
The style of the `Pay` button looks a little different in figma. Seems like it should be black and more rounded around the corners. Is that something for this story or is that coming later?

Demo/src/main/java/com/braintreepayments/demo/CardFieldsFragment.kt
```diff
@@ -18,9 +18,19 @@ class CardFieldsFragment : Fragment() {
...
+        val payButton = view.findViewById<Button>(R.id.pay_button)
+
+        // Card fields will tell the merchant if the card is valid, and they can enable the button
+        cardFields.setOnValidationChangedListener { isFormValid ->
+            payButton.isEnabled = isFormValid
+        }
+
+        // placeholder for tokenization
+        payButton.setOnClickListener {
+            Toast.makeText(requireContext(), R.string.card_fields_valid_toast, Toast.LENGTH_SHORT)
```
saralvasquez, line 31: So this will be removed when we hook up the back end processing?

noguier:
> The style of the Pay button looks a little different in figma. Seems like it should be black and more rounded around the corners. Is that something for this story or is that coming later?

I planned on addressing all the cosmetic changes to align 100% to figma in the Demo app ticket later. Right now i think it is more important to focus on functionality. But i can change it now if you think it is better?

buzzamus:
> > The style of the Pay button looks a little different in figma. Seems like it should be black and more rounded around the corners. Is that something for this story or is that coming later?
> 
> I planned on addressing all the cosmetic changes to align 100% to figma in the Demo app ticket later. Right now i think it is more important to focus on functionality. But i can change it now if you think it is better?

The merchant will actually provide and be in charge of their own pay button, so the one in the designs I think is just from a design standpoint on what it would have been if we did decide to own the button.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardFields.kt
```diff
@@ -0,0 +1,144 @@
...
+        scope.launch {
+            viewModel.detectedCardBrand.collect { brand ->
+                cvvView.updateCardBrand(brand)
+            }
+        }
+        scope.launch {
+            // drop(1) discards State Flow's replayed first value
+            viewModel.isFormValid.drop(1).collect { isValid ->
+                validationChangedListener?.onValidationChanged(isValid)
+            }
```
saralvasquez, line 111: I'm having a hard time understanding what this is doing. Could you please explain what this is needed for please? Thanks!

saralvasquez (approved):
This looks good to me! Great job! I just had a couple non-blocking questions

Demo/src/main/java/com/braintreepayments/demo/CardFieldsFragment.kt
noguier: yes, that's how the card flow will start :)

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardFields.kt
noguier: Great question it took me a little bit to wrap my head around it too.  
First, the setter stores the `validationChangedListener` and immediately calls it once with the current validity, so the merchant can put their Pay button in the right state right away. Then a StateFlow collector calls that same listener again every time validity actually changes (user types, fields become valid).
In case there is a time overlap for the two calls the merchant will get the same value twice. So we need to drop the first `validationChangedListener` callback value, so that they do not collide with each other. This is what drop(1) is for. So the setter owns __only__ the initial value, and the state collector get updated with __real__ state changes.

buzzamus (approved):
Looks good! Just had one quick question:

iOS does not add any margins to the width to the actual CardFields component itself, so when looking at just the component view it takes up 100% of the width of the device. This is to allow merchants to do what they want (either leave it this way or to add padding via their container view logic).

In our Demo app we add similar spacing to what is in your video, but not to the actual component view itself. I just wanted to clarify if that is what Android is planning on doing as well?

noguier:
@buzzamus great suggestion, I will add it as a new acceptance criteria to the Demo app ticket!

Issue #1572, opened 2026-04-15
Fatal Exception: java.lang.NoSuchMethodError


dcampogiani:
### Braintree SDK Version

5.25.0

### Environment

Both

### Android Version & Device

Galaxy S24 Android 16

### Braintree dependencies

braintree-card = "com.braintreepayments.api:card:5.25.0"
braintree-core = "com.braintreepayments.api:braintree-core:5.25.0"
braintree-dataCollector = "com.braintreepayments.api:data-collector:5.25.0"
braintree-dropIn = "com.braintreepayments.api:drop-in:6.17.0"
braintree-googlePay = "com.braintreepayments.api:google-pay:5.25.0"
braintree-payPal = "com.braintreepayments.api:paypal:5.25.0"
braintree-threeDSecure = "com.braintreepayments.api:three-d-secure:5.25.0"

### Describe the bug

On 9 April, Google released "com.google.android.gms:play-services-wallet:20.0.0", while trying to upgrade to the latest version I'm getting a crash on Braintree code.

```
Fatal Exception: java.lang.NoSuchMethodError: No static method resolveTask(Lcom/google/android/gms/tasks/Task;Landroid/app/Activity;I)V in class Lo/getRegisteredCallbackCookie; or its super classes (declaration of 'o.getRegisteredCallbackCookie' appears in /data/app/~~W0x6HpKew9cXhoanbzJT8w==/it.prima.staging-6jwumtI2ct_zXyvEFHsjRA==/base.apk!classes2.dex)
       at com.braintreepayments.api.googlepay.GooglePayActivity.onCreate(GooglePayActivity.kt:35)
       at android.app.Activity.performCreate(Activity.java:9519)
       at android.app.Activity.performCreate(Activity.java:9488)
       at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1524)
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:4692)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:4941)
       at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:222)
       at android.app.servertransaction.TransactionExecutor.executeNonLifecycleItem(TransactionExecutor.java:133)
       at android.app.servertransaction.TransactionExecutor.executeTransactionItems(TransactionExecutor.java:103)
       at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:80)
[...]

PR #1616, opened 2026-06-11
fix: remove deprecated GooglePay functionality in favour of TaskResultContracts

saralvasquez:
### Summary of changes
This ticket came in as a part of a [github issue](https://github.com/braintree/braintree_android/issues/1572). The problem was that we were still using a depreacted method that was later removed in newer versions of google wallet in the google pay component. The newer recommended method of gathering payment data from google pay allows for pretty big changes to the SDK. This includes getting rid of `GooglePayActivity` and `GooglePayActivityResultContract` since the result handling is covered by the new `TaskResultContracts`. More detailed summary below with documentation from google used to put this PR together

  - Replaces the internal GooglePayActivity and AutoResolveHelper pattern with `TaskResultContracts.GetPaymentDataResult()` API from Google Wallet                                                                                                                                                   
  - Merchants no longer need to declare GooglePayActivity in their AndroidManifest.xml — the SDK handles the payment sheet launch entirely through the activity result contract mechanism                                                                                                                                                   
  - Moves loadPaymentData() into GooglePayInternalClient so GooglePayLauncher drives the task directly, and result parsing is now handled inline in the launcher callback

Google documentation: 
https://developers.google.com/android/reference/com/google/android/gms/wallet/contract/TaskResultContracts
https://developers.google.com/pay/api/android/guides/tutorial#initiate-payment
Sample integration pattern: 
https://github.com/google-pay/android-quickstart/blob/468fc42e6d465e1b296bf2441835c3f053ef2c4b/java/app/src/main/java/com/google/android/gms/samples/pay/activity/CheckoutActivity.java

Note on deleted files - these changes feel like they are on the border of a major version change but all the changes are to internal files only so integrations shouldn't need to change. Let me know what you all think

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used claude code to research sources for replacements for the deprecated method and for suggestions on how to integrate the new preferred methods into our code base. Let it generate tests to cover the changed functionality

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry

### Authors
- @saralvasquez


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* GooglePay
+    * Remove requirement for `GooglePayActivity` to be declared in the Android manifest
```
jaxdesmarais: nit: we can add `(fixes #xxxx) to link it automatically to the GH issue

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayClient.kt
```diff
@@ -635,7 +619,10 @@ class GooglePayClient internal constructor(
     }
 
     companion object {
+        @Deprecated("This constant is no longer used and will be removed in a future release.")
```
jaxdesmarais: Take it or leave it, do we want to add anymore details to why this can be removed or is this fine? I am not sure the standard format of deprecation messages on Android.

jaxdesmarais: I know sometimes too we add a `NEXT_MAJOR_VERSION` note so we don't forget to actually remove it. 😄

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayInternalClient.kt
```diff
@@ -41,6 +42,14 @@ internal class GooglePayInternalClient {
         }
     }
 
+    fun loadPaymentData(context: Context, params: GooglePayPaymentAuthRequestParams): Task<PaymentData> {
```
jaxdesmarais, line 45: Is this public or internal because the class is internal?

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayLauncher.kt
```diff
@@ -1,27 +1,50 @@
...
     lifecycleOwner: LifecycleOwner,
+    context: Context,
+    private val internalGooglePayClient: GooglePayInternalClient = GooglePayInternalClient(),
     callback: GooglePayLauncherCallback
 ) {
 
-    private val activityLauncher: ActivityResultLauncher<GooglePayPaymentAuthRequestParams> = registry.register(
+    private val appContext: Context = context.applicationContext
+
+    private val activityLauncher: ActivityResultLauncher<Task<PaymentData>> = registry.register(
```
jaxdesmarais, line 28: It seems like this is the crux of the change. Are the any implications to older APIs or is this fully backwards compatible?

jaxdesmarais (commented):
I mainly just have general questions around my own understanding of Android and the new API. Overall it looks like this change cleans up a lot of old code which is a win in my book. It also looks to be non breaking, so seems like a good change to make.

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayClient.kt
saralvasquez: Ah yeah I can definitely add the NMV note and I'll rework the wording a bit

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayInternalClient.kt
saralvasquez: I think it should be internal since the class is but I'll double check

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayLauncher.kt
saralvasquez: Great question. I tested with older versions of the APIs on emulator and it worked just fine but I'll make sure to test on all supported API versions to make sure we won't run into issues

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayLauncher.kt
```diff
@@ -1,27 +1,50 @@
...
+        val result = when {
+            apiTaskResult.status.isSuccess ->
+                GooglePayPaymentAuthResult(apiTaskResult.result, null)
+            apiTaskResult.status.isCanceled ->
+                GooglePayPaymentAuthResult(null, UserCanceledException("User canceled Google Pay."))
+            else ->
+                GooglePayPaymentAuthResult(
+                    null,
+                    GooglePayException(
+                        "An error was encountered during the Google Pay " +
```
noguier, line 41: is there a way we can display the status object in our demo app, so that we can see what caused an error right away?

noguier (commented):
While i am still going through all the code changes, I just wanted to confirm that i tested the flow on the physical device, and API 23, and all are working as expected.

noguier:
I see the `EXTRA_ENVIRONMENT` and `EXTRA_PAYMENT_DATA_REQUEST` constants are being deprecated rather than deleted, but GooglePayActivity is being deleted. In what situations do we usually deprecate first and then remove, vs remove right away like the `GooglePayActivity`? Since we are changing the manifest, will it affect merchants with old API?

saralvasquez:
Great question! The general rule is that if the merchant can touch a piece of code you can't delete it until a major version. Those constants are technically publicly accessible by merchants. They shouldn't/wouldn't ever have used them for anything, but since they could be accessed it's safer to deprecate than completely remove them. The files that were deleted on the other hand were all internal so there is no chance that a merchant would be using any piece of them and are safe to completely delete. After some searching it looks like the changes that Google made to wallet that allowed this PR are backwards compatible all the way to version 21 which we don't even support so we should be fine for all API version we support in the SDK

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayLauncher.kt
saralvasquez: This is actually already covered in the demo app. This PR doesn't change how results are getting delivered, it only changes what's delivering them. If you take a look at the delete results contract file you can see that the handling of the payment status and sending of `GooglePayPaymentAuthResult` are very similar to how it looks in the Launcher now

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayInternalClient.kt
saralvasquez: Yeah confirmed it's internal since the class is

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayLauncher.kt
noguier: Thank you very much, I can see it now!

noguier:
This looks good, just a couple of clarifying questions, to get my understanding: 
Prior to this changes, we had to create a new activity, that was launching Google Pay, and then handles the result back with `onActivityResult`. Now we do not have to do it because `ActivityResultContract` solves the problem of creating an "invisible" just to handle results because we can use `TaskResultContracts` to launch and deliver results?

noguier:
Also from what I have read, and tested this changes are great and are backwards compatible!

noguier (approved):

saralvasquez:
> This looks good, just a couple of clarifying questions, to get my understanding: Prior to this changes, we had to create a new activity, that was launching Google Pay, and then handles the result back with `onActivityResult`. Now we do not have to do it because `ActivityResultContract` solves the problem of creating an "invisible" just to handle results because we can use `TaskResultContracts` to launch and deliver results?

Yeah exactly! They basically took on the responsibility of launching and delivering results so we don't need to do all the weird handling with the transparent activity or custom results contracts etc. I think you've got a firm understanding of the changes

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayLauncher.kt
saralvasquez: Ok so I couldn't get some of the older emulators to work well enough to test, but we do have confirmation that the new API is backwards compatible back to versions that are so old we don't even support. So I think we should be good

jaxdesmarais (approved):
Thanks for addressing all of my questions/comments. These changes look great and way less code for us to maintain long term. 🚀

PR #1620, opened 2026-06-16
Add UI Tests for CardFields

buzzamus:
### Summary of changes
- Adds UI Tests for CardFields UI

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used Claude to write iOS UI tests into Kotlin using Android libraries.

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsTest.kt
+++ b/UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsTest.kt
@@ -1,0 +1,305 @@
+package com.braintreepayments.api.uicomponents.cardfields
+
+import android.content.Intent
+import android.view.View
+import android.widget.EditText
+import android.widget.FrameLayout
+import android.widget.ImageView
+import android.widget.TextView
+import androidx.test.core.app.ActivityScenario
+import androidx.test.internal.runner.junit4.AndroidJUnit4ClassRunner
+import androidx.test.platform.app.InstrumentationRegistry
+import com.braintreepayments.api.core.BraintreeException
+import com.braintreepayments.api.uicomponents.R
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4ClassRunner::class)
+class CardFieldsTest {
+
+    private val instrumentation = InstrumentationRegistry.getInstrumentation()
+
+    private fun launchActivity(): ActivityScenario<CardFieldsTestActivity> {
+        val intent = Intent(
+            InstrumentationRegistry.getInstrumentation().targetContext,
+            CardFieldsTestActivity::class.java
+        )
+        return ActivityScenario.launch(intent)
+    }
+
+    private fun waitForMain() = instrumentation.waitForIdleSync()
+
+    private fun CardFields.cardNumberEditText(): EditText =
+        findViewById<CardNumberTextInputView>(R.id.card_fields_card_number_input)
+            .findViewById(R.id.text_input_edit_text)
+
+    private fun CardFields.expirationEditText(): EditText =
+        findViewById<ExpirationTextInputView>(R.id.card_fields_expiration_input)
+            .findViewById(R.id.text_input_edit_text)
+
+    private fun CardFields.cvvEditText(): EditText =
+        findViewById<CvvTextInputView>(R.id.card_fields_cvv_input)
+            .findViewById(R.id.text_input_edit_text)
+
+    private fun CardFields.cardNumberErrorLabel(): TextView =
+        findViewById<CardNumberTextInputView>(R.id.card_fields_card_number_input)
+            .findViewById(R.id.error_label)
+
+    private fun CardFields.expirationErrorLabel(): TextView =
+        findViewById<ExpirationTextInputView>(R.id.card_fields_expiration_input)
+            .findViewById(R.id.error_label)
+
+    private fun CardFields.cvvErrorLabel(): TextView =
+        findViewById<CvvTextInputView>(R.id.card_fields_cvv_input)
+            .findViewById(R.id.error_label)
+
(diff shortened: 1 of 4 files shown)


noguier:
Hi, would please update the branch with the current state of `card-fields-feature`? Also, would you please make sure new UI test run in CI? -- nevermind the last question, since they are under UIComponents subdirectory, they should automatically run in CI! 👍

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsTest.kt
```diff
@@ -0,0 +1,305 @@
...
+
+    private fun launchActivity(): ActivityScenario<CardFieldsTestActivity> {
+        val intent = Intent(
+            InstrumentationRegistry.getInstrumentation().targetContext,
+            CardFieldsTestActivity::class.java
+        )
+        return ActivityScenario.launch(intent)
+    }
+
+    private fun waitForMain() = instrumentation.waitForIdleSync()
```
noguier, line 33: out of curiosity, what do we need it for?

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsTest.kt
```diff
@@ -0,0 +1,305 @@
...
+            }
+        }
+    }
+
+    @Test
+    fun cardFields_payButton_isEnabledAfterAllFieldsAreValid() {
+        launchActivity().use { scenario ->
+            scenario.onActivity { activity ->
+                activity.cardFields.cardNumberEditText().setText("4111111111111111")
+                activity.cardFields.expirationEditText().setText("1245")
```
noguier, line 85: is 12/45 is the expiration date you use in iOS too?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardFields.kt
```diff
@@ -27,9 +27,11 @@ class CardFields internal constructor(
     defStyleAttr: Int = 0,
     private val viewModel: CardFieldsViewModel = CardFieldsViewModel(),
     private var cardClient: CardClient? = null,
-    private val analyticsClient: AnalyticsClient = AnalyticsClient.lazyInstance.value
+    private val analyticsClient: AnalyticsClient? = null
 ) : FrameLayout(context, attrs, defStyleAttr) {
 
+    private fun getAnalyticsClient(): AnalyticsClient = analyticsClient ?: AnalyticsClient.lazyInstance.value
```
noguier, line 33: Are these changes new or leftover from the analytics work you have done?

noguier:
This looks great, thank you for creating a test activity for easy testing! Just a couple of clarifying questions above 😄 
Approving ahead of time since the questions are non-blocking 👍

noguier (approved):

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsTest.kt
buzzamus: This is to avoid a potential race condition between the main and test thread: it lets the main thread complete the state change before allowing test thread to check the UI status

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsTest.kt
buzzamus: Yep! Unlike unit tests where we have more control this just uses the latest possible valid expiration date, so we have 19 years before this test goes bad.

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsTest.kt
noguier: ahhh, fun!

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsTest.kt
noguier: Great, thank you!

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardFields.kt
buzzamus: They are new in order to allow the analytics client to be lazily loaded in order to avoid an issue where `initialize()` isn't always called and gave an error of the analyticsClient being null

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsTest.kt
```diff
@@ -0,0 +1,305 @@
...
+            }
+        }
+    }
+
+    // endregion
+
+    // region CVV Length Revalidation
+
+    @Test
+    fun cardFields_showsCvvError_whenSwitchingFromVisaToAmex() {
```
saralvasquez, line 258: Just a thought. Would it make sense to have a test for the other direction as well, so the Amex 4 digit -> visa 3 digit?

saralvasquez (approved):
Looks good! Just a small thought but not blocking

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsTest.kt
buzzamus: This is a good idea! In fact, I went to add it and it fails due to a bug (we had a very similar one on iOS side), so going to merge this and then create a follow up ticket to address bug

PR #1622, opened 2026-06-22
feat: Merge PaymentButtons and CardFields under one screen in the demo app

noguier:
### Summary of changes
- Make PaymentButtons and CardFields appear under one Demo fragment to match iOS
- Merge individual xml files into shared xml file for PaymentButtons and CardFields
- Fix tests affected by this change
- Remove CardFields and PaymentButtons individual fragments
- Fix navigation screen

https://github.com/user-attachments/assets/a6752679-e3e7-4a06-b68c-3fd780a83f34

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used to fix navigation, tests and make toggle color buttons pretty. 
- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

@noguier


buzzamus (approved):
Looks good to me! Thanks for updating this

Demo/src/main/java/com/braintreepayments/demo/UIComponentsFragment.kt
```diff
@@ -0,0 +1,268 @@
...
+            false,
+            false
+        )
+
+        payPalButton.initialize(
+            this,
+            super.getAuthStringArg(),
+            "https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com/braintree-payments".toUri(),
+            "com.braintreepayments.demo.braintree"
+        )
```
mzlangreder, line 138: Maybe we can add property names here so it's more clear what we're setting up?

Demo/src/main/java/com/braintreepayments/demo/UIComponentsFragment.kt
```diff
@@ -0,0 +1,268 @@
...
+                storePayPalPendingRequest(request)
+            } else if (request is PayPalPendingRequest.Failure) {
+                handleError(request.error)
+            }
+        }
+    }
+
+    private fun setupVenmoButton() {
+        val activity = requireActivity()
+        activity.setProgressBarIndeterminateVisibility(true)
```
mzlangreder: NIT: I don't see activity used anywhere else in the method, we can drop the property and just use the value directly.
```suggestion
        requireActivity().setProgressBarIndeterminateVisibility(true)
```

Demo/src/main/java/com/braintreepayments/demo/UIComponentsFragment.kt
```diff
@@ -0,0 +1,268 @@
...
+            "com.braintreepayments.demo.braintree"
+        )
+
+        venmoButton.setVenmoRequest(venmoRequest)
+        venmoButton.venmoLaunchCallback = VenmoLaunchCallback { request: VenmoPendingRequest? ->
+            if (request is VenmoPendingRequest.Started) {
+                storeVenmoPendingRequest(request)
+            } else if (request is VenmoPendingRequest.Failure) {
+                handleError(request.error)
+            }
```
mzlangreder, line 182: This would be better as an exhaustive `when` condition so that we get compiler safety in case we add more sub-types to `VenmoPendingRequest`. This also exposed a missing case that could be handled for the `null` condition.

```suggestion
            when (request) {
                is VenmoPendingRequest.Started -> {
                    storeVenmoPendingRequest(request)
                }

                is VenmoPendingRequest.Failure -> {
                    handleError(request.error)
                }

                null -> {
                    // What should be done here if the request is null?
                }
            }
```

Demo/src/main/java/com/braintreepayments/demo/UIComponentsFragment.kt
```diff
@@ -0,0 +1,268 @@
...
+            "https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com/braintree-payments".toUri(),
+            "com.braintreepayments.demo.braintree"
+        )
+        payPalButton.setPayPalRequest(payPalRequest)
+        payPalButton.payPalLaunchCallback = PayPalLaunchCallback { request: PayPalPendingRequest? ->
+            if (request is PayPalPendingRequest.Started) {
+                storePayPalPendingRequest(request)
+            } else if (request is PayPalPendingRequest.Failure) {
+                handleError(request.error)
+            }
```
mzlangreder, line 144: Same comment as [here](https://github.com/braintree/braintree_android/pull/1622/changes#r3461134861)

Demo/src/main/java/com/braintreepayments/demo/UIComponentsFragment.kt
```diff
@@ -0,0 +1,268 @@
...
+
+    override fun onResume() {
+        super.onResume()
+
+        // PayPal flow after returning to app
+        val pendingRequest = getPayPalPendingRequest()
+        if (pendingRequest != null) {
+            payPalButton.handleReturnToApp(
+                pendingRequest, requireActivity().intent,
+                PayPalTokenizeCallback { payPalResult: PayPalResult? ->
```
mzlangreder: NIT: The SAM constructor here is redundant and could also use trailing lambda syntax and move it outside of the parenthesis.

But, maybe the callback type here works better for documentation?

Demo/src/main/java/com/braintreepayments/demo/UIComponentsFragment.kt
```diff
@@ -0,0 +1,268 @@
...
+            requireActivity().intent.data = null
+        }
+
+        // Venmo flow after returning to app
+        val venmoPendingRequest = getVenmoPendingRequest()
+        if (venmoPendingRequest != null) {
+            venmoButton.handleReturnToApp(
+                venmoPendingRequest,
+                requireActivity().intent,
+                VenmoTokenizeCallback { venmoResult: VenmoResult? ->
```
mzlangreder: Same comment as [this one](https://github.com/braintree/braintree_android/pull/1622/changes#r3461215118)

mzlangreder (approved):

Demo/src/main/java/com/braintreepayments/demo/UIComponentsFragment.kt
noguier: ahhh i see what you are saying it was leftover from the payments buttons (i had to convert from java to kotlin), but i think your suggestion makes total sense!

Demo/src/main/java/com/braintreepayments/demo/UIComponentsFragment.kt
noguier: will change it here as well

noguier: changed in d98ec41

Demo/src/main/java/com/braintreepayments/demo/UIComponentsFragment.kt
noguier: changed in d98ec41

Demo/src/main/java/com/braintreepayments/demo/UIComponentsFragment.kt
noguier: changed in d98ec41

Demo/src/main/java/com/braintreepayments/demo/UIComponentsFragment.kt
noguier: changed in d98ec41

Demo/src/main/java/com/braintreepayments/demo/UIComponentsFragment.kt
noguier: the createPayPalCheckoutRequest is a java function, so I couldn't add a property name to it, but I added the IDE comments for better readability: d98ec41

PR #1624, opened 2026-06-24
Fix cvv length issue

buzzamus:
### Summary of changes
- This is a similar issue that iOS encountered [in this PR](https://github.com/braintree/braintree_ios/pull/1822) where if the CVV is already entered and the card type changes to a type with a different length CVV then the error doesn't trigger unless the field is specifically focused and then moved off focus. This PR fixes that bug by re-triggering the error check:

[Screen_recording_20260624_114145.webm](https://github.com/user-attachments/assets/938329a0-e5f2-48d1-a27a-d4487b060a95)

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Had Claude analyze my iOS PR and then check for same issue on Android to get my unit test to pass that was created to test for this specific scenario

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsTest.kt
+++ b/UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsTest.kt
@@ -284,6 +284,37 @@ class CardFieldsTest {
         }
     }
 
+    @Test
+    fun cardFields_showsCvvError_whenSwitchingFromAmexToVisa() {
+        launchActivity().use { scenario ->
+            // Enter an Amex card with a valid 4-digit CVV
+            scenario.onActivity { activity ->
+                activity.cardFields.cardNumberEditText().setText("378282246310005")
+                activity.cardFields.cvvEditText().also { cvv ->
+                    cvv.requestFocus()
+                    cvv.setText("1234")
+                }
+            }
+            waitForMain()
+            // Switch to Visa — brand changes, 4-digit CVV is now too long (Visa requires 3)
+            scenario.onActivity { activity ->
+                activity.cardFields.cardNumberEditText().setText("4111111111111111")
+            }
+            waitForMain()
+            // Blur CVV to trigger validation against the new brand
+            scenario.onActivity { activity ->
+                activity.cardFields.expirationEditText().requestFocus()
+            }
+            waitForMain()
+            scenario.onActivity { activity ->
+                val errorLabel = activity.cardFields.cvvErrorLabel()
+                assertTrue(errorLabel.visibility == View.VISIBLE)
+                assertTrue(errorLabel.text.isNotEmpty())
+            }
+        }
+    }
+
+
     // endregion
 
     // region Submit before initialize
--- a/UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsTest.kt
+++ b/UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsTest.kt
@@ -343,6 +343,28 @@ class CardFieldsTest {
         }
     }
 
+    @Test
+    fun cardFields_payButton_remainsDisabled_afterConfigurationChange_whenFormIsInvalid() {
+        launchActivity().use { scenario ->
+            scenario.onActivity { activity ->
+                activity.cardFields.cardNumberEditText().setText("4111111111111111")
+                activity.cardFields.expirationEditText().setText("1245")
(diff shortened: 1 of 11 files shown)


jaxdesmarais (approved):
Thanks for adding the demo video and all the tests. These changes look good.

noguier:
would you please rebase this branch with the current `card-fields-feature`, since the demo UI changed, and I want to make sure your fix works with the current state of the branch :)

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsViewModel.kt
```diff
@@ -122,7 +122,13 @@ internal class CardFieldsViewModel(
...
     private fun revalidateCvv() {
         val result = cvvValidationUseCase(currentCvv, _detectedCardBrand.value)
-        suppressErrorDuringTyping(_cvvValidation, result)
+        if (_cvvValidation.value is ValidationResult.Valid && result is ValidationResult.Validating) {
+            _cvvValidation.value = ValidationResult.Invalid(R.string.cvv_error)
+        } else if (_cvvValidation.value is ValidationResult.Valid) {
+            _cvvValidation.value = result
+        } else {
+            suppressErrorDuringTyping(_cvvValidation, result)
+        }
```
noguier, line 131: i was wondering if we can use more Kotlin idiomatic `when` statement here?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/cardfields/CvvTextInputView.kt
```diff
@@ -34,13 +34,8 @@ internal class CvvTextInputView @JvmOverloads constructor(
...
     internal fun updateCardBrand(brand: CardBrand) {
         formatter.updateCvvLength(brand.cvvLength)
-        val currentCvv = getRawCvv()
-        if (currentCvv.length > brand.cvvLength) {
-            setText(currentCvv.take(brand.cvvLength))
-        } else {
-            editText.filters = (editText.filters.filterNot { it is InputFilter.LengthFilter } +
-                InputFilter.LengthFilter(brand.cvvLength)).toTypedArray()
-        }
+        editText.filters = (editText.filters.filterNot { it is InputFilter.LengthFilter } +
```
noguier, line 37: why do we no longer need if/else branching here?

noguier:
Hi Brent, while testing your changes, I noticed another edge case. 
Note: this is 99% not the bug caused by your PR, I should have checked that edge case in [this](https://github.com/braintree/braintree_android/pull/1621) PR.  Your bug fix looks great!
But here is what is happening: 
User types in good CC info, pay button is enabled. 
User rotates the phone --> pay button gets disables and **stays disabled** until you actually make a change to any of the input fields. So even if the CC info is still correct, the user has to go and manually do the change to the input field, to re-trigger the validation. 

Here is the video: 


https://github.com/user-attachments/assets/6acf3016-59c2-4001-bb21-faca702b7e37


Will you please consider looking into it as a part of this pr? Since we are waiting on the instrumentation tests fix anyways?

buzzamus:
Video of fix for submit button state not staying valid when phone rotated:

[Screen_recording_20260626_094729.webm](https://github.com/user-attachments/assets/51a6bed6-808d-477c-afda-66015ebef1ac)

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/cardfields/CardFieldsTest.kt
```diff
@@ -316,6 +316,35 @@ class CardFieldsTest {
 
     // endregion
 
+    // region Configuration change (rotation)
+
+    @Test
+    fun cardFields_payButton_remainsEnabled_afterConfigurationChange() {
```
noguier, line 322: should we create a "sister" test that ensures that the button is not being enabled on rotation change if it rotates with the "invalid" CardFields form?

buzzamus: good idea! Just added

noguier (approved):
Thank you for addressing both bug fixes! This looks great

buzzamus:
Merging in PR with red CI due to current sandbox issues in 3DS. Confirmed all tests failing are due to the ongoing 3DS sandbox issues.

PR #1635, opened 2026-07-09
feat: add instrumentation tests support for android 17

noguier:
### Summary of changes
- add a new job (`test_android_17`) to try and run out instrumentation tests with the new android 17 api - i decided to make it and keep it as a separate job since it needs extra set up for the emulator to boot correctly, but I am open to suggestions
- bump the versions of `androidxJunit`, `androidxTest`, `espresso` to make tests pass on the emulator that uses api 37

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [ ] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
NA

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

@noguier

### DIFF
--- a/.github/workflows/instrumentation_tests.yml
+++ b/.github/workflows/instrumentation_tests.yml
@@ -49,7 +49,7 @@ jobs:
       fail-fast: false
       matrix:
         api-level: [ "37.0" ]
-        target: [ google_apis_ps16k ]
+        target: [ playstore_ps16k ]
     steps:
       - name: checkout
         uses: actions/checkout@v4
@@ -69,6 +69,14 @@ jobs:
           target: ${{ matrix.target }}
           arch: x86_64
           profile: pixel_7_pro
+          emulator-options: >
+            -no-window
+            -gpu auto
+            -noaudio
+            -no-boot-anim
+            -camera-back none
+            -memory 4096
+            -feature GLDirectMem,HasSharedSlotsHostMemoryAllocator
           script: ./gradlew :BraintreeCore:connectedCheck :Card:connectedCheck :SharedUtils:connectedCheck :LocalPayment:connectedCheck :DataCollector:connectedCheck :GooglePay:connectedCheck :Demo:connectedCheck :PayPal:connectedCheck :PayPalMessaging:connectedCheck :AmericanExpress:connectedCheck :SEPADirectDebit:connectedCheck :ShopperInsights:connectedCheck :ThreeDSecure:connectedCheck :UIComponents:connectedCheck :Venmo:connectedCheck --stacktrace
 
       - name: retry tests
@@ -79,4 +87,12 @@ jobs:
           target: ${{ matrix.target }}
           arch: x86_64
           profile: pixel_7_pro
+          emulator-options: >
+            -no-window
+            -gpu auto
+            -noaudio
+            -no-boot-anim
+            -camera-back none
+            -memory 4096
+            -feature GLDirectMem,HasSharedSlotsHostMemoryAllocator
           script: ./gradlew :BraintreeCore:connectedCheck :Card:connectedCheck :SharedUtils:connectedCheck :LocalPayment:connectedCheck :DataCollector:connectedCheck :GooglePay:connectedCheck :Demo:connectedCheck :PayPal:connectedCheck :PayPalMessaging:connectedCheck :AmericanExpress:connectedCheck :SEPADirectDebit:connectedCheck :ShopperInsights:connectedCheck :ThreeDSecure:connectedCheck :UIComponents:connectedCheck :Venmo:connectedCheck --stacktrace
--- a/.github/workflows/instrumentation_tests.yml
+++ b/.github/workflows/instrumentation_tests.yml
@@ -48,7 +48,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        api-level: [ 37 ]
+        api-level: [ "37.0" ]
         target: [ google_apis_ps16k ]
     steps:
       - name: checkout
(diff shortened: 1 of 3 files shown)


.github/workflows/instrumentation_tests.yml
```diff
@@ -39,4 +39,60 @@ jobs:
...
+          script: ./gradlew :BraintreeCore:connectedCheck :Card:connectedCheck :SharedUtils:connectedCheck :LocalPayment:connectedCheck :DataCollector:connectedCheck :GooglePay:connectedCheck :Demo:connectedCheck :PayPal:connectedCheck :PayPalMessaging:connectedCheck :AmericanExpress:connectedCheck :SEPADirectDebit:connectedCheck :ShopperInsights:connectedCheck :ThreeDSecure:connectedCheck :UIComponents:connectedCheck :Venmo:connectedCheck --stacktrace
+
+  test_new_api:
+    runs-on: ubuntu-latest
+    env:
+      ORG_GRADLE_PROJECT_org.gradle.jvmargs: "-Xmx4096m -XX:MaxMetaspaceSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8"
+    strategy:
+      fail-fast: false
+      matrix:
+        api-level: [ "37.0" ]
```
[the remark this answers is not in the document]
noguier: new api is not stable yet, so we need to have it in "" for the emulator to start!

.github/workflows/instrumentation_tests.yml
mzlangreder, line 1: Should we name this file something more specific since it is (so far) only for API 37?

Demo/build.gradle
```diff
@@ -106,7 +106,7 @@ dependencies {
 
     androidTestImplementation libs.device.automator
     androidTestImplementation project(':TestUtils')
-    androidTestImplementation 'androidx.test.espresso:espresso-core:3.5.1'
+    androidTestImplementation libs.androidx.test.espresso.core
```
mzlangreder, line 109: ❤️

.github/workflows/instrumentation_tests.yml
```diff
@@ -39,4 +39,62 @@ jobs:
...
+      - name: run tests
+        id: run_tests
+        uses: reactivecircus/android-emulator-runner@v2
+        continue-on-error: true
+        with:
+          api-level: ${{ matrix.api-level }}
+          target: ${{ matrix.target }}
+          arch: x86_64
+          profile: pixel_7_pro
+          emulator-options: >
```
saralvasquez: Ah so I assume these are the extra emulator requirements to make it work for 17 right? Just to double check, these options don't work for the older versions of the API?

.github/workflows/instrumentation_tests.yml
saralvasquez: Ah good questions. So this file actually runs the instrumentation tests for a bunch of older versions of the APIs. Ana correct me if I'm wrong, the changes here are adding a new test job specifically for 17 that runs separately from the tests for the other API levels right?

Demo/build.gradle
saralvasquez: Love the cleanup!

.github/workflows/instrumentation_tests.yml
mzlangreder: Shoot, I bonked because @noguier and I had talked previously about a separate file/matrix for 37+ where they seem to be requiring a separate target. Looks like we instead just added a new job to this file, oops!

noguier: yes, Sara's comment is correct, I can create a separate file for 37+, though if it is making it more clear! what do you think?

.github/workflows/instrumentation_tests.yml
noguier: Great question, I actually haven't tried it with older versions, since the new emulator image still very new and requires a little more setup. We can, hopefully, remove extra setup once it becomes more stable, and tackle the clean up of this file as a follow up ticket. But I am open to suggestions! @saralvasquez @mzlangreder

.github/workflows/instrumentation_tests.yml
mzlangreder: I think it's ok as-is.

mzlangreder (approved):

.github/workflows/instrumentation_tests.yml
saralvasquez: Yeah I think keeping it all in one file is ideal IMO

.github/workflows/instrumentation_tests.yml
saralvasquez: Ah ok so it's not super stable. My thought was that if we could use the new emulator for the older versions we could keep the matrix setup so we wouldn't need the extra job which just keeps the file a bit simpler and smaller. But if it's not super stable maybe that's not the best idea. What's been your experience testing with it? Is the setup flaky or do these new requirements keep it pretty consistent?

noguier: so yeah, I tried before just adding a new api to the existing matrix, but it wasn't working correctly. I can try it just in case, but i think for now it will be best to leave it as a separate solution and then optimize later? 

The way it is setup right not the android 37 test is not flaky at all :) The only flaky tests are the usual 3DS suspects and most of the time for older APIs.

saralvasquez: Ok cool! Did you used the new emulator when you tried adding the new API to the matrix? I think it's worth trying using the new emulator for a matrix with all the versions plus the new one. But if it doesn't work immediately I'm happy to get this merged in

saralvasquez:
Aw bummer. It looks like it works for all of them besides 23. I think you're original solution is our best bet. Sorry for making you do the extra work for nothing! I'll give it a 👍 when the new stuff is reverted

agedd (approved):

saralvasquez (approved):
🎉

PR #1636, opened 2026-07-09
fix: Allow custom payment button sizing

saralvasquez:
### Summary of changes
- Removes restrictive button sizing
  - Allows setting button size via xml when integrating
  - Sets default button size if not otherwise decided
  - Enforces minimum button size to prevent logo clipping
- Adds instrumentation tests for button sizing

I also went ahead an rearranged the buttons and card fields in the UIComponents fragment to keep the buttons closer to the color selectors. Here's a screenshot of how it looks now. I know this differs from iOS but I think it looks more intuitive and I'd be happy to update iOS to match this

Screenshot_20260708_160229

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used claude to write tests and debugging an issue with logo clipping even after setting minimum

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry

### Authors
- @saralvasquez


UIComponents/src/main/java/com/braintreepayments/api/uicomponents/VenmoButton.kt
```diff
@@ -297,8 +296,17 @@ class VenmoButton @JvmOverloads constructor(
...
+    private fun setMinimumSize() {
+        minimumHeight = desiredHeight
+        val logoDrawable = logo ?: return
+        val focusPadding = 2 * resources.getDimension(R.dimen.pay_button_focus_border).toInt()
+        minimumWidth = logoDrawable.intrinsicWidth + 4 * focusPadding
+    }
+
     override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
-        setMeasuredDimension(desiredWidth, desiredHeight)
+        val width = resolveSize(desiredWidth, widthMeasureSpec).coerceAtLeast(minimumWidth)
```
noguier, line 308: just to clarify, will this resolve width size to a minimum if specified size is less than a minimum size correct?

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/PayPalButtonTest.kt
```diff
@@ -71,4 +75,91 @@ class PayPalButtonTest {
 
         assertTrue(button.isEnabled)
     }
+
+    @Test
+    fun onMeasure_withUnspecifiedSpec_usesDesiredSizeClampedToMinimum() {
```
noguier, line 80: can you please explain to me what this test is doing? What dimensions will the button be with MeasureSpec.Unspecified ?

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/PayPalButtonTest.kt
```diff
@@ -71,4 +75,91 @@ class PayPalButtonTest {
...
+
+        assertEquals(maxOf(desiredWidth(context), minWidth(context)), button.measuredWidth)
+        assertEquals(maxOf(desiredHeight(context), minHeight(context)), button.measuredHeight)
+    }
+
+    @Test
+    fun onMeasure_withExactSpec_usesProvidedSize() {
+        val context = ApplicationProvider.getApplicationContext<Context>()
+        val button = PayPalButton(context)
+        val exactWidth = desiredWidth(context) + LARGE_BOUND
```
noguier: what is `LARGE_BOUND`? I see below it is set to 2000, is it in pixels?

noguier (approved):
Just a couple of clarifying questions, but otherwise it looks great! thank you for doing this and thank you for rearranging the demo. I think this layout is more intuitive as well.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/VenmoButton.kt
saralvasquez: correct!

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/PayPalButtonTest.kt
saralvasquez: The unspecified means that there is not size set for the button so there is no dimension tied to that value. This test is for the case where the merchant doesn't pick a buttons size so it checks that the button is the default size set in the PPButton class

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/PayPalButtonTest.kt
saralvasquez: Yeah that's in pixels. The naming isn't very intuitive. It's supposed to just be an arbitrarily large number of pixels for the button size to test that it can be made bigger without the default size interfering. I can rename it if that would help

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/PayPalButtonTest.kt
noguier: great, thats what I thought!

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/PayPalButtonTest.kt
noguier: oh got it, I am always a fan of explicit naming, but it is up to you!

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
```diff
@@ -306,8 +305,17 @@ class PayPalButton @JvmOverloads constructor(
         }
     }
 
+    private fun setMinimumSize() {
+        minimumHeight = desiredHeight
+        val logoDrawable = logo ?: return
+        val focusPadding = resources.getDimension(R.dimen.pay_button_focus_padding).toInt()
+        minimumWidth = logoDrawable.intrinsicWidth + 4 * focusPadding
```
mzlangreder: Is the `4` why we needed to suppress magic number? Is there a semantic name we could use for it instead of suppressing?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/VenmoButton.kt
```diff
@@ -297,8 +296,17 @@ class VenmoButton @JvmOverloads constructor(
         }
     }
 
+    private fun setMinimumSize() {
+        minimumHeight = desiredHeight
+        val logoDrawable = logo ?: return
+        val focusPadding = 2 * resources.getDimension(R.dimen.pay_button_focus_border).toInt()
```
mzlangreder: What does `2` do?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
saralvasquez: Yeah the 4 is the culprit. I can try and come up with a reasonable name

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/VenmoButton.kt
saralvasquez: So `pay_button_focus_border` I _think_ (It's been forever since we built this) is additional padding to the button so that the focus when tabbing to the button properly outlines it without clipping. The value itself is only the padding for one side so the 2 is to account for the fact that the button's size includes padding out front and at the end of the button

saralvasquez: Ok correction. That value is the width of the focus indicator and we double it to give padding to an individual side of the button. Below the 4x was originally a 2x to account for both sides of the button but I increased it to 4x so that there was actual padding on either side of the logo instead of the minimum button width perfectly hugging the logo. I'll clean this up so it's actually clear and easy to understand in the variable names

saralvasquez:
Ok I tried to clear up the padding dimension naming by adding an additional variable that's the same size as the padding but specifically named to call out that it's there to pad the logo. It doesn't feel like the cleanest solution since the variable is strictly to clear up ambiguity and is the exact same value is the padding so please let me know if you all have a better solution

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
```diff
@@ -306,8 +305,18 @@ class PayPalButton @JvmOverloads constructor(
         }
     }
 
+    private fun setMinimumSize() {
+        minimumHeight = desiredHeight
+        val logoDrawable = logo ?: return
+        val fullButtonFocusPadding = resources.getDimension(R.dimen.pay_button_focus_padding).toInt() * 2
```
mzlangreder, line 311: Should this `2` also be updated?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/VenmoButton.kt
```diff
@@ -297,8 +296,18 @@ class VenmoButton @JvmOverloads constructor(
         }
     }
 
+    private fun setMinimumSize() {
+        minimumHeight = desiredHeight
+        val logoDrawable = logo ?: return
+        val fullButtonFocusPadding = resources.getDimension(R.dimen.pay_button_focus_padding).toInt() * 2
```
mzlangreder, line 302: Should this `2` also be updated?

mzlangreder (commented):
Sorry, I think I missed the couple of `2`s on my previous review :/

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/PayPalButton.kt
saralvasquez: That 2 should actually be there now. The pay_button_focus_padding is to add padding around the focus indicator but it only accounts for one side of the button. So the 2 is to account for padding on the leading and trailing sides of the button. Does that makes sense? I have a hard time explaining UI stuff without visuals

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/VenmoButton.kt
saralvasquez: same as above

mzlangreder (approved):

PR #1639, opened 2026-07-13
Add versioning to Android reference docs

buzzamus:
## Summary
- Wires up Dokka's official `versioning-plugin` (matching the `dokka` version already pinned in `gradle/libs.versions.toml`) so `dokkaHtmlMultiModule` renders a version picker instead of only ever showing the latest release's docs.
- Each release now publishes to `docs/versions/<version>/` instead of overwriting `docs/` in place; `docs/index.html` is a generated redirect to the latest version so the existing published URL (`braintree.github.io/braintree_android/`) keeps working unchanged.
- Adds a `pruneOldDocVersions` task (wired into the release workflow) that retains only the last `maxRetainedDocVersions` (5) versions, bounding repo/Pages growth.
- Seeds the archive with the already-published 5.29.0 docs (relocated via `git mv`, plus a hand-authored `version.json` marker matching what the plugin itself writes) so the site isn't broken until the next real release. No other historical versions are backfilled — versioning starts fresh from here per plan.

## Verified
- Confirmed `org.jetbrains.dokka:versioning-plugin:1.9.10` exists on Maven Central and inspected its bytecode directly to confirm the Gradle configuration name (`dokkaHtmlMultiModulePlugin`) and the `VersioningConfiguration`/`version.json` schema used in this change.
- Couldn't run a full `dokkaHtmlMultiModule` build in this sandbox (missing matching local Android SDK platform, a pre-existing environment gap unrelated to this change), so validated the custom Gradle logic directly instead:
  - `writeDocsRedirect` produces a correct `docs/index.html` redirect.
  - `pruneOldDocVersions` correctly keeps the 5 numerically-highest versions (verified it orders `5.9.0` before `5.10.0`, not lexicographically).
  - Served `docs/` locally over HTTP and confirmed the redirect, the archived `5.29.0` docs, their assets, and `version.json` all resolve with 200s.

## Test plan
- [ ] Run a real release through this workflow (or manually run `./gradlew dokkaHtmlMultiModule pruneOldDocVersions` in CI, where the correct Android SDK platform is installed) to confirm the versioning-plugin itself builds successfully end-to-end and renders the version dropdown against the seeded `5.29.0` archive.
- [ ] Confirm the published GitHub Pages site still resolves correctly after the next real release.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


buzzamus:
Closing this down since it was a POC and we are going with an alternate solution

PR #1634, opened 2026-07-09
feature(PayPal): merchant-independent foreground auto-link trigger

santugowda:
## Summary

PR 2 of 2 for the PayPal **auto-link on manual return** feature.

> [!IMPORTANT]
> **Stacked on #1633.** This branch is based on the PR 1 branch, so the diff currently shows both commits. Review/merge **after** #1633; once it merges, this PR reduces to its single commit (`ec303993c`). Only that commit is part of PR 2.

PR 1 delivered the deterministic `handleReturnToApp` `NoResult` path (covers `onResume` integrations) plus re-click. This PR adds the **merchant-independent** trigger for integrations where `handleReturnToApp` is never called on a manual return (`onNewIntent`/`singleTop`, which receive no new intent).

## What changed (this PR's commit only)

- **`AppForegroundDetector`** — a `DefaultLifecycleObserver` registered once via `ProcessLifecycleOwner` in `PendingPaymentStore.instance` (guarded; silently disabled if unavailable).
- **State machine** on `PendingPaymentStore`: `SWITCH_LAUNCHED` on launch → `AWAITING_RETURN` on `onStop` → on `onStart` the foreground trigger tokenizes the stored BA token.
- **Timer-free race resolution.** The foreground attempt is posted to the main-queue tail and re-checks state inside the coroutine. A real URL return's `Success` clears the store (state → `IDLE`) and wins the race — no suppression timer. The resolved nonce is delivered via `handleReturnToApp` (`NoResult → Success(nonce)`) or a re-click; the `CompletableDeferred` still guarantees exactly one BTGW call.
- **Dependency:** adds `androidx-lifecycle-process`.

## Analytics added
```
paypal:tokenize:auto-link:handle-return:succeeded   (nonce pre-resolved by the foreground trigger)
```

## New files
- `AppForegroundDetector.kt`
- `AppForegroundDetectorUnitTest.kt`

## Test plan
- [x] Unit tests: 79 passing across the PayPal auto-link suites, incl. new `AppForegroundDetectorUnitTest` (0 failures)
- [ ] Detekt / lint (not run locally — CI)
- [ ] Manual QA (`onNewIntent`/`singleTop` merchant): app-switch vault, uncheck "Open supported links", complete BA, manually return with no new intent → foreground trigger resolves the nonce; verify no duplicate BTGW call when a real App Link return also arrives.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1633, opened 2026-07-09
feature(PayPal): auto-link on manual return via handleReturnToApp

santugowda:
## Summary

PR 1 of 2 for the PayPal **auto-link on manual return** feature.

When a PayPal app-switch vault flow completes but the App Link return fails, the user manually returns to the merchant app. This PR adds the **deterministic** auto-link path so the SDK can tokenize the billing agreement directly with BTGW without a URL return.

## What changed

- **`handleReturnToApp` runs `completeRequest` first.** A real URL return is authoritative and clears any pending session; a `NoResult` with a valid pending billing-agreement session routes to auto-link. `NoResult` is the deterministic "manual return / App Link failed" signal — no timers or lifecycle guessing.
- **Auto-link tokenization** (`AutoLinkTokenizeUseCase`) — the stored BA token is sent as a standalone `billing_agreement_token` in the `paypal_account` payload (same field the Web SDK uses).
- **Re-click path folded in** — passing a `tokenizeCallback` to `createPaymentAuthRequest` delivers the nonce directly if the user re-taps PayPal after a failed return.
- **`PendingPaymentStore`** (minimal) persists the session across the round trip; a `CompletableDeferred` guarantees exactly one BTGW call across entry points.

## Compatibility

- **Non-breaking.** The only public API addition is an overload `createPaymentAuthRequest(context, request, callback, tokenizeCallback)`; the existing 3-arg method delegates to it with `tokenizeCallback = null`. All other new types are `internal`.

## Analytics events added

```
paypal:tokenize:auto-link:started | succeeded | failed
paypal:tokenize:auto-link:expired
paypal:tokenize:auto-link:handle-return:pending
paypal:tokenize:auto-link:reclick:started | succeeded | failed
```

## New files
- `PendingPaymentStore.kt`
- `AutoLinkTokenizeUseCase.kt`
- `PendingPaymentStoreUnitTest.kt`, `AutoLinkTokenizeUseCaseUnitTest.kt`

## Not in this PR (deferred to PR 2)
- `AppForegroundDetector` / `ProcessLifecycleOwner` merchant-independent foreground trigger
- `androidx-lifecycle-process` dependency and the `State` enum

## Test plan
- [x] Unit tests: 71 passing across `PayPalClientUnitTest`, `PayPalLauncherUnitTest`, `PendingPaymentStoreUnitTest`, `AutoLinkTokenizeUseCaseUnitTest` (0 failures)
- [ ] Detekt / lint (not run locally — will rely on CI)
[...]

PR #1640, opened 2026-07-14
feat: add PendingPaymentStore for auto-link session persistence

pedrofsn:
## Summary of changes

Adds `PendingPaymentStore`, the in-memory session store used by the PayPal auto-link on manual return feature to carry a pending billing agreement token across the App Switch round trip.

### Background

When a user approves a PayPal Billing Agreement (BA) via App Switch and returns to the merchant app, the return can fail to convert into a tokenized payment method in two cases:
- The universal link fails to reopen the merchant app (~5% of returns).
- The user manually switches back to the merchant app before the universal link launches.

Today, when `handleOpen(url)` is never called, the SDK has no reference to the approved BA token and cannot tokenize it — the BA is approved on PayPal's side but the merchant never receives a nonce.

This PR is **PR 1 of a multi-PR series** that adds "auto-link on manual return": the SDK persists the BA token before app switch and retries tokenization when the user returns, regardless of how they returned. This PR introduces only the storage primitive; no tokenization logic and no `PayPalClient` wiring are included here, so each PR in the series stays independently reviewable.

**Changes:**
- `PendingPaymentStore` (internal, in-memory, process-level singleton): holds a `PendingSession` (BA token + correlation ID + merchant account ID + intent + payment type + timestamp/TTL) and a `CompletableDeferred<PayPalAccountNonce>` used by later PRs to guarantee exactly one BTGW tokenization call across the multiple entry points (return-handling and re-click) that can race to trigger it.

### Steps to Test

This PR only adds a storage primitive with no wiring into the checkout flow yet, so it has no end-user-visible behavior to exercise manually. Verification is via the included unit tests:

1. Run `./gradlew :PayPal:testDebugUnitTest --tests "*.PendingPaymentStoreUnitTest"`
2. Confirm all tests pass: initial state is null, `PendingSession.isExpired()` correctly reflects TTL boundaries, `getOrCreateDeferred()` returns the same deferred to a second caller while the first remains the initiator, and `clear()` resets all fields and cancels any in-flight deferred.

### AI Usage

**Which AI Agent Was Used?**
- [x] Claude

**How was AI used?**
Code generation and unit test authoring for `PendingPaymentStore`, based on an existing HLD/LLD design and a prior reference implementation for this feature.

- [ ] less than 30%
- [x] 30 - 60%
- [ ] 60 - 100%

### Checklist
- [x] Added a changelog entry
- [x] Tested and confirmed payment flows affected by this change are functioning as expected (no flow wiring yet in this PR; unit-tested in isolation)

### Authors
- @pedrofsn

### Inner Source Process
Internal to PayPal contributors should fill out this section. All others can delete.

PR should follow these steps before codeowners review will begin:
1. Comment `/inner source` on this PR — this will automatically add the `inner source` and `tech lead review required` labels. Open the PR in a draft state.
2. PR should be reviewed by and approved by your team's technical lead, we do not allow LGTM reviews, there should be comments and feedback provided on all PR reviews
3. Once the above steps are completed, comment `/ready` on this PR — this will automatically remove the `tech lead review required` label. Move the PR to ready to review.
4. PR comments must be addressed within 24 hours, if you are unable to address within this timeframe, move the PR back to a draft state so our team knows not to review

### Inner Source Checklist
- [x] Added all labels to the PR
- [x] Provide steps to test the flows changed, if applicable in the summary
- [ ] Demo video of the functionality, if applicable — not applicable, this PR adds a storage primitive only, no user-visible behavior yet
- [ ] All upstream dependencies are merged in and this PR can be released at any time; PRs should not be opened until this is true — **this is PR 1 of 5 in a stacked series; PRs 2-5 will follow, each building on the previous one's branch, same pattern as the earlier #1633/#1634 split**
- [x] Unit tests and builds have been run locally and pass/compile as expected

## PR series

Part of a 5-PR series re-splitting the original "auto-link on manual return" feature (previously opened as #1633 and #1634, closed for being too large to review):

1. **This PR** - `PendingPaymentStore` (session persistence)
2. `AutoLinkTokenizeUseCase` (tokenization logic) - stacked on this PR
3. `PayPalClient` wiring (handleReturnToApp + re-click) - stacked on PR 2
4. `PayPalLauncher`/`PayPalPaymentAuthResult`/Demo wiring - stacked on PR 3
5. `AppForegroundDetector` merchant-independent foreground trigger - stacked on PR 4

Please review/merge in order 1 -> 5.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 * PayPal
     * Expose `createPaymentAuthRequest` as a public suspend function
     * Expose `tokenize` as a public suspend function
+    * Add internal `PendingPaymentStore` for auto-link on manual return (no public API changes)
 * SEPADirectDebit
     * Expose `createPaymentAuthRequest` as a public suspend function
     * Expose `tokenize` as a public suspend function
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * PayPal
     * Collect device information to improve PayPal app switch eligibility when it's enabled
+    * Add internal `PendingPaymentStore` for auto-link on manual return
 
 ## 5.30.0 (2026-07-21)
 
@@ -28,7 +29,6 @@
 * PayPal
     * Expose `createPaymentAuthRequest` as a public suspend function
     * Expose `tokenize` as a public suspend function
-    * Add internal `PendingPaymentStore` for auto-link on manual return (no public API changes)
 * SEPADirectDebit
     * Expose `createPaymentAuthRequest` as a public suspend function
     * Expose `tokenize` as a public suspend function
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PendingPaymentStore.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PendingPaymentStore.kt
@@ -1,0 +1,84 @@
+package com.braintreepayments.api.paypal
+
+import kotlinx.coroutines.CompletableDeferred
+
+/**
+ * Process-level singleton that holds state for the auto-link on manual return flow.
+ *
+ * When a PayPal app switch completes but the App Link return fails, the user manually
+ * switches back to the merchant app. This store persists the billing agreement token and
+ * session metadata across that round trip so the SDK can tokenize directly with BTGW
+ * without a URL return.
+ *
+ * A [CompletableDeferred] ensures exactly one BTGW tokenization call is made, even when
+ * multiple entry points (handle-return NoResult and user re-click) race to tokenize.
+ *
+ * This class must outlive individual [PayPalClient] instances because the client is
+ * per-Fragment and may be GC'd during the app switch.
+ */
+internal class PendingPaymentStore {
+
+    /**
+     * Captured session data needed to tokenize a billing agreement without a URL return.
+     *
+     * @property baToken the billing agreement token from the PayPal approval URL
+     * @property clientMetadataId correlation ID for fraud detection
+     * @property merchantAccountId optional merchant account override
+     * @property intent the PayPal payment intent string value (e.g. "authorize", "sale")
+     * @property paymentType always "billing-agreement" for auto-link flows
+     * @property timestampMs creation time for TTL expiry checks
+     * @property ttlMs time-to-live in milliseconds (default 30 minutes)
+     */
+    data class PendingSession(
+        val baToken: String,
+        val clientMetadataId: String?,
+        val merchantAccountId: String?,
+        val intent: String?,
+        val paymentType: String,
+        val timestampMs: Long = System.currentTimeMillis(),
+        val ttlMs: Long = TTL_MS
+    ) {
+        /** Returns true if this session has exceeded its time-to-live. */
+        fun isExpired(): Boolean = System.currentTimeMillis() - timestampMs > ttlMs
+    }
+
+    var pendingSession: PendingSession? = null
+    var tokenizeDeferred: CompletableDeferred<PayPalAccountNonce>? = null
+
+    /** Resolved nonce from auto-link tokenization. Volatile for safe cross-thread reads. */
+    @Volatile
+    var autoLinkNonce: PayPalAccountNonce? = null
+
+    /**
+     * Returns an existing or newly created [CompletableDeferred] along with a flag
+     * indicating whether this caller is the initiator (created the deferred).
+     *
+     * The initiator is responsible for making the BTGW call and completing the deferred.
+     * Subsequent callers receive the same deferred and should await it.
+     *
+     * @return pair of (deferred, isInitiator)
+     */
+    @Synchronized
+    fun getOrCreateDeferred(): Pair<CompletableDeferred<PayPalAccountNonce>, Boolean> {
+        val existing = tokenizeDeferred
+        if (existing != null) return Pair(existing, false)
+        val new = CompletableDeferred<PayPalAccountNonce>()
+        tokenizeDeferred = new
+        return Pair(new, true)
+    }
+
+    /** Resets all state and cancels any in-flight deferred. */
+    fun clear() {
+        pendingSession = null
+        tokenizeDeferred?.cancel()
+        tokenizeDeferred = null
+        autoLinkNonce = null
+    }
+
+    companion object {
+        /** Default session TTL: 30 minutes. */
+        internal const val TTL_MS = 30 * 60 * 1000L
+
+        val instance by lazy { PendingPaymentStore() }
+    }
+}
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PendingPaymentStore.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PendingPaymentStore.kt
@@ -42,7 +42,10 @@ internal class PendingPaymentStore {
         fun isExpired(): Boolean = System.currentTimeMillis() - timestampMs > ttlMs
     }
 
+    @Volatile
     var pendingSession: PendingSession? = null
+
+    @Volatile
     var tokenizeDeferred: CompletableDeferred<PayPalAccountNonce>? = null
 
     /** Resolved nonce from auto-link tokenization. Volatile for safe cross-thread reads. */
--- a/PayPal/src/main/java/com/braintreepayments/api/paypal/PendingPaymentStore.kt
+++ b/PayPal/src/main/java/com/braintreepayments/api/paypal/PendingPaymentStore.kt
@@ -7,10 +7,10 @@ import kotlinx.coroutines.CompletableDeferred
  *
  * When a PayPal app switch completes but the App Link return fails, the user manually
  * switches back to the merchant app. This store persists the billing agreement token and
- * session metadata across that round trip so the SDK can tokenize directly with BTGW
- * without a URL return.
+ * session metadata across that round trip so the SDK can tokenize directly with the
+ * Braintree Gateway (BTGW) without a URL return.
  *
- * A [CompletableDeferred] ensures exactly one BTGW tokenization call is made, even when
+ * A [CompletableDeferred] ensures exactly one Braintree Gateway (BTGW) tokenization call is made, even when
  * multiple entry points (handle-return NoResult and user re-click) race to tokenize.
  *
  * This class must outlive individual [PayPalClient] instances because the client is
@@ -56,7 +56,7 @@ internal class PendingPaymentStore {
      * Returns an existing or newly created [CompletableDeferred] along with a flag
      * indicating whether this caller is the initiator (created the deferred).
      *
-     * The initiator is responsible for making the BTGW call and completing the deferred.
+     * The initiator is responsible for making the Braintree Gateway (BTGW) call and completing the deferred.
      * Subsequent callers receive the same deferred and should await it.
      *
      * @return pair of (deferred, isInitiator)
--- a/PayPal/src/test/java/com/braintreepayments/api/paypal/PendingPaymentStoreUnitTest.kt
+++ b/PayPal/src/test/java/com/braintreepayments/api/paypal/PendingPaymentStoreUnitTest.kt
@@ -1,0 +1,115 @@
+package com.braintreepayments.api.paypal
+
+import io.mockk.mockk
+import kotlinx.coroutines.CompletableDeferred
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertNull
+import org.junit.Assert.assertSame
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+
+class PendingPaymentStoreUnitTest {
+
+    private lateinit var sut: PendingPaymentStore
+
+    @Before
+    fun setup() {
+        sut = PendingPaymentStore()
+    }
+
+    @Test
+    fun `initial pendingSession is null`() {
+        assertNull(sut.pendingSession)
+    }
+
+    @Test
+    fun `initial autoLinkNonce is null`() {
+        assertNull(sut.autoLinkNonce)
+    }
+
+    @Test
+    fun `PendingSession isExpired returns false when within TTL`() {
+        val session = PendingPaymentStore.PendingSession(
+            baToken = "BA-123",
+            clientMetadataId = "metadata-id",
+            merchantAccountId = null,
+            intent = null,
+            paymentType = "billing-agreement",
+            timestampMs = System.currentTimeMillis(),
+            ttlMs = PendingPaymentStore.TTL_MS
+        )
+        assertFalse(session.isExpired())
+    }
+
+    @Test
+    fun `PendingSession isExpired returns true when past TTL`() {
+        val session = PendingPaymentStore.PendingSession(
+            baToken = "BA-123",
+            clientMetadataId = null,
+            merchantAccountId = null,
+            intent = null,
+            paymentType = "billing-agreement",
+            timestampMs = System.currentTimeMillis() - PendingPaymentStore.TTL_MS - 1,
+            ttlMs = PendingPaymentStore.TTL_MS
+        )
+        assertTrue(session.isExpired())
+    }
+
+    @Test
+    fun `getOrCreateDeferred creates new deferred on first call`() {
+        val (deferred, isInitiator) = sut.getOrCreateDeferred()
+        assertTrue(isInitiator)
+        assertSame(deferred, sut.tokenizeDeferred)
+    }
+
+    @Test
+    fun `getOrCreateDeferred returns existing deferred on second call`() {
+        val (first, isFirstInitiator) = sut.getOrCreateDeferred()
+        val (second, isSecondInitiator) = sut.getOrCreateDeferred()
+
+        assertTrue(isFirstInitiator)
+        assertFalse(isSecondInitiator)
+        assertSame(first, second)
+    }
+
+    @Test
+    fun `clear resets all state`() {
+        val nonce = mockk<PayPalAccountNonce>()
+        sut.pendingSession = PendingPaymentStore.PendingSession(
+            baToken = "BA-123",
+            clientMetadataId = null,
+            merchantAccountId = null,
+            intent = null,
+            paymentType = "billing-agreement"
+        )
+        sut.tokenizeDeferred = CompletableDeferred()
+        sut.autoLinkNonce = nonce
+
+        sut.clear()
+
+        assertNull(sut.pendingSession)
+        assertNull(sut.tokenizeDeferred)
+        assertNull(sut.autoLinkNonce)
+    }
+
+    @Test
+    fun `clear cancels active deferred`() {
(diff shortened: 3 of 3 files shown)


saralvasquez:
I see this PR doesn't follow our standard template. Please take a look at [this PR](https://github.com/braintree/braintree_android/pull/1600) to see our inner sourcing process and follow the steps as listed. Thank you!

jaxdesmarais:
To add to @saralvasquez comment, this also is pointing to the main branch instead of a feature branch. Based on the PR description I am guessing claude nuked the template. Please open following the correct template and update the base branch to reflect that this will go into a feature branch before main.

pedrofsn:
Thanks for the feedback! Quick clarification on the base branch: I'm contributing via a fork (`pedrofsn/braintree_android`), so I only have push access to my own fork. I can't create a shared branch directly on this repo, since GitHub only lets a fork-based PR's base point to a branch that already exists upstream, and the only one available to me is `main`.

For reference, the original #1633 /#1634 split used the same approach: both targeted `main`, with each subsequent PR's head branch stacked on the previous one, so the diff naturally shrinks to just the new commit once the earlier PR merges.

If you'd prefer this series (5 PRs total) to target a dedicated integration branch instead of `main` directly, I'm happy to follow that but someone with write access would need to create it first (e.g. `feature/paypal-autolink-manual-return`), since I can't push branches here myself. Otherwise I'll keep following the `main` + stacked-head-branches pattern from the original split. Let me know which you'd prefer.

CC @jaxdesmarais @noguier

jaxdesmarais:
That's fair @pedrofsn - I have gone ahead and created `feature/paypal-autolink-manual-return` for you and pushed it up so you can use that as the base branch.

pedrofsn:
> That's fair @pedrofsn - I have gone ahead and created `feature/paypal-autolink-manual-return` for you and pushed it up so you can use that as the base branch.

Thanks @jaxdesmarais, appreciate you setting that up! Rebased this PR onto `feature/paypal-autolink-manual-return`.

Kicking off the inner source process now:

/inner source

PRs 2-5 in this series will also target this branch going forward.

pedrofsn:
Marking this as draft while it's pending tech lead review, per the inner source process.

PayPal/src/main/java/com/braintreepayments/api/paypal/PendingPaymentStore.kt
```diff
@@ -0,0 +1,84 @@
...
+        val intent: String?,
+        val paymentType: String,
+        val timestampMs: Long = System.currentTimeMillis(),
+        val ttlMs: Long = TTL_MS
+    ) {
+        /** Returns true if this session has exceeded its time-to-live. */
+        fun isExpired(): Boolean = System.currentTimeMillis() - timestampMs > ttlMs
+    }
+
+    var pendingSession: PendingSession? = null
```
santugowda, line 46: don't we need to make `pendingSession` and `tokenizeDeferred` @Volatile here similar to `autoLinkNonce` for for cross-thread visibility,.

PayPal/src/test/java/com/braintreepayments/api/paypal/PendingPaymentStoreUnitTest.kt
```diff
@@ -0,0 +1,115 @@
...
+        val (deferred, isInitiator) = sut.getOrCreateDeferred()
+        assertTrue(isInitiator)
+        assertSame(deferred, sut.tokenizeDeferred)
+    }
+
+    @Test
+    fun `getOrCreateDeferred returns existing deferred on second call`() {
+        val (first, isFirstInitiator) = sut.getOrCreateDeferred()
+        val (second, isSecondInitiator) = sut.getOrCreateDeferred()
+
```
santugowda, line 74: Nit : This test covers sequential calls, worth adding a test that races two concurrent callers into getOrCreateDeferred(). 
Let me your thought on this

PayPal/src/main/java/com/braintreepayments/api/paypal/PendingPaymentStore.kt
pedrofsn: Fixed. `@Volatile` to both, matching `autolinkNonce`. Good catch: `clear()` mutated them outside the `getOrCreateDeferred()` monitor, so there was a real visibility gap across threads.

PayPal/src/test/java/com/braintreepayments/api/paypal/PendingPaymentStoreUnitTest.kt
pedrofsn: Two threads racing via `CyclicBarrier`, repeated 500x, asserting same deferred + exactly one initiator. Verified it actually catches a broken lock (temporarily removed `@Synchronized`, confirmed the test fails, restored it).

santugowda:
LGTM - rebase with the latest once. Thanks

saralvasquez:
A lot of large commits went into main recently which is blowing up the changes for this branch. Could you please merge or rebase this branch so it's easier to see what changes are actually being made in this PR

pedrofsn:
> A lot of large commits went into main recently which is blowing up the changes for this branch. Could you please merge or rebase this branch so it's easier to see what changes are actually being made in this PR

I rebased my current branch with main. But we also need to rebase the target/feature branch.

I am trying to raise a new PR doing it, but I am facing the error `Pull request creation failed. Validation failed: must be a collaborator`. Could you help me @saralvasquez?

### Not synced
image

### PR desired
image

### Able to merge
image

### Error creating the PR
image

mzlangreder:
I've rebased the base branch, you should be able to rebase your PR's now.

pedrofsn:
> I've rebased the base branch, you should be able to rebase your PR's now.

Already rebased, @jaxdesmarais.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
 * PayPal
     * Expose `createPaymentAuthRequest` as a public suspend function
     * Expose `tokenize` as a public suspend function
+    * Add internal `PendingPaymentStore` for auto-link on manual return (no public API changes)
```
saralvasquez: Please move this under the unreleased section at the top of the file. Additionally I'd remove the `(no public API changes)`

PayPal/src/main/java/com/braintreepayments/api/paypal/PendingPaymentStore.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    /** Resolved nonce from auto-link tokenization. Volatile for safe cross-thread reads. */
+    @Volatile
+    var autoLinkNonce: PayPalAccountNonce? = null
+
+    /**
+     * Returns an existing or newly created [CompletableDeferred] along with a flag
+     * indicating whether this caller is the initiator (created the deferred).
+     *
+     * The initiator is responsible for making the BTGW call and completing the deferred.
```
saralvasquez: Please spell out the Braintree Gateway. Acronyms aren't always easy to understand

saralvasquez (commented):
I have a couple small nits so far, but before we go any further, are there any design docs that we can reference to get a full picture of the changes in all of the coming PRs? The reason I ask is because we make it a general rule to keep the SDK as stateless as possible and the changes made here so far are very counter to that. I think I need a better understanding of the problem this is trying to solve and clearer picture of how it will be accomplished before I give this any kind of sign-off

CHANGELOG.md
pedrofsn: Moved to the _unreleased_ section and removed the (no public API changes) note in `9f388eb2b`

PayPal/src/main/java/com/braintreepayments/api/paypal/PendingPaymentStore.kt
pedrofsn: Done

jaxdesmarais:
My main question is whether this state should live in an in-memory by lazy singleton at all. The rest of the PayPal Launcher/Client design deliberately holds no long-lived in-memory state, the merchant persists `pendingRequestString` to durable storage precisely because the Activity or whole process can be destroyed during the switch (`PayPalPendingRequest` calls this out explicitly, and `PayPalLauncher` keeps nothing). This change is a fundamental different in how the SDK is currently documented/used by merchants.

The concern: this feature exists to recover the ~5% of returns where the App Link fails or the user manually switches back and those cases correlate strongly with the OS reclaiming the backgrounded app under memory pressure, which is exactly when an in-memory singleton gets wiped. So `PendingPaymentStore` survives a lightweight app switch but not a process-death app switch, losing the BA token in the case the feature is meant to fix. The docstring ("persists … across that round trip") reads as stronger durability than an in-memory store actually provides.

We also, as I understand, added analytics tags around this. What is our data telling us on this use case thus far to ensure that this PR isn't adding additional state that do not actually resolve the issues? I think we also need to align on the storage model here early as storing this unencrypted at rest for 30 minutes on a client SDK is likely a security violation.

pedrofsn:
> My main question is whether this state should live in an in-memory by lazy singleton at all. The rest of the PayPal Launcher/Client design deliberately holds no long-lived in-memory state, the merchant persists `pendingRequestString` to durable storage precisely because the Activity or whole process can be destroyed during the switch (`PayPalPendingRequest` calls this out explicitly, and `PayPalLauncher` keeps nothing). This change is a fundamental different in how the SDK is currently documented/used by merchants.
> 
> The concern: this feature exists to recover the ~5% of returns where the App Link fails or the user manually switches back and those cases correlate strongly with the OS reclaiming the backgrounded app under memory pressure, which is exactly when an in-memory singleton gets wiped. So `PendingPaymentStore` survives a lightweight app switch but not a process-death app switch, losing the BA token in the case the feature is meant to fix. The docstring ("persists … across that round trip") reads as stronger durability than an in-memory store actually provides.
> 
> We also, as I understand, added analytics tags around this. What is our data telling us on this use case thus far to ensure that this PR isn't adding additional state that do not actually resolve the issues? I think we also need to align on the storage model here early as storing this unencrypted at rest for 30 minutes on a client SDK is likely a security violation. 

Good points, @jaxdesmarais.

@santugowda — since you're the original author of this design, I wanted to loop you in directly on two points that came up during the review. I'm mainly shepherding the PR process here, so you have much more context on the original decisions and tradeoffs.

### 1. In-memory-only `PendingPaymentStore` and process death

The HLD lists disk persistence (Keychain/EncryptedSharedPreferences) as an open question rather than a finalized decision. Since this feature specifically handles return scenarios where the App Link fails or the user manually switches back—and those scenarios may overlap with cases where the OS reclaims the backgrounded app under memory pressure—was there a specific reason for deferring disk persistence, or has it simply not been prioritized yet?

Do we have any data, or plans to gather data, on how often the in-memory state is lost before the recovery flow completes?

### 2. Storing the BA token unencrypted at rest for up to 30 minutes

Was encryption considered and intentionally ruled out—for example, because of the short TTL, the threat model, or existing SDK patterns—or is this still an open question?

It would be helpful to capture that rationale somewhere, such as the LLD or an ADR, so future reviewers have the context behind the decision.

Neither point necessarily needs to be addressed as part of this PR if these are intentional tradeoffs or deferred items. I mainly want to make sure the reasoning is documented before we merge, since both points came up during the review.

santugowda:
> I have a couple small nits so far, but before we go any further, are there any design docs that we can reference to get a full picture of the changes in all of the coming PRs? The reason I ask is because we make it a general rule to keep the SDK as stateless as possible and the changes made here so far are very counter to that. I think I need a better understanding of the problem this is trying to solve and clearer picture of how it will be accomplished before I give this any kind of sign-off

@saralvasquez  : not sure if we can post PayPal's docs here, but below is the HLD and LLD for this feature 
HLD -- https://paypal.atlassian.net/wiki/spaces/Checkout/pages/2862318705 
LLD -- https://paypal.atlassian.net/wiki/x/9yVDrQ

santugowda:
> > My main question is whether this state should live in an in-memory by lazy singleton at all. The rest of the PayPal Launcher/Client design deliberately holds no long-lived in-memory state, the merchant persists `pendingRequestString` to durable storage precisely because the Activity or whole process can be destroyed during the switch (`PayPalPendingRequest` calls this out explicitly, and `PayPalLauncher` keeps nothing). This change is a fundamental different in how the SDK is currently documented/used by merchants.
> > The concern: this feature exists to recover the ~5% of returns where the App Link fails or the user manually switches back and those cases correlate strongly with the OS reclaiming the backgrounded app under memory pressure, which is exactly when an in-memory singleton gets wiped. So `PendingPaymentStore` survives a lightweight app switch but not a process-death app switch, losing the BA token in the case the feature is meant to fix. The docstring ("persists … across that round trip") reads as stronger durability than an in-memory store actually provides.
> > We also, as I understand, added analytics tags around this. What is our data telling us on this use case thus far to ensure that this PR isn't adding additional state that do not actually resolve the issues? I think we also need to align on the storage model here early as storing this unencrypted at rest for 30 minutes on a client SDK is likely a security violation.
> 
> Good points, @jaxdesmarais.
> 
> @santugowda — since you're the original author of this design, I wanted to loop you in directly on two points that came up during the review. I'm mainly shepherding the PR process here, so you have much more context on the original decisions and tradeoffs.
> 

@jaxdesmarais, @pedrofsn  - please find my comments below
> ### 1. In-memory-only `PendingPaymentStore` and process death
> The HLD lists disk persistence (Keychain/EncryptedSharedPreferences) as an open question rather than a finalized decision. Since this feature specifically handles return scenarios where the App Link fails or the user manually switches back—and those scenarios may overlap with cases where the OS reclaims the backgrounded app under memory pressure—was there a specific reason for deferring disk persistence, or has it simply not been prioritized yet?
> 
> Do we have any data, or plans to gather data, on how often the in-memory state is lost before the recovery flow completes?

**Why in-memory, and why disk persistence isn't decided** :  
For the current implementation, we deliberately chose to keep this in memory rather than on disk, as discussed in the design review doc/comment.  56% of drop-offs at this stage retry within 5 minutes, that's the addressable population this phase targets. 
In-memory recovers the subset who stay in the same app session, it does not recover force-close/process-death, and the HLD calls that out directly and not claiming full coverage. 
Disk persistence was left as an open question because we didn't have data yet on how much of that 56% we're actually losing to process death vs recovering in-session. 

> ### 2. Storing the BA token unencrypted at rest for up to 30 minutes
> Was encryption considered and intentionally ruled out—for example, because of the short TTL, the threat model, or existing SDK patterns—or is this still an open question?

Since we decided to go with in-memory, there was no need to look into token encryption at rest. That said the token is only kept for a short time,  max 30 minutes and gets deleted right after tokenization succeeds/fails or the user returns normally. 

> It would be helpful to capture that rationale somewhere, such as the LLD or an ADR, so future reviewers have the context behind the decision.

Sure we can add this to the LLD's design/error-handling section so it's documented, not just in a PR comment.                                                                           

> Neither point necessarily needs to be addressed as part of this PR if these are intentional tradeoffs or deferred items. I mainly want to make sure the reasoning is documented before we merge, since both points came up during the review.

True, but I guess team is clarifying on implementation details before reviewing other PR's.

jaxdesmarais:
> For the current implementation, we deliberately chose to keep this in memory rather than on disk, as discussed in the design review doc/comment. 56% of drop-offs at this stage retry within 5 minutes, that's the addressable population this phase targets.

56% of what percent? It's really hard for teams to grock impact in these numbers and is misleading as it's a percent of a percent.

> Disk persistence was left as an open question because we didn't have data yet on how much of that 56% we're actually losing to process death vs recovering in-session.

The SDK is a platform and should act in a consistent way based on data. These PRs are a great departure from that idea, which is why there is some hesitation. Especially when we don't have data to yet prove that this 1) is the issue at all 2) scope of the issue. It feels pretty early to be pushing out a change when we don't know if the change will fix 100% of the issue of 0%. We also did not add analytics to iOS for this so over there we have even less information to make this change on. We need to better understand these pieces before making changes to ensure we are fixing the right thing.

> Since we decided to go with in-memory, there was no need to look into token encryption at rest. That said the token is only kept for a short time, max 30 minutes and gets deleted right after tokenization succeeds/fails or the user returns normally.

This is a long time security wise. We are moving from no storage and expose for fractions of a second to holding it for 30 minutes at rest in memory. If we have not gone through a security review on this yet we should close these PRs until we do. This represents a real customers payment method so any store need to be reviewed and approved.

santugowda:
> > For the current implementation, we deliberately chose to keep this in memory rather than on disk, as discussed in the design review doc/comment. 56% of drop-offs at this stage retry within 5 minutes, that's the addressable population this phase targets.
> 
> 56% of what percent? It's really hard for teams to grock impact in these numbers and is misleading as it's a percent of a percent.
                                                                                                                                                                    
> 
> > Disk persistence was left as an open question because we didn't have data yet on how much of that 56% we're actually losing to process death vs recovering in-session.
> 
> The SDK is a platform and should act in a consistent way based on data. These PRs are a great departure from that idea, which is why there is some hesitation. Especially when we don't have data to yet prove that this 1) is the issue at all 2) scope of the issue. It feels pretty early to be pushing out a change when we don't know if the change will fix 100% of the issue of 0%. We also did not add analytics to iOS for this so over there we have even less information to make this change on. We need to better understand these pieces before making changes to ensure we are fixing the right thing. 

> 
> > Since we decided to go with in-memory, there was no need to look into token encryption at rest. That said the token is only kept for a short time, max 30 minutes and gets deleted right after tokenization succeeds/fails or the user returns normally.
> 
> This is a long time security wise. We are moving from no storage and expose for fractions of a second to holding it for 30 minutes at rest in memory. If we have not gone through a security review on this yet we should close these PRs until we do. This represents a real customers payment method so any store need to be reviewed and approved.

@jaxdesmarais 
Fair callout, I'll follow up with product to get the actual numbers behind it (including whether process-death vs. same-session recovery is broken out), along with iOS instrumentation details and post back here before we continue this discussion.  
Regarding 30-minute window, since the BA token's own expiry is 30 minutes, using the same TTL for the in-memory store rather than inventing a separate value. That said, agreed this still needs a formal security review given it's a live payment credential

jaxdesmarais:
I am going to move these back to draft while we wait on product/security confirmation

pedrofsn:
> I am going to move these back to draft while we wait on product/security confirmation

@jaxdesmarais just to make sure this doesn't stall on an unclear owner: are you going to be the one following up with Security to get that review kicked off, or is that on @santugowda's plate alongside the product numbers he mentioned above? Happy to leave it with whichever of you two is driving it-just want to avoid it sitting in limbo with nobody officially owning the next step.

jaxdesmarais:
@pedrofsn this will be on the inner sourcing team to track down/own as this is not a platform project for our team.

santugowda:
Hi @jaxdesmarais,  I got a chance to talk with iOS engineering(Juan Carlos)  about this token storage

He confirmed the in-memory holding of the BA token isn't new on iOS: `contextID` has held the token in memory since the moment the Hermes call completes, and this is already in production today.        
Question : Do you have any details I can use/refer to check the security review process. 

On Android, the implementation is different, and with this change , using in-memory holding - we're closing the gap with iOS, not introducing a new pattern or a new risk.

jaxdesmarais:
Closing while we wait on resolution to clean up GH open PRs - please reopen once we have answers on open questions

PR #1650, opened 2026-07-16
feat: dokka plugin to avoid RestrictTo classes

saralvasquez:
### Summary of changes
- Adds a new module to house a custom dokka plugin that will avoid building docs for classes with `RestrictTo` annotations
- Applies the plugin to all sub-projects that use dokka

Changelog entry not needed since this plugin isn't published with the SDK and it's only internal tooling. I cleaned and regenerated the docs to manually confirm this plugin is working as intended, but I kept the docs output separate from this PR since it's absolutely enormous

As a note this plugin only targets classes that have the restrict annotation. It still generates docs for constructors or methods that have the annotation. I did this since the ticket called out classes with the annotation only but I also see value in restricting everything with the annotation, so let me know what you all think.

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
I guided claude through my requirements and explain what I wanted done using the functionality described in [this documentation](https://kotlin.github.io/dokka/1.6.10/developer_guide/introduction/). I then let claude build the plugin and tweaked the output based on what I thought was incorrect

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @saralvasquez


mzlangreder:
I like the convenience of basing this off of existing `@RestrictTo` annotations, but I did see that Dokka also supports [suppress tags in the comment block](https://kotlinlang.org/docs/kotlin-doc.html#suppress). I do think you solution is better since then developers wont forget to add the tag, but did you see this as a potential solution?

mzlangreder (approved):

DokkaRestrictToPlugin/settings.gradle
mzlangreder, line 1: Is this needed because it's `includeBuild` at the top level project so it doesn't inherit the root project settings file?

saralvasquez: Correct. I could have avoided it by including it as a subproject, but I figured keeping it as a separate build made more sense since it's not really related to the SDK

saralvasquez:
> I like the convenience of basing this off of existing `@RestrictTo` annotations, but I did see that Dokka also supports [suppress tags in the comment block](https://kotlinlang.org/docs/kotlin-doc.html#suppress). I do think you solution is better since then developers wont forget to add the tag, but did you see this as a potential solution?

Yeah I did come across this and we have done it in some places in the SDK. But since it's a manual process it gets missed more often then not which is why I think this solution was suggested in the first place

DokkaRestrictToPlugin/src/main/kotlin/com/braintreepayments/dokka/HideRestrictedTransformer.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.jetbrains.dokka.plugability.DokkaContext
+
+/**
+ * Suppresses any [Documentable] that carries an `androidx.annotation.RestrictTo` annotation.
+ *
+ * Runs as a pre-merge transformer so suppression happens per source set before the multi-module
+ * merge, matching how the built-in `@suppress` KDoc filter and the android-documentation-plugin
+ * operate.
+ */
+public class HideRestrictedTransformer(context: DokkaContext) :
```
jaxdesmarais: General question because I can never remember exactly. Does Kotlin default to public or do we typically annotate public?

jaxdesmarais (approved):
Just a general question, not a blocker. I think this will be a great additional and make our soon to be versioned docs better. 🎉

DokkaRestrictToPlugin/src/main/kotlin/com/braintreepayments/dokka/HideRestrictedTransformer.kt
saralvasquez: Ah great catch actually. I'm 99% sure kotlin defaults to public so this probably isn't necessary. @mzlangreder correct me if I'm wrong

jaxdesmarais: That's what I think I remember too, but always get a bit backwards because the Swift default is internal. :see_no_evil:

mzlangreder: Yep public is default. Something that I like the idea of from a previous project is using [explicit API mode](https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors), which does help with this question because the compiler will force us to apply visibility modifiers and types to the code.

PR #1651, opened 2026-07-16
feat: Drop Gradle Toolchains

mzlangreder:
### Summary of changes
- Drop Gradle Toolchains Usage

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [ ] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
N/A

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry

### Authors
- @mzlangreder

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## unreleased
 
+* Build
+  * Drop Gradle Toolchains usage
 * AmericanExpress
     * Expose `getRewardsBalance` as a public suspend function
 * Card
--- a/AmericanExpress/build.gradle
+++ b/AmericanExpress/build.gradle
@@ -39,12 +39,6 @@ android {
     kotlinOptions {
         jvmTarget = versions.javaTargetCompatibility.toString()
     }
-
-    kotlin {
-        jvmToolchain {
-            languageVersion.set(JavaLanguageVersion.of(versions.javaTargetCompatibility.toString()))
-        }
-    }
 }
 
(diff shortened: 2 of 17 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## unreleased
 
+* Build
+  * Drop Gradle Toolchains usage
```
saralvasquez, line 6: I'm waffling on whether we need this or not. I recall you saying there might be different behavior for merchants without this, could you refresh my memory again? I only ask because normally we don't add changelog entries for build/infra type things

saralvasquez (approved):
Small non-blocking question, otherwise looks good! 🎉

CHANGELOG.md
mzlangreder: I felt similarly. There should not be any consumer impact, but I was thinking maybe it's good to call out for developer awareness?

There could be bytecode optimizations applied by the newer compiler, but because we target a specific API version and with D8/R8 desugaring everything should be converted to compatible code and shouldn't be a problem.

agedd (approved):

CHANGELOG.md
saralvasquez: Yeah ok I think more information in this case is better than less. It might not have much dev impact but it is a reasonable nugget of info folks could find useful and it would be difficult to discover it on your own

PR #1653, opened 2026-07-17
fix: Remove Visa Checkout

buzzamus:
### Summary of changes
- This PR removes the Visa Checkout module from the SDK since it has officially been archived and no longer in use

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used Claude to find and remove the majority of the Visa Checkout code. I also used Claude to verify there were tests still in place for if a configuration response has Visa Checkout enabled that we will simply ignore. Finally, Claude was used to find a bug where the project would not build after everything seemed to be safely deleted. It turns out one of the payment buttons was actually pointing to a Visa Checkout "Blue" color reference instead of the correct one (They were the same color, so no change)

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [x] Added a changelog entry

- @buzzamus


Demo/src/main/res/layout/fragment_ui_components.xml
```diff
@@ -70,7 +70,7 @@
                     android:layout_width="wrap_content"
                     android:layout_height="wrap_content"
                     android:layout_gravity="center_horizontal"
-                    custom:buttonColor="blue" />
+                    custom:paymentButtonColor="blue" />
```
saralvasquez, line 73: This looks unrelated to visa checkout

saralvasquez (commented):
If we are removing the visa checkout module all together is it worth deprecating values that can't be used instead of removing them?

Demo/src/main/res/layout/fragment_ui_components.xml
buzzamus: It is semi-related in that it caused a build error after removing Visa Checkout, because it was referencing the buttonColor attribute that was actually for the Visa Checkout button

saralvasquez: Ah! That makes sense

buzzamus:
> If we are removing the visa checkout module all together is it worth deprecating values that can't be used instead of removing them?

I sort of took 2 approaches: delete where we could delete and deprecate where we needed to.

So, for example the VisaCheckout module was fully deleted since it is never actually released and published. For the Configuration properties those were left and given a `@Deprecated` tag to avoid breaking any integrations where merchants might be using any values in there

saralvasquez:
> > If we are removing the visa checkout module all together is it worth deprecating values that can't be used instead of removing them?
> 
> I sort of took 2 approaches: delete where we could delete and deprecate where we needed to.
> 
> So, for example the VisaCheckout module was fully deleted since it is never actually released and published. For the Configuration properties those were left and given a `@Deprecated` tag to avoid breaking any integrations where merchants might be using any values in there

Ah I didn't notice that commented out publishing for VisaCheckout. This approach makes sense to me!

saralvasquez (approved):
🎉

mzlangreder (approved):

PR #1654, opened 2026-07-17
feat: Add Payment Actions Base Interfaces and Types

mzlangreder:
### Summary of changes
Added the following interfaces and types to support Payment Actions:
- PaymentActionsPaymentMethod
- PaymentActionResult
- PaymentAction
- PaymentActionSelectedPaymentMethod
- PaymentActionSelectedPaymentMethodDetails
- PaymentActionStatus
- NextAction
- PaymentActionsAnalytics

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used primarily for review after ingesting integration guide, ticket, and ADR documents. Then any findings that needed action or were good suggestions were added to the code.

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @mzlangreder


BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
```diff
@@ -0,0 +1,143 @@
...
+    EXPIRED,
+    UNKNOWN
+}
+
+/**
+ * A sealed class that represents the different shapes a Next Action can take when returned from
+ * GraphQL.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+sealed class NextAction {
```
mzlangreder: Pretty much everything else is prefixed with PaymentAction, would that make sense here too?
```suggestion
sealed class PaymentActionNextAction {
```

agedd: yep i think so, to keep it in alignment with everything else

agedd (approved):
looking good! 🚀 just left a small nit

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
```diff
@@ -0,0 +1,143 @@
+package com.braintreepayments.api.core.paymentactions
+
+import androidx.annotation.RestrictTo
+import org.json.JSONObject
+
+/**
+ * Defines a payment method that can be used with [PaymentActionsService].
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+interface PaymentActionsPaymentMethod {
```
jaxdesmarais: Should this file name match the interface name? I think we generally break down each interface/class into their own file but would default to @saralvasquez there on best practices.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * A string representing the selection set for the GraphQL call in
+     * [PaymentActionsService.setPaymentActionPaymentMethod].
+     *
+     * `selectedPaymentMethod.details` is omitted because it's a union type. Override this
+     * function when the response must include `selectedPaymentMethod.details`.
+     *
+     * Known `selectedPaymentMethod.details` types:
+     * - `CreditCardDetails`
```
jaxdesmarais: 🔥  docstings, these are gonna be so helpful for future us.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
```diff
@@ -0,0 +1,143 @@
...
+     */
+    fun paymentActionSelectionSet(): String = """
+        id
+        status
+        nextAction {
+            type
+            songbirdUrl
```
jaxdesmarais: Are we able to omit things the mobile SDKs don't need here? I think this one is web specific since we use the Cardinal SDKs directly. But it may be necessary so wasn't sure.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
```diff
@@ -0,0 +1,143 @@
...
+sealed class PaymentActionsResult {
+    class Success(val paymentAction: PaymentAction) : PaymentActionsResult()
+    class Failure(val error: Exception) : PaymentActionsResult()
+}
+
+/**
+ * DTO reflecting the shape of a payment action response.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+data class PaymentAction(
```
jaxdesmarais, line 43: Maybe `PaymentActionResult` is a good name for this? What do y'all think?

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
```diff
@@ -0,0 +1,143 @@
...
+ * A sealed class that represents the different shapes a Next Action can take when returned from
+ * GraphQL.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+sealed class NextAction {
+    data object ProvideCvv : NextAction()
+
+    data class Redirect(val redirectUrl: String) : NextAction()
+
+    data class ThreeDSecure(
```
jaxdesmarais: Should we omit these for now until we are ready for 3DS to keep it just scoped to cards?

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
mzlangreder: This is part of normal [Kotlin conventions](https://kotlinlang.org/docs/coding-conventions.html#source-file-organization)

> Placing multiple declarations (classes, top-level functions or properties) in the same Kotlin source file is encouraged as long as these declarations are closely related to each other semantically, and the file size remains reasonable (not exceeding a few hundred lines).

If we want to enforce or stick to one definition per-file as a team pattern, I can break this down no problem, but that is more of Java convention.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
mzlangreder: Yea we can, my original thinking here was to try and make the default act like a bit of documentation around what is available from GraphQL, and then each individual payment method can override the selection set based on what that payment method requires. Since the contract is still not finalized I figured I would add them here and then as we determine what is/isn't needed we can iterate. 

But, I wasn't aware we wouldn't need the songbirdUrl, so we can omit it here.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
mzlangreder: Current convention across the repository is that the `Result` types are wrappers for a given result object, so I stuck with a plain name for this object since it represents the shape of a PaymentAction on the GraphQL side. I think this matches up with what we want.

e.g. [CardResult](https://github.com/braintree/braintree_android/blob/main/Card/src/main/java/com/braintreepayments/api/card/CardResult.kt)

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
mzlangreder: This is a good question and something I waffled on a bit myself. If we want to limit everything we're adding to _only_ M3, we would:

- Remove the `nextAction` block from the selection set and related docstring notes
- Remove `PaymentActionNextAction` entirely
- Remove the nextAction field from `PaymentAction`
- Remove most of the enum types from `PaymentActionStatus` that don't apply
- Remove the ready for confirmation analytics definition

I like to try and fully model things we know at the data layer, even if unused for now because it acts as the truth for what's available from Payment Actions, especially since the Payment Method layer will insulate clients from being exposed to anything they don't need. I think if they are things we won't need, we shouldn't model them here like the `songbirdUrl` you mentioned. Thoughts?

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
jaxdesmarais: Interesting - I'll defer to Sara here on the current SDK structure.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
jaxdesmarais: Up to you, I am not sure what pattern exists, but I think on iOS at least we only add what we need and omit others. I am fine deferring to the pattern we have in the rest of the SDK. If we want to come back to it later maybe a `TODO` would be helpful so we don't forget what we do/don't need to come back to.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
jaxdesmarais: Ah, okay I missed this was the GQL shape, that makes sense. Should have been obvious by the scope but I missed it initially. :)

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
```diff
@@ -0,0 +1,142 @@
...
+ * Wrapper result type for [PaymentAction].
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+sealed class PaymentActionResult {
+    class Success(val paymentAction: PaymentAction) : PaymentActionResult()
+    class Failure(val error: Exception) : PaymentActionResult()
+}
+
+/**
+ * DTO reflecting the shape of a payment action response.
```
saralvasquez: Just for my knowledge, what does DTO stand for?

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
saralvasquez: Just since it's been on my mind recently, this class is marked `@RestrictTo` which will prevent dokka from building docs for this class. I assume this will remain restricted for the future right?

saralvasquez (approved):
I _think_ this looks good based on what I know about Payment Actions. Just a couple questions for my info

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
jaxdesmarais: Yeah, I guess my thinking is more of a say we get in a world where we want to release M3 but M4 is delayed. We would then have to pull some code out to avoid unused code being in the SDKs in that case. Up to y'all though - I think the most important thing is that we are consistent across the 2 platforms/PRs so will let you and @agedd decide on that.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
mzlangreder: Yep. The idea here is that what we're defining in core is an internal layer for Payment Actions, where our Payment Method modules will then make use of them internally and consumers will be totally isolated from them so that the API surface they're exposed to is minimal.

For the default selection set below I wanted it for documentation purposes, but wasn't able to include the details within `selectedPaymentMethod` since it is a union type that will expand as we add more payment method support. Because of this I added some details to the comments to hopefully make things more clear.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
mzlangreder: Data Transfer Object. In our case they are here to model and give us static types for the incoming data from GraphQL. But this raises a good point, I should not use acronyms :grin:

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
saralvasquez: As the SDK stands now we break everything out into it's own file unless it's an isolated helper object just for one class or something. You hit the nail right on the head with the Java convention part of that though. I think the structure now is a hold over from when we converted from java to kotlin. It seems like we just converted each file and called it a day without modifying structure along with content.

Considerations for how we want to move forward, I think the repo should definitely adhere to convention and we should make a Next Major Version (NMV) ticket to make that switch over to Kotlin conventions. But for now, do we follow kotlin convention for new additions or keep everything broken out until we make the switch in the NMV?

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
agedd: i am also okay with removing anything unrelated to m3 here (i.e. `nextAction` and `selectedPaymentMethod`) and having a `TODO` marker in-place so we can add those m4 specific types later on

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
agedd: agreed — think it may benefit us to trim now given the scenarios where m3 ships and then m4 slips well pas that - we'd have `PaymentActionNextAction`, the unused `PaymentActionStatus `cases, and the `nextAction` selection set sitting in a released SDK version with zero consumers for however long M4 takes.

Card/src/main/java/com/braintreepayments/api/card/paymentactions/CardPaymentActionsAnalytics.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import androidx.annotation.RestrictTo
+
+/**
+ * Payment Actions Analytics Keys.
+ *
+ * TODO: These keys are currently placeholder values that must be changed when we get the official values.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+object CardPaymentActionsAnalytics {
```
jaxdesmarais, line 11: Do we want to align these with the iOS PR: https://github.com/braintree/braintree_ios/pull/1830

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
```diff
@@ -0,0 +1,57 @@
...
+data class PaymentAction(
+    val id: String,
+    val status: PaymentActionStatus,
+)
+
+/**
+ * Enum of possible payment action status values.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+enum class PaymentActionStatus {
```
jaxdesmarais, line 52: Do we also need `READY_FOR_CONFIRMATION`?

mzlangreder: We will need `READY_FOR_CONFIRMATION`, but it is out of scope for M3 so I've removed it based on our earlier decision not to model everything we know about up front but instead to only include what's needed for M3.

Card/src/main/java/com/braintreepayments/api/card/paymentactions/CardPaymentActionsAnalytics.kt
mzlangreder: Updated the keys.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
jaxdesmarais: Ah - that makes sense! @agedd can we open a follow up iOS PR to align with @mzlangreder on this?

jaxdesmarais (approved):

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsDeclarations.kt
mzlangreder: This was discussed outside of GitHub comments and we agreed adopt standard conventions going forward.

PR #1656, opened 2026-07-22
fix: clean up and regenerate stale dokka docs

saralvasquez:
### Summary of changes
This PR is a follow up to the dokka plugin PR to clean up stale docs that never got removed. The docs should look exactly the same as what is currently being served on the pages site. This just cleans up old docs that aren't served anymore

Note: I know this PR is absolutely ENORMOUS. You can't really see the difference this PR makes without looking at the docs themselves. I had Claude set up a way for me to view the new docs side by side with the current ones. Upon a glace over everything two directory levels in it looks that same. I recommend folks who review do that as well so they can actually see the differences (or lack there of)

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
I asked claude to spin up a local server and forward my docs changes to it so that I could look at the newly generated docs

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @saralvasquez


mzlangreder (approved):
Checked out the doc sites locally and everything looks good to me.

jaxdesmarais (approved):
Also pulled down locally and scrolled through - all looks correct. Thanks for including the claude tip to do this, helped me a ton with the preview.

PR #1658, opened 2026-07-27
feat: Create payment actions service

mzlangreder:
### Summary of changes
- Create PaymentActionsService
- Create unit tests

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Review ADR to match up expected graphQL shapes and writing unit tests.

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @mzlangreder

### DIFF
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
@@ -1,0 +1,94 @@
+package com.braintreepayments.api.core.paymentactions
+
+import androidx.annotation.RestrictTo
+import com.braintreepayments.api.core.BraintreeClient
+import com.braintreepayments.api.core.BraintreeException
+import com.braintreepayments.api.core.GraphQLConstants
+import org.json.JSONException
+import org.json.JSONObject
+import java.io.IOException
+
+/**
+ *
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class PaymentActionsService(
+    private val braintreeClient: BraintreeClient,
+) {
+
+    suspend fun setPaymentActionPaymentMethod(
+        paymentMethod: PaymentActionPaymentMethod,
+    ): PaymentActionResult {
+        return try {
+            braintreeClient
+                .sendGraphQLPOST(buildQuery(paymentMethod))
+                .toPaymentActionsResult()
+        } catch (exception: IOException) {
+            PaymentActionResult.Failure(exception)
+        }
+    }
+
+    private fun buildQuery(paymentMethod: PaymentActionPaymentMethod): JSONObject {
+        // TODO: SetPaymentActionPaymentMethodInput also requires a paymentActionId. Determine
+        //  where this should be sourced from and thread it through once available.
+        val input = JSONObject().put(PAYMENT_METHOD_KEY, paymentMethod.toGraphQLVariables())
+        val variables = JSONObject().put(GraphQLConstants.Keys.INPUT, input)
+
+        val mutation = """
+            mutation SetPaymentActionPaymentMethod(${'$'}input: SetPaymentActionPaymentMethodInput!) {
+                setPaymentActionPaymentMethod(input: ${'$'}input) {
+                    paymentAction {
+                        ${paymentMethod.paymentActionSelectionSet()}
+                    }
+                }
+            }
+        """.trimIndent()
+
+        return JSONObject()
+            .put(GraphQLConstants.Keys.QUERY, mutation)
+            .put(GraphQLConstants.Keys.VARIABLES, variables)
+    }
+
+    private fun String.toPaymentActionsResult(): PaymentActionResult {
+        return try {
+            val responseJson = JSONObject(this)
+            if (responseJson.has(GraphQLConstants.Keys.ERRORS) &&
+                responseJson.getJSONArray(GraphQLConstants.Keys.ERRORS).length() > 0
+            ) {
+                return PaymentActionResult.Failure(BraintreeException(responseJson.toString()))
+            }
+
+            responseJson
+                .getJSONObject(DATA_KEY)
+                .getJSONObject(SET_PAYMENT_ACTION_PAYMENT_METHOD_KEY)
+                .getJSONObject(PAYMENT_ACTION_KEY)
+                .let { paymentActionJson ->
+                    PaymentAction(
+                        id = paymentActionJson.getString(ID_KEY),
+                        status = paymentActionJson.getString(STATUS_KEY).toPaymentActionStatus()
+                    ).let { paymentAction ->
+                        PaymentActionResult.Success(paymentAction)
+                    }
+                }
+        } catch (jex: JSONException) {
+            PaymentActionResult.Failure(jex)
+        }
+    }
+
+    private fun String.toPaymentActionStatus(): PaymentActionStatus {
+        return try {
+            PaymentActionStatus.valueOf(this.uppercase())
+        } catch (_: IllegalArgumentException) {
+            PaymentActionStatus.UNKNOWN
+        }
+    }
+
+    companion object {
+        private const val PAYMENT_METHOD_KEY = "paymentMethod"
+        private const val DATA_KEY = "data"
+        private const val SET_PAYMENT_ACTION_PAYMENT_METHOD_KEY = "setPaymentActionPaymentMethod"
+        private const val PAYMENT_ACTION_KEY = "paymentAction"
+        private const val ID_KEY = "id"
+        private const val STATUS_KEY = "status"
+    }
+}
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
@@ -9,26 +9,37 @@ import org.json.JSONObject
 import java.io.IOException
 
 /**
- *
+ * This is a centralized service for handling all Payment Actions related backend requests.
+ * It acts as a thin client in order to insulate our payment method specific clients from the
+ * networking layer.
  */
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 class PaymentActionsService(
     private val braintreeClient: BraintreeClient,
 ) {
 
+    /**
+     * Submits a payment method to `setPaymentActionPaymentMethod` GraphQL mutation.
+     *
+     * On success returns a [PaymentActionResult.Success] wrapping a [PaymentAction] is returned.
+     * On failure returns a [PaymentActionResult.Failure] wrapping an [Exception] is returned.
+     *
+     * @param paymentMethod the [PaymentActionPaymentMethod] to submit to GraphQL.
+     * @return a [PaymentActionResult] wrapper denoting success or failure.
+     */
     suspend fun setPaymentActionPaymentMethod(
         paymentMethod: PaymentActionPaymentMethod,
     ): PaymentActionResult {
         return try {
             braintreeClient
-                .sendGraphQLPOST(buildQuery(paymentMethod))
+                .sendGraphQLPOST(buildSetPaymentActionPaymentMethodQuery(paymentMethod))
                 .toPaymentActionsResult()
         } catch (exception: IOException) {
             PaymentActionResult.Failure(exception)
         }
     }
 
-    private fun buildQuery(paymentMethod: PaymentActionPaymentMethod): JSONObject {
+    private fun buildSetPaymentActionPaymentMethodQuery(paymentMethod: PaymentActionPaymentMethod): JSONObject {
         val input = JSONObject().put(PAYMENT_METHOD_KEY, paymentMethod.toGraphQLVariables())
         val variables = JSONObject().put(GraphQLConstants.Keys.INPUT, input)
 
--- a/BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
+++ b/BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
@@ -29,8 +29,6 @@ class PaymentActionsService(
     }
 
     private fun buildQuery(paymentMethod: PaymentActionPaymentMethod): JSONObject {
-        // TODO: SetPaymentActionPaymentMethodInput also requires a paymentActionId. Determine
-        //  where this should be sourced from and thread it through once available.
         val input = JSONObject().put(PAYMENT_METHOD_KEY, paymentMethod.toGraphQLVariables())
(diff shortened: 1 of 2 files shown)


BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
```diff
@@ -0,0 +1,103 @@
...
+                .getJSONObject(PAYMENT_ACTION_KEY)
+                .let { paymentActionJson ->
+                    PaymentAction(
+                        id = paymentActionJson.getString(ID_KEY),
+                        status = paymentActionJson.getString(STATUS_KEY).toPaymentActionStatus()
+                    ).let { paymentAction ->
+                        PaymentActionResult.Success(paymentAction)
+                    }
+                }
+        } catch (jex: JSONException) {
```
saralvasquez, line 82: Interesting naming convention `jex`. Just out of curiosity, is that a personal preference or style standard?

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
```diff
@@ -0,0 +1,103 @@
...
+     *
+     * @param paymentMethod the [PaymentActionPaymentMethod] to submit to GraphQL.
+     * @return a [PaymentActionResult] wrapper denoting success or failure.
+     */
+    suspend fun setPaymentActionPaymentMethod(
+        paymentMethod: PaymentActionPaymentMethod,
+    ): PaymentActionResult {
+        return try {
+            braintreeClient
+                .sendGraphQLPOST(buildSetPaymentActionPaymentMethodQuery(paymentMethod))
```
saralvasquez, line 35: nit: since the method name is so long this is starting to feel a bit difficult to parse in a single line, but I don't feel super strongly

BraintreeCore/src/test/java/com/braintreepayments/api/core/paymentactions/PaymentActionsServiceUnitTest.kt
```diff
@@ -0,0 +1,527 @@
...
+import kotlin.test.assertIs
+
+@Suppress("MaxLineLength")
+class PaymentActionsServiceUnitTest {
+
+    private val testDispatcher = StandardTestDispatcher()
+
+    private val paymentMethodVariables = JSONObject().apply {
+        put(
+            "paymentMethodDetails",
```
saralvasquez, line 29: Just double checking it should be `paymentMethodDetails` and not `paymentMethod`. I don't think I've seen what the payment method input is supposed to look like yet

BraintreeCore/src/test/java/com/braintreepayments/api/core/paymentactions/PaymentActionsServiceUnitTest.kt
```diff
@@ -0,0 +1,527 @@
...
+
+    private fun mockPaymentMethod(
+        variables: JSONObject = paymentMethodVariables,
+        selectionSet: String = "id\nstatus",
+    ) = mockk<PaymentActionPaymentMethod> {
+        every { toGraphQLVariables() } returns variables
+        every { paymentActionSelectionSet() } returns selectionSet
+    }
+
+    @OptIn(ExperimentalCoroutinesApi::class)
```
saralvasquez: Is this something we can put on the class so it doesn't need to be added to every test?

BraintreeCore/src/test/java/com/braintreepayments/api/core/paymentactions/PaymentActionsServiceUnitTest.kt
```diff
@@ -0,0 +1,527 @@
...
+    @Test
+    fun `when responseBody status is requires_payment_method, Success is returned with REQUIRES_PAYMENT_METHOD status`() =
+        runTest(testDispatcher) {
+            val responseBody = """
+                {
+                    "data": {
+                        "setPaymentActionPaymentMethod": {
+                            "paymentAction": {
+                                "id": "pa123",
+                                "status": "requires_payment_method"
```
saralvasquez: I think this is more of a general feature question but I can't seem to answer it using the technical decision docs I've seen. What use case will the `setPaymentActionPaymentMethod` query be called and return a success but also have the status `requires_payment_method`. Isn't that status to signify that `setPaymentActionPaymentMethod` has to be called?

Again I'm probably under informed so please let me know if there's a doc I need to be referring to

saralvasquez (commented):
This looks great! Just a couple questions and nits

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
```diff
@@ -0,0 +1,103 @@
...
+            }
+
+            responseJson
+                .getJSONObject(DATA_KEY)
+                .getJSONObject(SET_PAYMENT_ACTION_PAYMENT_METHOD_KEY)
+                .getJSONObject(PAYMENT_ACTION_KEY)
+                .let { paymentActionJson ->
+                    PaymentAction(
+                        id = paymentActionJson.getString(ID_KEY),
+                        status = paymentActionJson.getString(STATUS_KEY).toPaymentActionStatus()
```
agedd, line 77: same thing came up on iOS: `getString` throws if `id` is missing, but not if it's an empty string — so that currently returns a Success with an unusable id. worth adding a check so empty fails the same way as missing, matching iOS.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
```diff
@@ -0,0 +1,103 @@
...
+     * @return a [PaymentActionResult] wrapper denoting success or failure.
+     */
+    suspend fun setPaymentActionPaymentMethod(
+        paymentMethod: PaymentActionPaymentMethod,
+    ): PaymentActionResult {
+        return try {
+            braintreeClient
+                .sendGraphQLPOST(buildSetPaymentActionPaymentMethodQuery(paymentMethod))
+                .toPaymentActionsResult()
+        } catch (exception: IOException) {
```
agedd, line 37: wondering if we should widen the exception type or if it's okay as-is 👀 in the case `sendGraphQLPOST` can throw anything besides `IOException`

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
```diff
@@ -0,0 +1,103 @@
...
+                    }
+                }
+        } catch (jex: JSONException) {
+            PaymentActionResult.Failure(jex)
+        }
+    }
+
+    private fun String.toPaymentActionStatus(): PaymentActionStatus {
+        return try {
+            PaymentActionStatus.valueOf(this.uppercase())
```
noguier, line 89: is status required to be uppercase in out implementation?

BraintreeCore/src/test/java/com/braintreepayments/api/core/paymentactions/PaymentActionsServiceUnitTest.kt
noguier, line 29: Seconding Sara's question, I am assuming that this is how the response will be returned from the backend, but just wanted to double check

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
jaxdesmarais: ➕ 1 - haven't seen this pattern before

mzlangreder: It's pretty common in Java/Kotlin land where you're often catching different types of exceptions. In this specific case it's not really needed since there aren't any other catch blocks. It's definitely not a hard and fast rule though.

Happy to change it if we like.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
mzlangreder: Yea it is long. Initially I had something like "buildQuery", but eventually we'll have other calls such as confirmCustomerActionComplete so I went with a heuristic of `build` + `mutationName` + `Query` and it ended up long. 

Any suggestions as to how we could improve it?

BraintreeCore/src/test/java/com/braintreepayments/api/core/paymentactions/PaymentActionsServiceUnitTest.kt
mzlangreder: Yep, that portion of the query is a `@oneof` where either it accepts a `paymentMethodId` or a `paymentMethodDetails`. It's not build in this PR yet because we haven't added the CreditCard related things yet.

BraintreeCore/src/test/java/com/braintreepayments/api/core/paymentactions/PaymentActionsServiceUnitTest.kt
mzlangreder: Nah this is a good thing to call out. Even though the mocked value here is not important _for this test_ specifically when it comes to the actual expected values, we should still keep our tests modeled after what we'd expect.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
mzlangreder: Good catch, thanks! More broadly this block is wrapped in a try catch, where if any JSON parsing fails we return a `PaymentActionResult.Failure`. Do you think that covers us or did iOS implement a different type of fallback? In my head if we fail to get an id or status back, something has probably gone quite wrong and we should tell the consumer to bail out.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
mzlangreder: I think this is OK as it is unless, as an SDK, we tend to prefer catching everything all in one go just in case.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
mzlangreder: What's going on here is that enum values coming from graphQL are lower_snake_case, but Enums should be SCREAMING_SNAKE_CASE e.g. `READY_FOR_CONFIRMATION`.

What this is doing is taking that string value and upcasing it so that when we use `valueOf`, the value will be able to match our Enum values.

BraintreeCore/src/test/java/com/braintreepayments/api/core/paymentactions/PaymentActionsServiceUnitTest.kt
mzlangreder: `paymentMethodDetails` is something we will pass in our requests, yep. In the case of Credit Cards it would be an object containing the required Credit Card data needed for submitPaymentActionPaymentMethod mutation.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
jaxdesmarais: Yeah looks like other places in the SDK we just use `e` which isn't terribly descriptive either. I feel like these types of shorthands are a lot more common on Android vs iOS tends to be more verbose and would likely do something like `jsonException` or `exception`. I don't feel strongly either way, more just an observation. :)

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
jaxdesmarais: Yeah, that's what the iOS PR does: throws an error if status/id does not exist

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
jaxdesmarais: It looks like in most places we catch the `Exception` and sometimes cast it to a `BraintreeException`. But graphQL is only used in a few modules (venmo, card, shopper insights). I think we can defer to @mzlangreder @noguier and @saralvasquez on what's best for Android.

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
saralvasquez: Yeah I don't feel strongly, but I do like that it's more descriptive than just `e` which is what I've seen most of the time

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
saralvasquez: Yeah I hadn't thought about other mutations coming down the line. This might be a bit heavy handed for a readability improvement, but what if there was a data model/spec of some kind holding the data for the different mutations and then a more generalized `buildQuery` that the spec is fed to. So it would be more like `buildQuery(SET_PAYMENT_METHOD_SPEC, paymentMethod)`. Not sure if that even simplifies things since I'm not sure what future mutations will look like. But that's the only thing that came to mind besides just shortening the method name by removing the `PaymentAction` part since that's implied by it's location in the codebase

BraintreeCore/src/main/java/com/braintreepayments/api/core/paymentactions/PaymentActionsService.kt
mzlangreder: I'm not opposed to full names either, I am a fan of more explicitness everywhere :grin:. Maybe something we can settle on outside of this PR?

jaxdesmarais: Oh for sure, not a blocker here at all. More of a "huh this is a style thing we should be consistent on".

agedd (approved):
🚀

jaxdesmarais (approved):
None of my questions are blockers, but we should likely align iOS and Android on the check for ID and status. That can be in a follow up though if needed. Approving ahead of time in any case as most of the comments here are related to smaller style things vs blocking.

PR #1663, opened 2026-08-04
Feat: Add Reference Docs Versioning

buzzamus:
### Summary of changes
- This PR moves Dokka output off of the main branch (docs/) and into a git ignored build directory, then publishes it to a separate `reference-docs` branch via the new publish-docs.sh script.

A follow up PR to generate the docs will follow once this is merged to remove docs/ and then switch over to new docs.

- to test/preview docs you can follow these steps locally:

1. run the publish script with `./scripts/publish-docs.sh 5.30.0-test`
2. cd ../braintree_android-reference-docs
3. preview the docs in the browser (an easy way to set up a local server is with `python3 -m http.server 8000` and then go to http://localhost:8000/ → should redirect to current/index.html and then also try http://localhost:8000/5.30.0-test/

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used for initial research, creating bash script to mimic iOS docs behavior

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/scripts/publish-docs.sh
+++ b/scripts/publish-docs.sh
@@ -1,15 +1,15 @@
 #!/usr/bin/env bash
-# Publishes Dokka reference docs for a single version to the reference-docs branch,
+# Publishes Dokka reference docs for a single version to the gh-pages branch,
 # mirroring how braintree_ios publishes Jazzy docs: one folder per version plus
 # a `current` symlink to the latest, no version-picker UI.
 #
-# This script only commits locally to the reference-docs branch checked out in a
+# This script only commits locally to the gh-pages branch checked out in a
 # worktree next to the repo. It intentionally does not push - the caller
 # (e.g. the release workflow) is responsible for pushing when ready.
 set -euo pipefail
 
 VERSION="${1:?Usage: publish-docs.sh <version>}"
-BRANCH="reference-docs"
+BRANCH="gh-pages"
 
 REPO_ROOT="$(git -C "$(dirname "${BASH_SOURCE[0]}")" rev-parse --show-toplevel)"
 DOCS_SOURCE="${REPO_ROOT}/build/dokkaDocs"
@@ -25,7 +25,7 @@ if ! git -C "${REPO_ROOT}" worktree list --porcelain | grep -qx "worktree ${WORK
     git -C "${REPO_ROOT}" worktree add "${WORKTREE_DIR}" "${BRANCH}"
   else
     git -C "${REPO_ROOT}" worktree add --orphan -b "${BRANCH}" "${WORKTREE_DIR}"
-    git -C "${WORKTREE_DIR}" commit --allow-empty -m "Initialize reference-docs branch" --quiet
+    git -C "${WORKTREE_DIR}" commit --allow-empty -m "Initialize gh-pages branch" --quiet
   fi
 fi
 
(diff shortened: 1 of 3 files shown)


noguier:
Can you give me a context why the decision was made to move the docs into a separate branch?

buzzamus:
> Can you give me a context why the decision was made to move the docs into a separate branch?

Yep! I sent you the document on the outlined possible solutions. This idea that was settled on was to match what iOS already does, which moves docs off of `main`

scripts/publish-docs.sh
```diff
@@ -0,0 +1,67 @@
...
+REPO_ROOT="$(git -C "$(dirname "${BASH_SOURCE[0]}")" rev-parse --show-toplevel)"
+DOCS_SOURCE="${REPO_ROOT}/build/dokkaDocs"
+WORKTREE_DIR="$(dirname "${REPO_ROOT}")/$(basename "${REPO_ROOT}")-${BRANCH}"
+
+if [ ! -d "${DOCS_SOURCE}" ]; then
+  echo "No generated docs found at ${DOCS_SOURCE}. Run ./gradlew dokkaHtmlMultiModule first." >&2
+  exit 1
+fi
+
+if ! git -C "${REPO_ROOT}" worktree list --porcelain | grep -qx "worktree ${WORKTREE_DIR}"; then
```
noguier, line 23: The script looks great, but i have difficulty parsing through this conditions. 
As far as I understand we first check if this directory already exist, 
if not we check if the local branch with that name exists, 
if it does we add it to the new folder,
if not we create a orphaned new branch. 

Is that roughly correct?

noguier (approved):

scripts/publish-docs.sh
buzzamus: Pretty much yep! It checks to see if it already exists in the git worktree and if it is just use it, otherwise create it. This is to ensure it is idempotent and really only have to worry about setup on first run then after that a new version folder is added on each run

saralvasquez (commented):
This looks great! Just to cover all bases. I know there were cases where old docs files that weren't served anymore would build up since the docs weren't cleaned before they were built again for the new version. It seems like that would also be the case here since we building into the same directory each release and then copying that to a new version directory. Have you run into that problem with the iOS docs before with this method?

Would it make sense to add an extra step to clean stuff up before building or something? Or I think if you call `clean` before the dokka task it might take care of that

buzzamus:
> This looks great! Just to cover all bases. I know there were cases where old docs files that weren't served anymore would build up since the docs weren't cleaned before they were built again for the new version. It seems like that would also be the case here since we building into the same directory each release and then copying that to a new version directory. Have you run into that problem with the iOS docs before with this method?
> 
> Would it make sense to add an extra step to clean stuff up before building or something? Or I think if you call `clean` before the dokka task it might take care of that

It will look the same as this branch on iOS: https://github.com/braintree/braintree_ios/tree/gh-pages

Right now I have the branch named `reference-docs`, but going to change name to match iOS by using `gh-pages`.  The reason there aren't any docs yet is that I was waiting for reviews before running the script in order to avoid a bunch of noise, so there will be a 2nd PR opened up for that

buzzamus:
Updated the branch name for branch that will serve docs to `gh-pages` to match iOS docs branch

buzzamus:
@saralvasquez you can preview what it will look like here -> https://github.com/braintree/braintree_android/tree/gh-pages

saralvasquez (approved):
🎉

PR #1667, opened 2026-08-06
fix: handle shopper insights errors correctly

saralvasquez:
### Summary of changes
We noticed an issue with Shopper Insights V2 in the demo app and after digging into that we noticed that error handling for the responses to create and update the customer session was broken. We were getting a JSON exception instead of the proper GQL error in response. This PR fixes the error handling so that the correct error is surfaced

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used claude to find GQL error handling pattern and for writing unit tests

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry

### Authors
- @saralvasquez

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Braintree Android SDK Release Notes
 
+## unreleased
+
+* ShopperInsights
+    * Fix response handling bug that buries customer session create and update errors
+
 ## 5.31.0 (2026-08-03)
 
 * PayPal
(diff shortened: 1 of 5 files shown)


jaxdesmarais (commented):
We probably want a changelog for this one. Feels like a bug fix.

saralvasquez:
Ok I cannot get this PR out of this state waiting for checks to pass for the life of me. I'm gonna close this permanently and open a new PR

PR #1668, opened 2026-08-06
fix: handle shopper insights errors correctly

saralvasquez:
### Summary of changes
We noticed an issue with Shopper Insights V2 in the demo app and after digging into that we noticed that error handling for the responses to create and update the customer session was broken. We were getting a JSON exception instead of the proper GQL error in response. This PR fixes the error handling so that the correct error is surfaced

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used claude to find GQL error handling pattern and for writing unit tests

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry

### Authors
- @saralvasquez


saralvasquez:
And it's still stuck...

jaxdesmarais (approved):

ShopperInsights/src/test/java/com/braintreepayments/api/shopperinsights/v2/internal/GenerateCustomerRecommendationsApiUnitTest.kt
```diff
@@ -196,6 +197,38 @@ class GenerateCustomerRecommendationsApiUnitTest {
         assertEquals(error, (result as GenerateCustomerRecommendationsResult.Error).error)
     }
 
+    @OptIn(ExperimentalCoroutinesApi::class)
```
noguier, line 200: If errors is an array, can there be more than one error in that array, should we have a test case for that?

saralvasquez: great catch! yeah there should be. I'll add that

noguier (approved):
Thank you for adding an extra test case! this looks great!

PR #1671, opened 2026-08-07
feat(PaymentActions): Update and add data models

mzlangreder:
### Summary of changes
Updating and adding data models to support the single client Payment Actions design.

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Reviewing docs and helping to set up models to match expected object models from the designs.

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @mzlangreder

---

### DIFF
--- a/PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
+++ b/PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
@@ -1,46 +1,116 @@
 package com.braintreepayments.api.paymentactions
 
-import androidx.annotation.RestrictTo
 import org.json.JSONObject
 
 /**
  * Defines a payment method that can be used with [PaymentActionsService].
  */
-@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-interface PaymentActionPaymentMethod {
+sealed class PaymentActionRequest {
+
     /**
      * Produces the set of GraphQL variables needed to satisfy the GraphQL call in
      * [PaymentActionsService.setPaymentActionPaymentMethod].
      */
-    fun toGraphQLVariables(): JSONObject
+    internal abstract fun toGraphQLVariables(): JSONObject
 
     /**
      * A string representing the minimal selection set for the GraphQL call in
      * [PaymentActionsService.setPaymentActionPaymentMethod].
      */
-    fun paymentActionSelectionSet(): String = """
+    internal open fun paymentActionSelectionSet(): String = """
         id
         status
     """.trimIndent()
 }
 
 /**
- * Wrapper result type for [PaymentAction].
+ * Unvaulted credit card input for a payment action.
  */
-@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-sealed class PaymentActionResult {
-    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-    class Success(val paymentAction: PaymentAction) : PaymentActionResult()
+class CreditCard(
+    val number: String,
+    val expirationMonth: String,
+    val expirationYear: String,
+    val cvv: String? = null,
+    val cardholderName: String? = null,
+    val streetAddress: String? = null,
+    val extendedAddress: String? = null,
+    val locality: String? = null,
+    val region: String? = null,
+    val postalCode: String? = null,
+    val countryCodeAlpha2: String? = null,
+) : PaymentActionRequest() {
 
-    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-    class Failure(val error: Exception) : PaymentActionResult()
+    override fun toGraphQLVariables(): JSONObject {
+        return JSONObject().apply {
+            streetAddress?.let { put(PaymentActionsGraphQLConstants.Keys.BILLING_ADDRESS_STREET_ADDRESS, it) }
+            extendedAddress?.let { put(PaymentActionsGraphQLConstants.Keys.BILLING_ADDRESS_EXTENDED_ADDRESS, it) }
+            locality?.let { put(PaymentActionsGraphQLConstants.Keys.BILLING_ADDRESS_LOCALITY, it) }
+            region?.let { put(PaymentActionsGraphQLConstants.Keys.BILLING_ADDRESS_REGION, it) }
+            postalCode?.let { put(PaymentActionsGraphQLConstants.Keys.BILLING_ADDRESS_POSTAL_CODE, it) }
+            countryCodeAlpha2?.let {
+                put(PaymentActionsGraphQLConstants.Keys.BILLING_ADDRESS_COUNTRY_CODE_ALPHA_2, it)
+            }
+        }.let { billingAddressFields ->
+            JSONObject().apply {
+                put(PaymentActionsGraphQLConstants.Keys.CREDIT_CARD_NUMBER, number)
+                put(PaymentActionsGraphQLConstants.Keys.CREDIT_CARD_EXPIRATION_MONTH, expirationMonth)
+                put(PaymentActionsGraphQLConstants.Keys.CREDIT_CARD_EXPIRATION_YEAR, expirationYear)
+                cvv?.let { put(PaymentActionsGraphQLConstants.Keys.CREDIT_CARD_CVV, it) }
+                cardholderName?.let { put(PaymentActionsGraphQLConstants.Keys.CREDIT_CARD_CARDHOLDER_NAME, it) }
+                if (billingAddressFields.length() > 0) {
+                    put(PaymentActionsGraphQLConstants.Keys.CREDIT_CARD_BILLING_ADDRESS, billingAddressFields)
+                }
+            }
+        }.let { creditCardFields ->
+            JSONObject().put(PaymentActionsGraphQLConstants.Keys.CREDIT_CARD, creditCardFields)
+        }.let { paymentMethodDetails ->
+            JSONObject().put(PaymentActionsGraphQLConstants.Keys.PAYMENT_METHOD_DETAILS, paymentMethodDetails)
+        }
+    }
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is CreditCard) return false
+        return number == other.number &&
+            expirationMonth == other.expirationMonth &&
+            expirationYear == other.expirationYear &&
+            cvv == other.cvv &&
+            cardholderName == other.cardholderName &&
+            streetAddress == other.streetAddress &&
+            extendedAddress == other.extendedAddress &&
+            locality == other.locality &&
+            region == other.region &&
+            postalCode == other.postalCode &&
+            countryCodeAlpha2 == other.countryCodeAlpha2
+    }
+
+    override fun hashCode(): Int {
+        var result = number.hashCode()
+        result = 31 * result + expirationMonth.hashCode()
+        result = 31 * result + expirationYear.hashCode()
+        result = 31 * result + (cvv?.hashCode() ?: 0)
+        result = 31 * result + (cardholderName?.hashCode() ?: 0)
+        result = 31 * result + (streetAddress?.hashCode() ?: 0)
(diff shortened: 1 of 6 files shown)


jaxdesmarais:
Looks like some tests are failing if we want to address those

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
```diff
@@ -4,36 +4,109 @@ import androidx.annotation.RestrictTo
...
-    fun paymentActionSelectionSet(): String = """
+    internal open fun paymentActionSelectionSet(): String = """
         id
         status
     """.trimIndent()
 }
 
 /**
- * Wrapper result type for [PaymentAction].
+ * unvaulted credit card input for a payment action.
```
agedd: small nit: 
```suggestion
 * Unvaulted credit card input for a payment action.
```

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
```diff
@@ -52,9 +125,73 @@ data class PaymentAction(
...
+ */
+sealed class PaymentActionResult {
+    class Success(val outcome: PaymentActionOutcome) : PaymentActionResult()
+    class Failure(val error: Exception) : PaymentActionResult()
+}
+
+/**
+ * Normalized outcome of a payment action, derived from a [PaymentAction]'s status.
+ */
+sealed class PaymentActionOutcome {
```
agedd: qq: do we need to define this class and are these `PaymentActionOutcome` types confirmed 👀

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
```diff
@@ -52,9 +125,73 @@ data class PaymentAction(
...
+    class PaymentMethodRequired(val paymentAction: PaymentAction) : PaymentActionOutcome()
+    class CustomerActionRequired(
+        val paymentAction: PaymentAction,
+    ) : PaymentActionOutcome()
+}
+
+/**
+ * A server-driven action the merchant/SDK must perform to advance a payment action.
+ */
+enum class ServerAction {
```
agedd, line 180: same question as above - do we need to expose this in here 👀

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
mzlangreder: My thinking is this will be our "Result" type. A "what do we need to communicate to the consumer so that it's clear to them what the status is and also what information they need for a given outcome. I reckon these types could use some refinement and also some sync up with Web to see if they were thinking similarly

In a later commit I refined it by dropping `PaymentActionResult` completely and replacing it with this class to reduce nesting.

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
mzlangreder: At the moment it's part of `PaymentActionOutcome`, so yes. 

Here it would work like:
1. `submitForPaymentAction`
2. Get a `PaymentActionOutcome.ServerActionRequired("some id", ServerAction.Confirm`
3. Merchant performs confirmation.

Do you think this makes sense as a usage pattern?

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
```diff
@@ -4,36 +4,109 @@ import androidx.annotation.RestrictTo
...
+                }
+            }
+        }.let { creditCardFields ->
+            JSONObject().put(PaymentActionsGraphQLConstants.Keys.CREDIT_CARD, creditCardFields)
+        }.let { paymentMethodDetails ->
+            JSONObject().put(PaymentActionsGraphQLConstants.Keys.PAYMENT_METHOD_DETAILS, paymentMethodDetails)
+        }
+    }
+
+    override fun equals(other: Any?): Boolean {
```
jaxdesmarais, line 71: not blocking just for my own knowledge what is `equals` and `hashCode` used for? I have not seen us use these before.

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
jaxdesmarais: `Outcome` is not something we've used before as language in the SDK (at least for public things returned to merchants). Do we want to consider using `Result` instead to stay consistent or is there a more explicit reason we want to move to `Outcome`?

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
```diff
@@ -52,9 +125,73 @@ data class PaymentAction(
...
+    class Success(val outcome: PaymentActionOutcome) : PaymentActionResult()
+    class Failure(val error: Exception) : PaymentActionResult()
+}
+
+/**
+ * Normalized outcome of a payment action, derived from a [PaymentAction]'s status.
+ */
+sealed class PaymentActionOutcome {
+    class Completed(val paymentAction: PaymentAction) : PaymentActionOutcome()
+    class Cancelled(val paymentAction: PaymentAction) : PaymentActionOutcome()
```
jaxdesmarais: nit: the SDKs use the 1 l canceled so we can use the same pattern here.

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
mzlangreder: I've iterated on the types a bit in the next PR. I'm going to bring those changes down in to this PR so it's more clear what the shape is.

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
mzlangreder: Without overriding `equals` the method falls back to checking reference equality.
Without overriding `hashCode` two instances of the same class that appear equal will still have different hash codes to tell them apart.

I added them because normally we'd get these implementations from free by using `data class` instead of `class`, but using `data class` in libraries [should be avoided](https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api).

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
```diff
@@ -4,36 +4,109 @@ import androidx.annotation.RestrictTo
 import org.json.JSONObject
 
 /**
- * Defines a payment method that can be used with [PaymentActionsService].
+ * Base type for merchant-facing input submitted to a payment action via
+ * `PaymentActionsClient` (added in a follow-up task).
  */
-@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-interface PaymentActionPaymentMethod {
+sealed class PaymentActionRequest {
```
noguier, line 8: just out of curiosity why did we switch interface with `RestrictTo` to an sealed class?

PaymentActions/src/test/java/com/braintreepayments/api/paymentactions/CreditCardUnitTest.kt
```diff
@@ -0,0 +1,120 @@
...
+                }
+            }
+            """.trimIndent()
+        )
+
+        JSONAssert.assertEquals(expected, creditCard.toGraphQLVariables(), true)
+    }
+
+    @Test
+    fun `paymentActionSelectionSet returns the default id and status selection`() {
```
noguier, line 109: according to the KDoc in SelectionSet : `the minimal set can be overridden` , should we create a test for it?

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
jaxdesmarais: Nice! Thanks for the explanation. :)

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
jaxdesmarais: I think that makes sense - though my main question is still around a new result type, but not sure if those changes you mentioned have been pulled into this PR yet.

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsGraphQLConstants.kt
```diff
@@ -0,0 +1,21 @@
+package com.braintreepayments.api.paymentactions
+
+internal object PaymentActionsGraphQLConstants {
+
+    object Keys {
```
jaxdesmarais, line 5: I noticed some places we use this `Constants.Keys` pattern and others we use `const val`. Is there an advantage/disadvantage to either. Again, non blocking just curious.

jaxdesmarais (approved):
Mainly just general questions but no blockers on the approach aside from the `Outcome` naming decision. And obviously tests too we will want passing. :) Approving ahead of time so we can make those decisions async.

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
mzlangreder: Originally we were using the interface as an internal marker interface coming from BraintreeCore module. The idea was our public API would rely on pre-existing types (like `Card`) and then wrap them with a new internal type that conformed to `PaymentActionPaymentMethod`. In the new design, we have an all-in-one module and so we can switch to using a sealed set of request types available publicly, and our `PaymentActionsClient` will be able to take in the top level `PaymentActionRequest` type so that the API stays simple but sub types give the Merchant the details they need about what information to provide for a given request.

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsGraphQLConstants.kt
mzlangreder: It's mostly stylistic, but I think it helps when sets of constants are grouped up together so the line reads explicitly, for example you'd have `PaymentActionsGraphQLConstants.Keys.PAYMENT_METHOD_DETAILS` at the usage site.

PaymentActions/src/test/java/com/braintreepayments/api/paymentactions/CreditCardUnitTest.kt
mzlangreder: This is a good question, but for now we do not need extra testing since nothing is currently overriding the default selection set. 

If/when we require more fields on the `PaymentAction` DTO based on the selection set being overridden,  we can add more tests for those request types.

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
noguier: Ah, thank you for this explanation, I really like this design decisions you all came up with!

noguier (approved):
i think this is a great clean design, thank you for answering my questions too!

PR #1672, opened 2026-08-07
feat: Add PaymentActionsClient

mzlangreder:
### Summary of changes
- Create the PaymentActionsClient and make associated changes.
- Updated and improved KDoc

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Largely as reviewer, with some code gen mixed in for tests and fixing up issues.

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @mzlangreder

---

### DIFF
--- a/PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsClient.kt
+++ b/PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsClient.kt
@@ -1,0 +1,127 @@
+package com.braintreepayments.api.paymentactions
+
+import android.content.Context
+import com.braintreepayments.api.core.BraintreeClient
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import org.json.JSONObject
+
+/**
+ * Callback for receiving a [PaymentActionResult] from a [PaymentActionsClient] operation.
+ */
+fun interface PaymentActionCallback {
+    fun onResult(result: PaymentActionResult)
+}
+
+/**
+ * Used to submit and drive a payment method through the Payment Actions flow.
+ *
+ * Once a [PaymentActionRequest] has been submitted, consumers should utilize [handleNextAction] to
+ * drive the Payment Action through to completion. The basic interaction loop is:
+ * 1. Submit a [PaymentActionRequest] to [submitForPaymentAction].
+ * 2. Read [PaymentActionResult] and perform any required actions.
+ * 3. Call [handleNextAction] after required actions are complete.
+ * 4. Goto 2 unless a terminal [PaymentActionResult] is received.
+ *
+ */
+class PaymentActionsClient internal constructor(
+    private val service: PaymentActionsService,
+    private val dispatcher: CoroutineDispatcher = Dispatchers.Main,
+    private val coroutineScope: CoroutineScope = CoroutineScope(dispatcher),
+) {
+
+    /**
+     * Initializes a new [PaymentActionsClient] instance.
+     *
+     * @param context an Android [Context]
+     * @param authorization a Client Token used to authenticate
+     */
+    constructor(context: Context, authorization: String) : this(
+        PaymentActionsService(BraintreeClient(context, authorization))
+    )
+
+    /**
+     * Submits a [PaymentActionRequest] to initiate a payment action flow.
+     *
+     * On success the [PaymentActionCallback.onResult] method will be invoked with a
+     * [PaymentActionResult] variant describing the next required step.
+     *
+     * On failure the [PaymentActionCallback.onResult] method will be invoked with a
+     * [PaymentActionResult.Failure] containing an exception.
+     *
+     * @param request  [PaymentActionRequest]
+     * @param callback [PaymentActionCallback]
+     */
+    fun submitForPaymentAction(request: PaymentActionRequest, callback: PaymentActionCallback) =
+        coroutineScope.launch { callback.onResult(submitForPaymentAction(request)) }
+
+    /**
+     * Submits a [PaymentActionRequest] to initiate a payment action flow.
+     *
+     * On success a [PaymentActionResult] variant describing the next required step is returned.
+     *
+     * On failure [PaymentActionResult.Failure] containing an exception is returned.
+     *
+     * @param request [PaymentActionRequest]
+     * @return [PaymentActionResult]
+     */
+    suspend fun submitForPaymentAction(request: PaymentActionRequest): PaymentActionResult {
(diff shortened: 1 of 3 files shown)


PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsClient.kt
```diff
@@ -0,0 +1,127 @@
...
+                    PaymentActionResult.PaymentMethodRequired(paymentAction.id)
+                PaymentActionStatus.REQUIRES_CUSTOMER_ACTION -> PaymentActionResult.CustomerActionRequired(
+                    paymentAction.id,
+                )
+                PaymentActionStatus.READY_FOR_CONFIRMATION ->
+                    PaymentActionResult.ServerActionRequired(paymentAction.id, ServerAction.CONFIRM)
+                PaymentActionStatus.REQUIRES_CAPTURE ->
+                    PaymentActionResult.ServerActionRequired(paymentAction.id, ServerAction.CAPTURE)
+                PaymentActionStatus.SUCCEEDED -> PaymentActionResult.Completed(paymentAction.id)
+                PaymentActionStatus.CANCELED, PaymentActionStatus.EXPIRED ->
```
jaxdesmarais, line 118: Ah interesting, so expired returns a canceled result?

jaxdesmarais (approved):
Just had 1 non-blocking question. This looks great!

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsClient.kt
```diff
@@ -0,0 +1,127 @@
...
+                    PaymentActionResult.Canceled(paymentAction.id)
+                PaymentActionStatus.PROCESSING -> PaymentActionResult.Processing(paymentAction.id)
+                PaymentActionStatus.UNKNOWN ->
+                    PaymentActionResult.Failure(NotImplementedError("Received unrecognized status: ${paymentAction.status}"))
```
mzlangreder, line 122: This should be a different throwable/exception

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsService.kt
```diff
@@ -37,6 +37,15 @@ internal class PaymentActionsService(
         }
     }
 
+    /**
+     * Fetches the current state of a payment action.
+     *
+     * Stubbed for now — the GraphQL query for fetching a payment action by id is not yet defined.
+     */
+    suspend fun getPaymentAction(): PaymentActionServiceResult {
+        throw NotImplementedError("getPaymentAction is not yet implemented")
```
mzlangreder, line 46: This is coming in a follow up PR.

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsClient.kt
mzlangreder: This is a good catch here and I think I need to go over some of these mappings/names a bit more and if we're communicating clearly/the right thing to the Merchant/consumer.

PaymentActions/src/test/java/com/braintreepayments/api/paymentactions/PaymentActionsClientUnitTest.kt
```diff
@@ -0,0 +1,180 @@
...
+
+        val result = buildClient().handleNextAction()
+
+        val outcome = assertIs<PaymentActionResult.Processing>(result)
+        assertEquals("pa123", outcome.id)
+    }
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    @Test
+    fun `handleNextAction with UNKNOWN returns Failure since no outcome mapping exists yet`() = runTest(testDispatcher) {
```
noguier, line 131: will this test be removed once the mapping is finalized?

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsClient.kt
```diff
@@ -0,0 +1,127 @@
...
+            }
+            is PaymentActionServiceResult.Failure -> PaymentActionResult.Failure(error)
```
noguier, line 124: so the failure result will only happen if there is a server/fetching error? Cancelled and expired will come as a part of `PaymentActionServiceResult.Success`?

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsClient.kt
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Used to submit and drive a payment method through the Payment Actions flow.
+ *
+ * Once a [PaymentActionRequest] has been submitted, consumers should utilize [handleNextAction] to
+ * drive the Payment Action through to completion. The basic interaction loop is:
+ * 1. Submit a [PaymentActionRequest] to [submitForPaymentAction].
+ * 2. Read [PaymentActionResult] and perform any required actions.
+ * 3. Call [handleNextAction] after required actions are complete.
+ * 4. Goto 2 unless a terminal [PaymentActionResult] is received.
+ *
```
noguier, line 27: Thank you very much for these detailed KDocs, makes it very easy to understand the flow and the architecture!

noguier (approved):

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsClient.kt
buzzamus: Plus 1 to this. This makes it super easy to follow even for an iOS dev to understand 😛

PaymentActions/src/test/java/com/braintreepayments/api/paymentactions/PaymentActionsClientUnitTest.kt
mzlangreder: I reckon we would keep UNKNOWN type, just in case the backend returns a status value we can't parse.

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsClient.kt
mzlangreder: Yep. Originally I had the same success/failure wrapper in the client too, but I flattened it into a single Result type that also contains a Failure status to make it simpler to consume.

agedd (approved):
this looks great! clean design 🚀

PR #1678, opened 2026-08-14
feat: add card fields compose expiration field

saralvasquez:
### Summary of changes
- Adds expiration date field with expected functionality
  - auto "/" formatting
  - leading 0 padding
  - validation on focus change
- unit test coverage

Just as a note the demo app has it taking up the whole width of the screen for now. I will change it to share the column with the cvv in the changes for the cvv field

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used claude for debugging with the leading 0 padding and formatting. Also used it to write unit tests

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @saralvasquez


noguier:
Hi there, I was wondering if we can have the expiration input field description to be on one line: 
Here is the current compose : 
Screenshot 2026-08-18 at 10 55 13 AM
and here is the XML UI components: 
Screenshot 2026-08-18 at 10 56 51 AM

i think it looks a bit nicer with the description on one line :)

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardExpirationField.kt
```diff
@@ -0,0 +1,35 @@
...
+    errorText: String? = null,
+    focusRequester: FocusRequester? = null,
+    onFocusChanged: (Boolean) -> Unit = {}
+) {
+    val accessibilityLabel = stringResource(R.string.expiration_hint_accessibility)
+    CardFieldBaseTextInputField(
+        value = value,
+        onValueChange = onValueChange,
+        hint = stringResource(R.string.expiration_hint),
+        modifier = modifier.semantics { contentDescription = accessibilityLabel },
```
noguier: once the contentDescription change from #1677 will be merged we do not need to use semantics anymore

```suggestion
        modifier = modifier,
```

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardExpirationField.kt
```diff
@@ -0,0 +1,35 @@
...
+        errorText = errorText,
+        visualTransformation = ExpirationDateVisualTransformation(),
+        keyboardOptions = KeyboardOptions(keyboardType = KeyboardType.Number),
+        focusRequester = focusRequester,
+        onFocusChanged = onFocusChanged
```
noguier: then we can set contentDescription here

```suggestion
        onFocusChanged = onFocusChanged, 
        contentDescription = accessibilityLabel
```

UIComponents/src/test/java/com/braintreepayments/api/uicomponents/compose/ExpirationDateVisualTransformationUnitTest.kt
```diff
@@ -0,0 +1,42 @@
+package com.braintreepayments.api.uicomponents.compose
+
+import androidx.compose.ui.text.AnnotatedString
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class ExpirationDateVisualTransformationUnitTest {
```
noguier: nit: since  you renamed `CardNumberVisualTransformation` to `CardFieldsVisualTransformations` should we move the tests together too?

saralvasquez: yeah I think that makes sense

saralvasquez:
For the expiration date two line picture you posted, was that an emulator or physical device? I only ask because mine doesn't look like that so I don't know how to test that. But I can see if we can have the font size down automatically instead of useing two lines. But I think this should be on the cvv PR since that's the one with the two fields side by side

noguier:
@saralvasquez that was on my physical device! and yes i agree it can leave on cvv PR, or even on a later PR when we add all to the demo app , just wanted to comment before I forget :)

saralvasquez:
Make sense! Can you make the comment over on that PR though so it doesn't fall through the cracks?

noguier (approved):
thank you for addressing all my comment, this looks great 🥇

UIComponents/src/test/java/com/braintreepayments/api/uicomponents/compose/CardFieldsUnitTest.kt
```diff
@@ -71,4 +71,74 @@ class CardFieldsUnitTest {
     }
 
     // endregion
+
+    // region sanitizeCardExpirationInput
```
mzlangreder, line 75: Discussion: How do you feel about region comments? I don't think I've seen them on other projects before.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsVisualTransformations.kt
```diff
@@ -14,6 +15,13 @@ internal class CardNumberVisualTransformation(private val formatGaps: IntArray)
     }
 }
 
+internal class ExpirationDateVisualTransformation : VisualTransformation {
+    override fun filter(text: AnnotatedString): TransformedText {
+        val formatted = ExpirationDateFormatter.formatExpiration(text.text)
+        return TransformedText(AnnotatedString(formatted), digitOffsetMapping(formatted))
```
mzlangreder, line 21: Style nit:
```suggestion
        return ExpirationDateFormatter.formatExpiration(text.text).let {
            TransformedText(AnnotatedString(it), digitOffsetMapping(it))
        }
```

mzlangreder (approved):
Nothing blocking

agedd (approved):
🚀

UIComponents/src/test/java/com/braintreepayments/api/uicomponents/compose/CardFieldsUnitTest.kt
saralvasquez: I'm honestly a bit 50/50 on them. I see them being helpful as test classes grow, but the naming of the test cases are pretty descriptive already so the regions feel a bit redundant. Personally I never added them myself but the robots seem to like them so since moving that task to them they've been working their way in

PR #1682, opened 2026-08-17
feat: Add Compose-friendly initializer for Google Pay

buzzamus:
### Summary of changes
- Created a Compose-Friendly Initializer for Google Pay

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used Claude to analyze the existing compose initializer patterns and follow same pattern for Google Pay, create tests

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/GooglePayFragment.kt
+++ b/Demo/src/main/java/com/braintreepayments/demo/GooglePayFragment.kt
@@ -51,7 +51,11 @@ class GooglePayFragment : BaseFragment() {
         savedInstanceState: Bundle?
     ): View {
         googlePayClient = GooglePayClient(requireContext(), args.authString)
-        googlePayLauncher = GooglePayLauncher(this) { paymentAuthResult ->
+        googlePayLauncher = GooglePayLauncher(
+            registry = requireActivity().activityResultRegistry,
+            lifecycleOwner = viewLifecycleOwner,
+            context = requireContext()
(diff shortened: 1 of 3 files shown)


Demo/src/main/java/com/braintreepayments/demo/GooglePayFragment.kt
```diff
@@ -51,7 +51,11 @@ class GooglePayFragment : BaseFragment() {
         savedInstanceState: Bundle?
     ): View {
         googlePayClient = GooglePayClient(requireContext(), args.authString)
-        googlePayLauncher = GooglePayLauncher(this) { paymentAuthResult ->
+        googlePayLauncher = GooglePayLauncher(
+            registry = requireActivity().activityResultRegistry,
+            lifecycleOwner = viewLifecycleOwner,
+            context = requireContext()
+        ) { paymentAuthResult ->
```
noguier, line 58: so technically here we could have left the old code the way it was here since it is still a Fragment, but we can leave it as is as an example for Compose use :)

noguier (approved):
This looks good, thank you for providing the detailed Kdocs too 👍

agedd (approved):

PR #1683, opened 2026-08-19
feat: add compose pay button and processing

saralvasquez:
### Summary of changes
- adds pay button
- adds backend processing like card client init and submit on pay button click
- resolves a UI bug for the hint float height after recomposition

[Screen_recording_20260819_115504.webm](https://github.com/user-attachments/assets/807107d3-30d4-4b14-8af8-c5213a5bb123)

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used the find cause of UI bug, wire back end processing, and write unit tests

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @saralvasquez

### DIFF
--- a/Demo/src/main/java/com/braintreepayments/demo/ComposeCardFieldsFragment.kt
+++ b/Demo/src/main/java/com/braintreepayments/demo/ComposeCardFieldsFragment.kt
@@ -5,11 +5,24 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.Button
+import androidx.compose.material3.ButtonDefaults
+import androidx.compose.material3.Text
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.getValue
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.ComposeView
 import androidx.compose.ui.platform.ViewCompositionStrategy
 import androidx.compose.ui.unit.dp
+import androidx.navigation.fragment.NavHostFragment
+import com.braintreepayments.api.card.Card
+import com.braintreepayments.api.uicomponents.cardfields.CardFieldsResult
 import com.braintreepayments.api.uicomponents.compose.CardFields
 import com.braintreepayments.api.uicomponents.compose.rememberCardFieldsController
 
@@ -24,9 +37,50 @@ class ComposeCardFieldsFragment : BaseFragment() {
         return ComposeView(requireContext()).apply {
             setContent {
                 val cardFieldsController = rememberCardFieldsController()
+                val isFormValid by cardFieldsController.isFormValid.collectAsState()
+
+                LaunchedEffect(Unit) {
+                    cardFieldsController.initialize(context, authStringArg)
+                    cardFieldsController.setPaymentRequest(
+                        Card(
+                            cardholderName = "John Doe",
+                            postalCode = "12345"
+                        )
+                    )
+                }
 
                 Column(modifier = Modifier.padding(16.dp)) {
                     CardFields(controller = cardFieldsController)
+                    Button(
+                        modifier = Modifier
+                            .height(68.dp)
+                            .fillMaxWidth()
+                            .padding(top = 16.dp),
+                        shape = RoundedCornerShape(12.dp),
+                        colors = ButtonDefaults.buttonColors(
+                            containerColor = Color.Black,
+                            contentColor = Color.White,
+                            disabledContainerColor = Color.Gray,
+                            disabledContentColor = Color.White
+                        ),
+                        enabled = isFormValid,
+                        onClick = {
+                            cardFieldsController.submit { result ->
+                                when (result) {
+                                    is CardFieldsResult.Success -> {
+                                        onPaymentMethodNonceCreated(result.nonce)
+                                        val action = ComposeCardFieldsFragmentDirections
+                                            .actionComposeCardFieldsFragmentToDisplayNonceFragment(result.nonce)
+                                        NavHostFragment.findNavController(this@ComposeCardFieldsFragment)
+                                            .navigate(action)
+                                    }
+                                    is CardFieldsResult.Failure -> handleError(result.error)
+                                }
+                            }
+                        }
+                    ) {
+                        Text(getString(R.string.card_fields_pay))
+                    }
                 }
             }
             setViewCompositionStrategy(ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed)
--- a/Demo/src/main/java/com/braintreepayments/demo/ComposeCardFieldsFragment.kt
+++ b/Demo/src/main/java/com/braintreepayments/demo/ComposeCardFieldsFragment.kt
@@ -12,7 +12,6 @@ import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material3.Button
 import androidx.compose.material3.ButtonDefaults
 import androidx.compose.material3.Text
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Modifier
@@ -36,18 +35,15 @@ class ComposeCardFieldsFragment : BaseFragment() {
         super.onCreateView(inflater, container, savedInstanceState)
         return ComposeView(requireContext()).apply {
             setContent {
-                val cardFieldsController = rememberCardFieldsController()
-                val isFormValid by cardFieldsController.isFormValid.collectAsState()
-
-                LaunchedEffect(Unit) {
-                    cardFieldsController.initialize(context, authStringArg)
-                    cardFieldsController.setPaymentRequest(
-                        Card(
-                            cardholderName = "John Doe",
-                            postalCode = "12345"
-                        )
+                val cardFieldsController = rememberCardFieldsController(
+                    authorization = authStringArg,
+                    // optional customer data
+                    request = Card(
+                        cardholderName = "John Doe",
+                        postalCode = "12345"
                     )
-                }
+                )
+                val isFormValid by cardFieldsController.isFormValid.collectAsState()
 
                 Column(modifier = Modifier.padding(16.dp)) {
                     CardFields(controller = cardFieldsController)
--- a/UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFields.kt
+++ b/UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFields.kt
@@ -5,7 +5,6 @@ import androidx.compose.foundation.interaction.MutableInteractionSource
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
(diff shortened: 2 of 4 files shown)


Demo/src/main/java/com/braintreepayments/demo/ComposeCardFieldsFragment.kt
```diff
@@ -24,9 +37,50 @@ class ComposeCardFieldsFragment : BaseFragment() {
         return ComposeView(requireContext()).apply {
             setContent {
                 val cardFieldsController = rememberCardFieldsController()
+                val isFormValid by cardFieldsController.isFormValid.collectAsState()
+
+                LaunchedEffect(Unit) {
+                    cardFieldsController.initialize(context, authStringArg)
+                    cardFieldsController.setPaymentRequest(
+                        Card(
```
noguier, line 46: should we add a little comment that this is an optional customer data ? and we do not have to set it, just like we have in UIComponentsFragment

saralvasquez: Hm I would normally put that in the docs instead of the demo but couldn't hurt to have both! I'll add that

UIComponents/src/test/java/com/braintreepayments/api/uicomponents/compose/CardFieldsControllerUnitTest.kt
```diff
@@ -60,4 +83,136 @@ class CardFieldsControllerUnitTest {
...
+    @Test
+    fun `submit before initialize does not tokenize`() {
+        // No client is injected (null cardClient), so submit() hits the pre-init guard and
+        // skips tokenization entirely — cardClient.tokenize is never reached.
+        val controller = createCardFieldsController()
+
+        controller.submit { }
+
+        verify(exactly = 0) { cardClient.tokenize(any(), any()) }
+    }
```
noguier: do we really need both of these tests? I think test `submit before initialize delivers a Failure to the callback` already covers all the necessary behavior?

noguier (approved):
Just a couple of non-blocking comments. Everything looks and functions great :)

UIComponents/src/test/java/com/braintreepayments/api/uicomponents/compose/CardFieldsControllerUnitTest.kt
saralvasquez: I think they are technically testing different behavior but I think they can be combined

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFields.kt
```diff
@@ -51,7 +50,7 @@ fun CardFields(controller: CardFieldsController, modifier: Modifier = Modifier)
 
     Column(
         modifier = modifier
-            .fillMaxSize()
+            .fillMaxWidth()
             .clickable(
```
buzzamus, line 54: Just out of curiosity for me trying to learn Android nuances: What was the reason for size vs width here. Was it incorrectly filling up too much space in the height?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
```diff
@@ -1,25 +1,115 @@
...
+    fun setPaymentRequest(card: Card?) {
+        request = card
+    }
+
+    /**
+     * Tokenizes the card details entered by the user, along with any additional data provided in
+     * [setPaymentRequest]. The result is delivered to [callback].
+     *
+     * If called before [initialize], delivers a [CardFieldsResult.Failure].
+     */
```
buzzamus, line 34: These comments are very helpful!

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
mzlangreder, line 1: NIT: We're a bit inconsistent with the application of the `@param`'s in our comment blocks, let's apply them or make sure all the params are covered.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
```diff
@@ -1,25 +1,115 @@
...
+    val isFormValid: StateFlow<Boolean> = viewModel.isFormValid
+
+    /**
+     * Initializes the card tokenization flow. Must be called before [submit].
+     * @param authorization a tokenization key or client token.
+     */
+    fun initialize(context: Context, authorization: String) {
+        cardClient = CardClient(context, authorization)
+    }
+
```
mzlangreder: Should this be a constructor instead? That way there is no user error/ambiguity about needing to call this first?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
```diff
@@ -1,25 +1,115 @@
...
+    /**
+     * Optionally accepts a [Card] with additional data to be included in the tokenization request, such as
+     * cardholder name or billing address. The card number, expiration, and CVV entered by the user will
+     * override any values set on the [Card] object.
+     */
+    fun setPaymentRequest(card: Card?) {
```
mzlangreder: Could/Should this be included in `initialize` as an optional param?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
```diff
@@ -1,25 +1,115 @@
...
+        }
+
+        client.tokenize(buildCard()) { cardResult ->
+            val result = when (cardResult) {
+                is CardResult.Success -> CardFieldsResult.Success(cardResult.nonce)
+                is CardResult.Failure -> CardFieldsResult.Failure(cardResult.error)
+            }
+            callback.onCardFieldsResult(result)
+        }
+    }
```
mzlangreder, line 43: I think this would be a lot simpler if we decide to implement the initialize -> constructor change.

Otherwise we could make this more Kotliny to help:


```suggestion
fun submit(callback: CardFieldsResultCallback) {
    cardClient?.let { client ->
        client.tokenize(buildCard()) { cardResult ->
            val result = when (cardResult) {
                is CardResult.Success -> CardFieldsResult.Success(cardResult.nonce)
                is CardResult.Failure -> CardFieldsResult.Failure(cardResult.error)
            }
            callback.onCardFieldsResult(result)
        }
    } ?: run {
        callback.onCardFieldsResult(
            CardFieldsResult.Failure(
                BraintreeException("CardFieldsState must be initialized by calling initialize() before submit()")
            )
        )
    }
}
```

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFields.kt
saralvasquez: Yeah if I recall correctly this was taking up too much space so I couldn't position the pay button on the screen close to the card fields which made a weird gap

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
saralvasquez: Yeah that would make sense to me, but I chose this to mirror how the original card fields integration looks. So I think the question is do we keep consistency across the xml and compose versions? I could see arguments for both, especially since compose is already deviating a bit with the controller passed to the composable

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
saralvasquez: Ah yeah agreed. I'll go through and clean those up. Or we could wait until everything is merged and clean up all the kdoc comments in the PR for the feature branch to go into main. I think there are probably some inconsistencies across the PRs that have already been merged

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
saralvasquez: Although looking back at the demo app as an example of integration `rememberCardFieldsController` sort of complicated the constuctor idea since it houses the call to the internal constructor. I think since it wouldn't be the cleanest constructor model and the current method mirrors the XML version I think I'm inclined to keep it as it is. What do you think? @noguier do you have any opinions here?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
saralvasquez: This was another example of mirroring the xml version so I think we should decide if we want to keep things consistent across the two or if we'd prefer to deviate for some cleaner solutions here

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
mzlangreder: Either is all good with me.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
mzlangreder: My main thinking is that if we can enforce something versus hoping someone reads the docs, it will eliminate frustration when it comes to using this feature. They get IDE/compiler help straight away and there's no ambiguity. Would it make the code more difficult/cumbersome to use for a merchant if we made the change?

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
mzlangreder: I think it's ok to diverge if it means having a better developer experience.

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
noguier: Sorry for the delay, I didnt see the notification that I was tagged in the comment. I think I agree with Matt here, because it does provide better developer experience and less error prone. I think if we can make the integration be as easy as possible it is definitely a win. I think we should try Matt's suggestions. Plus i think matching XML is not that necessary since it is a dying UI language.

noguier: Although, as I was writing this I realized that me and Brent wanted to make sure we have some parity across the platforms, so I checked on ios and we do not have initialize, I think we needed initialize in XML version because of XML specifics. So i think we are good to proceed with trying Matt's suggestion

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
noguier: I think here I also agree with Matt

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
saralvasquez: Ok yeah that sounds good. Let me experiment with moving to the constructor idea and see what's possible

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
saralvasquez: I'll put it in the readme updates PR since it's all kind of docs related

saralvasquez:
Okie dokie! Take a look at the latest commit. I amended it so there's a lot in there but I made some changes to how `rememberCardFieldsController` works. It mostly functions as a constructor this way. I was going to rename it to `CardFieldsController` but apparently the `remember` keyword is a compose standard naming convention to signal that it's remembered on the back end so the robot suggested I don't change it

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFieldsController.kt
```diff
@@ -1,25 +1,87 @@
...
+
+    private fun buildCard(): Card {
+        val rawExpiration = expiration.value.text
+        return (request ?: Card()).copy(
```
mzlangreder: Could request be nonNullable and default to `Card()` instead of `null?

saralvasquez: Op! Yeah definitely. I like that

mzlangreder (approved):

PR #1689, opened 2026-08-24
Return External ID On Venmo Vault Flows

buzzamus:
### Summary of changes
- This PR returns the External ID that is normally returned on normal Venmo flows on flows where a Client token + vault flow are used to trigger the flow. Confirmed with Venmo team that common ID is external ID in the response we get 

### AI Usage

**Which AI Agent Was Used?**
- [x] Copilot 
- [ ] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used to take iOS fix and apply it to Android

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [x] Added a changelog entry

- @buzzamus


jaxdesmarais (approved):

mzlangreder (approved):

agedd (approved):
🚀

PR #1691, opened 2026-08-26
feat: add compose card fields README entry

saralvasquez:
### Summary of changes
- Adds a new entry to the README under the Card Fields header to describe how to integrate with the compose card fields

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [ ] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
It wasn't

- [ ] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @saralvasquez

### DIFF
--- a/UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFields.kt
+++ b/UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFields.kt
@@ -5,7 +5,6 @@ import androidx.compose.foundation.interaction.MutableInteractionSource
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
@@ -51,7 +50,7 @@ fun CardFields(controller: CardFieldsController, modifier: Modifier = Modifier) 
 
     Column(
         modifier = modifier
-            .fillMaxSize()
+            .fillMaxWidth()
             .clickable(
                 indication = null,
                 interactionSource = remember { MutableInteractionSource() }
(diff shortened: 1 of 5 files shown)


noguier (approved):
This looks great, thank you!

saralvasquez:
Converted this to draft since the controller might change

saralvasquez:
Alrighty! I went ahead and added some missing kdoc comments. Matt mentioned that we were inconsistent with naming `@params` in these comments. I agree but ultimately opted to only add them when the comment was for a publicly facing class because those are the only classes that will have docs generated. The internal classes for this feature have a whole ton of params and it made the comment a lot harder to read. I figured since internal devs were likely to be the only ones reading the comment in these files it was reasonable that they could see what each param was for

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFields.kt
```diff
@@ -27,6 +27,12 @@ import com.braintreepayments.api.uicomponents.cardfields.CardField
 import com.braintreepayments.api.uicomponents.cardfields.ExpirationDateFormatter
 import com.braintreepayments.api.uicomponents.cardfields.ValidationResult
 
+/**
+ * Renders the card number, expiration, and CVV fields for card tokenization. Pair with a
+ * [CardFieldsController] created via [rememberCardFieldsController].
```
jaxdesmarais, line 32: Super basic q but what's a `rememberCardFieldsController`?

jaxdesmarais: Oh, nvm, just needed to keep scrolling. 😄 Seems like a Compose pattern used for process death?

jaxdesmarais (approved):

UIComponents/src/main/java/com/braintreepayments/api/uicomponents/compose/CardFields.kt
saralvasquez: Yup! Remember just keeps it across process death and re-composition of the composable elements

saralvasquez:
The failures here are a known problem with local payments that is being looked at by the alt pay team and are unrelated to the changes on this branch. The changes are also comments and readme changes only

PR #1692, opened 2026-08-31
Compose CardFields Instrumentations Tests

noguier:
### Summary of changes
- Created tests for each individual input field: cc number, expiration date, cvv
- Created tests for all fields put together to test autoadvance and the form validation

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Brainstroming and getting familiar with syntax to use to test Compose

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

@noguier


UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/compose/CardCvvFieldTest.kt
```diff
@@ -0,0 +1,80 @@
...
+
+    @Test
+    fun hintPopup_isNotShown_beforeIconIsTapped() {
+        setCvvField()
+
+        composeTestRule.onNodeWithText(str(R.string.cvv_overlay_body)).assertDoesNotExist()
+    }
+
+    @Test
+    fun tappingCloseButton_dismissesHintPopup() {
```
saralvasquez, line 70: Do we also want a test that clicking anywhere on the screen outside the popup will dismiss it? I recall that being important functionality

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/compose/CardFieldsFormTest.kt
```diff
@@ -0,0 +1,134 @@
...
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
+import androidx.compose.ui.test.performTextInput
+import androidx.test.platform.app.InstrumentationRegistry
+import com.braintreepayments.api.uicomponents.R
+import org.junit.Rule
+import org.junit.Test
+
+/**
+ * Cross-field behavior owned by the [CardFields] composable + `CardFieldsViewModel`,
```
saralvasquez, line 26: I noticed we don't have a test to see that the pay button becomes enabled after all fields are valid. Is that something that's testable since it's not technically part of card fields?

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/compose/CardNumberFieldTest.kt
```diff
@@ -0,0 +1,118 @@
...
+        setCardFields()
+
+        cardNumberField().performTextInput(prefix)
+
+        composeTestRule.onNodeWithContentDescription(str(iconContentDescriptionRes))
+            .assertIsDisplayed()
+    }
+
+    @Test
+    fun typingDigits_updatesFieldText() {
```
saralvasquez, line 59: Do we want a test here that passes a value that's too long like we do for expiration?

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/compose/CardCvvFieldTest.kt
```diff
@@ -0,0 +1,80 @@
...
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
+import androidx.compose.ui.test.performTextInput
+import androidx.compose.ui.text.input.TextFieldValue
+import androidx.test.platform.app.InstrumentationRegistry
+import com.braintreepayments.api.uicomponents.R
+import org.junit.Rule
+import org.junit.Test
+
+class CardCvvFieldTest {
```
saralvasquez, line 22: Do we also want a test here that passes a value that's too long like for expiration? We also might want to check that the max value changes with the right card brand. But that might be better in the CardFieldsFormTest. What do you think?

saralvasquez (commented):
Just a couple test cases I think could be good to add but otherwise these look great!

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/compose/CardFieldsFormTest.kt
```diff
@@ -0,0 +1,134 @@
+package com.braintreepayments.api.uicomponents.compose
+
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.assertIsFocused
```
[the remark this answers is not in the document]
noguier: the file compiles as is

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/compose/CardCvvFieldTest.kt
```diff
@@ -0,0 +1,80 @@
+package com.braintreepayments.api.uicomponents.compose
+
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.test.assert
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.hasText
```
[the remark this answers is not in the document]
noguier: the file compiles as is

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/compose/CardCvvFieldTest.kt
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Test
+    fun typingDigits_updatesFieldText() {
+        setCvvField()
+
+        composeTestRule.onNodeWithContentDescription(str(R.string.cvv_accessibility))
+            .performTextInput("123")
+
+        composeTestRule.onNodeWithContentDescription(str(R.string.cvv_accessibility))
```
buzzamus, line 48: Possibly dumb question: What is `R`?

noguier: Great question, R is autogenerated class, that assigns an id to your resources: https://developer.android.com/guide/topics/resources/providing-resources

so here we are just grabbing the string we defined that has an accessibility text from `strings_card_fields.xml`

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/compose/CardFieldsFormTest.kt
noguier: So I was contemplating about that myself as well. Since the button is technically owned by a merchant, but i see that XML cardFields do have a test for the button, so I can add it here as well.

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/compose/CardCvvFieldTest.kt
noguier: I think both of these scenarios are good to add. I think I can add a test for long value here, but the test for length change according to the brand in `CardFieldsFormTest`

UIComponents/src/androidTest/java/com/braintreepayments/api/uicomponents/compose/CardExpirationFieldTest.kt
```diff
@@ -0,0 +1,91 @@
...
+            composeTestRule.onNodeWithContentDescription(str(R.string.expiration_hint_accessibility))
+        expirationInput.performTextInput("1")
+        expirationInput.performTextInput("2")
+        expirationInput.performTextInput("2")
+        expirationInput.performTextInput("8")
+        expirationInput.performTextInput("8")
+
+        composeTestRule.onNodeWithContentDescription(str(R.string.expiration_hint_accessibility))
+            .assert(hasText("1228"))
+    }
```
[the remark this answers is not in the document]
noguier: I believe the test is fine as is

gradle/libs.versions.toml
```diff
@@ -69,6 +70,7 @@ androidx-test-runner = { group = "androidx.test", name = "runner", version.ref =
 androidx-test-rules = { group = "androidx.test", name = "rules", version.ref = "androidxTest" }
 androidx-junit = { group = "androidx.test.ext", name = "junit", version.ref = "androidxJunit" }
 androidx-test-espresso-core = { group = "androidx.test.espresso", name = "espresso-core", version.ref = "espresso" }
+androidx-test-uiautomator = { group = "androidx.test.uiautomator", name = "uiautomator", version.ref = "uiautomator" }
```
mzlangreder, line 73: Was this needed to test Compose features/functionality? It seems like a new/different way of writing Android tests so I'm curious how the team usually goes about adopting new dependencies like this one.

I also see that there is a [newer version](https://developer.android.com/training/testing/other-components/ui-automator) out, albeit an alpha. The docs do say:

> The API is under development, and we strongly recommend using it for any new development with UI Automator.

Maybe we should consider using the newer version in the style that Google is going to be developing for moving forward since we're introducing here for the first time.

noguier: I needed to add this to test the behavior of tapping anywhere on the screen to close the cvv info popup, and I did tried to use 2.4.0, but that version resulted in 
`e:  Class 'kotlin.reflect.KClass' was compiled with an incompatible version of Kotlin. The actual metadata version is 2.1.0, but the compiler version 1.9.0 can read versions up to 2.0.0.
`
So in order to use a new version, we would need to update Kotlin to at least 2.1.0. So I opted to using the olde version. Thoughts?

mzlangreder: Yea I guess we'll be stuck using the older one until we can determine if/when we can upgrade to Kotlin 2+

mzlangreder (approved):

saralvasquez (approved):

saralvasquez:
The test failures here are a local payments issue on the back end unrelated to the changes on this branch

PR #1694, opened 2026-09-03
feat: remove unused internal enums

noguier:
### Summary of changes
- Removed internal `ConfirmationMethod` enum
- Removed internal `CaptureMethod` enum
- These enums were unused and the PaymentActions mode is not going to be surfaced to merchants 

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [ ] Claude 
- [x] Other (Type Name Here)

**How was AI used?**
Didn't use AI

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

@noguier

### DIFF
--- a/PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
+++ b/PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
@@ -131,24 +131,6 @@ internal enum class PaymentActionStatus {
 }
 
 /**
- * Merchant gateway configuration for whether a payment action is confirmed automatically by the
- * server or requires an explicit confirmation step.
- */
-internal enum class ConfirmationMethod {
-    AUTOMATIC,
-    MANUAL,
-}
-
-/**
- * Merchant gateway configuration for whether a payment action is captured automatically by the
- * server or requires an explicit capture step.
- */
-internal enum class CaptureMethod {
-    AUTOMATIC,
-    MANUAL,
-}
-
-/**
  * Wrapper result type for [PaymentActionsService], carrying the raw [PaymentAction]
  * returned by GraphQL.
  */


PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
```diff
@@ -130,24 +130,6 @@ internal enum class PaymentActionStatus {
...
-
-/**
- * Merchant gateway configuration for whether a payment action is captured automatically by the
- * server or requires an explicit capture step.
- */
-internal enum class CaptureMethod {
-    AUTOMATIC,
-    MANUAL,
-}
-
```
buzzamus, line 150: Is there a different mechanism to differentiate between Automatic and Manual capturing?

buzzamus (approved):

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsDeclarations.kt
noguier: As far as I understand, we will read the status from the backend here: https://github.com/braintree/braintree_android/blob/5d96ff4d5b9f33ad0a86b7bf02755dfef980cec3/PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsService.kt#L85, , which is saved into `PaymentActionServiceResult.Success` the status should have all the needed information to proceed. And then, I think, we process the data here : https://github.com/braintree/braintree_android/blob/5d96ff4d5b9f33ad0a86b7bf02755dfef980cec3/PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsClient.kt#L114 for manual capturing, and automatic never stops at `REQUIRED_CAPTURE`.  But I might be wrong?

buzzamus: Great! Thanks for the additional context

noguier: Similarly, `READY_FOR_CONFIRMATION` will signify manual confirmation path 😄

mzlangreder (approved):

PR #1695, opened 2026-09-08
feat(payment-actions): Remap Unknown status to Unknown result type

mzlangreder:
### Summary of changes
Remapped the UNKNOWN status from emitting a Failure to emitting an Unknown result type.

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [ ] Claude 
- [x] Other (GLM 5.2)

**How was AI used?**
Local review and coding.

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @mzlangreder

---


PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsClient.kt
```diff
@@ -118,7 +118,7 @@ class PaymentActionsClient internal constructor(
                     PaymentActionResult.Canceled(paymentAction.id)
                 PaymentActionStatus.PROCESSING -> PaymentActionResult.Processing(paymentAction.id)
                 PaymentActionStatus.UNKNOWN ->
-                    PaymentActionResult.Failure(NotImplementedError("Received unrecognized status: ${paymentAction.status}"))
+                    PaymentActionResult.Unknown(paymentAction.id)
             }
```
buzzamus, line 122: Just to make sure I am understanding this: Before `Unknown` was a type of `Failure` and now it is just a completely new type correct?

buzzamus (approved):
Looks good! I just had a question for my own understanding that I think is very obvious, but it is the end of the day and my brain is fried 😄

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsClient.kt
mzlangreder: Yea, the idea is to match up with what iOS is doing by having Unknown and Expired types exposed directly instead of indicating a Failure.

noguier (approved):

PR #1697, opened 2026-09-08
feat(payment-actions): Fire analytics events

mzlangreder:
### Summary of changes
Implement firing of analytics events

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [ ] Claude 
- [x] Other (GLM 5.2)

**How was AI used?**
Implementation review and test generation.

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @mzlangreder

---


PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsClient.kt
```diff
@@ -67,7 +69,17 @@ class PaymentActionsClient internal constructor(
...
-        return service.setPaymentActionPaymentMethod(request).handleNextAction()
+        braintreeClient.sendAnalyticsEvent(PaymentActionsAnalytics.CARD_SET_PAYMENT_METHOD_STARTED)
+        return service.setPaymentActionPaymentMethod(request).handleNextAction().also { result ->
+            if (result is PaymentActionResult.Failure) {
+                braintreeClient.sendAnalyticsEvent(
+                    PaymentActionsAnalytics.CARD_SET_PAYMENT_METHOD_FAILED,
+                    AnalyticsEventParams(errorDescription = result.error.message),
+                )
+            } else {
+                braintreeClient.sendAnalyticsEvent(PaymentActionsAnalytics.CARD_SET_PAYMENT_METHOD_SUCCEEDED)
```
buzzamus: Just out of curiosity how does `also` tied to a return statement work?

mzlangreder: `also` is like a secondary effect block. It takes in a value from the upstream chain and returns that same value. Within the block we have access to the value for side-effect purposes, in this case we want to fire an analytics event based on what the result ended up being without effecting the result.

buzzamus: Gotcha that makes sense! Thanks for the explanation

buzzamus (approved):
Looks good!

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsClient.kt
```diff
@@ -67,7 +69,17 @@ class PaymentActionsClient internal constructor(
      * @return [PaymentActionResult]
      */
     suspend fun submitForPaymentAction(request: PaymentActionRequest): PaymentActionResult {
-        return service.setPaymentActionPaymentMethod(request).handleNextAction()
+        braintreeClient.sendAnalyticsEvent(PaymentActionsAnalytics.CARD_SET_PAYMENT_METHOD_STARTED)
```
jaxdesmarais: I think we changed some of the names/values of these events on iOS. Do we want to update them on Android so it matches across platforms?

mzlangreder: Yes, good catch ty.

mzlangreder: Updated

agedd (approved):
🚀

jaxdesmarais (approved):

PaymentActions/src/main/java/com/braintreepayments/api/paymentactions/PaymentActionsClient.kt
```diff
@@ -67,7 +69,17 @@ class PaymentActionsClient internal constructor(
...
-        return service.setPaymentActionPaymentMethod(request).handleNextAction()
+        braintreeClient.sendAnalyticsEvent(PaymentActionsAnalytics.SET_PAYMENT_ACTION_PAYMENT_METHOD_STARTED)
+        return service.setPaymentActionPaymentMethod(request).handleNextAction().also { result ->
+            if (result is PaymentActionResult.Failure) {
+                braintreeClient.sendAnalyticsEvent(
+                    PaymentActionsAnalytics.SET_PAYMENT_ACTION_PAYMENT_METHOD_FAILED,
+                    AnalyticsEventParams(errorDescription = result.error.message),
+                )
+            } else {
+                braintreeClient.sendAnalyticsEvent(PaymentActionsAnalytics.SET_PAYMENT_ACTION_PAYMENT_METHOD_SUCCEEDED)
```
noguier, line 80: so we submit `SET_PAYMENT_ACTION_PAYMENT_METHOD_SUCCEEDED` event regardless of payment action status, or am I missing something?

mzlangreder: This is a great question! In this case, we fire the succeeded event as long as the `PaymentActionResult` is not a `Failure`. And in our case a `Failure` can come from 3 places:
1. We got an exception during the graphQL call
2. The graphQL call returned errors
3. We failed to parse the result of the graphQL call

The PaymentAction does not have a `Failure` status, so our `PaymentActionResult.Failure` is there to indicate a failure in the system. It's analogous to our other clients having a `Failure` result type but in our case we flattened it in with the other Result statuses.

noguier: Great, thank you for clarification, and we do not think it will be useful to log different payment action statuses?

mzlangreder: I could see that being valuable actually. Originally we just followed the existing analytics pattern we have today, but I can see how with payment actions, it's different and more complex. We don't have a One call One Result type of behavior in Payment Actions. 

For instance we're going to have flows here we submit for payment action and then our SDK is going to perform _multiple_ actions based on the status of the Payment Action. Those are all things that are going to be valuable to log I think but....we also don't have true logging/observability going on in our SDK either. It's mostly just posting events that a thing happened.

@agedd @jaxdesmarais do y'all have any thoughts/opinions on our analytics strategy and what extras we _could_ add here?

agedd: great catch @noguier — yeah right now "succeeded" just means we didn't hit our failure case, not that the payment action itself fully succeeded. agreed with @mzlangreder that this can get confusing once we have multiple possible actions that could spin off a single submission.

I think we can keep this PR as-is and address this in a fast-follow ticket, where we can dig into what granularity we'd need to accurately reflect the status of the PA at any given time.

PR #1700, opened 2026-09-09
fix: add per-test tokenization key to prevent cache collission

saralvasquez:
### Summary of changes
These changes grew out of an increase in test flake for flaky tests that we were already tracking. There was such a steep increase in the number of fails that it slowed development effort. The tests causing the issue were a couple tests in the `VenmoClientTest` class and the `threeDSecure_authenticates` test in the `ThreeDSecureCardinalTest` class

For the venmo tests the issue that seemed to be causing the flake was cache collision or race conditions of the configuration settings since we were using the same tokenization key for every test. To fix that I added per-test tokenization keys

For the cardinal test, the flake was due to timeouts on the emulator UI that threw null exceptions. This was particularly noticeable in the older emulator tests. The only real way to get around this was to increase the timeout length to give the slower emulators more time to load cardinal functionality. This fixes the flake most of the time but there's not other way I could find around this besides just making the timeout longer and longer

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used claude to pull runs of the instrumentation test runs to give me data on which tests were flaking and what the error were that were shown on failed runs

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @saralvasquez


saralvasquez:
I ran integration tests on this branch 10 times and 8 were completely green. Two of them did have the flaky Cardinal test still fail, but I don't think there is much to do for it other than keep increasing the timeout, which feels like the wrong approach. Let me know if any of you see something I'm missing to completely resolve the flake, otherwise I think this will have to be good enough

jaxdesmarais (approved):
Thank you for the detailed description on the PR for what needed to change here and why!

noguier:
Thank you for adding these and looking into it. I also was unable to find other way to fix 3DS, we can also potentially disable those tests for older APIs, since newer are passing?

noguier (approved):

mzlangreder (approved):

saralvasquez:
> we can also potentially disable those tests for older APIs, since newer are passing?

I think I'd hesitate to disable tests permanently to avoid flake. Flake is annoying but the test coverage is there for a reason 🤷‍♀️ 

Why don't we keep an eye on the flaky test tracker and if the flake stays as bad as it has been we can disable them. I'll also add an extra field to the test tracker to keep track of the android version that the flake is seen on

PR #1702, opened 2026-09-09
feat: combine payment buttons and Card Fields into one Demo fragment

noguier:
### Summary of changes
- Combine Payment Buttons and Card Fields under one fragment in the Demo app to match XML
- Remove separate Payment Buttons Compose and Card Fields Compose fragments
- Update Nav fragment
- Add `CHANGELOG` entry

Demo Video 

https://github.com/user-attachments/assets/e223c4f6-60c3-4471-84f4-8ecab18ae676

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Claude was used as a first reviewer on the combined fragments work

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry

@noguier


Demo/src/main/java/com/braintreepayments/demo/ComposeUIComponentsFragment.kt
```diff
@@ -0,0 +1,284 @@
...
+                            cardFieldsController.submit { result ->
+                                when (result) {
+                                    is CardFieldsResult.Success -> handleNonce(result.nonce)
+                                    is CardFieldsResult.Failure -> handleError(result.error)
+                                }
+                            }
+                        }
+                    ) {
+                        Text("Pay")
+                    }
```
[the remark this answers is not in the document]
saralvasquez: This seems reasonable

saralvasquez (approved):
This looks great! The only thing I noticed during testing that might be worth coming back to is that the color selected for the buttons doesn't survive recomposition so it defaults back to blue whenever the screen is rotated. But I don't know if that's worth digging into now since this is just about bringing the UI components together into one screen

saralvasquez:
Op I missed the lint failure. Looks like a couple things detekt is angry about

noguier:
> The only thing I noticed during testing that might be worth coming back to is that the color selected for the buttons doesn't survive recomposition so it defaults back to blue whenever the screen is rotated. But I don't know if that's worth digging into now since this is just about bringing the UI components together into one screen

Great catch, I confirmed that this behavior was present in the Compose PaymentButtons and is not caused by this change, since it feels a little out of scope for this PR I will create a separate ticket to address that! Thank you!

Demo/src/main/java/com/braintreepayments/demo/ComposeUIComponentsFragment.kt
```diff
@@ -0,0 +1,284 @@
...
+import com.braintreepayments.api.venmo.VenmoPaymentMethodUsage
+import com.braintreepayments.api.venmo.VenmoRequest
+import com.braintreepayments.api.venmo.VenmoResult
+import com.braintreepayments.api.venmo.VenmoTokenizeCallback
+
+private const val APP_LINK_RETURN_URL =
+    "https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com/braintree-payments"
+private const val DEEP_LINK_FALLBACK_SCHEME = "com.braintreepayments.demo.braintree"
+
+@Suppress("LongMethod")
```
buzzamus: Should this just be placed on the offending method, or are there several causing issues in this file?

Demo/src/main/java/com/braintreepayments/demo/ComposeUIComponentsFragment.kt
```diff
@@ -0,0 +1,284 @@
...
+class ComposeUIComponentsFragment : BaseFragment() {
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View {
+        super.onCreateView(inflater, container, savedInstanceState)
+        val payPalRequest = paypalRequest(requireContext())
```
buzzamus: nit: variable name is `payPal` camelcase while `paypalRequest` has both P's lower

Demo/src/main/java/com/braintreepayments/demo/ComposeUIComponentsFragment.kt
```diff
@@ -0,0 +1,284 @@
...
+                            style = MaterialTheme.typography.labelMedium,
+                            maxLines = 1
+                        )
+                    }
+                )
+            }
+        }
+    }
+
+    private fun handleNonce(nonce: PaymentMethodNonce) {
```
buzzamus, line 235: nice! Love the consolidation of the separate result handling 😎

buzzamus (approved):
Looks good! I had one tiny nit and another take it or leave it, but nothing blocking here.

Demo/src/main/java/com/braintreepayments/demo/ComposeUIComponentsFragment.kt
noguier: Great catch thank you!

Demo/src/main/java/com/braintreepayments/demo/ComposeUIComponentsFragment.kt
noguier: Good eye!

Issue #1662, opened 2026-08-03
GooglePayLauncher for compose


sjqnice:
### Is your feature request related to a problem? Please describe.

GooglePayLauncher works without an issue for activities/fragments, but it will be very nice to consider support it an alternative creation method for JetpackCompose usage.

### Describe the solution you'd like.

A way to create the GooglePayLauncher with just context, without activity/fragment parameter

noguier:
Hi @sjqnice, thank you for the suggestion. Our team will look into this request, and will prioritize it as soon as the capacity  allows. We will update this thread with more a more specific timeline as soon as we can. 

Additionally, since the SDK is open source you are welcome to create a submit a PR, if you see something that needs improvement, in the future :) 

Thank you again!

agedd:
hi @sjqnice - our team has developed a solution for using GooglePayLauncher in a compose-friendly format and the enhancement is included as part of [v5.32.0](https://github.com/braintree/braintree_android/releases/tag/5.32.0) release. please check it out when you have a chance. thank you!

PR #1704, opened 2026-09-11
feat: add a compose friendly constructor to `SEPADirectDebit`

noguier:
### Summary of changes
- Create  compose-friendly constructor to SEPADirectDebitLauncher
- wire the new constructor in the demo app to ensure it works as expected
- add a changelog

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
test creation

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry

@noguier


SEPADirectDebit/src/main/java/com/braintreepayments/api/sepadirectdebit/SEPADirectDebitLauncher.kt
```diff
@@ -20,6 +21,15 @@ class SEPADirectDebitLauncher internal constructor(private val browserSwitchClie
         browserSwitchClient = if (caller != null) BrowserSwitchClient(caller) else BrowserSwitchClient()
     )
 
+    /**
+     * Used to launch the SEPA mandate in a web browser and deliver results. Use it in Compose flows
+     * or anywhere else an [ActivityResultCaller] is not available.
+     * @param registry ActivityResultRegistry parameter. It will be passed to BrowserSwitchClient.
+     */
```
buzzamus, line 28: This is a nice, helpful comment!

SEPADirectDebit/src/test/java/com/braintreepayments/api/sepadirectdebit/SEPADirectDebitLauncherUnitTest.kt
```diff
@@ -32,6 +34,15 @@ class SEPADirectDebitLauncherUnitTest {
...
 
+    @Test
+    fun `constructor with ActivityResultRegistry creates a valid launcher`() {
+        val registry: ActivityResultRegistry = mockk(relaxed = true)
+
+        val launcher = SEPADirectDebitLauncher(registry)
+
+        assertNotNull(launcher)
+    }
+
```
buzzamus, line 45: This is a question coming from my lack of compose knowledge, but is this type of test good enough to confirm the compose constructor is working as intended?

noguier: great question, i think I was mostly trying to test here if the constructor _can_ be constructed with ActivityResultCaller without crashing. Most of the hard work is passed to `BrowserSwitchClient` so it is hard to test anything besides launcher creation. I can possibly mock the creation of BrowserSwitchClient with the newly created launcher, do you think it will be better?

buzzamus: I think that question is probably best answered by you and other Android platform specific folks since I don't know the common best practices for compose UI tests and how much is necessary

mzlangreder: When I read the text of the test, "creates a valid launcher", I think here I would at least expect some more assertions on the launcher itself to verify it is "valid". If all we wanted was a test that the launcher gets created when being passed the expected params, that kind of feels like a low value test to me unless there is something specific you're looking for having been set up from calling the constructor.

buzzamus: Yeah that makes sense and sort of what I was curious about: Are there additional things that could be asserted to see if any attributes can be verified, or is this comprehensive enough

saralvasquez (commented):
These changes seem simple enough, but could I get a bit of context on why we are adding this? The ticket for this referenced the acceptance criteria but didn't give much explanation otherwise

I think I also second the comments on testing. I wonder if there would be a way to do extra testing on this by possibly adding to the demo app or something

noguier:
hey @saralvasquez , thank you for your question! this ticket came as a follow up for this ask: https://github.com/braintree/braintree_android/issues/1662. Here we are asked to do the compose for GooglePay, but upon investigation I found that SEPADirectDebit and some other launchers can also use compose friendly constructor :) 

I think adding that compose for the demo might be a good middle ground in testing the constructor out, I will add that!

SEPADirectDebit/src/test/java/com/braintreepayments/api/sepadirectdebit/SEPADirectDebitLauncherUnitTest.kt
noguier: Hi all, i removed this test, and instead wired a new constructor in the demo app, I dont think we can add a useful unit test for this change since all the constructor is doing is passing the registry to browserSwitchClient, that then does all the job. I feel like wiring the new constructor is comprehensive enough to ensure the flow works as expected. Let me know what you think :)

SEPADirectDebit/src/main/java/com/braintreepayments/api/sepadirectdebit/SEPADirectDebitLauncher.kt
```diff
@@ -20,6 +21,15 @@ class SEPADirectDebitLauncher internal constructor(private val browserSwitchClie
...
+    constructor(registry: ActivityResultRegistry) : this(
+        browserSwitchClient = BrowserSwitchClient(registry)
+    )
```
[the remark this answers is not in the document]
noguier: the constructor was wired in the Demo app.

SEPADirectDebit/src/main/java/com/braintreepayments/api/sepadirectdebit/SEPADirectDebitLauncher.kt
```diff
@@ -20,6 +21,15 @@ class SEPADirectDebitLauncher internal constructor(private val browserSwitchClie
         browserSwitchClient = if (caller != null) BrowserSwitchClient(caller) else BrowserSwitchClient()
     )
 
+    /**
+     * Used to launch the SEPA mandate in a web browser and deliver results. Use it in Compose flows
+     * or anywhere else an [ActivityResultCaller] is not available.
+     * @param registry ActivityResultRegistry parameter. It will be passed to BrowserSwitchClient.
+     */
+    constructor(registry: ActivityResultRegistry) : this(
+        browserSwitchClient = BrowserSwitchClient(registry)
```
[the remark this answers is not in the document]
noguier: These constructors match the same pattern that Venmo and PayPal Launchers have today.

saralvasquez (approved):
Thanks for adding the change to the demo app! I think this is a good approach since testing these changes is challenging

SEPADirectDebit/src/main/java/com/braintreepayments/api/sepadirectdebit/SEPADirectDebitLauncher.kt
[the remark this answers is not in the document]
mzlangreder: We should still validate (and they may be broken on those other Launchers if this ends up being true)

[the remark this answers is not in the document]
noguier: Added unit tests to test the behavior outlined by copilot. The code didnt not crash, so I believe we can continue with the merge.

agedd (approved):
looks good! 🚀

PR #1707, opened 2026-09-16
Fix: allow a custom ActivityResult registration key

noguier:
### Summary of changes
We received a merchant inbound that stated if the merchant have two fragments initialized at the same time with the same GooglePayLauncher, for example 2 fragments run in the web view in their own Fragment, and both require GooglePayLauncher, one of the flows fails with : IllegalStateException. This is caused internally because the ActivityResultContract is being registered in a previously (non-destroyed) Fragment in the stack, which is causing a collision in the map and registration to not be completed.

Merchant ask: If possible, being able to pass in a custom key value would be appreciated.

After the analysis the only affected launchers are ThreeDSecureLauncher and GooglePayLauncher, as they each registered their ActivityResultContract under a hardcoded key.

Summary of changes

- GooglePayLauncher  added a new constructor with a resultKey - a custom key for initialization. Added 3 new constructor overloads to accept a new parameter
- ThreeDSecureLauncher  added a new constructor with a resultKey - a custom key for initialization. Added 2 new constructor overloads to accept a new parameter
- Fix affected tests

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
debugging and fixing tests

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry

@noguier


GooglePay/src/test/java/com/braintreepayments/api/googlepay/GooglePayLauncherUnitTest.kt
```diff
@@ -81,6 +81,24 @@ class GooglePayLauncherUnitTest {
         }
     }
 
+    @Test
+    fun `when GooglePayLauncher is constructed with a custom result key, that key is used to register`() {
```
mzlangreder, line 85: Should we add some tests around multiple registrations like same key/different key?

ThreeDSecure/src/test/java/com/braintreepayments/api/threedsecure/ThreeDSecureLauncherUnitTest.kt
```diff
@@ -58,12 +58,29 @@ class ThreeDSecureLauncherUnitTest {
         }
     }
 
+    @Test
+    fun `when constructed with a custom result key, registers an activity result launcher with that key`() {
```
mzlangreder, line 62: Same question here RE: multiple registration testing.

noguier: great catch will add !

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayLauncher.kt
```diff
@@ -48,8 +49,9 @@ class GooglePayLauncher internal constructor(
     }
 
     /**
-     * Used to launch the Google Pay payment sheet from within an Android Fragment. This class must be
-     * instantiated before the Fragment is created.
+     * Used to launch the Google Pay payment sheet from within an Android Fragment. This class must
+     * be instantiated after the Fragment's view has been created (i.e., in onCreateView or
```
saralvasquez, line 53: Just double checking. I remember someone asking about the change from the comment saying "must be instantiated before" to "must be instantiated after". We confirmed that the comment was wrong before right?

GooglePay/src/test/java/com/braintreepayments/api/googlepay/GooglePayLauncherUnitTest.kt
```diff
@@ -81,6 +81,90 @@ class GooglePayLauncherUnitTest {
...
+            registry.register(
+                any(), eq(lifecycleOwner),
+                any<TaskResultContracts.GetPaymentDataResult>(),
+                any()
+            )
+        }
+    }
+
+    @Test
+    fun `when two launchers are constructed with the same custom key, construction does not throw`() {
```
saralvasquez, line 141: I thought that the keys had to be unique. Shouldn't there be some kind of error thrown if the keys are the same?

ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureLauncher.kt
```diff
@@ -30,8 +31,9 @@ class ThreeDSecureLauncher internal constructor(
     }
 
     /**
-     * Used to launch the 3DS authentication flow to tokenize a 3DS card. This class must be
-     * instantiated before your Fragment is created.
+     * Used to launch the 3DS authentication flow to tokenize a 3DS card. This class must
+     * be instantiated after the Fragment's view has been created (i.e., in onCreateView or
```
saralvasquez, line 35: Same comment wording question as above

ThreeDSecure/src/test/java/com/braintreepayments/api/threedsecure/ThreeDSecureLauncherUnitTest.kt
```diff
@@ -58,12 +58,93 @@ class ThreeDSecureLauncherUnitTest {
...
+            registry.register(
+                any(), eq(lifecycleOwner),
+                any<ActivityResultContract<ThreeDSecureParams?, Any>>(),
+                any()
+            )
+        }
+    }
+
+    @Test
+    fun `when two launchers are constructed with the same custom key, construction does not throw`() {
```
saralvasquez, line 116: Same behavior question as above

GooglePay/src/test/java/com/braintreepayments/api/googlepay/GooglePayLauncherUnitTest.kt
```diff
@@ -81,6 +81,90 @@ class GooglePayLauncherUnitTest {
...
+        val registry = mockk<ActivityResultRegistry>(relaxed = true)
+
+        every {
+            registry.register(
+                eq(sameKey),
+                any(),
+                any<TaskResultContracts.GetPaymentDataResult>(),
+                any()
+            )
+        } returns activityResultLauncher
```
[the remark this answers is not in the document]
saralvasquez: If we do want to allow the same key this feels like it could be a valid concern. What do you think?

saralvasquez (commented):
This looks good to me. I just have a couple questions

GooglePay/src/test/java/com/braintreepayments/api/googlepay/GooglePayLauncherUnitTest.kt
noguier: Great question, 

The whole premise of this issue was not that the keys were colliding, it was because if the Fragment is registering one key, then another fragment is created using that key, then the first Fragment is getting destroyed, and since it was the original creator of the key, the key gets destroyed too. The second fragment now has the launcher without the key. When we try to invoke the launcher without the key, that is when the Exception gets thrown. 

The scenario is pretty specific, but providing a custom key in this scenario will solve it. The test is just to prove that the launcher __can__ be constructed with two keys of the same value.

I hope it makes sense.

GooglePay/src/main/java/com/braintreepayments/api/googlepay/GooglePayLauncher.kt
noguier: Yes, the comment was misleading before because we cannot access fragment.viewLifeCycleOwner, as it is not available prior to fragment creation!

GooglePay/src/test/java/com/braintreepayments/api/googlepay/GooglePayLauncherUnitTest.kt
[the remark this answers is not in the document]
noguier: i tested it manually by creating an app from scratch and mimicking the behavior, but i will see if I can programmatically add tests that test this specific scenario, thank you!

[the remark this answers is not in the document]
noguier: I have added a test that will mimic the actual behavior we fixed with allowing custom keys.

saralvasquez (approved):
Ok I think this looks good! Thanks for making those test changes, they look great!

mzlangreder (approved):

PR #1708, opened 2026-09-17
feat(payment-actions) Demo App Integration

mzlangreder:
### Summary of changes
Add Demo App integration of Payment Actions. An MVVM set of fragments have been added as a new section to the demo app.

* Create Client Token and payment method selection
* Credit Card Payment Method entry screen with submit
* Result screen

#### Recording
CleanShot 2026-09-21 at 12 52 32

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [x] Other (GLM, Deepseek, Kimi)

**How was AI used?**
Implementation planning, execution, and reviewing.

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @mzlangreder

---


saralvasquez:
Small nit: I noticed there was button to take you to the credit card submission page after creating the payment action. Are we intending on adding other payment types as option on that page? If not, would it make more sense to just put the card info and submission all on the same page as the "create payment action" button?

Demo/src/main/java/com/braintreepayments/demo/AutofillHelper.java
```diff
@@ -15,6 +15,7 @@ public class AutofillHelper {
     void fillCardNumber(String cardNumber) {
         cardForm.getCardEditText().requestFocus();
         Editable editable = cardForm.getCardEditText().getText();
+        editable.clear();
```
saralvasquez, line 18: I assume this is to clear the field if it has stuff in it before auto-filling?

Demo/src/main/java/com/braintreepayments/demo/PaymentActionsMerchant.kt
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Locale
+import retrofit2.Call
+import retrofit2.Callback
+import retrofit2.Response
+
+object PaymentActionsMerchant {
+
+    fun createPaymentAction(
+        context: Context,
+        confirmationMethod: ConfirmationMethod = ConfirmationMethod.AUTOMATIC,
```
saralvasquez, line 14: Just out of curiosity because I haven't been following this feature very well, there are other confirmation and capture methods right? Would those result in different behavior in the SDK, and if so will those be added to the demo later too? Or is this the only combo of methods that means anything to the SDK?

Demo/src/main/java/com/braintreepayments/demo/PaymentActionsMerchant.kt
```diff
@@ -0,0 +1,85 @@
...
+
+                override fun onFailure(call: Call<CreatePaymentActionResponse>, throwable: Throwable) {
+                    callback.onResult(CreatePaymentActionResult.Error(Exception(throwable)))
+                }
+            })
+    }
+
+    enum class ConfirmationMethod {
+        AUTOMATIC,
+        MANUAL,
```
saralvasquez, line 73: Op yeah there's the other methods. So will other demo additions come later?

Demo/src/main/java/com/braintreepayments/demo/PaymentActionsFragment.kt
```diff
@@ -0,0 +1,136 @@
...
+                )
+            }
+
+            state.clientToken?.let { token ->
+                Text(
+                    text = stringResource(
+                        R.string.payment_actions_client_token_placeholder,
+                        token.take(CLIENT_TOKEN_PREVIEW_LENGTH)
+                    ),
+                    style = MaterialTheme.typography.bodySmall
```
saralvasquez, line 88: Any reason for the bodySmall instead of medium? Just curious

Demo/src/main/java/com/braintreepayments/demo/PaymentActionMethodViewModel.kt
```diff
@@ -0,0 +1,129 @@
...
+        const val STATE_UNKNOWN = "UNKNOWN"
+        const val STATE_SERVER_ACTION_REQUIRED = "SERVER_ACTION_REQUIRED"
+    }
+}
+
+sealed class PaymentActionMethodEvent {
+    class PaymentMethodRequired(val id: String) : PaymentActionMethodEvent()
+    class CustomerActionRequired(val id: String) : PaymentActionMethodEvent()
+    class ServerActionRequired(val id: String, val serverAction: ServerAction) : PaymentActionMethodEvent()
+    class Terminal(val state: String, val id: String, val detail: String = "") : PaymentActionMethodEvent()
```
saralvasquez, line 128: Another "just for my info", this seems like the final state after everything has been completed. Is that going to be something the merchant will have to pay attention to? Or is that only something the SDK will need to pay attention to?

saralvasquez (approved):
Just a few "gaining knowledge" questions but this looks pretty darn great to me! I tested this on emulator and physical device and everything runs smoothly

saralvasquez (commented):
Op I spoke too soon. I did a bit more testing with screen rotation and such and it looks like the credit card screen won't let you scroll so when you rotate it you can't access the submit or autofill buttons

mzlangreder:
> Small nit: I noticed there was button to take you to the credit card submission page after creating the payment action. Are we intending on adding other payment types as option on that page? If not, would it make more sense to just put the card info and submission all on the same page as the "create payment action" button?

Yep, I am trying to plan ahead a bit. As of now we will support Credit Card, but in the future we plan to support many (most?) payment methods via Payment Actions.

Demo/src/main/java/com/braintreepayments/demo/AutofillHelper.java
mzlangreder: Yes, or if the user pressed autofill a second time.

Demo/src/main/java/com/braintreepayments/demo/PaymentActionsMerchant.kt
mzlangreder: This is a good point. As far as I can tell we don't handle/manage any confirm/capture flows in the Demo app today. We _should_ add the ability to alter confirmation/capture method from the demo app though, so that we can see the expect status on the result screen.

Demo/src/main/java/com/braintreepayments/demo/PaymentActionsMerchant.kt
mzlangreder: No I should add toggles for them now. I will do this.

Demo/src/main/java/com/braintreepayments/demo/PaymentActionsFragment.kt
mzlangreder: Nope, the robot picked for me here. Do we have a preference?

agedd (approved):
this is looking great! tested it out and everything runs smoothly 🚀

Demo/src/main/java/com/braintreepayments/demo/PaymentActionsFragment.kt
saralvasquez: Ah ok! I just figured keeping it consistent across all the fields would make sense. It looks a lil odd having it be slightly smaller

Demo/src/main/java/com/braintreepayments/demo/PaymentActionMethodViewModel.kt
mzlangreder: Yep, these are handoff/terminal states that a merchant would need to handle. A seam where we transfer control to the merchant for various actions.

Demo/src/main/java/com/braintreepayments/demo/PaymentActionsMerchant.kt
mzlangreder: Done and added recording to PR description

Demo/src/main/java/com/braintreepayments/demo/MainFragment.kt
```diff
@@ -226,6 +227,12 @@ class MainFragment : BaseFragment() {
         }
     }
 
+    private fun launchPaymentActions() {
+        // Payment Actions does not use the demo app authString; the flow creates its own
+        // payment action scoped client token via the merchant server.
+        findNavController().navigate(MainFragmentDirections.actionMainFragmentToPaymentActionsFragment())
```
buzzamus, line 233: Is the comment here talking about how a client token must specifically be generated in order to verify that Payment Actions is enabled for the account?

Demo/src/main/java/com/braintreepayments/demo/PaymentActionFlowState.kt
```diff
@@ -0,0 +1,69 @@
...
+        is PaymentActionFlowState.PaymentMethodRequired -> {
+            Text(
+                modifier = Modifier.padding(vertical = 8.dp),
+                text = stringResource(R.string.payment_actions_id_placeholder, state.id)
+            )
+            Text(text = stringResource(R.string.payment_actions_select_method))
+        }
+        is PaymentActionFlowState.CustomerActionRequired -> {
+            Text(
+                modifier = Modifier.padding(vertical = 8.dp),
```
buzzamus, line 42: I see the value of `8` used a lot in multiple places....would it make sense to set it as a constant, to use as a vertical padding default?

Demo/src/main/java/com/braintreepayments/demo/PaymentActionFlowState.kt
```diff
@@ -0,0 +1,69 @@
...
+            }
+        }
+        is PaymentActionFlowState.Processing -> {
+            Text(
+                modifier = Modifier.padding(vertical = 8.dp),
+                text = stringResource(R.string.payment_actions_id_placeholder, state.id)
+            )
+            Button(
+                modifier = Modifier
+                    .padding(vertical = 4.dp)
```
buzzamus, line 61: Same with 4...maybe `VERTICAL_PADDING_MED` and `VERTICAL_PADDING_SM`? Probably overkill

Demo/src/main/java/com/braintreepayments/demo/PaymentActionMethodViewModel.kt
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Owns the method agnostic submit -> branch -> handleNextAction loop driven by [PaymentActionResult].
+ * Subclasses capture payment method details, build a [PaymentActionRequest], and call [submit].
+ * Future integrations (e.g. 3D Secure) hook into the CustomerActionRequired event and drive
+ * [advance] from their result callbacks.
+ *
+ * The client's callback overloads are intentionally not used: they launch into the client's own
+ * scope without error handling and the service's getPaymentAction is still a stub that throws
+ * NotImplementedError (an Error, not an Exception).
+ */
```
buzzamus, line 30: Thanks for the helpful comments!

Demo/src/main/java/com/braintreepayments/demo/PaymentActionsCardFragment.kt
```diff
@@ -0,0 +1,125 @@
...
+                        .padding(end = 4.dp),
+                    enabled = state !is PaymentActionFlowState.Submitting,
+                    onClick = { onAutofill() }
+                ) {
+                    Text(text = stringResource(R.string.autofill))
+                }
+                Button(
+                    modifier = Modifier
+                        .weight(1f)
+                        .padding(start = 4.dp),
```
buzzamus, line 91: more `4` padding values 🤨

Demo/src/main/java/com/braintreepayments/demo/PaymentActionsFragment.kt
```diff
@@ -0,0 +1,202 @@
...
+            if (state.isCreating) {
+                Text(text = stringResource(R.string.payment_actions_creating))
+            }
+
+            state.errorMessage?.let { message ->
+                Text(text = message)
+            }
+
+            state.paymentActionId?.let { id ->
+                Spacer(modifier = Modifier.height(8.dp))
```
buzzamus, line 99: 🤨

buzzamus (approved):
Looks good! Left a few comments that are mostly take it or leave it and not really a concern and a general question, but overall nice work!

Demo/src/main/java/com/braintreepayments/demo/MainFragment.kt
mzlangreder: Yes. Payment Actions needs a client token scoped to the specific payment action, so the demo calls a dedicated create-payment-action endpoint on the merchant server that returns both the paymentActionId and a scoped client_token (that token is what PaymentActionsClient gets constructed with later in the flow). One extra wrinkle: that endpoint cannot fall back to the gateway default merchant account the way /client_token does, so an explicit merchant account ID has to be resolved first. See PaymentActionsMerchant.kt.

Demo/src/main/java/com/braintreepayments/demo/PaymentActionFlowState.kt
mzlangreder: I looked into it and decided to leave as-is: every other Compose screen in the demo (MainFragment, ShopperInsightsFragmentV2, ComposeUIComponentsFragment, GooglePayFragment) also hardcodes dp literals in Modifier.padding(), and dimens.xml is only referenced by the legacy XML layouts. Extracting constants only in the new screens would be inconsistent. If we want to go that direction, a follow-up cleanup across the whole demo app would make more sense.

Demo/src/main/java/com/braintreepayments/demo/PaymentActionFlowState.kt
mzlangreder: Agreed, probably overkill. Leaving as-is for consistency with the rest of the demo Compose screens (same rationale as the 8dp thread).

Demo/src/main/java/com/braintreepayments/demo/PaymentActionMethodViewModel.kt
mzlangreder: Glad they helped!

Demo/src/main/java/com/braintreepayments/demo/PaymentActionsCardFragment.kt
mzlangreder: Same rationale as the padding threads above. Leaving as-is to match the rest of the demo Compose screens.

Demo/src/main/java/com/braintreepayments/demo/PaymentActionsFragment.kt
mzlangreder: Same as above. Leaving the hardcoded dp values for consistency with the other Compose screens.

